PR #2, opened 2017-05-27
Enhance mapper certificates_enddate.py

JoySnow:
Add method expiration_date() to replace get_expiration_date(),
  this new method will return a namedtuple which contains the
  raw data and parsed datetime object of expiration date.


bfahr:
@JoySnow please ensure your commit message includes a description of the change, see [this reference](https://chris.beams.io/posts/git-commit/) for more information.

insights/mappers/certificates_enddate.py
```diff
@@ -67,3 +72,18 @@ def get_expiration_date(self, path):
...
+                  expiration date in string and datetime format.
+        :raises Exception: unparsable expiration data
+        """
+        path_date = self.data.get(path)
+        if path_date:
+            try:
+                path_datetime = datetime.strptime(path_date, '%b %d %H:%M:%S %Y')
+                return self.ExpirationDate(path_date, path_datetime)
+            except:
+                raise Exception("Unparsable expiration data of %s" % path)
```
bfahr: Since this exception is raised here the calling code would need to handle an exception.  Is there a better way to handle this, maybe by checking for valid data in `parse_content` instead of storing `None`.  Should the path be present in the `data` attribute if there is no valid `datestamp`?  If it is useful to know that the `datestamp` is invalid then the better approach would be to keep only valid data in this mapper, and create a separate mapper for invalid `datestamp` data.

JoySnow: > maybe by checking for valid data in parse_content instead of storing None
We'd better not check the data is valid or not in parse_content, that will cost too much and is not necessary. 
And about the `Exception`, we can return `return self.ExpirationDate(path_date, None)` instead of raise an `Exception`.

insights/mappers/certificates_enddate.py
```diff
@@ -29,19 +29,24 @@
...
+from collections import namedtuple
+from .. import Mapper, mapper, LegacyItemAccess
 
 
 @mapper("certificates_enddate")
-class CertificatesEnddate(Mapper):
+class CertificatesEnddate(LegacyItemAccess, Mapper):
     "Class to parse the expiration dates"
 
+    ExpirationDate = namedtuple('ExpirationDate', ['str', 'datetime'])
```
JoySnow: Yes, sure.

insights/mappers/certificates_enddate.py
```diff
@@ -67,3 +72,18 @@ def get_expiration_date(self, path):
...
         except:
             raise Exception("Unable to parse the expiration data of %s" % path)
+
+    def expiration_date(self, path):
+        """
+        :param path: the certificate file path
+        :returns: return a namedtuple(['str', 'datetime']) contains the
+                  expiration date in string and datetime format.
+        :raises Exception: unparsable expiration data
+        """
```
JoySnow: Done.

JoySnow:
@bfahr  I put the wanted commit message body into the first comment cell. It seems the right place for it.

insights/mappers/certificates_enddate.py
```diff
@@ -67,3 +73,22 @@ def get_expiration_date(self, path):
...
+    def expiration_date(self, path):
+        """This will return a namedtuple(['str', 'datetime']) contains the
+        expiration date in string and datetime format. If the expiration date
+        is unparsable, the ExpirationDate.datetime should be None.
+
+        Args:
+            path(str): the certificate file path
+
+        Returns:
+            A ExpirationDate for available path. None otherwise.
```
xiangce: Actually, Not "None" here, but an ExpirationDate only the `datetime` is None, the "RAW" format of the datestamp `.str` is still returned.

insights/mappers/certificates_enddate.py
JoySnow: > If it is useful to know that the datestamp is invalid then the better approach would be to keep only valid data in this mapper, and create a separate mapper for invalid datestamp data.    

@bfahr  Hi, I am sorry, I don't understand the meaning of `a separate mapper for invalid datestamp data`, why & how ? :question:

xiangce:
@JoySnow - You'd better submit this PR to 1.x branch in this week, otherwise the `insights-plugins` cannot get use this modification.

JoySnow:
@bfahr  @xiangce  Should I also submit this PR to 1.x branch if we don't need to change this mapper related insights-plugins in one or two weeks ?

PR #4, opened 2017-05-29
Added Shared Mapper for mariadb service parameters check

vishwanathjadhav:
Add Mapper for systemctl show <Service Name> 

Trello: https://trello.com/c/1lz5colL

KCS:  https://access.redhat.com/solutions/2995891


insights/mappers/systemctl_show.py
```diff
@@ -1,11 +1,62 @@
...
 from .. import mapper
-from .. import get_active_lines
-
+from .. import LegacyItemAccess
+from ..mappers import split_kv_pairs
 
-@mapper('systemctl_cinder-volume')
-class SystemctlShowCinderVolume(Mapper):
 
+class SystemctlShow(LegacyItemAccess, Mapper):
```
bfahr: @vishwanathjadhav this class needs to be documented since the mapper classes are based on this class.  Take a look at the generated documentation and you'll see that the base class `SystemctlShow` but there is no documentation for this class.

bfahr: You could move the `parse_content` docstrings to the class level and have most of what you need.

insights/mappers/systemctl_show.py
```diff
@@ -23,10 +74,19 @@ def parse_content(self, content):
         We will also suppressed empty properties in return data.
         """
         data = {}
+        data = split_kv_pairs(content, use_partition=False, ordered=True)
+
+        """Remove empty key"""
+        self.data = {k: v for k, v in data.iteritems() if not v == ''}
```
bfahr: If the sequence of the data is important, indicated by `ordered=True` above, you may want to consider creating a list of tuples and then using that to initialize an `OrderedDict`.

bfahr: You could also look at looping through the values and using `.pop()` to remove empty values on the original `data`.

vishwanathjadhav: Doesn't matter even if it out of order.  We need just key value pair

insights/mappers/systemctl_show.py
```diff
@@ -1,20 +1,74 @@
+"""
+/bin/systemctl show <Services_Name> - Command
+=============================================
+
+This command shows all the paramters for this service, these parameters can be set
+through unit files corresponding to that service or from command line.
+
+Typical output of ``/bin/systemctl show mariadb`` command is::
```
xiangce: One last comment: 
Would it be better to shift each "Typical output" and "Examples" to each corresponding class?

vishwanathjadhav: Yes, good Idea

xiangce:
@vishwanathjadhav - since the master branch changed as per https://github.com/RedHatInsights/insights-core/pull/27, we have to resolve the conflicts before to merge this PR.

Please refer to below steps to resolve the conflicts:
~~~
Step 1: From your project repository, bring in the changes and test.

git fetch origin
git checkout -b osp_mariadb_plugin origin/osp_mariadb_plugin
git merge master   <---- please resove the conflicts manuall

Step 2: Merge the changes and update on GitHub.

git checkout master
git merge --no-ff osp_mariadb_plugin
git push origin osp_mariadb_plugin    <--- not master, but osp_maroadb_plugin
~~~

PR #5, opened 2017-05-30
XFS info improvements

PaulWay:
* Write documentation.
* Extract mount point from file name if possible.
* Store device name in attribute as well.
* Throw errors if the mapper didn't find expected things in the info, like a 'meta-data' and 'data' sections.
* Less verbose ``__repr__`` method
* Test all of the above.  100% code coverage in tests.


bfahr:
@PaulWay there's a problem accessing this pull request, it may be a problem accessing it from your fork. Maybe its related to the project going public today and your fork possibly having different permissions?

bfahr:
Making your fork public may fix the problem.

PaulWay:
Hi @bfahr - yes, it was still private.  I've made it public :-)

PaulWay:
Not sure why that final commit (to fix the path in the docs rst) hasn't come to here.

PaulWay:
Moving this over to https://github.com/RedHatInsights/insights-core/pull/19 since that does seem to be working.

PR #19, opened 2017-05-31
XFS info improvements

PaulWay:
* Write documentation.
* Extract mount point from file name if possible.
* Store device name in attribute as well.
* Throw errors if the mapper didn't find expected things in the info, like a 'meta-data' and 'data' sections.
* Less verbose `__repr__` method
* Test all of the above. 100% code coverage in tests.


insights/mappers/xfs_info.py
```diff
@@ -1,16 +1,72 @@
...
+        '16777184 blks' as a string.
+    mount(str): If the mount point can be derived from the file name of the
+        original output, then this attribute contains the reconstructed
+        mount point name.
+    device(str): The device name immediately after the '``meta-data``' section
+        heading.
+    data_size(int): The size of the data segment in bytes, from multiplying
+        the ``blocks`` and ``bsize`` values of the ``data`` section together.
+    log_size(int): The size of the log segment in bytes, from multiplying
+        the ``blocks`` and ``bsize`` values of the ``log`` section together.
```
bfahr: This attributes docstring should be moved to the `XFSInfo` class, which needs to have a docstring to be included in the documentation.  Having it in the module causes them to appear to be module attributes instead of class attributes.

There's also a warning generated from the doc build that needs to be fixed:
```
insights-core/insights/mappers/xfs_info.py:docstring of insights.mappers.xfs_info:46: WARNING: Unknown target name: "xfs_info".
```

insights/mappers/xfs_info.py
```diff
@@ -47,19 +109,46 @@ def parse_content(self, content):
...
         # Calculate a few things
         self.data_size = xfs_info['data']['blocks'] * xfs_info['data']['bsize']
         self.log_size = xfs_info['log']['blocks'] * xfs_info['log']['bsize']
 
+        # Determine mount from file name if possible:
+        mount_name_re = re.compile('xfs_info_(?P<mount>.*)')
+        match = mount_name_re.search(self.file_name)
+        if match:
+            # Try to put slashes back...
+            self.mount = match.group('mount').replace('.', '/')
```
bfahr: `self.mount` should have a documented default value so that it will be present in the object regardless of whether the mount information is parsed.

xiangce:
@PaulWay, Please re-submit this improvement to 1.x branch after this PR is merged, Thanks

PR #20, opened 2017-05-31
Netstat mapper improvements

PaulWay:
* While parsing lines, unpack 'PID/Program name' into 'PID' and 'Program name' fields, and unpack 'Local Address' (if found) into 'Local IP' and 'Port' fields, for convenience.
* Supply a ``rows_by()`` search method which takes a section name and a dictionary, and returns all rows which have matching key/value pairs. The original line for this row is stored under the raw line key.
* Several further test improvements to test all branch paths
* Improved documentation and added documentation for new features.


insights/mappers/netstat.py
```diff
@@ -285,22 +307,40 @@ def add_data(self, line):
...
+        if '/' in pidprogram:
+            pid, program = pidprogram.split('/')
+            self.datalist[-1]['PID'] = pid
+            self.datalist[-1]['Program name'] = program
+        # For convenience, unpack 'Local Address' into 'Local IP' and 'Port'
+        if 'Local Address' in self.datalist[-1]:
+            local_addr = self.datalist[-1]['Local Address']
+            # Assert rather than throw exception here because Netstat's format
+            # isn't going to change, since it's deprecated.
+            assert ':' in local_addr
```
bfahr: @PaulWay can you explain this a bit further? Exceptions should be used instead of asserts in production.

bfahr: There's also another assert on line 278 in the `__init__` method.

PaulWay: Basically, the format of 'netstat -neopa' isn't going to change, and it guarantees that the Local Address is always in address:ip format.  Since netstat is deprecated in favour of 'ss', and when we start parsing 'ss' output we're going to almost definitely have to rewrite this mapper, I consider the assert a reasonable short-term way to make sure that we can find out if something's wrong with the parser in testing, but we don't need to make up fictional test cases with made-up bad data to test the parser properly.  I came to that conclusion while writing up fictional test cases with made-up bad data :-)

I'm happy to go with whatever you'd prefer though.  If you think that the assert is useless, or that we need to check more thoroughly, then let me know how you want me to proceed :-)

insights/parsers/netstat.py
```diff
@@ -285,22 +307,40 @@ def add_data(self, line):
...
+            self.datalist[-1]['Port'] = parts[-1]
+        # Unix socket information doesn't have Local Address.
 
     def _merge_data_index(self):
         merged_data = {}
         component_len = {}
-        i = 0
-        while i < len(self.indexes):
+        for i, index in enumerate(self.indexes):
             index = self.indexes[i]
```
xiangce: After `enumerate` being used, line-340 is not necessary, `index` is self.indexes[i]. 
also in line-343, the "self.indexes[i]" can be replaced with `index`

insights/mappers/netstat.py
bfahr: We are trying to standardize on exceptions and I don't think its a good idea to leave something hanging around, even if its going to change in the future, because we don't know when the future change will occur.  It would be better to just raise a ParseException here instead of the asserts.

insights/parsers/netstat.py
bfahr: @xiangce once you are satisified, I approve.

bfahr: @xiangce once asserts are changes to raise ParseException, and you are satisfied I approve.

PaulWay: @xiangce Dangit, missed that obvious one.  Thanks!

PaulWay: @bfahr Raising a ParseException now, thanks for bearing with me on that.

xiangce: @bfahr - Okay, I will go ahead and merge this BR, since @PaulWay has already modified it.

xiangce: @PaulWay - You'd better submit this improvement to 1.x branch after this one is merged.

PR #21, opened 2017-05-31
Enhance grub_conf.py

xiangce:
- Move Grub1Config and Grub2Config to grub_conf.py
  Remove redundant functions among these 3 shared mappers
- Remove some useless variables
- Add get_grub_cmdlines(search_text) to shared reducer GrubConf
- Modify the 2 associated reducers that used the old GrubConfig
  - selinux.py and user_namespaces.py
- Fix a bug in _parse_script to skip nested 'if'
- Refine some docstring
- Enhance coverage of grub_conf.py to 100%


xiangce:
This PR is from falafel (RedHatInsights/falafel#214).

insights/reducers/grub_conf.py
```diff
@@ -71,55 +67,54 @@
...
+            contains the `search_text` in below format:
+            - 'name': "Red Hat Enterprise Linux Server"
+            - 'cmdline': "kernel /vmlinuz-2.6.32-431.11.2.el6.x86_64 crashkernel=128M rhgb quiet"
         """
-
-        self.is_kdump_iommu_enabled = False
-        self.kernel_initrds = self.grub2.kernels_initrds
+        if not search_text:
+            raise Exception("None or empty `search_text` is not allowed.")
+        return [entry for entry in self.grub.boot_entries if search_text in entry.cmdline]
```
bfahr: @xiangce would it be better to just return an empty list here rather than raise a generic exception?  If `search_text` is the wrong time than a different exception will be raised, and if `search_text` is `''` then all lines will be returned which doesn't seem to be the correct answer.

xiangce: Make sense! will modify it.

insights/combiners/grub_conf.py
```diff
@@ -60,66 +59,78 @@
...
+        # get grub configuration
+        self.grub = None
+        if Grub1Config in shared:
+            self.grub = shared.get(Grub1Config)
+        elif Grub2Config in shared:
+            self.grub = shared.get(Grub2Config)
+        elif Grub2EFIConfig in shared:
+            self.grub = shared.get(Grub2EFIConfig)
+        else:
+            raise Exception("No Grub Configuration.")
```
bfahr: This should never happen, but in general we don't want to raise exceptions to rules.

PR #27, opened 2017-05-31
Use new naming conventions.

csams:
Components supporting mappers, shared reducers, regular reducers, etc. have been renamed to parser, combiner, rule, etc. This includes decorators, class names, package names, and documentation references. "Reducer" is still referenced in some places where it couldn't be removed without affecting evaluator logic. Those references will be removed in a PR containing evaluator changes.


kylape:
@csams any idea why a shippable build wasn't triggered?

csams:
@kylape I'm not sure. They were having heroku issues around the time I submitted the PR, and I had some builds fail with statuses like "couldn't find code in repo" and "couldn't find shippable.yml/master." I also had some OAuth token issues with them at the time. Travis picked up and successfully tested my last commit, though.

PR #30, opened 2017-05-31
Prepare for release on pypi

kylape:
- Clean up dependencies
- Add package metadata

Please rebase instead of squash, thanks.


setup.py
```diff
@@ -43,6 +42,16 @@
         name=name,
         version=package_info["VERSION"],
         description="Insights Application Programming Interface",
+        author_email="insights@redhat.com",
+        license="Apache 2",
+        url="https://access.redhat.com/insights",
+        classifiers=[
+            "Development Status :: 5 - Production/Stable",
+            "License :: OSI Approved :: Apache Software License",
```
bfahr: @kylape this may not be necessary but should the license be specified as `Apache License, Version 2.0`, here and above for `license`?

kylape: I wasn't sure what the "official" terms where. I found this one [here](https://pypi.python.org/pypi?%3Aaction=list_classifiers), but I didn't find an official list of license terms.

bfahr: I looked at a few on PyPI and there inconsistency in the examples, even those from the Apache Foundation.

PR #33, opened 2017-06-01
Enhance mapper certificates_enddate.py

JoySnow:
Add method expiration_date() to replace get_expiration_date(),
  this new method will return a namedtuple which contains the
  raw data and parsed datetime object of expiration date.


xiangce:
@JoySnow - Please update plugins in insights-plugins that used the get_expiration_date to expiration_date after this PR is merged. Then back here remove the deprecated get_expiration_date.

JoySnow:
@xiangce Got it! :+1:

PR #17, opened 2017-05-31
cpuinfo_max_freq mapper is no longer in use - removing mapper files entirely

PaulWay:

PR #34, opened 2017-06-01
cpuinfo_max_freq mapper is no longer in use - removing mapper files entirely

PaulWay:
Not sure what went wrong with https://github.com/RedHatInsights/insights-core/pull/17 but the files were not actually removed.  This change does so.


xiangce:
This PR is the same to #17 which is already approved.

PR #36, opened 2017-06-01
Enhance httpd_conf mapper and shared reducer

skontar:
Mapper:
- improved full_data field to store not only last found
  value, but also all previous ones in a list

Reducer:
- store parsed data in data field not only the last value,
  but also all found values as the list.
- added methods get_active_setting, get_setting_list to get
  more complex data structures with more data than backward
  compatible method get_valid_setting

https://bugzilla.redhat.com/show_bug.cgi?id=1455516


insights/combiners/httpd_conf.py
```diff
@@ -114,42 +117,64 @@ def __init__(self, local, shared):
...
-        if section:
-            return self.data.get(section, {}).get(directive)
-        return self.data.get(directive)
+        values_list = self.get_setting_list(directive, section)
+        if values_list is None:
+            return None
+        else:
+            return values_list[-1]
+
+    def get_valid_setting(self, directive, section=None):
```
jsvob: The difference between get_active_setting and get_valid_setting is confusing and not explained.

jsvob: Also, the difference, if any, between "active" and "valid" is not explained.

insights/parsers/httpd_conf.py
```diff
@@ -104,11 +104,13 @@ class HttpdConf(LegacyItemAccess, Parser):
 
     Attributes:
         data (dict): Dictionary of parsed data in format of {option: value} or
-                     {section: {option: value}}.
-        full_data (dict): Dictionary of parsed data with key being option and value a named tuple
-                          with the following properties:
+                     {section: {option: value}}. Stores last found value.
+        full_data (dict): Dictionary of parsed data with key being option and value a list of named
```
jsvob: ...with key being *the* option...

Without *the*, I couldn't parse the sentence for a half a minute.

insights/combiners/httpd_conf.py
skontar: "get_valid_setting" is inferior legacy method used in older rules, so it is left unchanged. Difference between "get_active_setting" and "get_valid_setting" is explained in their respective docstrings. What do you find confusing and not explained there?

insights/parsers/httpd_conf.py
```diff
@@ -119,9 +121,9 @@ class HttpdConf(LegacyItemAccess, Parser):
 
     def __init__(self, *args, **kwargs):
         self.data = {}
-        self.full_data = {}
-        self.first_half = {}
-        self.second_half = {}
+        self.full_data = defaultdict(list)
+        self.first_half = defaultdict(list)
+        self.second_half = defaultdict(list)
```
bfahr: @skontar since these are type `defaultdict` they could inadvertently be modified if the rule checks for a key that does not exist.  One option is to convert these to `dict` once you have finished building them.

skontar: @bfahr good point and great catch. I will fix that.

insights/combiners/httpd_conf.py
jsvob: The explanation makes sense.
I didn't find a comment about get_valid_setting() being a legacy method.
While get_valid_setting and get_active_setting are so similar, there's no explanation whether there is any human meaning difference between "active" and "valid" and the only way for me to find out is to spot the different representation of the output data.
I suggest adding a short comment about it being a legacy method.

PR #42, opened 2017-06-01
Modify 1.x for rule analysis

songuyen:
- Modify settings.py so it will work with falafel as egg file
- Adding extract_dir option when extract tar file from buffer


falafel/settings.py
```diff
@@ -1,32 +1,43 @@
...
 INSTALL_DIR = os.path.dirname(os.path.abspath(__file__))
 NAME = "falafel.yaml"
 DEFAULTS_NAME = "defaults.yaml"
 
 PATHS = [
-    os.path.join(INSTALL_DIR, DEFAULTS_NAME),  # Defaults
-    os.path.join("/etc", NAME),  # System-wide config
-    os.path.join(os.path.expanduser("~/.local"), NAME),  # User-specific config
-    "." + NAME  # Directory-specific config
+    ("resource", DEFAULTS_NAME),  # Defaults
```
bfahr: @songuyen I just wanted to confirm that you no longer need to join `INSTALL_DIR` to `DEFAULTS_NAME`?

songuyen: @bfahr yes, we don't need it because pkgutil is taking care of that

Issue #14, opened 2017-05-30
Stacked Decorators are Bad


jhjaggars:
_From @kylape on May 4, 2017 21:1_

We shouldn't let developers stack insights decorators. Currently, they only work if all specs in a stack are pattern specs or if only one spec in a stack is present, and even then, the expected output is hard to reason about.

Example:

    @mapper('foo')
    @mapper('bar')
    class FooBar(Mapper):
        pass

Here's the current code that attempts to handle the pattern spec case. Note this forces the evaluator to assume all mapper outputs are lists.

    def collect_results(results_dict):
        plugin_output = defaultdict(dict)
        shared_output = {}
        for producer, output in results_dict.iteritems():
            if producer.shared:
                if not producer._reducer:
                    is_pattern = pattern_file(producer.symbolic_names[0])
                    shared_output[producer] = output if is_pattern else output[0]
                else:
                    shared_output[producer] = output
            else:
                plugin = sys.modules[producer.__module__]
                plugin_output[plugin].update(autobox(producer.symbolic_names, output))
        return plugin_output, shared_output

Current stacked mappers:

    (falafel-mark2) [10026 csams@localhost mappers]$ ag '@mapper.*\n@mapper'
    lvm.py
    77:@mapper('pvs')
    78:@mapper('pvs_noheadings')
    141:@mapper('vgs')
    142:@mapper('vgs_noheadings')
    211:@mapper('lvs')
    212:@mapper('lvs_noheadings')

    grub_conf.py
    100:@mapper('grub2.cfg')
    101:@mapper("grub.conf")

    xinetd_conf.py
    99:@mapper("xinetd.conf")
    100:@mapper("xinetd.d")

    limits_conf.py
    43:@mapper("limits.conf")
    44:@mapper("limits.d")
    85:@mapper("limits.conf")
    86:@mapper("limits.d")

    httpd_conf.py
    91:@mapper('httpd.conf', filters=['IncludeOptional'])
    92:@mapper('httpd.conf.d')

    modprobe.py
    5:@mapper('modprobe.conf')
    6:@mapper('modprobe.d')

`pvs`, `vgs`, and `lvs` work, but only because of the `NoneGroup` hack. `grub` works because both versions are simple file specs and neither should show up at the same time.  `httpd_conf` works because its `httpd.conf` spec was changed to a pattern file spec.  `xinetd`, `limits_conf`, and `modprobe` are currently broken.

_Copied from original issue: RedHatInsights/falafel#190_

[...]

PR #46, opened 2017-06-02
MultiFileParser class experiment

PaulWay:
* A MultiFileParser class, like the Parser class, that provides the base for other parsers that take all the related files given by the specs and parses them using a 'parse_context' method.  The 'parse_context' method is given the context so it can see the file name of the the content source, because rules will generally need to report the actual line and/or which file the information came from.
* The ``@parser`` decorator takes an optional 'multifile' boolean parameter, which normally defaults to ``False``.  When set to ``True`` this remembers this parser as one that takes multiple files.
* The ``@parser`` decorator can also take a list as its first argument, which is treated as a list of spec names.  This saves having multiple decorators on one parser class.
* The SingleEvaluator turns all its files into contexts first before iterating through its parsers, rather than repeatedly turning each file into a context for every parser in turn.
* The SingleEvaluator then checks to see if the given parser accepts multiple files, and if so it is given the entire list of file contexts.  Otherwise, each parser is called one after another with each file context in turn.
* The SpecMapper object now has a ``__repr__`` method so that it can be printed when debugging.

Also included is a sample 'ModprobeConfig' parser class to demonstrate how the MultiFileParser class can be used.


bfahr:
@PaulWay I discussed this with @kylape and @jhjaggars and based on that discussion and the discussion in https://github.com/RedHatInsights/insights-core/issues/14 we don't want to had this change at this time.  Over time this problem may go away due to other changes, and the burden of having a 1:1 ratio of specs to mappers is not significant enough to make these changes at this time.  We should close this PR.

PaulWay:
If it's just the list-of-specs-as-first-argument that is a problem, then I can easily remove that.  However, it sounds like I'm outvoted on this so I'll look forward to these other changes :-)

PR #47, opened 2017-06-02
AlternativesOutput mapper class definition

PaulWay:
The AlternativesOutput parser class handles the output of the ``alternatives --display (program)`` command, which can only ever output one program.
    
* The output is self-naming, so subclasses don't have to declare anything.
* The object now has 'program', 'status', 'link', 'best', and 'paths' as attributes rather than as dictionary keys.
* If more than one 'status is' line is found, this raises a ParseException.
* All attributes are initialised to None (or [] for paths).

The main behaviour of this is now tested in tests/test_alternatives_output.py

The JavaAlternatives mapper now uses this class.

All tests pass and have 100% code coverage.

Once merged, rules will be rewritten to use the JavaAlternatives class and the mapper function removed.


kylape:
My opinion is that we should have an `alternatives.py` module that has both `AlternativesOutput` and `JavaAlternatives` in it as well as any additional "alternatives" mappers created in the future.

PaulWay:
I'm more than happy to rename the display_java.py file to alternatives.py and put all the alternatives in it.  But that's not how any of the SysconfigOptions-based  parsers are done, that's not how any of the ls-based file listing parsers are organised, and we have other inconsistencies in both parser naming and file organisation.  So I don't see this as a big problem at the moment.

If you want, I can submit a separate merge request to clean up that technical debt :-)

PR #50, opened 2017-06-05
tomcat_web_xml at 100% code coverage in tests

PaulWay:
Adding test for missing 'session-timeout' value to test other value setting branch path.


PaulWay:
OK, that's weird - why is Travis reporting flake8 errors when I'm not seeing any in my local testing?

xiangce:
@PaulWay - I can also reproduce the flake8 issue on my environment. You'd better use another branch name and do not use `.py` at the end, otherwise, the git-references of this branch will be treated as a `python` file and will be checked by flake8.

bfahr:
@PaulWay, @xiangce is correct that your branch name ending in `.py` triggered this failure.  I have changed the flake8 configuration to ignore the `.git` directory in the future but it would be better to not use `.py` in your branch name.  

@xiangce this PR is ready to merge.

xiangce:
@PaulWay, Please re-submit this improvement to 1.x branch after this PR is merged, Thanks

PR #51, opened 2017-06-05
SMARTctl parser at 100% code coverage in tests

PaulWay:
If device name not found in context, use (and test for) '' as the device name.


insights/parsers/smartctl.py
```diff
@@ -35,12 +35,13 @@ class SMARTctl(Parser):
...
         super(SMARTctl, self).__init__(context)
+        self.data = {}
         filename_re = re.compile(r'smartctl_-a_\.dev\.(?P<device>\w+)$')
         match = filename_re.search(context.path)
-        self.data = {}
         if match:
             self.device = match.group('device')
-        # Else warn?  fail?
+        else:
+            self.device = ''
```
bfahr: @PaulWay any idea if this occurs?  Seems like it might be better to indicate this as `Unknown` or some specific value to make it more explicit.  Have you seen any examples of this in actual data?

miclark: @PaulWay It seems like a failure of this match would indicate a bug in either filling in the `context`s `path` or a bug in the associated `CommandSpec` or the referenced `pre-command`.  As such, it seems that if the path doesn't match, we have a case where we probably should not trust data from the mapper at all.  If that's all true, then I think we should raise an `Error`.

jhjaggars: Also, what would you do with this data if you don't know what device it is related to?

PaulWay: Good points!  I'm a bit cautious about bailing out with an exception if it doesn't match the device name, because if the CommandSpec format changes or we add extra FileSpecs or something then that could cause more things to break.

However, I think Jesse's hit the nail on the head - it's useless information if you don't know which device its attached to.  So let's go with throwing a ParseException.

PR #57, opened 2017-06-06
Certificates enddate parser at 100% code coverage in tests

PaulWay:
* Raise a ParseException, not an Exception, in (deprecated) get_expiration_date method.
* Explicitly return None if path given to expiration_date method not found in data.
* Tests for same


xiangce:
@PaulWay - Sorry for not informing you that the interface `get_expiration_date` is deprecated and is replaced by the newly added `expiration_date`, it will be removed from both the master and 1.x branches.

@JoySnow will remove it once all the `get_expiration_date` are replaced.

I suggest we close this PR and let JoySnow remove it in the near future.

PaulWay:
Hi @xiangce @JoySnow - yep, I saw that and thought that'd be the case; no worries, will close this one.  It'd be good to get 100% code coverage in testing - feel free to copy my code here to do so.

PR #58, opened 2017-06-06
Bond module tests at 100% code coverage

PaulWay:
* Turn the bond prefix list into a map for faster look-ups - I checked all my
  sosreports and the names in this list are all exactly what appears in the
  available bond interface information.
* Raise a ParseException if we don't recognise this map type, because it's
  a bonding type we don't know about.
* Test the above - all branches tested and 100% code coverage.
* Documentation improvements to move the documentation into the parser itself.


insights/parsers/bond.py
```diff
@@ -1,69 +1,31 @@
 """
-bonding - File /proc/net/bonding
-================================
+``bonding`` - network bonding information
+==========================================
```
xiangce: The underline has one more "=", all others look good to me.

PR #62, opened 2017-06-07
Fix netstat rows_by() method to implement AND rather than OR search

PaulWay:
I realised in thinking about it that the ``rows_by()`` search algorithm would find all rows where *any one* of the search criteria was found.  This was confirmed in testing.  This merge fixes this behaviour to only match rows where *all* the search criteria were met.

The code's logic would then match all rows if given an empty dictionary, but it is preferable to return no results given no search criteria.

Tests now also check this behaviour and retain 100% code coverage.


insights/parsers/netstat.py
```diff
@@ -525,12 +525,19 @@ def rows_by(self, section_id, search_dict):
...
+            found = True
             for key, value in search_dict.iteritems():
-                if key in row and row[key] == value:
-                    copy = dict(row)
-                    copy['raw line'] = self.lines[section_id][line]
-                    results.append(copy)
+                if not (key in row and row[key] == value):
+                    found = False
+            if found:
+                copy = dict(row)
```
bfahr: @PaulWay you might consider optimizing this a bit by replacing 533-537 with this:

```python
if all([k in row and row[k] == v for k, v in search_dict.iteritems()]):
```

PaulWay: That is so much neater!  I didn't think of a list comprehension of booleans.  Nice!  Thanks, @bfahr

xiangce:
Since Paul has already updated the code as per Bob's comment, I go ahead and merge this PR 。

PR #64, opened 2017-06-07
Fix exception 'unexpected keyword argument' when calling a rule.

matysek:
This fixes exception that prevents running insights rules.


kylape:
@csams how does this change look to you?

PR #69, opened 2017-06-09
Added ``keyword_search`` function

PaulWay:
One technique used in libraries such as Django is to allow keyword-based searching of data, using the keyword arguments to a function as the search terms and their values as the required values.

This implements that process in a simple fashion on lists of dictionaries representing records in a table.

* No keywords to search for == no data returned.
* No data available == no data returned.
* Only exact matching is supported - relations such as 'is none' or 'greater than' cannot be tested.
* Types are significant - comparison of different types will always be false.
* If a field being searched for is not found in the row, then the row doesn't match.

The function returns a list of the rows that match, as they were in the data, or an empty list if nothing is found.  This function can then be used in parsers to provide a keyword search method on their content, e.g.:

```
    def find_limit(self, **kwargs):
        return keyword_search(self.data, **kwargs)
```

Added tests for this - code coverage at 100%.


insights/parsers/__init__.py
```diff
@@ -277,3 +277,17 @@ def parse_fixed_table(table_lines,
         table_data.append(col_data)
 
     return table_data
+
+
+def keyword_search(rows, **kwargs):
+    """
+    Takes a list of dictionaries and finds all the dictionaries where the
+    keys and values match those found in the keyword arguments.
```
bfahr: @PaulWay since this is in the API would you mind to include more complete documentation and examples to make it easier for developers to understand and use?

bfahr: This should also be added to the `docs/api_index.rst` file to include it in the documentation.

bfahr: BTW, I added some additional documentation to `parsers/__init__`, to follow my own recommendation.

insights/parsers/__init__.py
```diff
@@ -277,3 +277,17 @@ def parse_fixed_table(table_lines,
...
+    """
+    results = []
+    if not kwargs:
+        return results
+    for line, row in enumerate(rows):
```
xiangce: `line` is not used below, `enumerate` is necessary? If no, better iterate the `rows` directly.

insights/parsers/__init__.py
PaulWay: Sure thing - will do.

insights/parsers/__init__.py
PaulWay: That's probably been carried over from my more convoluted logic - good catch, @xiangce, will remove that.

insights/parsers/__init__.py
bfahr: @PaulWay you will still need to add your function to `docs/api_index.rst` in order to include it in the built documentation.

bfahr:
@PaulWay you'll need to rebase your branch against master to fix this conflict.

insights/parsers/__init__.py
PaulWay: Ah, of course!  Thanks @bfahr

bfahr:
Approved for merge once PR has been submitted for 1.x per Issue #86.

PaulWay:
OK, I think I've got this sorted out - try https://github.com/RedHatInsights/insights-core/pull/93 for the 1.x PR.

PR #70, opened 2017-06-09
Limits conf combiner

PaulWay:
At the moment anything using the limits_conf module LimitsConf parser has the same problem previously identified in modprobe.conf - that the SimpleFileSpec returns a scalar but the PatternSpec returns a list.  Even if this were fixed, rules would still have to manually iterate across all files in order to find rules that match.

The preferred solution at the moment is to use a combiner to merge all this data into one source.  This is provided here.

The AllLimitsConf combiner provides the same features that LimitsConf exposes - the ``rules`` and ``domains`` attributes and a ``find_all()`` method for searching.

All tests for this combiner pass and give 100% code coverage.


insights/combiners/limits_conf.py
```diff
@@ -0,0 +1,60 @@
...
+        limits(list): a list of the original LimitsConf parser results.
+        rules(list): the entire list of rules.
+    """
+    def __init__(self, local, shared):
+        rules = []
+        domains = set([])
+        self.limits = shared[LimitsConf]
+        for limits in shared[LimitsConf]:
+            # The rule dictionary contains the file_path as a key for later.
+            rules.extend(limits.rules)
```
bfahr: @PaulWay does order in the conf files matter, meaning should you check which file limits is from first?  Can there be duplicates, for instance could lines in `limits.conf` and the `limits.d/file` be duplicates or override?

insights/combiners/limits_conf.py
```diff
@@ -0,0 +1,60 @@
...
+            domains.update(limits.domains)
+
+        self.rules = rules
+        self.domains = domains
+        super(AllLimitsConf, self).__init__()
+
+    def find_all(self, **kwargs):
```
xiangce: Would it be better for this combiner to provide an interface to return the `active/valid` setting? It seems the `limits.d/file` can override the same entry in `limits.conf`, https://access.redhat.com/solutions/199993.


just like what HttpdConfAll does?

insights/combiners/limits_conf.py
PaulWay: Good question.  There's no mention of how this is handled in the limits.conf man page, apart from:

> In general, individual limits have priority over group limits, so if you impose no limits for admin group, but one of the members in this group have a limits line, the user will have its limits set according to this line.

I believe that the files are read in name order, which is why some limits files are numbered - e.g. '90-nproc.conf'.  However, there's no requirement for files to not contain duplicate limits or for particular limits to be in any particular file - e.g. 'nproc' limits can be defined anywhere, not just in the '90-nproc.conf' file.

We can enforce that order in that loop by making line 28 something like:

`for limits in sorted(shared[LimitsConf], key=lambda x: getattr(x, 'file_path')):`

But @jhjaggars or @kylape may know if the files are presented in name order, which would mean that we default to the 'correct' way of working.

De-duplicating this list is going to be complicated, because we can't compare on the entire row, we have to compare on only the domain, type and item of each row.  A temporary list of that subset would make that possible.

However, the `find_all()` method is going to give people all the limits that match the query.  In most cases that should be a list with a single row, but the matching function is designed to handle the matching rules defined in `man limits.conf` - e.g matching on user, group, user ID range, and more.  Rules which use this output will need to handle multiple rows, which means that the will need to handle duplicates.

Aside - we could provide a method like find_all called, say, `find_most_specific` which attempts to implement the logic of selecting the most specific limit that affects a particular domain and type.  But I'm not sure yet if this is going to be used by any rule so I'm loath to add that complexity.

insights/combiners/limits_conf.py
PaulWay: Nice work finding that, @xiangce - thanks.

chenlizhong:
@PaulWay 

Did you fix the order of the limits files issue that xiangce metioned ?https://access.redhat.com/solutions/199993

PaulWay:
On 14/06/17 11:37, Chenlizhong wrote:
>
> @PaulWay <https://github.com/paulway>
>
> Did you fix the order of the limits files issue that xiangce metioned
> ?https://access.redhat.com/solutions/199993
>
Yes, I checked that and that logic is in both the master and 1.x branches.

HTH,

Paul

-- 
Paul Wayper - Senior Software Engineer - Insights Team

insights/combiners/limits_conf.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    def __init__(self, local, shared):
+        rules = []
+        domains = set([])
+        self.limits = shared[LimitsConf]
+        # Store the tuple of domain, type and item against the row of that
+        # limit in our full list of rules.
+        rule_tuples_found = {}
+        rule_number = 0
+        for limits in shared[LimitsConf]:
```
chenlizhong: the code just iterates shared[LimitsConf], we should sort by filenames.  It should be load limits.conf firstly, the load other files by filename's order

PaulWay: Hi @chenlizhong - from what I can see the items in shared[LimitsConf] will automatically be in file name order by spec, so limits.conf will be listed first and then all the files in limits.d in name order.  This is because they are read into the spec handling system in name order.

chenlizhong: @PaulWay  

Great. the spec handling system handles the order.  I suggest that we add tests for the file name order to sure this in limits conf combiner. One day, someone change the spec system behavior. We can find the issue.

bfahr: @PaulWay relying on the specs to stay in a particular order is probably not a good idea.  If order is important it is probably better to enforce that where it is important.

PaulWay: OK, let's sort explicitly then :-)

bfahr:
@PaulWay if you can go ahead and incorporate these changes to #95 then I'll check that branch first and try to get this merged today.

PR #82, opened 2017-06-12
Enhance httpd_conf mapper and shared reducer (#36)

xiangce:
* Enhance httpd_conf mapper and shared reducer

Mapper:
- improved full_data field to store not only last found
  value, but also all previous ones in a list

Reducer:
- store parsed data in data field not only the last value,
  but also all found values as the list.
- added methods get_active_setting, get_setting_list to get
  more complex data structures with more data than backward
  compatible method get_valid_setting

https://bugzilla.redhat.com/show_bug.cgi?id=1455516

* Resolved review comments

* Remove unused import


xiangce:
Just cherry-pick from #36

PR #84, opened 2017-06-13
Enhance httpd_conf to support nested sections

xiangce:
- IfModule section is nest-able
- Support all the sections (includes nested sections)
- Re-define section name with a tuble in format of:('type', 'name')
  e.g.: ('IfModule', 'prefork.c')
- Remove the useless data and rename the `full_data` to `data`
- Remove the deprecated `get_valid_setting` from master branch
  `insights-plugin` does not depend on the master branch
- Re-design the `get_active_setting`
- Remove the filter: Do not filter httpd_conf from now on


xiangce:
Required by https://trello.com/c/PxWG7Cbx/34-ifmodule-sections-are-nest-able and https://trello.com/c/nnjdeZrM/29-enhance-shard-mapper-httpdconf-and-reducer-httpdconfall

insights/combiners/httpd_conf.py
```diff
@@ -107,78 +130,57 @@ def __init__(self, local, shared):
...
+                if option in self.data and isinstance(parsed_data, dict):
+                    self.data[option].update(parsed_data)
                 else:
-                    # For directive
-                    values = [self.ParsedData(a.value, a.line, file_name, file_path)
-                              for a in parsed_data]
-                    self.data[option] = values
+                    self.data[option] = parsed_data
 
-    def get_setting_list(self, directive, section=None):
```
skontar: The methods 'get_setting_list' and 'get_active_setting' serve different purposes and both of them should be retained. "get_active_setting" is meant to get the last value of directive which does value overwriting, "get_setting_list" is meant for parsing either directives which does value "addition" as explained in case of HTTPD_CONF_MORE test, or for rules which warn even on currently overwritten values - in case that the config file with overwriting is removed in the future. 

Current implementation of "get_active_setting" does not get the active value of the directive so the name is misleading.

We are planing to use both in https://bugzilla.redhat.com/show_bug.cgi?id=1436224 .

insights/combiners/tests/test_httpd_conf.py
```diff
@@ -153,21 +217,8 @@ def test_httpd_one_file_overwrites():
...
     result = HttpdConfAll(None, shared)
 
-    assert result.get_valid_setting('UserDir') == ('enable bob', 'httpd.conf')
-
     active_setting = result.get_active_setting('UserDir')
     assert active_setting.value == 'enable bob'
     assert active_setting.line == 'UserDir enable bob'
     assert active_setting.file_path == '/etc/httpd/conf/httpd.conf'
     assert active_setting.file_name == 'httpd.conf'
-
```
skontar: Please do not remove method tests for valid methods. This is expected behavior. See https://bugzilla.redhat.com/show_bug.cgi?id=1455516 .

skontar:
Adding @kgrant-rh for review, as he is an author of future Apache hardening rule.

insights/combiners/tests/test_httpd_conf.py
```diff
@@ -153,21 +217,8 @@ def test_httpd_one_file_overwrites():
...
-    setting_list = result.get_setting_list('UserDir')
```
skontar: Please do not remove method tests for valid methods (get_setting_list). This is expected behavior. See https://bugzilla.redhat.com/show_bug.cgi?id=1455516 .

insights/parsers/httpd_conf.py
```diff
@@ -89,39 +77,39 @@
...
-    sections are parsed into 'MPM_prefork' and 'MPM_worker' sub-dictionaries
-    respectively.
+    and trailing spaces.
 
     If the file is ``httpd.conf``, it also stores first half, before
     ``IncludeOptional conf.d/*.conf`` line, and the rest, to the ``first_half``
     and ``second_half`` attributes respectively.
 
     Attributes:
-        data (dict): Dictionary of parsed data in format of {option: value} or
```
kgrant-rh: As @skontar mentioned, it is important for some configuration directives to have more than just the last parsed value, because some directives change their meaning based on previous definitions.  Thus the inclusion of `full_data`, for rules which examine those types of directives, as well as `data`, for rules which use only last-parsed types of directives.

skontar: @kgrant-rh I think he actually got rid of what was `data` before and only retains `full_data` and renamed it to `data` (as it is actually a super-set). As `master` branch can break backwards compatibility, I think it is a good move - if I understood the change correctly.

kgrant-rh: If all of the configuration is going to be stored in one parameter, that's fine, as long as it stores what was previously called `full_data`.  In that case, however, `get_setting_list` and `get_active_setting` should provide a clearer interface than referring to the setting list implicitly or referring to the `[-1]` index of that list.

insights/combiners/httpd_conf.py
xiangce: Sorry for that, will revert it in next commit.

xiangce:
@bfahr @kgrant-rh @skontar, would you please to review this PR one more time? Thanks.

xiangce:
https://github.com/RedHatInsights/insights-core/pull/133 to 1.x branch is merged.

PR #81, opened 2017-06-10
Sync with 1.x branch

xiangce:

PR #92, opened 2017-06-15
Remove private URL from uname's docstring

xiangce:
- Remove private URL as per Bob's comment in https://github.com/RedHatInsights/insights-core/pull/81


xiangce:
Hey @bfahr, this PR is used to sync the uname part in https://github.com/RedHatInsights/insights-core/pull/81.

Here I have a question, for this kind of situation that author has already submitted the same PR on both `master` and `1.x` branches (and linked them well) themselves before opening issue, is it necessary to open an issue for it?

Issue #85, opened 2017-06-13
Limits conf combiner #70


bfahr:
@PaulWay you need to create a back-port PR for the 1.x branch for #70

xiangce:
Both #70 and #95 are merged

Issue #86, opened 2017-06-13
Added keyword_search function #69


bfahr:
@PaulWay you need to create a PR for the 1.x branch for #69

bfahr:
1.x branch PR is #93.

xiangce:
Both #93 and #69 were merged, close this issue.

PR #95, opened 2017-06-15
Limits_conf combiner for 1.x branch

PaulWay:
Let's try that again, shall we?  Requesting a merge of the work on https://github.com/RedHatInsights/insights-core/pull/70 for https://github.com/RedHatInsights/insights-core/issues/85


bfahr:
@PaulWay this PR appears to break the rule `nproc_common_entries`, would you mind to check it?

PaulWay:
@bfahr That's because nproc_common_entries used a different set of parsers (in nproc.py) but the same specs as limits_conf.py.

The solution to this is to fix nproc_common_entries so that it uses the limits_conf parsers, and so that it can cope with these changes.  I've submitted a MR for that in https://github.com/RedHatInsights/insights-plugins/pull/86 .  Once that's done we can apply this, then look at removing the nproc parser.

xiangce:
https://github.com/RedHatInsights/insights-plugins/pull/86 is not merged, trying to push #86  forward.

xiangce:
@PaulWay - This PR will cause insights-plugins broken, so we have to wait for your corresponding PR to plugins being merged.

PR #96, opened 2017-06-15
Simplify 'Policy booleans' handling

PaulWay:
Since we always start with an empty dictionary for the policy_booleans data,
and booleans are found in the output by the absence of a colon, we can
simplify the code by adding booleans directly to that dictionary.

This also restores 100% code coverage in testing.


PaulWay:
Sure thing - good thinking @bfahr, should have added that as a matter of course :-)

PR #108, opened 2017-06-19
Improve docs, test coverage and parsing of the system_time parser 

PaulWay:
* Improving the `ntptime` parser to get all the data not just the few bits previously used.
* Code test coverage at 100%
* Improving documentation in line with https://trello.com/c/20o5fhAE/


insights/parsers/system_time.py
```diff
@@ -174,26 +193,45 @@ class NtpTime(Parser):
...
-            'ntp_gettime',
-            'ntp_adjtime'
-        ]
         result = {}
-        return_code_re = re.compile(r'returns code (\d+) ')
-        status_code_re = re.compile(r'\((.*)\)')
+        return_code_re = re.compile(r'(?P<func>ntp_(?:get|adj)time)\(\) returns code (?P<code>\d+) ')
+        status_code_re = re.compile(r'status (?P<status>0x[0-9a-f]{4}) \((?P<flags>.*)\)')
+        time_data_re = re.compile(r'time (?P<timecode>[0-9a-f]{8}\.[0-9a-f]{8}) +(?P<timestamp>\w{3}, \w{3} \d+ .*),')
+        value_re = re.compile(r'(?P<keyword>[A-Za-z][A-Za-z ]+) (?P<value>\d+(?:\.\d+|x[0-9a-f]+)? ?(?:us|ppm|s|\(.*\))?)')
```
lonicerae: Dedicated units for output data are well defined. Can we return the numeric value instead? This can ease the latter comparison in plugin. For example:
~~~
assert result['frequency'] == '4.201'
assert result['interval'] == '1'
assert result['precision'] == '0.001'
assert result['tolerance'] == '500'
~~~

insights/parsers/system_time.py
```diff
@@ -174,26 +193,45 @@ class NtpTime(Parser):
...
-                        result[func] = g.group(1)
-            if "status " in line:
-                frags = line.split()
-                if len(frags) == 3:
-                    result['status'] = frags[1]
-                    g = status_code_re.search(frags[2])
-                    if g and g.lastindex == 1:
-                        result['flags'] = g.group(1).split(',')
+            # Check for function line
+            # print "line:", line
```
lonicerae: Better to remove the debug lines when commit.

PaulWay:
Hi @bfahr - I've updated the documentation here (as well) to move the subheadings into the .rst directly - let me know what you think!

docs/shared_parsers_catalog/system_time.rst
```diff
@@ -1,3 +1,38 @@
...
+
+command ``ntptime``
+-------------------
+.. autoclass:: insights.parsers.system_time.NtpTime
+   :members:
+   :show-inheritance:
+
+file ``/etc/sysconfig/ntpd``
+----------------------------
+.. autoclass:: insights.parsers.system_time.NTPDService
```
bfahr: @PaulWay the `NTPConfParser` class is missing from the documentation in this PR and I just noticed it was also missing from #109 .   Please update this PR and create a new PR in include this class in the 1.x documentation.

PaulWay: Hmmm - it's not supposed to be.  It's a base class for NTP_conf and ChronyConf, so they both work the same way.  It doesn't have a parser decorator for that reason.  I certainly agree that both NTP_conf and ChronyConf need better documentation if they're going to refer to a class that doesn't exist, but I'd rather not mention the base class so as to prevent rule writers thinking they need to use that class.

How about I update the NTP_conf and ChronyConf documentation with a copy of the documentation in NTPConfParser?  What's the best way of getting good documentation into the parsers that we want people to use?

bfahr: @PaulWay it does need to be included in the documentation so that it will be linked from the child class and users will be able to understand implementation of the child class.

PaulWay:
OK, I've added the NTPConfParser base class to the documentation.  How does that look now?

xiangce:
@PaulWay -  The NTPConfParser class is also missing in the 1.x branch, would please create a new PR to add this class to the 1.x documentation?

xiangce:
@PaulWay - Sorry. I got it https://github.com/RedHatInsights/insights-core/pull/215

PR #109, opened 2017-06-19
Improve docs, test coverage and parsing of the system_time parser for the 1.x branch

PaulWay:
See https://github.com/RedHatInsights/insights-core/pull/108 for the master branch


xiangce:
@lonicerae - would you like take a look at this PR?

insights/parsers/system_time.py
```diff
@@ -174,26 +193,45 @@ class NtpTime(Parser):
...
-            'ntp_gettime',
-            'ntp_adjtime'
-        ]
         result = {}
-        return_code_re = re.compile(r'returns code (\d+) ')
-        status_code_re = re.compile(r'\((.*)\)')
+        return_code_re = re.compile(r'(?P<func>ntp_(?:get|adj)time)\(\) returns code (?P<code>\d+) ')
+        status_code_re = re.compile(r'status (?P<status>0x[0-9a-f]{4}) \((?P<flags>.*)\)')
+        time_data_re = re.compile(r'time (?P<timecode>[0-9a-f]{8}\.[0-9a-f]{8}) +(?P<timestamp>\w{3}, \w{3} \d+ .*),')
+        value_re = re.compile(r'(?P<keyword>[A-Za-z][A-Za-z ]+) (?P<value>\d+(?:\.\d+|x[0-9a-f]+)? ?(?:us|ppm|s|\(.*\))?)')
```
lonicerae: Dedicated units for output data are well defined. Can we return the numeric value instead? This can ease the latter comparison in plugin. For example:
~~~
assert result['frequency'] == 4.201
assert result['interval'] == 1
assert result['precision'] == 0.001
assert result['tolerance'] == 500
~~~

insights/parsers/system_time.py
```diff
@@ -174,26 +193,45 @@ class NtpTime(Parser):
...
-                        result[func] = g.group(1)
-            if "status " in line:
-                frags = line.split()
-                if len(frags) == 3:
-                    result['status'] = frags[1]
-                    g = status_code_re.search(frags[2])
-                    if g and g.lastindex == 1:
-                        result['flags'] = g.group(1).split(',')
+            # Check for function line
+            # print "line:", line
```
lonicerae: Better to remove the debug lines when commit.

xiangce:
Since @lonicerae has already posted his comment in #108 (I missed it:( sorry), please ignore above reminder and go on with your discussion on #108.

insights/parsers/system_time.py
PaulWay: I was initially a little worried that the units might change but on reflection that seems unlikely.  Thanks for that @lonicerae, will do.

insights/parsers/system_time.py
PaulWay: Good point, will do.

insights/parsers/system_time.py
PaulWay: OK, question - should we convert hex values to integers or leave them as strings?

PaulWay: Going with converting them to integer.  However, we're now a bit inconsistent with the other parts of the code, where the adjtime and gettime results and the stats value are returned as strings.  I'm going to leave them as is but this is a little awkward.

insights/parsers/system_time.py
```diff
@@ -3,7 +3,26 @@
...
+
+NTP_conf - file ``/etc/ntp.conf``
+---------------------------------
+
+NtpTime - command ``ntptime``
+-----------------------------
+
+NTPDService - file ``/etc/sysconfig/ntpd``
+------------------------------------------
+
```
bfahr: @PaulWay this is not compliant with the documentation style guide.  I think we discussed alternatives to achieve the same effect that would be better.

PaulWay: Yep, I'll fix this.  Thanks @bfahr

PaulWay: Hi @bfahr - I've updated the documentation to move the subheadings into the .rst directly - let me know what you think!

PR #111, opened 2017-06-19
Simplify 'Policy booleans' handling

PaulWay:
Since we always start with an empty dictionary for the policy_booleans data, and booleans are found in the output by the absence of a colon, we can simplify the code by adding booleans directly to that dictionary.

This is the push request for the 1.x branch.


PaulWay:
Working on adding documentation...

PR #114, opened 2017-06-20
Unitfiles parser enhanced to deal with all unit states and enhanced w…

jsvob:
…ith a new method.


insights/parsers/systemd/unitfiles.py
```diff
@@ -45,6 +48,19 @@ def is_on(self, service_name):
...
 
         Returns:
-            bool: True if service is enabled, False otherwise
+            Union[bool, None]: True if service is enabled, False if it is disabled. None if the
+                               service doesn't exist.
         """
-        return self.services.get(service_name, False)
+        return self.services.get(service_name, None)
+
+    def exists(self, service_name):
```
PaulWay: I like this differentiation - good one Jakub.

insights/parsers/systemd/unitfiles.py
```diff
@@ -45,6 +48,19 @@ def is_on(self, service_name):
             service_name (str): service name including '.service'
 
         Returns:
-            bool: True if service is enabled, False otherwise
+            Union[bool, None]: True if service is enabled, False if it is disabled. None if the
+                               service doesn't exist.
         """
-        return self.services.get(service_name, False)
+        return self.services.get(service_name, None)
```
PaulWay: I'm a little wary of tri-state booleans, but I'm happy enough since code that does 'if obj.is_on(service)' will still work correctly.  Nothing should be checking the exact return value for False.

jsvob: The problem with the former solution was that neither the docstring nor the returned data made any indication of the behavior when the service doesn't exist at all, which turned out to be a slight complication in one of my rules, so I fixed it in 2.x for posterity and left it at its original behavior in 1.x for maximum compatibility.

PR #116, opened 2017-06-20
Parsers for the 389 Directory Server access and errors log files

PaulWay:
These use the standard LogFileOutput parser class, including the `get_after()` method.


insights/parsers/dirsrv_logs.py
```diff
@@ -0,0 +1,95 @@
+'''
+389 Directory Server logs
+=========================
+
+'''
+
+from insights import parser, LogFileOutput
+
+
+@parser('dirsrv_access')
```
bfahr: @PaulWay you use the spec `dirsrv_access` here, and you added the spec `dirsrv_logs` in the specs, which one should it be?

PaulWay: Dangit!  How did that ever work?  We need something to check that all specs defined in decorators correspond to specs in the spec file :-)

OK, I've submitted an update for this and have fixed the 1.x branch in #173

PR #118, opened 2017-06-20
LogFileOutput parser class `get_after()` method

PaulWay:
This implements a `get_after` method for the LogFileOutput parser and demonstrates how it would be used by subclassed parsers.

This method is used to find all the log lines after a given date, by extracting the timestamp from each log line based on a `strptime()` time format string. Lines without time stamps are considered to continue the previous log line. The method implemented in LogFileOutput is normally considered 'private' and the individual mapper supplies a public `get_after()` method which supplies the time format template appropriate to that log.

A few log file parsers have been updated to provide the `get_after` method as a demonstration of its use.  Once this is accepted we can look at adding this to the other LogFileOutput derived parsers.


PaulWay:
OK, this time for sure!

PR #119, opened 2017-06-20
LogFileOutput parser class `get_after()` method - 1.x branch

PaulWay:
This implements a `get_after` method for the LogFileOutput parser and demonstrates how it would be used by subclassed parsers.

This method is used to find all the log lines after a given date, by extracting the timestamp from each log line based on a `strptime()` time format string. Lines without time stamps are considered to continue the previous log line. The method implemented in LogFileOutput is normally considered 'private' and the individual mapper supplies a public `get_after()` method which supplies the time format template appropriate to that log.

A few log file parsers have been updated to provide the `get_after` method as a demonstration of its use.  Once this is accepted we can look at adding this to the other LogFileOutput derived parsers.


insights/core/__init__.py
```diff
@@ -426,6 +427,120 @@ def _scan(self):
...
+        }
+
+        # Construct the regex from the time string
+        timefmt_re = re.compile(r'%(\w)')
+
+        def replacer(match):
+            if match.group(1) in format_conversion_for:
+                return format_conversion_for[match.group(1)]
+            else:
+                raise ParseException(
```
bfahr: @PaulWay you need to import `ParseException`.

insights/core/__init__.py
```diff
@@ -426,6 +427,120 @@ def _scan(self):
...
+
+        Params:
+            timestamp(datetime.datetime): lines before this time are ignored.
+            time_format(str): the ``strptime()`` string of the format of the
+                dates in this log.
+            lines(list): an optional list of lines from this parser to search.
+
+        Returns:
+            (list): a list of lines with timestamps after this date.
+        """
```
bfahr: You should add a `Raises:` section to describe how this method raises a `ParseException`.

insights/core/__init__.py
```diff
@@ -426,6 +427,120 @@ def _scan(self):
 
         cls.scan(result_key, _scan)
 
+    def _get_after(self, timestamp, time_format='%Y-%m-%d %H:%M:%S', lines=None):
```
bfahr: Not clear why this needs to be a private method? If the defaults are approriate for the logfile then the base method should work right?  Also, because this is private it is not included in the documentation so a parser developer would have to look at the code to know that it is available.

insights/core/__init__.py
```diff
@@ -426,6 +427,120 @@ def _scan(self):
...
+                    output.append(line)
+                else:
+                    # Earlier - start excluding
+                    including_lines = False
+            else:
+                # If we're including lines, add this continuation line
+                if including_lines:
+                    output.append(line)
+
+        return output
```
bfahr: Would it make more sense for this to be a generator method?

insights/parsers/ceilometer_log.py
```diff
@@ -32,7 +32,21 @@ class CeilometerCentralLog(LogFileOutput):
     Note:
         Please refer to its super-class ``LogFileOutput``
     """
-    pass
```
bfahr: FYI, the `Note:` above is not necessary as Sphinx is configured to identify the base class and provides a link to the base class documentation as part of this class documentation.

insights/parsers/ceilometer_log.py
```diff
@@ -32,7 +32,21 @@ class CeilometerCentralLog(LogFileOutput):
...
+
+        Parameters:
+            timestamp(datetime.datetime): log lines after this time are
+                returned.
+            lines(list): the list of log lines to search (e.g. from a get).
+                If not supplied, all available lines are searched.
+
+        Returns:
+            (list): The list of log lines with time stamps after the given
+                date and time.
```
bfahr: Subsequent lines for `Returns:` should be indented at the same level as the first line.

insights/core/__init__.py
PaulWay: Ah, yes, had forgotten that bit of the documentation.

insights/core/__init__.py
PaulWay: Hmmmm.  I suppose I wanted to require the subclass to provide the 'correct' method and I wanted to prevent the rule writer thinking about overriding the timestamp format.

But ... I can see that that's probably unnecessary.  And if it saves a bunch of copy-and-paste code, then so much the better.  In the cases where the subclass needs to change the timestamp format, then it can use super(ThisLogClass, self).get_after() ...

Let me try putting that together and see what I get.

bfahr: Since you already provide a default time format, I think that you could just implement in child parsers where its different.  I think you might want to move the args around so that the least changed is last.  I don't think that rules would have to deal with the format.

insights/core/__init__.py
PaulWay: Dunno how that got left out - will fix.

insights/core/__init__.py
PaulWay: That _is_ a good idea!  Thanks, bfahr!

insights/parsers/ceilometer_log.py
PaulWay: Ah, OK, will fix.

insights/parsers/ceilometer_log.py
PaulWay: Yep - that was already in the documentation for that parser, I didn't include it :-)

PaulWay:
I realised in looking at the Messages parser that it's `get()` method already parses each line into a dictionary with a 'timestamp' key.  I like that idea, and it would mean that parsers that implemented a similar method or a `parse_line()` method that similarly turned a line string into a dictionary of fields and values, then that would save the convoluted process of constructing a regular expression and searching for it in each line.

So the idea would be that if the first entry in the 'lines' list is a dict, then the 'timestamp' value is thrown directly to `strptime`.  If it doesn't have a `timestamp` key then we throw a ParseException (or maybe a ValueError?)  Is this going too far?

insights/parsers/foreman_log.py
```diff
@@ -16,7 +16,21 @@
...
+
+        Parameters:
+            timestamp(datetime.datetime): log lines after this time are
+                returned.
+            lines(list): the list of log lines to search (e.g. from a get).
+                If not supplied, all available lines are searched.
+
+        Returns:
+            (list): The list of log lines with time stamps after the given
+                date and time.
```
bfahr: @PaulWay you need to unindent the second line on the `Returns:` so that it will format correctly.

insights/parsers/messages.py
```diff
@@ -86,3 +86,23 @@ def get(self, s):
...
+
+        Parameters:
+            timestamp(datetime.datetime): log lines after this time are
+                returned.
+            lines(list): the list of log lines to search (e.g. from a get).
+                If not supplied, all available lines are searched.
+
+        Returns:
+            (list): The list of log lines with time stamps after the given
+                date and time.
```
bfahr: You need to unindent the second line on the `Returns:` so that it will format correctly.

insights/parsers/messages.py
```diff
@@ -86,3 +86,23 @@ def get(self, s):
...
+
+        *NOTE*: Because /var/log/messages timestamps by default have no year,
+        you must supply the year in your timestamp as 1900.  This may also
+        cause confusion around the end of the year.
```
bfahr: Should there be some sort of test case for this?

PaulWay: Do you mean 'test that `get_after()` handles timestamps around 1900 correctly'?  Or '`get_after()` should test that the tested-for timestamp has a year reasonably close to the log's timestamp'?  I'm not sure how to do the former, and I don't think we should do the latter, so obviously I don't understand what you mean here :-)

insights/parsers/messages.py
PaulWay: Yeah - I keep on thinking Sphinx needs that when it doesn't.  Will fix.

In general what I should have done is to get the whole patch completed and okayed, then submit the PR for the 1.x code.  Much easier that way.

insights/parsers/messages.py
bfahr: Sorry, what I really should have said is how should the messages parser handle the case of where log messages go from Dec 31 to Jan 1 if we are hard coding the year?  Should the parser do something to handle that?

PaulWay: I'm really not sure how.

The main problem is that if we see a skip back in time - let's say over 11 months - it indicates that all the logs read up until now are from the previous year.  But in the current implementation we've yielded those logs, so we can't go back and re-evaluate those logs because we now know that they're in the previous year.

I'm trying to avoid just focusing on the specific use-case of finding all the logs after the last collection run.  There are other dates, such as the date the configuration file was changed, that a rule could use in get_after() that may be longer than 24 hours since the last update, and may be earlier than that roll-over from December to January.

One possible approach might be to use the year given in the timestamp argument itself as the assumed year for logs without a year.  When the date is read from the log line, we check whether the log date is greater than 11 months ahead of the sought timestamp.  If it is, this indicates that the log line is from December and the sought date is in January, so we shift that log date back a year to put it in the previous year relative to the sought date.  That's done for every log line, so even if for some reason (e.g. Apache log lines) the logs go backward in time, log dates are always corrected as above.

I'm not super keen on adding this kind of specialised processing into what is already a large and complicated function, but I'll give it a go and see how it turns out.

PaulWay: I think I worked out the logic there so it's relatively simple - let me know what you think @bfahr

bfahr: I assume that you do it this way because you can't assume the log is in chronological order?  If it was in order then you could just check the month value for a end-of-year rollover.

PaulWay: Yeah - some log files, like Apache's access log, get the date stamp from when the request started but are written to the log when the request finishes; long requests can make the log look like it went backward in time.  The other problem with just checking an end-of-year rollover is that a search for dates in January 2017 will give logs written in December the year of 2017 - which makes them appear as if they're in the log range when they are most likely to be from 2016.  Because this method is a generator we can't go back and retroactively re-parse the logs now that we know that they probably started in December 2016.

insights/core/__init__.py
```diff
@@ -426,6 +429,159 @@ def _scan(self):
...
+        by syslog.
+
+        No attempt is made to order the lines by date.  This also means that
+        lines can be out of order, as each timestamp is evaluated individually.
+
+        Params:
+            timestamp(datetime.datetime): lines before this time are ignored.
+            lines(list): an optional list of lines from this parser to search.
+            time_format(str): the ``strptime()`` string of the format of the
+            dates in this log.
```
bfahr: @PaulWay your params are not rendered correctly in the output.  The following changes appear to fix it
```
diff --git a/insights/core/__init__.py b/insights/core/__init__.py
index 976e145..47c00cf 100644
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -475,20 +475,20 @@ class LogFileOutput(Parser):
         No attempt is made to order the lines by date.  This also means that
         lines can be out of order, as each timestamp is evaluated individually.
 
-        Params:
+        Parameters:
             timestamp(datetime.datetime): lines before this time are ignored.
             lines(list): an optional list of lines from this parser to search.
             time_format(str): the ``strptime()`` string of the format of the
-            dates in this log.
+                dates in this log.
 
         Yields:
-            (string): a list of lines with timestamps after this date.
+            string: a list of lines with timestamps after this date.
 
         Raises:
             ParseException: If the format conversion string contains a
-            format that we don't recognise.  In particular, no attempt is
-            made to recognise or parse the time zone or other obscure
-            values like day of year or week of year.
+                format that we don't recognise.  In particular, no attempt is
+                made to recognise or parse the time zone or other obscure
+                values like day of year or week of year.
         """
         if lines is None:
             lines = self.lines
```

insights/core/__init__.py
```diff
@@ -426,6 +429,161 @@ def _scan(self):
...
+
+        No attempt is made to order the lines by date.  This also means that
+        lines can be out of order, as each timestamp is evaluated individually.
+
+        Parameters:
+
+            timestamp(datetime.datetime): lines before this time are ignored.
+            lines(list): an optional list of lines from this parser to search.
+            time_format(str): the ``strptime()`` string of the format of the
+            dates in this log.
```
bfahr: @PaulWay you need to indent this line too.

insights/parsers/foreman_log.py
bfahr: This line needs to be unindented.

insights/parsers/messages.py
bfahr: Still need to fix this one too.  If you look at the generated docs you can see the problem.

PaulWay:
How's all that documentation looking now @bfahr ?

insights/core/__init__.py
bfahr: @PaulWay you missed this one.

PaulWay:
OK, this time for sure!

insights/parsers/ceph_osd_log.py
```diff
@@ -27,4 +27,18 @@ class CephOsdLog(LogFileOutput):
...
-    pass
+    def get_after(self, timestamp, lines=None):
+        """
+        Get a list of lines after the given time stamp.
+
+        Parameters:
+            timestamp(datetime.datetime): log lines after this time are
+            returned.
+            lines(list): the list of log lines to search (e.g. from a get).
+            If not supplied, all available lines are searched.
```
bfahr: @PaulWay subsequent lines should be indented here.  You can see this in the generated documentation.

insights/parsers/foreman_log.py
bfahr: The `Returns` is still broken.

insights/parsers/messages.py
bfahr: This one still needs to be fixed.

insights/parsers/ceph_osd_log.py
PaulWay: OK, fixed.

Issue #90, opened 2017-06-14
Uname not properly parsed for RHEL 6.3 sosreports


miclark:
For rhel 6.3 (sos-2.2) sosreport, the files of command output under sos_commands folders are all end with newline. Below two examples for you to compare(you can check `hostname`, `blkid`, `uname_-a`):

Example rhel 6.3 sosreport : https://api.access.redhat.com/rs/cases/01857736/attachments/bd1496bd-317f-43af-ad9f-fb802751667a
Example rhel 6.8 sosreport https://api.access.redhat.com/rs/cases/01857704/attachments/39b557e2-22ee-41bd-876e-a961d2985d65

Some mappers rise error for this.

Example `uname.py` error:

```
ERROR:eval:Mapper failed
Traceback (most recent call last):
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/core/evaluators.py", line 197, in run_mappers
    self.add_result(self._execute_mapper(plugin, context),
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/core/evaluators.py", line 76, in _execute_mapper
    return mapper(context)
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/mappers/uname.py", line 195, in __init__
    super(Uname, self).__init__(context)
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/core/__init__.py", line 112, in __init__
    self.parse_content(context.content)
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/mappers/uname.py", line 223, in parse_content
    raise UnameError("Uname string appears invalid", uname_line)
UnameError: Uname string appears invalid:''
```

Root cause is the `uname.py` does not take account in the newline at the end of file.

The error raised by below code, you will see the code `content[-1` to get the last line.  It is empty line.

```
uname_line = content[-1]  # read the last line instead of the first
        uname_parts = uname_line.split(' ')
        if len(uname_parts) < 3:
            ver_rel_match = re.match("[0-9](\.[0-9]+){2}-[0-9]+", uname_parts[0])
            if not ver_rel_match:
                raise UnameError("Uname string appears invalid", uname_line)
            data['kernel'] = uname_parts[0]
```

For rhel 6.4 (`sos-2.2-38.el6.noarch`) and above, the output of commands do not include the newline at the end of the file.

According above, the insights-cli does not work well for rhel 6.3 and previous version.

wushiqinlou:
[...]

PR #123, opened 2017-06-21
Enhance parser uname so that this parser can handle the case that the…

wushiqinlou:
… last line is blank


wushiqinlou:
This PR is for https://github.com/RedHatInsights/insights-core/issues/90

insights/parsers/uname.py
```diff
@@ -209,7 +210,11 @@ def parse_content(self, content):
             raise UnameError('Empty uname line', '')
 
         data = {}
-        uname_line = content[-1]  # read the last line instead of the first
+        uname_line = None
+        for line in content:
```
xiangce: @wushiqinlou - for tarball collected `foreman-debug`, the uname's output looks like

~~~
COMMAND> uname -a

Linux slolmeprdrhsv01.ipa.lme.co.uk 3.10.0-327.10.1.el7.x86_64 #1 SMP Sat Jan 23 04:54:55 EST 2016 x86_64 x86_64 x86_64 GNU/Linux
~~~

So, you'd better get the line that starts with "Linux" as the `uname_line`.

BTW, you can implement it in a list comprehension.

PR #125, opened 2017-06-21
Add new shared parser "teamdctl_state_dump.py"

xiaoyu74:


insights/parsers/teamdctl_state_dump.py
```diff
@@ -0,0 +1,69 @@
...
+    ...             "ifname": "team0"
+    ...         }
+    ...     }
+    ... }
+    ... '''.strip()
+
+    >>> from insights.parsers.teamdctl_state_dump import TeamdctlStateDump
+    >>> from insights.tests import context_wrap
+    >>> shared = {TeamdctlStateDump: TeamdctlStateDump(context_wrap(teamdctl_state_dump_content))}
+    >>> result = TeamdctlStateDump(context_wrap(teamdctl_state_dump_content)).data
```
bfahr: @xiaoyu74 since the parser is based on `LegacyItemAccess` you can skip this step and access the dictionary directly.  If you want to keep this example then you should document the `data` class attribute.

insights/parsers/teamdctl_state_dump.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def parse_content(self, content):
+        """
+        Parse the output of the ``teamdctl {team interface} state dump``.
+        """
+        self.data = json.loads(''.join(content))
+
+
+@parser("teamdctl_state_dump")
+class TeamdctlStateDump(TeamdJsonParsing):
```
bfahr: Is there a need for this extra class?  If you are not planning on using the base `TeamdJsonParsing` class for other parsers then it makes more sense to just have one class here.

insights/parsers/tests/test_teamdctl_state_dump.py
```diff
@@ -0,0 +1,59 @@
...
+            "ifname": "team0"
+        }
+    }
+}
+""".strip()
+
+
+class TestTeamdctlStateDump():
+    def test_teamdctl_state_dump(self):
+        result = TeamdctlStateDump(context_wrap(TEAMDCTL_STATE_DUMP_INFO))
```
bfahr: You could implement this as test functions instead of adding the complexity of a class, unless you need the class for some other reason.

insights/parsers/teamdctl_state_dump.py
```diff
@@ -0,0 +1,64 @@
...
+@parser("teamdctl_state_dump")
+class TeamdctlStateDump(Parser, LegacyItemAccess):
+    """
+    Class to parse the output of ``teamdctl {team interface} state dump``.
+    """
+
+    def parse_content(self, content):
+        self.data = json.loads(''.join(content))
+
+    @property
```
xiangce: Better add one more decorator `@defaults()` to avoid any other implicit errors (e.g. KeyError)

insights/parsers/teamdctl_state_dump.py
```diff
@@ -0,0 +1,64 @@
...
+    ...     }
+    ... }
+    ... '''.strip()
+
+    >>> from insights.parsers.teamdctl_state_dump import TeamdctlStateDump
+    >>> from insights.tests import context_wrap
+    >>> shared = {TeamdctlStateDump: TeamdctlStateDump(context_wrap(teamdctl_state_dump_content))}
+    >>> result = shared[TeamdctlStateDump]
+    >>> result['runner']['active_port']
+    eno1
```
xiangce: you might forget the quotation marks

PR #131, opened 2017-06-22
Parser for systemctl status service_name

vishwanathjadhav:
Added Parser to parse  **_systemctl status service_name_**  command output.
this parser will show status, vendor preset, service name, PID, of ovs and neutron-opflex-agent, the parser can be enhanced for other services as well


insights/parsers/systemctl_status.py
```diff
@@ -0,0 +1,86 @@
...
+            'Main PID' : '673'}
+        """
+
+        self.data = {}
+        for line in content:
+            if "Main PID" in line or "Loaded" in line or "Active" in line:
+                line = line.replace('   ', '')
+                for spch in [';', '(', ')']:
+                    line = line.replace(spch, '\n')
+                for st in line.split('\n'):
```
bfahr: You might consider using `re.split` replacing the above four lines with something like:
```
for st in re.split('[;\(\)\s]*', line):
```
The above example may not produce exactly what you need.

insights/parsers/tests/test_systemctl_status_opflex.py
```diff
@@ -0,0 +1,28 @@
...
+neutron-opflex-agent - Open vSwitch Internal Unit
+   Loaded: loaded (/usr/lib/systemd/system/neutron-opflex-agent; static; vendor preset: disabled)
+   Active: inactive (dead) since Fri 2017-04-14 00:23:48 UTC; 3min 17s ago
+   Main PID: 940 (code=exited, status=0/SUCCESS)
+""".strip()
+
+
+def test_systemctl_status_opflex():
+    context = context_wrap(SYSTEMCTL_STATUS_OPFLEX_ACTIVE)
+    result = SystemctlStatusOpflex(context).data
```
xiangce: Since "SystemctlStatus" is also based on "LegacyItemAccess", we can access its content directly without via ".data"

insights/parsers/tests/test_systemctl_status_ovs.py
```diff
@@ -0,0 +1,49 @@
...
+    context = context_wrap(SYSTEMCTL_STATUS_OVS_INACTIVE)
+    result = SystemctlStatusOVS(context).data
+    assert result['service'] == "openvswitch-nonetwork.service"
+
+    context = context_wrap(SYSTEMCTL_STATUS_OVS_ACTIVE)
+    result = SystemctlStatusOVS(context).data
+    assert result['Active'] == "active"
+
+    context = context_wrap(SYSTEMCTL_STATUS_OVS_EMPTY)
+    result = SystemctlStatusOVS(context).data
```
xiangce: You might forget an `assert` check on this result

xiangce:
@vishwanathjadhav - how is this PR going?

vishwanathjadhav:
I will delete this pr as parser is discarded

PR #145, opened 2017-06-25
Neutron plugin documentation for 1.x branch

PaulWay:


docs/shared_parsers_catalog/neutron_plugin.rst
```diff
@@ -0,0 +1,9 @@
+Neutron Plugin Configuration
+============================
+
+file ``/etc/neutron/plugin.ini``
```
xiangce: This is indeed a new method to write the `.rst` file, however, it will override the `title` and `header` (line 5 & 6) in file `neutron_plugin.py`. Do you really want to hide the content (line 2~6 of file "neutron_plugin.py") in the documentation?

bfahr: I think specifying the classes in the `.rst` file is more useful when there are multiple mappers in module.  It doesn't seem to provide much benefit here.

xiangce: @PaulWay - just like @bfahr commented in #147, would you like to change it back to the current documentation standard?

PaulWay: OK, following from the docs discussion elsewhere let's get this sorted out before we start trying things out randomly :-) I'll go back to the previous docs system.

xiangce:
Will merge it after it's re-submitted to the master branch, see #151.

Issue #132, opened 2017-06-22
Use consistent heading and class style for parser catalogue


PaulWay:
The proposal is to eventually replace the current .rst file for parsers with one of the form:

```
{{ module title }}
==================

{{ for each class in the module:}}
{{ file or command }}
---------------------
.. autoclass:: insights.parsers.{{module}}.{{parser class name}}
   :members:
   :show-inheritance:
{{ endfor }}
```

This provides a more structured way to organise the information currently presented in an ad-hoc fashion within the parsers catalogue.

PaulWay:
Something a bit more similar to the current format would be:

```
{{ for each class in the module:}}
{{parser class name} - {{ file or command }}
------------------------------------------
.. autoclass:: insights.parsers.{{module}}.{{parser class name}}
   :members:
   :show-inheritance:
{{ endfor }}

```

Would that be OK?

PaulWay:
@bfahr in looking at this - does this mean we end up moving the file documentation out of the parser's main docstring and into the .rst file?  That'd be a step backward to me - I'd like to keep as much documentation as possible in the parser module. I'll continue to experiment but any advice you have will be invaluable here :-)

PaulWay:
The key thing I want to aim for in this is that every parser module lists every parser class it provides, in a format that makes it easy for a rule writer to find the parser they need based on the file they want to read.

I'd also like consistency - i.e. that every module listing looks the same, regardless of whether it has one or six parsers in it.

I'm prepared for the .rst file to require more management as a result, but OTOH I would really like to keep the automatic functionality and single place for documentation changes that we currently have.

bfahr:
@PaulWay I agree that we don't want have have the docs in both the code and the .rst file.  The .rst file should just be the map for building the doc, and not the doc itself.  Otherwise we've got a disconnect and have to maintain it in two places.

Its a bit hard to visualize, I think it might be easier to discuss if it was mocked up.

miclark:
Closing due to age.  We can revisit with a larger conversation about style of the catalog at a future date.

PR #146, opened 2017-06-25
Neutron server log documentation for 1.x branch

PaulWay:


docs/shared_parsers_catalog/neutron_server_log.rst
```diff
@@ -0,0 +1,9 @@
+Neutron Server Log
+==================
+
+NeutronServerLog - ``/var/log/neutron/server.log``
+--------------------------------------------------
+.. autoclass:: insights.parsers.neutron_server_log.NeutronServerLog
+   :members:
+   :show-inheritance:
+
```
bfahr: @PaulWay I don't believe there is really much benefit in adding this complexity when there is only a single mapper in the module.

insights/parsers/neutron_server_log.py
```diff
@@ -3,14 +3,33 @@
...
     def get(self, keywords):
         """
-        Returns all lines that contain all keywords. keywords can be str or str list.
-        -- Sample --
-        keywords is a list. Example: ["WARNING","Could not find token", "404"]
-        ------------
-        2016-09-13 06:06:46.131 30586 WARNING keystonemiddleware.auth_token [-] Identity response: {"error":
-        {"message": "Could not find token: bc029dbe33f84fbcb67ef7d592458e60", "code": 404, "title": "Not Found"}}
+        Returns all lines that contain all keywords, supplied either as a
+        single string or a list of strings.
```
bfahr: Do you think it would be useful to use the `Parameters:` and `Returns:` keywords here to define `keywords` and the return value?

docs/shared_parsers_catalog/neutron_server_log.rst
PaulWay: I sort of agree, especially since it means that new classes don't get automatically added to the documentaiton.  But I really want the consistency of presentation this gives us.  I'm not sure what a good solution looks like - we should talk about this in the documentation issue https://github.com/RedHatInsights/insights-core/issues/132 .

insights/parsers/neutron_server_log.py
PaulWay: Yes, that'd be much better - sorry!  Thanks Bob.

docs/shared_parsers_catalog/neutron_server_log.rst
PaulWay: OK, following from the docs discussion elsewhere let's get this sorted out before we start trying things out randomly :-)  I'll go back to the previous docs system.

xiangce:
Will merge it after it's re-submitted to the master branch, see #152.

PR #147, opened 2017-06-25
Kerberos KDC log parser for 1.x branch

PaulWay:
Includes tests and documentation

* `get()` method parses logs into fields
* `get_after()` method


docs/shared_parsers_catalog/krb5kdc_log.rst
```diff
@@ -0,0 +1,9 @@
+Kerberos KDC Logs
+=================
+
+KerberosKDCLog - ``/var/log/krb5kdc.log``
+-----------------------------------------
+.. autoclass:: insights.parsers.krb5kdc_log.KerberosKDCLog
+   :members:
+   :show-inheritance:
+
```
bfahr: @PaulWay until we decide on how we want to revamp the module documentation, this should be changed back to the current documentation standard just listing the module.

PaulWay:
@bfahr Yeah, you're right, I was too keen on that update :-)  Let's get the documentation looking right and then deploy separately.

xiangce:
Since Paul has already updated the documentation as per @bfahr 's request, I'll go ahead and merge this PR.

PR #157, opened 2017-06-27
update the parser to define an API to return the team device name

xiaoyu74:
define an API to return team device name.


xiaoyu74:
@PaulWay - If the command `teamdctl {team iface} state dump` is executed in the teaming environment, the JSON format output is fixed and the keys such as "team_device", "info" and "ifname" will be definitely existed in the collected data.  Furthermore, since I have used the decorator `@default()`, it will return `None` if the specified key does not exist in the case.

PaulWay:
@xiaoyu74 - OK - does the `@default()` decorator handle a KeyError exception?  In general I'd like to see this scenario tested to make sure we know how it handles bad input.

xiaoyu74:
close the PR due to the messy commits.

PR #159, opened 2017-06-28
making logging_file overridable

jhjaggars:
also adding an entry point to call the client

### DIFF
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -15,8 +15,8 @@ handler = None
 class InsightsClientApi(object):
 
     def __init__(self,
-            options=None,
-            config=None):
+                 options=None,
+                 config=None):
 
         InsightsClient.config, InsightsClient.options = self.parse_options()
         if options:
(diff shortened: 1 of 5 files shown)


insights/client/client_config.py
```diff
@@ -191,6 +192,8 @@ def set_up_options(parser):
                      action="store",
                      dest="run_specific_specs",
                      default=None)
+    group.add_option('--logging-file',
```
brantleyr: This is missing the rest of the option definition.

setup.py
```diff
@@ -13,7 +13,8 @@
     'console_scripts': [
         'insights-run = insights.core:main',
         'gen_api = insights.tools.generate_api_config:main',
-        'insights-perf = insights.tools.perf:main'
+        'insights-perf = insights.tools.perf:main',
+        'client = insights.client:run'
```
brantleyr: Does this mean we can now invoke the core and client with args such as 'insights-core client --args1 --args2 --etc'?

insights/client/__init__.py
```diff
@@ -237,3 +231,8 @@ def delete_archive(self, path):
             returns (bool): successful archive deletion
         """
         return client.delete_archive(path)
+
+
+def run():
+    c = InsightsClientApi()
+    c.run()
```
brantleyr: C:\DOS
C:\DOS\RUN
RUN\DOS\RUN

brantleyr: I don't think we want to instantiate and simply 'run' the Client API. I think we want to have some options/args logic here, like the old client does in 'handle_startup.' Thoughts?

PR #160, opened 2017-06-28
clean up client utilities and add test cases

SteveHNH:


insights/tests/client/test_utilities.py
```diff
@@ -0,0 +1,46 @@
...
+        with open(tf.name, 'r') as f:
+            written_file = f.read()
+
+        self.assertEquals(machine_id, written_file)
+
+    def test_get_time(self):
+        import re
+        time_regex = re.match('\d{4}-\d{2}-\d{2}\D\d{2}:\d{2}:\d{2}\.\d+',
+                              util.get_time())
+        if time_regex.group(0) is not None:
```
PaulWay: This can be better put as:
```
assert time_regex.group(0)
```
(or `assert time_regex.group(0) is not None` if you want to just follow the logic there)

bfahr:
@SteveHNH when I try to run unit tests, I get the following error from `test_client.py`:
```
insights/tests/test_client.py:7: in <module>
    class TestClient(unittest.TestCase):
insights/tests/test_client.py:9: in TestClient
    client = InsightsClientApi()
insights/client/__init__.py:31: in __init__
    client.set_up_logging()
insights/client/client.py:44: in set_up_logging
    os.makedirs(log_dir, 0700)
lib64/python2.7/os.py:157: in makedirs
    mkdir(name, mode)
E   OSError: [Errno 13] Permission denied: '/var/log/insights-client'
```
Is there a way to run the tests without access to protected dirs and files?

SteveHNH:
@bfahr That got fixed with PR #159. Jesse added the ability to override the log location and updated the client test to reflect that change.

PR #161, opened 2017-06-29
Add 'macpools' field into RHEV metadata

JoySnow:
This 'macpools' API exists in RHEV3.6 and RHV4.x,  RHEV3.5 don't have such thing. 
So the insights-client needs to handle this collection error condition for RHEV3.5.


PaulWay:
Can you please also add complete tests for this parser?

PaulWay:
Just looking at the Multinode parser, I'd really like to see this converted to a class-based mapper.

PaulWay:
It's also worth noting that it's perfectly possible to get the macpools data from the multinode object by using the data property dictionary, e.g. `obj.data['macpools']`.

JoySnow:
@PaulWay I will convert the _metadata function to class and add a test file to it. Thanks for the suggestions.

bfahr:
I'm not familiar with how this code works but the changes look OK.  It would be really helpful to have documentation with this code so that developers can understand if they need it, and how to use it.

JoySnow:
@bfahr Sure, I will add documentations to it later.

JoySnow:
According to https://github.com/RedHatInsights/insights-core/issues/221#issuecomment-315093049, this parser had also been revised in master branch.   
@PaulWay I will not convert the _metadata function to class or add a test file to it either.

xiangce:
It's still necessary to add the `macpools` to the master brach, to avoid losing it when meget the latest code from master to 1.x.

JoySnow:
@xiangce Yes, and I will add and only add this `macpools` filed only to master.

PR #168, opened 2017-07-03
Enhance2 parser openshift_get 1x

wushiqinlou:


wushiqinlou:
Import class `LegacyItemAccess` so that parser can use function `get` to get data instead of using `.data` directly.

insights/parsers/openshift_get.py
```diff
@@ -22,15 +22,15 @@
     True
 """
 
-from .. import YAMLParser, parser
+from .. import YAMLParser, parser, LegacyItemAccess
```
xiangce: @wushiqinlou - better add the `LegacyItemAccess` to `YAMLParser` directly, by doing this, we can avoid forgetting to add the `LegacyItemAccess` to the newly created class(YAMLParser).

And I do suggest you remove the `LegacyItemAcess` from "CobblerSettings" which already added it in this PR.

xiangce:
@wushiqinlou -  Better explain your updating in `git commit` a little more detailed. I updated it this time as per your modification points.

PR #181, opened 2017-07-06
Add pre-command for shared-mapper dumpe2fs-h

zhangyi733:


insights/parsers/dumpe2fs_h.py
```diff
@@ -53,13 +53,13 @@
 
 """
```
xiangce: Since "LegacyItemAccess" is added, better update the example in documentation (docstring) and the unit test: remove the ".data" and access the items with index "[key]" or get('key')

xiangce:
@zhangyi733 - Approved, will merge it after you re-submit it to the master branch.

PR #183, opened 2017-07-06
Add pre-command for dumpe2fs-h and LegacyItemAccess to parser DumpE2fs

zhangyi733:


xiangce:
@zhangyi733 - To keep the two branches synchronized and for reviewing convenience, you'd better use `cherry-pick` or `git format-patch and git am` to shift the modifications from 1.x to master.

PR #184, opened 2017-07-06
Change the wildcard * to pre_command for sysctl.conf_initramfs spec

xiangce:
- The * wildcard doesn't work in insights-client env
- And modify some indent in specs.py


xiangce:
@mhuth - I didn't add the command for RHEL6, because the only 1 rule that used this spec is only for RHEL7.

PR #186, opened 2017-07-06
Log file output get after improvements for 1.x branch

PaulWay:
* To save considerable copy-and-paste boilerplate, the time_format is now a property of the object.  Subclasses should override this - it is not advised for the end user (i.e. a rule writer) to change the time_format property.
* Sometimes different log lines in the same file have different timestamp formats.  Sometimes that's because the log format changes between versions (and our parser has to read the different versions), and sometimes that's because e.g. error messages get a different timestamp format from regular messages.  In order to cope with this, we now accept a list of `strptime()` formats as well as a string.  The formats can also be given as the values in a dictionary, which allows the keys to be names indicating the different versions or purposes of the formats.
  If given a list of time formats, the regular expression turns into an alternation between the different format patterns and `strptime()` is tried on each one in turn to find the correct pattern.
* Some patterns were not tested and these were found to have errors - in particular %p.
* Some patterns needed to be extended to cope with different variations that `strptime()` accepts.
* The old '_get_after' name was used in some places - this has been changed.


insights/core/__init__.py
```diff
@@ -585,15 +595,50 @@ def replacer(match):
...
 
         # Please do not attempt to be tricky and put a regular expression
         # inside your time format, as we are going to also use it in
         # strptime too and that may not work out so well.
-        time_re = re.compile('(' + timefmt_re.sub(replacer, time_format) + ')')
+
+        # Check time_format - must be string or list.  Set the 'logs_have_year'
+        # flag and timestamp parser function appropriately.
+        # Grab values of dict as a list first
+        if type(time_format) == dict:
```
xiangce: Except this, all others look good to me.

PaulWay: Fair enough, although I really wonder why anyone would be subclassing a dict for this argument :-)  `isinstance()` isn't perfect either, but it's better than `type()`.

xiangce:
@PaulWay - This PR is ready to merge after you re-submit it to the master branch.

PR #187, opened 2017-07-06
Catalina_out parser add get_after for 1.x branch

PaulWay:
Updating the parser to support the `get_after()` method.

This requires https://github.com/RedHatInsights/insights-core/pull/186 to fix the %p pattern.

### DIFF
--- a/insights/parsers/nova_api_log.py
+++ b/insights/parsers/nova_api_log.py
@@ -1,17 +1,26 @@
+"""
+NovaApiLog - file ``/var/log/nova/nova-api.log``
+================================================
+"""
 from .. import LogFileOutput, parser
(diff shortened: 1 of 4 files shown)


insights/parsers/nova_api_log.py
```diff
@@ -1,17 +1,26 @@
+"""
```
xiangce: The "nova_api_log" is included in "https://github.com/RedHatInsights/insights-core/pull/188", better remove it from this PR.

PaulWay:
OK, something's definitely wrong with this branch, it should only have three commits and should not touch nova_api_log.  Let me close this and make a new PR.

PR #188, opened 2017-07-06
Nova_api_log parser add get_after for 1.x branch

PaulWay:
* Add documentation for NovaApiLog
* Add `get_after()` method and tests


bfahr:
@xiangce hold off on this one until we discuss with @PaulWay regarding whether this PR is necessary.

PR #189, opened 2017-07-06
Ceilometer_log parser add get_after for 1.x branch

PaulWay:


insights/parsers/ceilometer_log.py
```diff
@@ -45,4 +59,18 @@ class CeilometerCollectorLog(LogFileOutput):
...
+            timestamp(datetime.datetime): log lines after this time are
+                returned.
+            lines(list): the list of log lines to search (e.g. from a get).
+                If not supplied, all available lines are searched.
+
+        Yields:
+            Log lines with time stamps after the given date and time, in the
+            same format they were supplied.
+        """
+        return super(CeilometerCollectorLog, self).get_after(timestamp, lines, '%Y-%m-%d %H:%M:%S')
```
bfahr: @PaulWay this is using the same timestamp format as the base class so why implement this here?

PR #194, opened 2017-07-07
Secure log parser add get_after for 1.x branch

PaulWay:
* Added documentation
* Added get_after() method and tests
* Tests now cover 100% of code


insights/parsers/secure.py
```diff
@@ -1,41 +1,69 @@
+"""
+Secure -  file ``/var/log/secure``
+==================================
+"""
 from .. import LogFileOutput, parser
 
 
 @parser('secure')
 class Secure(LogFileOutput):
+    """Class for parsing the ``rhn_taskomatic_daemon.log`` file.
```
bfahr: @PaulWay is this log file reference correct?

PR #198, opened 2017-07-07
Dmesg_log parser add get_after for 1.x branch

PaulWay:


insights/parsers/dmesg.py
```diff
@@ -30,27 +30,75 @@
...
     Class for reading output of ``dmesg`` using the LogFileOutput parser class.
     """
+    _line_re = re.compile(r'^(?:\[\s+(?P<timestamp>\d+\.\d+)\]\s+)?(?P<message>.*)$')
+
     def has_startswith(self, prefix):
         """
         Parameters:
-            prefix (str): The prefix of the line to look for.
+            prefix (str): The prefix of the line to look for.  Ignores any
+            timestamp before the message body.
```
bfahr: @PaulWay you need to indent this continuation line.

insights/parsers/dmesg.py
```diff
@@ -30,27 +30,75 @@
...
         """
-        return any(line.startswith(prefix) for line in self.lines)
+        def yield_messages(lines):
+            for line in lines:
+                match = self._line_re.search(line)
+                if match:
+                    yield match.group('message')
+                else:
+                    yield line
+        return any(msg.startswith(prefix) for msg in yield_messages(self.lines))
```
bfahr: Do you gain anything using the yield here, since you can't return until a line triggers the test?  Could you simplify this?

PaulWay: OK, you made me look at this and I figured out a better way.

1. The regex always matches, since the message subgroup matches .* and everything else is optional.
2. Therefore we can always check that subgroup for the prefix.
3. Therefore we can remove the generator entirely and expand the test into the rather longwinded `self._line_re.search(line).group('message').startswith(prefix)`.

I've also tested for negative matches - i.e. that things that aren't in the log or aren't prefixes should not generate a match in `has_startswith()` - and fixed the documentation glitch.

xiangce:
[Dmesg_log parser add get_after for 1.x branch #198](https://trello.com/c/aGh4hx4g/4-dmesglog-parser-add-getafter-for-1x-branch-198)

PR #199, opened 2017-07-07
Cinder_log parser add get_after for 1.x branch

PaulWay:


PaulWay:
Thanks @bfahr - I grepped for 'cinder' in the shared_parsers_catalog and didn't think that the one listed was for a different parser :-/  Fixed now.

PR #213, opened 2017-07-09
Add new shared parser irqbalance_conf.py

lonicerae:


insights/parsers/irqbalance_conf.py
```diff
@@ -0,0 +1,46 @@
+"""
+irqbalance_conf - Contents of ``/etc/sysconfig/irqbalance``
+=============================================================
```
xiangce: Normally we use `File` instead of `Contents of` here.
Better keep the "=" in the same length with the title.

insights/parsers/tests/test_irqbalance_conf.py
```diff
@@ -0,0 +1,41 @@
...
+# IRQBALANCE_ARGS
+# append any args here to the irqbalance daemon as documented in the man page
+#
+IRQBALANCE_ARGS="-d"
+""".strip()
+
+
+def test_irqbalance_conf():
+    ret = irqbalance_conf.SysconfigIrqbalance(context_wrap(IRQBALANCE_SYSCONF_TEST))
+    assert keys_in(["IRQBALANCE_BANNED_CPUS", "IRQBALANCE_ARGS"], ret.data)
```
xiangce: Since its a mixin class with LegacyItemAccess, better remove the ".data" from the test.

PR #214, opened 2017-07-09
Vdsm_log parser gets full documentation and get_after() method for 1.x branch

PaulWay:
* Adding full documentation
* Adding parse() method
* Adding get_after() method
* Full testing


insights/parsers/vdsm_log.py
```diff
@@ -1,14 +1,94 @@
...
+            If the line is too short, for some reason, then as many fields as
+            possible are pulled from the line.
+        """
+
+        # funcname is in brackets before message, so parse those separately
+        fieldnames = ('thread', 'level', 'timestamp', 'module', 'line', 'logname')
+        for line in lines:
+            parts = line.split('::', 6)
+            fields = {'raw_line': line}
+            fields.update({k: v for k, v in zip(fieldnames, parts)})
```
bfahr: @PaulWay based on your comment on line 79 above, is there any chance that there would be missing fields that could cause this `zip` to fail, or to mislabel fields?

PaulWay: Well, we can't know for sure why a log line has less than seven fields - the usual reason would be because it got truncated.  `zip()` will only zip up to the shorter array's last element, so if we get fewer than six fields the fields are simply named in the order they appear.  As you can see from the tests this means we can capture the timestamp even if the line is partly truncated.

Only if we got seven fields - with the last being the function name and message - do we capture those two.  We don't handle the situation where the line is truncated within the function name - i.e. that the seventh field does not contain a space - because I think that's extremely unlikely.

PR #216, opened 2017-07-10
Catalina_out parser add get_after() method for 1.x branch - This Time For Sure

PaulWay:
This time we only modify the catalina_out parser.

Requires #186 to work correctly.


PaulWay:
@wcmitchell Absolutely right - that was fixed in #186 which is now merged.  Should be good to go now :-)

PR #218, opened 2017-07-10
add schedule tests and relocate test_client

SteveHNH:
Had to change a couple scheduler functions to make them testable.


insights/client/schedule.py
```diff
@@ -13,27 +13,28 @@
...
+    def already_linked(self, cronfile):
         """
         Determine if we are already scheduled
         """
-        if os.path.isfile(CRON_WEEKLY + APP_NAME):
+        if os.path.isfile(cronfile):
             logger.debug('Found cron.weekly')
             return True
-        elif os.path.isfile(CRON_DAILY + APP_NAME):
+        elif os.path.isfile(cronfile):
```
brantleyr: You will no longer need this second 'elif' check as it does the same thing as the first. I would remove this one and change the verbiage to 'Found %s' or something along those lines.

insights/client/schedule.py
```diff
@@ -13,27 +13,28 @@
...
             logger.debug('Found cron.daily')
             return True
         else:
             return False
 
-    def set_daily(self):
+    def set_daily(self, cronfile):
```
brantleyr: I would rename this function as it no longer only does 'daily' scheduling. Maybe 'set_schedule?' Idunno.

insights/tests/client/test_schedule.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def test_already_linked(self):
+        open(fakefile, 'a').close()
+        self.assertTrue(schedule.already_linked(fakefile))
+        self.assertFalse(schedule.already_linked('/tmp/foo'))
+        os.remove(fakefile)
+
+    def test_set_daily(self):
+        cronfile = '/tmp/crontest.cron'
+        schedule.set_daily(cronfile)
```
brantleyr: I think this changed to .set_schedule()

PR #220, opened 2017-07-11
INSIGHTS_OSP: Added parser for systemctl list-units

vishwanathjadhav:
Modified shared mapper **_unitfiles.py_** for parsing **_systemctl list_units_**

This parser will collect the state of service, sub state of service and and its unit name.


insights/parsers/systemd/unitfiles.py
```diff
@@ -81,3 +93,76 @@ def is_on(self, service_name):
...
+        neutron-openvswitch-agent.service   loaded active running   OpenStack Neutron Open vSwitch Agent
+
+    Example:
+
+        >>> conf = shared[ListUnits]
+        >>> conf.get_service_details('swap.target')
+        {'LOAD': 'loaded', 'ACTIVE': 'active', 'SUB': 'active', 'UNIT': 'swap.target'}
+    """
+    def __init__(self, *args, **kwargs):
+        self.unit_list = {}
```
bfahr: @vishwanathjadhav it would be helpful to show usage of the `unit_list` attribute in the examples assuming you intend for it to be public.

insights/parsers/systemd/unitfiles.py
```diff
@@ -81,3 +93,76 @@ def is_on(self, service_name):
...
+        Returns:
+            Service details
+
+            >>> service_name = "neutron-dhcp-agent.service"
+            >>> self.data.get(service_name)
+            {'LOAD': 'loaded', 'ACTIVE': 'active', 'SUB': 'running', 'UNIT': 'neutron-dhcp-agent.service'}
+
+        Return Type:
+            Dictionary format
+        """
```
bfahr: Although this renders, a better way to format this docstring would be:
```
Returns:
    dict: Dictionary containing details for the service.  If service is not present
    dictionary values will be `None`::

        {'LOAD': loaded', 'ACTIVE': 'active', 'SUB': 'running', 'UNIT': 'neutron-dhcp-agent.service'}
```
Sphinx will automatically provide the `Return Type` section.  I don't believe an examples section is necessary since you have already shown this method in the class examples section.  Also this example uses the `data` attribute which is no longer present.

insights/parsers/systemd/unitfiles.py
```diff
@@ -81,3 +93,76 @@ def is_on(self, service_name):
...
+        valid_states = {'invalid', 'loaded', 'inactive', 'active',
+                        'exited', 'running', 'failed', 'mounted', 'waiting', 'plugged'}
+
+        valid_units = {'service', 'socket', 'device', 'mount', 'automount', 'swap', 'target',
+                       'path', 'timer', 'slice', 'scope'}
+
+        for line in get_active_lines(content):
+            parts = line.split(None)  # AWK like split, strips whitespaces
+            if any(part in valid_states for part in parts):
+                service_details = {}
```
PaulWay: I'd tend to do this as:
```
service_details = {
    'LOAD': parts[1],
    'ACTIVE': parts[2],
    'SUB': parts[3]
}
```
or, if I was feeling too clever:
```
service_defaults = {k: v for k, v in zip(('LOAD', 'ACTIVE', 'SUB'), parts[1:4])}
```

But I'm happy with what you've got there.

insights/parsers/systemd/unitfiles.py
```diff
@@ -81,3 +93,76 @@ def is_on(self, service_name):
...
+        # man systemctl - "is-enabled" knows these states
+        valid_states = {'invalid', 'loaded', 'inactive', 'active',
+                        'exited', 'running', 'failed', 'mounted', 'waiting', 'plugged'}
+
+        valid_units = {'service', 'socket', 'device', 'mount', 'automount', 'swap', 'target',
+                       'path', 'timer', 'slice', 'scope'}
+
+        for line in get_active_lines(content):
+            parts = line.split(None)  # AWK like split, strips whitespaces
+            if any(part in valid_states for part in parts):
```
PaulWay: I've always been somewhat unhappy with this lazy matching of states in any part of the line - it seems inefficient to me.  I'm not sure that the 'is-enabled- states are relevant, either.  But I'm happy to leave it as is if the tests work.

insights/parsers/systemd/unitfiles.py
```diff
@@ -81,3 +93,76 @@ def is_on(self, service_name):
...
+                service_details = {}
+                service_details['LOAD'] = parts[1]
+                service_details['ACTIVE'] = parts[2]
+                service_details['SUB'] = parts[3]
+                if any(unit in parts[0] for unit in valid_units):
+                    service_details['UNIT'] = parts[0]
+                    self.unit_list[parts[0]] = service_details
+
+    def get_service_details(self, service_name):
```
PaulWay: I'd really like to see 'is_loaded', 'is_active' and 'is_running' methods, which would do things like:
```
def is_loaded(self, service_name)
    """ Is the service loaded? """
    return self.unit_list.get(service_name, 'unloaded') == 'loaded'
```
These helper methods make accessing the actual state we're interested in much easier.  But I'm happy with this as is - we can add those later.

bfahr:
@vishwanathjadhav please remove any links to sites that are not publically accessible such as the trello in your inital comment.

insights/parsers/systemd/unitfiles.py
```diff
@@ -166,3 +166,47 @@ def get_service_details(self, service_name):
...
+        """
+        Return the LOAD state of service managed by systemd.
+
+        Args:
+            service_name (str): service name including its extension.
+
+        Returns:
+            bool: True if service is loaded False if not loaded
+        """
+        empty_details = {'LOAD': None, 'ACTIVE': None, 'SUB': None, 'UNIT': None}
```
PaulWay: Since you already have more or less this logic in get_service_details, you should be able to conflate these three lines to:
```
return self.get_service_details(service_name)['LOAD'] == 'loaded'
```

insights/parsers/systemd/unitfiles.py
```diff
@@ -166,3 +166,47 @@ def get_service_details(self, service_name):
...
+        Return the ACTIVE state of service managed by systemd.
+
+        Args:
+            service_name (str): service name including its extension.
+
+        Returns:
+            bool: True if service is active False if inactive
+        """
+
+        empty_details = {'LOAD': None, 'ACTIVE': None, 'SUB': None, 'UNIT': None}
```
PaulWay: Same logic here - use:
```
return self.get_service_details(service_name)['ACTIVE'] == 'active'
```

insights/parsers/systemd/unitfiles.py
```diff
@@ -166,3 +166,47 @@ def get_service_details(self, service_name):
...
+        Return the SUB state of service managed by systemd.
+
+        Args:
+            service_name (str): service name including its extension.
+
+        Returns:
+            bool: True if service is running False in all other states.
+        """
+
+        empty_details = {'LOAD': None, 'ACTIVE': None, 'SUB': None, 'UNIT': None}
```
PaulWay: Same logic here - use:
```
return self.get_service_details(service_name)['SUB'] == 'running'
```

PR #225, opened 2017-07-12
Log file output get after improvements for master branch

PaulWay:
* To save considerable copy-and-paste boilerplate, the time_format is now a property of the object. Subclasses should override this - it is not advised for the end user (i.e. a rule writer) to change the time_format property.
* Sometimes different log lines in the same file have different timestamp formats. Sometimes that's because the log format changes between versions (and our parser has to read the different versions), and sometimes that's because e.g. error messages get a different timestamp format from regular messages. In order to cope with this, we now accept a list of strptime() formats as well as a string. The formats can also be given as the values in a dictionary, which allows the keys to be names indicating the different versions or purposes of the formats.
* If given a list of time formats, the regular expression turns into an alternation between the different format patterns and strptime() is tried on each one in turn to find the correct pattern.
* Some patterns were not tested and these were found to have errors - in particular %p.
* Some patterns needed to be extended to cope with different variations that strptime() accepts.
* The old '_get_after' name was used in some places - this has been changed.


xiangce:
Corresponding to #186

PR #253, opened 2017-07-13
Systemd list units class master

vishwanathjadhav:
Modified shared mapper unitfiles.py for parsing systemctl list_units

This parser will collect the state of service, sub state of service and and its unit name.


xiangce:
@vishwanathjadhav - Since the `insights-core` repo is already public, we'd better do not add any private information into the `git commit` message or PR comments. For detail information please refer to the mail I forwarded from Bob at Jun 28.

xiangce:
Removed the trello link from the `git commit` message

PR #258, opened 2017-07-14
Add 'macpools' field into RHEV metadata

JoySnow:


JoySnow:
1.x branch PR:  https://github.com/RedHatInsights/insights-core/pull/161

JoySnow:
Since the 2.x/master is changed a lot about these stuffs, and maybe they will still undergo other changes, I suppose no need to add test for master now.

PR #261, opened 2017-07-17
Add get_dir() method for the mount point associated with a path for 1.x branch

PaulWay:
One common test is 'what is the disk space associated with this path?', which finds the mount point that underlies that path. This routine tries to find that path, iteratively removing path segments from the end of the given path to find a mount point.  Absolute paths will always resolve at '/', which will always be mounted, so a mount point will always be found.  Relative paths will never resolve to an absolute mount and will return None.

Also tidied the logic in the line parsing, which restores 100% code coverage in testing.


insights/parsers/df.py
```diff
@@ -160,6 +156,23 @@ def get_mount(self, name):
         """Record: Returns Record obj for mount point ``name``."""
         return self.mounts.get(name)
 
+    def get_dir(self, path):
```
xiangce: It's a useful method, after it's merged, I can replace the `get_mount_fs` in sat_diskfree_check with this one.

PR #262, opened 2017-07-18
Enhance virt-what

xiangce:
- add "Red Hat" to DMIDecode.virt_what
- fix typos in dmidecode and refactor tests to pytest style
- add combiner VirtWhat
- add documentation


xiangce:
Need to refactor the DMIDecode.virt_what further as per @lonicerae's suggestion.

insights/combiners/virt_what.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Raises:
+        ParseException: If no ``virt-what`` or ``dmidecode`` information can be found.
+    """
+
+    def __init__(self, local, shared):
+        self.is_virtual = self.is_pyhsical = self.generic = self.specific = None
+        vw = shared.get(VW)
+        if vw and vw.generic != 'Failed':
+            self.is_pyhsical = (vw.generic == 'Baremetal')
```
bfahr: @xiangce I believe this should be `is_physical`.

insights/parsers/virt_what.py
```diff
@@ -1,17 +1,55 @@
...
+    Currently ``virt-what`` fails when running without an environment (or very
+    restricted environment) on RHEL7, and reports below error::
+
+        virt-what: virt-what-cpuid-helper program not found in $PATH
+
+'''
 from .. import Parser, parser
 
+# FIXME:
+# The logic to handle below error should be updated after the bug being fixed.
```
bfahr: @xiangce iI assume that the bug in the Insights client.  The code at line 56 looks for this in the content, are the results valid if this is present in the input?

I think it makes sense to open an issue for this to ensure that we go back and fix it once the bug has been fixed.

mhuth: Even when the bug is fixed, there may be still many systems running the broken version of virt-what.  Afterall, virt-what works fine when run outside of our collector so most people will be unaware of this problem at all.  Its just the way in which our collector runs virt-what (without any environment vars) that the problem presents itself.

xiangce: @bfahr , @mhuth - Thanks, I refactored this parser and removed the "FIXME" description, please check it again when you are available.

insights/parsers/dmidecode.py
```diff
@@ -110,7 +110,8 @@ class DMIDecode(Parser, LegacyItemAccess):
     """
     Class for DMI information.
     """
-
+    # TODO:
+    # Remove below 4 dicts
```
bfahr: This comment is not very clear about when and why to remove.  Please add some context here to explain to another developer why this TODO is here.

xiangce: @bfahr - Thanks a lot. I updated the code according to your suggestions, please check it again.

PR #265, opened 2017-07-18
correct error in log file creation

gavin-romig-koch:
The original code was a problem because if you were not running as 'root' you couldn't create '/var/log/insights-client/' which is what constants.log_dir was set to by default.


brantleyr:
I don't know that this is an issue. You can override the default logging file with --logging-file or pass in custom options when instantiating the API through Python.

gavin-romig-koch:
The existing code will always use, check for and create, the directory defined in constants.log_dir, _not_ the directory specified by --logging-file/configuration file.  Even if you specify --logging-file it will still check for and try to create the directory in constants.log_dir.

The actual log will be written to the right place, using --logging-file/configuration, but the directory check and make are always on the directory in constants.log_dir.

brantleyr:
I believe this has been resolved in the latest logging handlers by @kylape.

kylape:
Hey @gavin-romig-koch just to finish this off -- I did a whole bunch of refactoring that got merged right before you made your changes in this PR.  They definitely make sense, but it would have been a good chunk of work to take your changes and merge them into what I had done.  So I just took the issue you raised and included a fix (very similar to yours) on top of the logging changes I already made.  So you might check out `master` and verify that it works as you expect now.

PR #266, opened 2017-07-19
Logging refactor

kylape:
The goal here is to clean up the code and hopefully make it more approachable, yet keep the exact same logging behavior as before.


kylape:
Feel free to squash all these commits together.  I was really just wanting to break it all out so that you could understand all the individual steps I took to refactor for the review.

PR #270, opened 2017-07-20
Add component metadata and subgraph evaluation.

csams:
Component types and declarations can be enriched with arbitrary
metadata by declaring a type like component = new_component_type(type_metadata=)
or directly in the component declaration with @component(metadata=...).
The metadata in the component declaration inherits from the type and overrides
keys in case of collision.

dr now includes get_subgraphs which takes a dependency graph and yields all
of the disjoint subgraphs within it. This functionality is used within
run_incremental to yield brokers for each independent subgraph given a starting
graph. This should help with memory consumption when executing large numbers of
mostly independent components.


insights/core/dr.py
```diff
@@ -139,7 +163,7 @@ def stringify_requirements(requires):
 
 
 def register_component(component, delegate, component_type,
-        requires=None, optional=None, group=GROUPS.single, alias=None):
+        requires=None, optional=None, group=GROUPS.single, alias=None, metadata={}):
```
jhjaggars: the use of `{}` rather than `dict()` is inconsistent here.

insights/core/dr.py
```diff
@@ -46,6 +50,26 @@ def get_name(component):
     return str(component)
 
 
+def get_metadata(component):
+    return COMPONENT_METADATA.get(component)
+
+
+def get_subgraphs(graph=DEPENDENCIES):
+    keys = set(graph.keys())
```
jhjaggars: I think `set(graph)` behaves the same and saves you a copy.

insights/core/dr.py
```diff
@@ -46,6 +50,26 @@ def get_name(component):
...
+    seen = set()
+    while keys:
+        frontier = {keys.pop()}
```
jhjaggars: I think you'll have to use `set([keys.pop()])` to support python 2.6.

csams: Good catch.

Issue #112, opened 2017-06-19
chkconfig.py can not parsed RHEL 7.3 'chkconfig --list'


miclark:
Transferred from [trello](https://trello.com/c/oIERHjUJ)

The mapper can not handle below content

~~~
Note: This output shows SysV services only and does not include native
      systemd services. SysV configuration data might be overridden by native
      systemd configuration.

      If you want to list systemd services use 'systemctl list-unit-files'.
      To see services enabled on particular target use
      'systemctl list-dependencies [target]'.

netconsole     	0:off	1:off	2:off	3:off	4:off	5:off	6:off
network        	0:off	1:off	2:on	3:on	4:on	5:on	6:off
rhnsd          	0:off	1:off	2:on	3:on	4:on	5:on	6:off

xinetd based services:
	chargen-dgram: 	off
	chargen-stream:	off
	daytime-dgram: 	off
	daytime-stream:	off
	discard-dgram: 	off
	discard-stream:	off
	echo-dgram:    	off
	echo-stream:   	off
	rsync:         	on
	tcpmux-server: 	off
	time-dgram:    	off
	time-stream:   	off
~~~

Error:
~~~
Traceback (most recent call last):
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/core/evaluators.py", line 197, in run_mappers
    self.add_result(self._execute_mapper(plugin, context),
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/core/evaluators.py", line 76, in _execute_mapper
    return mapper(context)
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/mappers/chkconfig.py", line 66, in __init__
    super(ChkConfig, self).__init__(*args, **kwargs)
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/core/__init__.py", line 112, in __init__
    self.parse_content(context.content)
  File "/opt/insights-plugins/lib/python2.7/site-packages/falafel/mappers/chkconfig.py", line 104, in parse_content
    states = self.level_states['xinetd']
KeyError: 'xinetd'
~~~

miclark:
Hey @PaulWay, assigning this to you. Trying to prioritize bugs over other tasks. If you don't feel it's appropriate to work this one, let me know and we'll move it around.

PaulWay:
Sure thing @miclark - no problem!

PaulWay:
[...]

PR #275, opened 2017-07-24
Handle RHEL 7.3 `chkconfig` output

PaulWay:
RHEL 7 uses systemd, which replaces the previous System V init system.  The `chkconfig` command still exists, but only outputs data for System V services.  Since systemd does not use or need `xinetd`, the output of `chkconfig` does not include the `xinetd` service.  However, previous versions of this parser used the state information of the `xinetd` service as the state of any xinetd-based services (since such services would only be enabled when `xinetd` was enabled).

This update allows the `ChkConfig` parser to read the `chkconfig` output on RHEL 7 systems that do not list `xinetd` as a service but include `xinetd`-like services.


PaulWay:
This also fixes #112

PR #91, opened 2017-06-15
Remove the deprecated method: get_expiration_date()

JoySnow:

PR #18, opened 2017-05-31
fd_total_limit mapper is unused - remove it and improve tests 

PaulWay:
Tests of insights-core and insights-plugins all pass.

PR #277, opened 2017-07-24
Document the ``nproc`` and ``get_limits`` as deprecated

xiangce:
- update the spec of limits.d to collect only `.conf` file
- remove the parser nproc.py and its specs
- update the unit test of limits_conf.py to pytest style


jhjaggars:
Hey @xiangce since this is an API change we shouldn't merge this in to the 1.x branch.

xiangce:
@jhjaggars - Yep, I'm working on refactoring the plugins to use the new `AllLimitsConf`, will not merge this PR before that refactorings finished. Thanks.

insights/parsers/limits_conf.py
```diff
@@ -2,86 +2,16 @@
...
-        for line in get_active_lines(context.content):
-            parsed = parse_line(line)
-            lines.append(parsed)
-        result[cfg_file] = lines
-        return result
-
-
 @parser("limits.conf")
 @parser("limits.d")
 class LimitsConf(Parser):
```
bfahr: @xiangce is there a plan to split these out into separate parsers?  This is related to issue #14.

jhjaggars:
@xiangce what I mean is, we can't be sure that other projects aren't using the parsers you aim to remove and thus it isn't safe to do so in this branch, even after you convert plugins to use the new parser.

insights/parsers/limits_conf.py
```diff
@@ -2,86 +2,16 @@
...
-        "value": value
-    }
-
-
-@parser("limits.conf")
-@parser("limits.d")
-def get_limits(context):
-    """
-    Returns a dictionary with one item, keyed to the file name, which is a
-    list of dictionaries corresponding to each rule in the file in order.
```
bfahr: @xiangce per @jhjaggars comment we should document this function as deprecated and leave it in the module.  Here's an example of how to document this function as deprecated:

```
"""
    .. warning::
        Deprecated function, please use :class:`LimitsConf` instead.

    Returns a dictionary...
```

This will render similar to the following image:
deprecation_rendered

xiangce: @bfahr - I see, thank you for the detailed explanation. I will update the document in the next commit.

@jhjaggars - Got it.

xiangce:
@jhjaggars - Are there any other projects have used the "insights-core" framework except the "insights-plugins" and "diag-insights-rules"? If I make sure that those deprecated APIs are never used in the two repos, can I remove them from the 1.x branch? AFAIK, we have removed some deprecated API before, such as https://github.com/RedHatInsights/insights-core/pull/91 and https://github.com/RedHatInsights/insights-core/pull/18.

insights/parsers/limits_conf.py
xiangce: @bfahr - I added the warning into the docsting, would you please check it again? Thanks.

xiangce:
Hi @bfahr - I updated the docsting and added `warning` for the deprecated APIs, would you please to check this PR again? Thanks.

insights/parsers/nproc.py
```diff
@@ -1,6 +1,9 @@
 """
 limits : Files ``limits.conf`` and ``*-nproc.conf``
 ===================================================
+.. warning::
+    Deprecated parser, please use the :class:`AllLimitsConf` combiner instead.
```
bfahr: @xiangce to reference the combiner use the full path, `insights.combiners.limits_conf.AllLimitsConf`, so that the link is functional in the documentation.

xiangce: Cool, I modified it. Thanks @bfahr

PR #279, opened 2017-07-24
Install the Core to /var/lib

brantleyr:


bfahr:
@brantleyr should this PR be against the new_client branch or master?

brantleyr:
The new_client branch has been retired and merged into master.

insights/client/__init__.py
```diff
@@ -255,22 +255,54 @@ def install(self, new_egg):
...
-
+            if not os.path.isdir(constants.insights_core_lib_dir):
+                logger.debug("Creating directory %s for the Core." % (constants.insights_core_lib_dir))
+                os.mkdir(constants.insights_core_lib_dir)
+        except OSError:
+            message = "There was an error creating %s for Core installation." % (constants.insights_core_lib_dir)
+            raise OSError(message)
+
+        # Copy the NEWEST egg to /var/lib/insights/last_stable.egg
+        old_newest_egg = constants.insights_core_rpm
```
kylape: This block of code is confusing to me.  Let me just lay out how I think the logic should be and see what you think:

- If there is a `last_stable` and `newest` egg: Copy `newest` to `last_stable`, overwriting the current egg at `last_stable`.
- If there is a `last_stable` but no `newest`: Do nothing (though I think this state will be very rare)
- If there is no `last_stable` but a `newest`: Copy `newest` to `last_stable`
- If there is neither `last_stable` nor `newest`: Do nothing (since there's nothing to do)

So the logic can be reduced to this:  If there's a `newest` egg, copy it to `last_stable`.  I don't think anything should be done with the RPM egg.  It's built in to the fallback mechanism in the client script already.

One question that comes to mind:  How can the script determine if the `newest` egg's last run was actually successful?

insights/client/__init__.py
```diff
@@ -255,22 +255,54 @@ def install(self, new_egg):
         """
         returns (dict): {'success': True if the new core was installed successfull else False}
         """
+        import os
+        from shutil import copyfile
+
         success = False
```
brantleyr: Done.

insights/client/__init__.py
brantleyr: I had installed the 'rpm.egg' as 'last_stable.egg' as an additional fallback mechanism but that logic can be removed. I believe the logic should be (as you stated), if there is a 'newest.egg' then copy/move that to 'last_stable.egg,' then copy/move the egg that was fetched and verified in /tmp to 'newest.egg.' I'm not sure exactly how we track if the 'newest.egg's last run was actually successfully. First we would need to define what successful was, I assume that would be that it made it through all essential steps (collecting and uploading). Then maybe after a successful upload we keep track of a dot file in /etc/insights-client. Then, however, we would have to add some logic to link that dot file to that actual egg file.

kylape: Yeah, I think the logic to copy rpm to last_stable can be removed because the insights-client script falls back to the rpm egg at its original location, thus no need to copy it anywhere.

The idea that was mentioned yesterday was to let the egg move itself from 'newest' to 'last_stable' at the end of the upload function, but I don't know if that's really better than the dot file solution you described.

kylape:
We took this branch and did a lot more work and then merged it manually.  So this PR is outdated.

PR #272, opened 2017-07-24
Improvements to VgDisplay for `vgdisplay -vv` command handling for 1.x branch

PaulWay:
Substantial rewrite of the parsing of the output, both to cope with the extended information from ``vgdisplay -vv`` and to reduce the number of corner cases.

This keeps the previous structure of a list of volume groups in the 'vginfo_dict' key of a 'data' dictionary.  All previous tests pass.  However, we significantly extend the keys stored, we store logical and physical volume data inside each volume group structure, and we ignore informational and verbose-status lines more easily.

Documentation for the VgDisplay parser class has also been added, with sample data and usage code.

PR #285, opened 2017-07-25
Improvements to VgDisplay for `vgdisplay -vv` command handling for master branch

PaulWay:
Substantial rewrite of the parsing of the output, both to cope with the extended information from `vgdisplay -vv` and to reduce the number of corner cases.

All previous tests pass. However, we significantly extend the keys stored, we store logical and physical volume data inside each volume group structure, and we ignore informational and verbose-status lines more easily.

The 'data' dictionary property has been broken up into a `vg_list` property for the list of volume groups and the existing `debug_info` property.

Documentation for the VgDisplay parser class has also been added, with sample data and usage code.

This is the master branch version of #272


insights/parsers/vgdisplay.py
```diff
@@ -1,83 +1,154 @@
...
     """
 
+    _FILTER_INFO = [
+        "Couldn't find device with uuid",
+        "physical volumes missing",
+    ]
+
     def parse_content(self, content):
         self.data = {}
-        vginfo_dict_list = []
```
bfahr: @PaulWay do you need to also remove the `self.data` declaration above?

insights/parsers/vgdisplay.py
```diff
@@ -1,83 +1,154 @@
...
+
+        --- Physical volumes ---
+        PV Name               /dev/mapper/luks-96c66446-77fd-4431-9508-f6912bd84194
+        PV UUID               EfWV9V-03CX-E6zc-JkMw-yQae-wdzp-Je1KUn
+        PV Status             allocatable
+        Total PE / Free PE    118466 / 4036
+
+    Lines containing 'Couldn't find device with uuid' and 'missing physical
+    volume' are stored in a ``debug_info`` property, which is also available
+    in the ``debug_info`` key in the ``data`` dictionary.
```
bfahr: Reference to `data` dictionary.

PR #295, opened 2017-07-31
Enhance hammer_ping for master

xiangce:
- now can handle some unexpected outputs
- update interfaces


insights/parsers/hammer_ping.py
```diff
@@ -1,21 +1,13 @@
 """
-hammer_ping - command ``/usr/bin/hammer ping``
-==============================================
+HammerPing - command ``/usr/bin/hammer ping``
+=============================================
 
 The hammer ping parser reads the output of ``hammer ping`` and turns it into
 a dictionary of service statuses in the ``data`` property.  Each service has
```
bfahr: @xiangce if you could add the `data` and `error` attributes to the documentation I believe this is ready to merge.

insights/parsers/hammer_ping.py
```diff
@@ -30,69 +22,75 @@
...
+        self.status_of_service = {}
+        """dict: The status of each service, converted to lower case."""
+        self.response_of_service = {}
+        """dict: The response of each service, as is."""
+        self.service_list = []
+        """list: The list of service names in order."""
+        self.errors = []
+        """list: Any error messages encountered during parsing."""
+        self.is_ok = False
+        """bool: Were all services read correctly and are they all in 'ok' state?"""
```
bfahr: @xiangce could you move these docstrings to the class level so they will be included in the documentation?

PR #296, opened 2017-08-01
handle unicode characters in RPM name when accessing InstalledRpm property

bobcallaway:
Ensure that if RPM name contains an unicode character that the 'package' property is formatted using an unicode format string; without this, the call fails with a UnicodeEncodeError

Signed-off-by: Bob Callaway <bob.callaway@redhat.com>


bobcallaway:
#297 just created

PR #298, opened 2017-08-02
Remove unused parsers 1.x

PaulWay:
There are two parsers that aren't used in the rules and that do not have specs listed in `config/specs.py`: cpuinfo_max_freq (deprecated in favour of CpuInfo) and OracleConfig.  These are proposed to be removed.

Two other parsers - PamConf and OcGetPvc - have specs ('pam.conf' and 'oc_get_pvc' respectively) that were not mention in `config/specs.py` and are now included.


xiangce:
@PaulWay - As per the new policy of `branch 1.x`, we do not suggest to remove any APIs from now on although these API are never used in `insights-pulgins`, since can't be sure that other projects aren't using them. We should mark them as deprecated in 1.x know and remove them from 2.0 branch.

See: https://github.com/RedHatInsights/insights-core/pull/277

insights/config/specs.py
```diff
@@ -246,6 +246,7 @@
     "ovirt_engine_server.log"   : SimpleFileSpec("var/log/ovirt-engine/server.log"),
     "ovs-vsctl_show"            : CommandSpec("/usr/bin/ovs-vsctl show"),
     "pacemaker.log"             : SimpleFileSpec("var/log/pacemaker.log"),
+    'pam.conf'                  : SimpleFileSpec("etc/pam.conf"),
```
xiangce: This doesn't seem related to the "Remove...". Better open a new PR for these specs adding.

PaulWay:
OK, I've split this up into two separate MRs, one (#356) for the removal of the deprecated parsers and one (#357) for adding the two missing specs.  I'll close this one.

PR #299, opened 2017-08-03
SMARTctl documentation and improved functionality for 1.x branch

PaulWay:
* Added documentation for SMARTctl parser class.
* Hiding the regexes from documentation
* Instead of having an 'info' dict, make the information from each section available as a separate property.
* Get attributes directly from group dictionary in regex match.
* Device names are of the form '/dev/sda', not just 'sda'.
* If we can't get the device name from the path, raise a ParseException.
* Test for ParseExceptions - 100% code coverage.
* Put test data next to test cases.


insights/parsers/smartctl.py
```diff
@@ -1,55 +1,111 @@
...
-            'full_line': '',  # hack for persistent line storage
-        }
+        self.info = {}
+        self.health = 'not parsed'
+        self.values = {}
+        self.attributes = {}
+        # hack for persistent line storage in parse_content context -
+        # otherwise it gets treated as a local variable within the sub-
+        # functions
+        self.full_line = ''
```
bfahr: @PaulWay you can add the new functionality but you need to maintain the existing structure of `self.info` for backward compatibility in this branch.

PaulWay: Hi Bob - A good point, and one I will make sure to remember in the future!  On the other hand, part of the reason I did this change was that there are no rules in either the insights-plugins or diag-insights-rules repositories that use the SMARTctl parser.  Therefore I felt it was OK to change an interface that has not yet been used :-)

bfahr: @PaulWay while we may not be using this on our projects, we don't know how it is being used by other internal or external users.  So we need to minimize the impact to those users we may not know about now that the project is publicly available.  You can make all of these changes on the master branch but we need to maintain backward compatibility on this branch.

PaulWay: I'm not sure who else could possibly be using it, internally or externally, but OK.

insights/parsers/smartctl.py
```diff
@@ -1,55 +1,118 @@
...
-         * 'attributes' - the SMART attributes (-A) - run time data
-        This function gets called once per detected file, so we don't have to
-        deal with file changes within this process.
+    Parser for output of ``smartctl -a`` for each drive in system.
+
+    This stores the information from the output of `smartctl` in the
+    following properties:
+
+     * ``device`` - the name of the device after /dev/ - e.g. sda
+     * ``info`` - the -i info (vendor, product, etc)
```
bfahr: @PaulWay I believe this should be `information` shouldn't it?  You should also add information about `info` being deprecated in a future version if that change was made in the master branch.

insights/parsers/smartctl.py
```diff
@@ -1,55 +1,118 @@
...
+          1 Raw_Read_Error_Rate     0x000f   118   099   034    Pre-fail  Always       -       179599704
+          3 Spin_Up_Time            0x0003   098   098   000    Pre-fail  Always       -       0
+          4 Start_Stop_Count        0x0032   100   100   020    Old_age   Always       -       546
+          5 Reallocated_Sector_Ct   0x0033   100   100   036    Pre-fail  Always       -       0
+        ...
+
+    Examples:
+        >>> for drive in shared[SMARTctl]:
+        ...     print "Device:", drive.device
+        ...     print "Model:", drive.info['Device Model']
```
bfahr: Should this be `information` as well?

PaulWay: A pox upon me - sorry, completely forgot to update the documentation :-(  Will do!

PaulWay:
@bfahr is this ready to go now?

bfahr:
Approved pending PR for the master branch.

PR #301, opened 2017-08-04
New parser katello_service_status

xiangce:
- Since the output of `katello- service status` is very different
  between RHEL6 and RHEL7 and we only care about if there is any failed
  services, we only analyze the last line of this command at present.


xiangce:
@PaulWay 
Below is the output of `pulp_workers` and `foreman-tasks` which are part of the output of the `katelllo-service status on RHEL6.
~~~
[root@liuxc-rhel6-sat6 ~]# service pulp_workers status
celery init v10.0.
Using config script: /etc/default/pulp_workers
node reserved_resource_worker-0 (pid 902) is running...
node reserved_resource_worker-1 (pid 921) is running...
node reserved_resource_worker-2 (pid 938) is running...
node reserved_resource_worker-3 (pid 959) is running...
[root@liuxc-rhel6-sat6 ~]# service foreman-tasks status
dynflow_executor is running.
dynflow_executor_monitor is running.
[root@liuxc-rhel6-sat6 ~]# 
~~~

Especially for `pulp_worker` its output would be different corresponding to the server's configuration. That's why I do not analyze each of them in the parser, on RHEL6 it's difficult to collect all the information.

AND, the most important is,  we only need the result the command (that is: if there are some services are failed), it's enough for us to analyze the last line which would tell us the result. And I prefer to get the result via the command itself instead of to analyze its uncertain output ourselves.

`katello-service status` collected all the necessary services' status and analyzed their result and reported at the last line, it's the best way to get what we want via the command itself.

For other purposes, when we want to get more detail information about the service's `status`, it will be the best way to analyze the output of the `service SVC_NAME status` or `systemctl status SVC_NAM` directly. Because it's good to get the result via the command itself, we can make another parser based on `service {SVC_NAME} status, SVC_NAME=['httpd', 'tomcat', ....]` in that time.

BTW, At present, there is no parser to analyze the output of service's status. the `Service` combiner only checks if the specific service is enabled or not, but it didn't analyze their running `status`.

PaulWay:
If you need the status of the service from the Services combiner, then I'd recommend giving it that feature.

I'm a bit puzzled by your logic.  You say you're only interested in which services have failed and you throw away everything but the last line of the `katello-service status` output, and then you say that we can't just check those services directly because we need to check the actual status of pulp-worker?  Can you explain how what you need cannot be retrieved from the output of `systemctl` or `chkconfig` but can only be retrieved from the last line of `katello-service status`?

I'm trying to imagine what rule you're constructing that needs this output but cannot get it from anywhere else.  I would really prefer to not add parsers, especially of command output, if we can get that information from existing parsers.  For example, it would be fairly simple to write a rule that checks these services' status:

```
running = shared[PsAuxcww].running
if not all(svc in running for svc in ('celery', 'pulp', tomcat6', 'httpd')):
    ...
```

(Just choose the appropriate process names to search for.)

Overall I still don't see a reason to add this command when we already have existing parsers to check process and service status.

xiangce:
The example of `pulp_workers` and `foreman-tasks` is just want to explain that the output of some services on RHEL6 is different with the output of  `systemctl status xxx` on RHEL7. And it's obvious that there isn't explicit information about `pulp_workers` and `foreman-tasks` processes in the example output. It's not an easy task to add this feature into the Servcies combiner (and I do not suggest to add it because on one plugin needs it till now).

AND I do not need to analyze the output of the "status" because I only care about the result that the last line of the `katello-service status` can tell me directly.

SO, it's not necessary to analyze the status of each service in `katello-service`.

Although, I can get if the service is running by checking if the daemon process is in PS, but it's not the best way to do that, just like the example that I gave above, we can not check the daemon process for some services (e.g. pulp_workers, you can't even find any processes for this service, on RHEL7, it has already existed; on RHEL6 it's related to server configuration). 

Using the most simple tool to get the most accurate result is what I want. The last line of the `katello-service status` command just gives me that. The command itself already analyzed all of the necessary services, no matter the OS is RHEL6 or RHEL7 and no matter the Sat is 6.0 or 6.2. It just gives out the final result that I wanted. That's why I make this parser.

PaulWay:
I'm sorry, @xiangce, but I can't agree with creating another parser that retrieves data that is already available in other parsers.  I think this would be best as a combiner that pulls the information from the Services combiner and can report on which Satellite 6 services are not running.

I'll be guided by @bfahr on this.

xiangce:
@PaulWay - As I said, it's not an easy task to parse the output of "service xxx status" (RHEL6), especially for services like `pulp_workers` or to check it in `PsAux*`. Even if we made parsers/combiners for them, I do not think it could be more accurate than to parse the `katello-service status` directly :-(.

@bfahr would you please to take a look at this parser?

bfahr:
@xiangce and @PaulWay, I discussed this with @kylape and we agreed that having additional sources of information adds to reliability of the information in that it helps to insulate against changes across release of the OS, and different distributions.  We also agree that having combiners that can simplify access to this information is a good idea.  Regarding how much information should be parsed from an output, it is difficult to anticipate every use of the parser information, so parsing should target first the information needed by the rule, and second the information that is easy/obvious to parse.  Beyond that the value of parsing information that does not have a current use has diminishing value.  Parsers should be reliable and useful, and if additional use cases are identified in the future those parsers can be modified.

PR #314, opened 2017-08-09
Make smb.conf parser evaluate the ini file the same as samba.

jsvob:
Options outside any section belong to [global].
Section names are lowercased, stripped, and merged.

Signed-off-by: Jakub Svoboda <jsvoboda@redhat.com>


insights/parsers/samba.py
```diff
@@ -24,4 +24,30 @@ class SambaConfig(IniConfigFile):
 
     def parse_content(self, content):
         lstripped = [line.lstrip() for line in content]
```
PaulWay: I think you're better off putting the two 'lstripped' expressions into one here.

insights/parsers/samba.py
```diff
@@ -24,4 +24,30 @@ class SambaConfig(IniConfigFile):
...
+        for old_key, old_section in self.data._sections.iteritems():
+            new_key = old_key.strip().lower()
+            if new_key in new_dict:
+                new_section = new_dict[new_key]
+            else:
+                new_section = self.data._dict()
+                new_dict[new_key] = new_section
+            # Merge same-named sections just as samba's `testparm` does.
+            for s_k, s_v in old_section.iteritems():
+                new_section[s_k] = s_v
```
xiangce: Hi @jsvob - It would be better to use the `update()` method of `dict` to simplify the code, e.g.:

~~~
        for old_key, old_section in self.data._sections.iteritems():
            new_key = old_key.strip().lower()
            if new_key not in new_dict:
                new_dict[new_key] = self.data._dict()
            # Merge same-named sections just as samba's `testparm` does.
            new_dict[new_key].update(old_section)
~~~

PR #317, opened 2017-08-10
Update args to kwargs

brantleyr:


insights/client/__init__.py
```diff
@@ -68,94 +68,21 @@ def test_connection(self):
...
-              verbose=False):
+    def fetch(self, **kwargs):
         """
+            parms:
+                core_url (str): URL to retrieve egg from
+                force (bool): force fetch egg (dont check etags)
             returns (str): path to new egg.  None if no update.
         """
         # was a custom egg url passed in?
         if config['core_url']:
```
kylape: Why not just combine this into one line:

```
egg_url = kwargs.get('core_url', config['core_url'])
```

And just make sure the default in the config for `core_url` is `constants.egg_path`

insights/client/__init__.py
```diff
@@ -68,94 +68,21 @@ def test_connection(self):
...
-            logger.debug('Skipping upload, running offline, or system is not properly registered.')
-            logger.debug('Skipping upload.')
-            logger.debug('Insights results: %s', results)
-            return results
-
-    def fetch(self,
-              egg_url=constants.egg_path,
-              force=False,
-              verbose=False):
+    def fetch(self, **kwargs):
```
kylape: I think it's more clear to have the kwargs listed out here.  But I don't think the verbose kwarg should be here.  So `fetch(self, egg_url=None, force=False)` sounds reasonable to me.

insights/client/__init__.py
```diff
@@ -281,13 +208,13 @@ def install(self, new_egg):
         logger.debug("The new Insights Core was installed successfully.")
         return {'success': True}
 
-    def update_rules(self, options=None, config=None):
```
kylape: These kwargs make sense to remove, though.

insights/client/__init__.py
brantleyr: Genius.

PR #320, opened 2017-08-11
All parsers are now automatically available for use in Fava

gavin-romig-koch:
declare the parser name to Fava when register_parser runs
     remove the explicit calls to fava.add_shared_parser
     change the tests to test a wider range of parsers and non-parsers
     rationalize how imports of parsers are done in test_fava


insights/tests/test_fava.py
```diff
@@ -808,16 +815,57 @@ def test_yaml_tests(yaml2jsonTest):
...
+    with pytest.raises(fava.FavaTranslationError):
+        fava.FavaRule(PARSER_NAMES_TEST_TEMPLATE.substitute({"name": "Nosuchparser"}))
+
+
+def test_parser_names_combiner():
+    # combiner names should not work in Fava
+    with pytest.raises(fava.FavaTranslationError):
+        fava.FavaRule(PARSER_NAMES_TEST_TEMPLATE.substitute({"name": hostname.__name__}))
+
+
```
bfahr: @gavin-romig-koch would you mind to convert code below from unittest to pytest?

gavin-romig-koch: I would rather covert the whole file as a separate PR, than just the code below in this PR.  

Would that be acceptable?

bfahr: Yes it would, thanks.

PaulWay:
Looks good to me.

PR #321, opened 2017-08-11
Move and translate Fava*.md files to docs/fava.rst

gavin-romig-koch:


docs/fava.rst
```diff
@@ -1,10 +1,26 @@
...
+:Date: Oct 27, 2016
 
-### Third Party Rule Plugins ###
+.. contents:: Table of Contents
+    :depth: 6
+
+##############
+Fava Rationale
+##############
+
```
bfahr: @gavin-romig-koch I think it might present better if the sections *Fava Rationale* and *Fava Reference* were subheadings under *Fava*.  This would be consistent with the other major sections of the doc.  See this reference: http://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html#headings

docs/fava.rst
```diff
@@ -23,33 +39,28 @@ The important parts of a rule plugin:
...
-```yaml
----
-rule:
-  name: "UNAME_TEST"
-  pydata:
-    kernel: "{{ Uname.kernel }}"
-  when: Uname.fixed_by('2.6.32-431.11.2.el6', introduced_in='2.6.32-431.el6')
-```
-
+  ---
```
bfahr: You can use the following for code blocks if you want syntax highlighting:
```
.. code block:: yaml
```
See https://github.com/RedHatInsights/insights-core/blob/master/docs/customtut_parsers.rst for some examples.  Also see Sphinx reference: http://www.sphinx-doc.org/en/stable/markup/code.html

docs/fava.rst
```diff
@@ -59,36 +70,38 @@ The value of the pydata key is a dictionary of the pydata to be added to the res
 
 ### Fava and Loops ###
 
+**************
```
bfahr: You have an orphaned markdown line at 71 above.

docs/fava.rst
```diff
@@ -59,36 +70,38 @@ The value of the pydata key is a dictionary of the pydata to be added to the res
...
-  # stuff skipped....
-```
+  ---
+  rule:
+    # stuff skipped....
+    vars:
+      up: "{{ eth|selectattr('link_detected')|list }}"
+    # stuff skipped....
 
 This is a Jinja2 expression, it says to apply the 'selectattr' filter to 'eth', and then to apply the 'list' filter to its result.  The 'selectattr' filter selects items from its argument (the thing it is filtering), including in its result only those for which the given attribute ('link_detected') is true.  The 'list' filter converts it's argument to an actual list if it isn't already.  Jinja2 will optimize a series of fiters to avoid creating intermediate lists for each step.
```
bfahr: Did you mean to emphasize terms using backticks like this `selectattr` instead of single quotes 'selectattr'?

gavin-romig-koch: I should have used double backticks for these.

docs/fava.rst
gavin-romig-koch: woops.

docs/fava.rst
gavin-romig-koch: yes.

docs/fava.rst
gavin-romig-koch: yes.

PR #322, opened 2017-08-11
Fava - add when with a list

gavin-romig-koch:
if the value of a when condition is a list, then AND it's elements so you can
   write a long list of 'when' conditions as a list


PaulWay:
Looks good to me.

bfahr:
@gavin-romig-koch a newer PR for your test file was merged ahead of this and is causing merge conflicts with this PR.  Could you update this PR to fix that problem.

gavin-romig-koch:
@bfahr fixed the problem.

PR #323, opened 2017-08-11
Standard Import Hooks for Fava

csams:
fava files can be loaded as python modules with standard import statements:
`from insights.plugins import basic_one`

All files with ".fava" extension are imported with the FavaImporter & Loader.

This is done at the sys.path_hooks level using the PEP-302 Importer protocol.
The Loader is compatible with pkgutil.walk_packages, iter_modules, etc. and
so works with the load_components function in master.

For parser or combiner dependencies to resolve, just ensure they've
been loaded: `plugins.load("insights.parsers")`, etc. They get added
to the fava machinery as they register themselves.

Note this will not work if fava files are contained in compressed files,
because those file types are handled by the builtin zipimporter, and importers
currently aren't stackable. We'd have to reimplement archive handling as well.


jhjaggars:
@csams can you rebase with master and try this against the new ci scripts. They cover 2.6.

insights/core/plugins.py
```diff
@@ -413,6 +414,8 @@ def register_component(
         EMITTERS.add(component)
         register_consumer(component)
 
+    fava.add_shared_parser(component.__name__, component)
```
gavin-romig-koch: Not sure about this, but you @csams would know better than me.  Do all components end up with key/value pairs in the 'shared' argument? Or only some components?   Fava can only handle components that show up in the 'shared' argument.   Maybe this line needs an 'if component.shared:' protecting it?

insights/core/fava.py
```diff
@@ -614,3 +617,150 @@ def create_module(full_module_name, the_ast, the_globals):
...
+    def find_module(self, fullname, paths=None):
+        """ Returns a FavaLoader for fava files and a pkgutil.ImpLoader
+            for standard python files. The ImpLoader is a PEP-302 wrapper
+            for python's regular import machinery.
+        """
+        name = fullname.split(".")[-1]
+        filename = os.path.join(self.path, name + self.ext)
+        if os.path.exists(filename):
+            return FavaLoader(filename)
+        return pkgutil.ImpLoader(fullname, *imp.find_module(name, [self.path]))
```
kylape: Do you need to return a custom loader here?  Based on PEP 302, you can let the original import machinery handle the import if you raise `ImportError` here.

bfahr:
@csams it appears that @kylape made the last comment, are you still working on this PR?

insights/core/fava.py
csams: Raising `ImportError` from `find_module` would abort the import entirely, which isn't what we want: "If find_module() raises an exception, it will be propagated to the caller, aborting the import." When `FavaImporter` successfully constructs, that tells the import machinery that we can handle the path. Since I wanted to support .fava and .py files in the same package, find_module has to return the right loader for either of them.

insights/core/plugins.py
csams: @gavin-romig-koch You're right, we should protect that. The registration function has a boolean for whether the component is shared, so I'll use that to guard.

csams:
@bfahr @kylape @gavin-romig-koch I think I've addressed everything, and I merged 1.x into the branch to get it back up to date. Let me know if I need to change anything else.

insights/core/fava.py
kylape: I see, thanks for clarifying.

PR #326, opened 2017-08-14
More features added to keyword_search

PaulWay:
* Allow searching on keys in the data that have spaces and dashes (both impossible to use in a keyword parameter), by copying the data in the searched row and mashing these characters in the key into underscores.  So 'fix-up path' can be found with the search key 'fix_up_path'.
* Searches can be refined using the __contains and __startswith suffixes, which do exactly what they say.


insights/parsers/__init__.py
```diff
@@ -329,6 +329,17 @@ def keyword_search(rows, **kwargs):
...
+    match the keyword argument parsing.  For example, the keyword argument
+    'fix_up_path' will match a key named 'fix-up path'.
+
+    In addition, several suffixes can be added to the key name to do partial
+    matching of values:
+
+    * '__contains' will test whether the data value contains the given
+      value.
+    * '__startswith' tests if the data value starts with the given value
+
```
bfahr: @PaulWay do you think it would be useful to include examples using these suffixes in the *Examples* section?

bfahr: Also the `Returns` docstring should not have the second line indented.

bfahr:
Would it be useful to have a `case_insensitive` option to match irrespective of case?

insights/parsers/__init__.py
```diff
@@ -353,7 +364,24 @@ def keyword_search(rows, **kwargs):
...
+    # __startswith
+    def key_match(row, key, value):
+        # Translate ' ' and '-' of keys in dict to '_' to match keyword arguments.
+        my_row = {
+            key.replace(' ', '_').replace('-', '_'): val
+            for key, val in row.iteritems()
+        }
+        if key.endswith('__contains'):
+            key = key[:-len('__contains')]
+            return key in row and value in row[key]
```
xiangce: Should it be `key in my_row and value in my_row[key]`?

PaulWay:
OK, I think I've addressed all the issues - how does this look?

xiangce:
Currently, the "lower_value" mode only supports “full" match (like "default" mode, but it is case-insensitive). I'm OK if you think it's enough. However, do you think it would be better to add an option (maybe  @bfahr also think so) to the "keyword_search" function, e.g. `keyword_search(rows, ignorecase=False, **kwargs)`. In this way, all the three modes ('default', 'startswith' and 'contains') would support this feature. How do you think?

insights/parsers/__init__.py
```diff
@@ -329,6 +329,19 @@ def keyword_search(rows, **kwargs):
...
+
+    In addition, several suffixes can be added to the key name to do partial
+    matching of values:
+
+    * '__contains' will test whether the data value contains the given
+      value.
+    * '__startswith' tests if the data value starts with the given value
+    * '__lower_value' compares the lower-case version of the data and given
+      values.
+
```
gavin-romig-koch: This is useful and needed functionality, but I think there are better ways to implement this added functionality. 

If there are users asking for this functionality, and this is how they wanted it implemented, or if there are other reasons for doing it this way, then the following is not so important.  But if not, Paul would you consider changing the way you did this to something along the lines ....

First the spaces and dashes in keys part of this PR.  Python already has a syntax that allows one to put spaces and dashes and anything else one might need into mapping keys.  

Instead of:
    result = keyword_search(rows, key_with_spaces='value')

just do:
   result = keyword_search(rows, **{ 'key with spaces' : 'value' })

Yes, it a few extra characters over keyword arguments, but it doesn't require any extra code in keyword_search, and will never conflate spaces, dashes, or underscores.

Second, the value match functions, '__contains', etc.  I think it would be better to allow boolean functions for the values to match would be better than a fixed set of suffixes on the keys.  

Instead of:
    result = keyword_search(rows, item__contains='c')

do:
    result = keyword_search(rows, item=lambda v: 'c' in v)

To implement this, you would change keyword_search to look at the type of the value in each kwarg key/value pair.  If it's a string compare it equality, if it's a function (a callable) then call it.

insights/parsers/tests/test_parsers_module.py
```diff
@@ -313,3 +346,50 @@ def test_keyword_search():
...
     assert len(keyword_search([], role='server')) == 0
+
+    # Search with contains
+    results = keyword_search(DATA_LIST, role__contains='e')
+    assert len(results) == 4
+    assert results[0] == DATA_LIST[0]
+    assert results[1] == DATA_LIST[1]
+    assert results[2] == DATA_LIST[2]
+    assert results[3] == DATA_LIST[3]
+
```
PaulWay: @gavin-romig-koch - I'm always happy to use a more compact representation, so I'll use that :-)

insights/parsers/__init__.py
PaulWay: @gavin-romig-koch - I like those ideas.

I'm borrowing from the Django QuerySet syntax, which uses the double-underscore syntax to do both match specification (e.g. name__less_than='M' or date__year=1956) and key-relation referencing (e.g. author__name='Banks, Iain M.', where the 'author' field is a foreign key into the author table, which has a 'name' field).

In general I wanted to avoid rule code having to write Python expressions, precisely because I thought the keyword searching would be a useful feature in Fava, where you can't use Python expressions to specify matching.  After all, these functions are relatively easy to implement as list comprehensions; the keyword_search functionality just generalises the search capability out of methods that are specific to the data (e.g. "all_xfs_filesystems()" as a made-up example).

We could take a callable as a separate matcher suffix, e.g.:

```
keyword_search(fs_vfstype__matchfn=lambda x: x in ('ext2', 'ext3', 'ext4'))
```

For those tests that aren't easy to encapsulate in one match.

Does that sound like a reasonable compromise?

bfahr:
@PaulWay if you will fix the merge conflict I can then go ahead and merge this PR.

insights/parsers/__init__.py
gavin-romig-koch: Yes.  Thank you.

PaulWay:
OK, should be good to go.

PR #328, opened 2017-08-14
Parser for `getcert list` for 1.x branch

PaulWay:
This parser handles the output of 'getcert list', including catching the (incomplete) IPA variant 'ipa-getcert list' stored in sosreports.

Waiting on merging of #326 for full search functionality but tests fine as is.


insights/parsers/getcert_list.py
```diff
@@ -0,0 +1,138 @@
...
+
+"""
+
+from insights.core import Parser
+from insights.parsers import ParseException, keyword_search
+from insights.core.plugins import parser
+
+
+@parser('getcert_list')
+class Cert_List(Parser):
```
bfahr: @PaulWay normally we would use just camelcase for class names, `CertList` instead of `Cert_List`.

insights/parsers/getcert_list.py
```diff
@@ -0,0 +1,138 @@
...
+    def parse_content(self, content):
+        """
+        We're only interested in lines that contain a ':'.  Special lines
+        start with 'Request ID' and 'Number of certificates...'; we handle
+        those separately.  All other lines are stripped of surrounding white
+        space and stored as a key-value pair against the last request ID.
+        """
+        self._data = {}
+        self.num_tracked = 0
+        self.requests = []
```
bfahr: You need to define the class attributes `num_tracked` and `requests`.

insights/parsers/getcert_list.py
```diff
@@ -0,0 +1,138 @@
...
+                track: yes
+                auto-renew: yes
+
+    Examples:
+
+        >>> certs = shared[Cert_List]
+        >>> certs.num_tracked  # number of certificates tracked from first line
+        2
+        >>> len(certs)  # number of requests stored - may be smaller than num_tracked
+        >>> certs.requests
```
bfahr: There should be an output for `len(certs)`.

insights/parsers/getcert_list.py
PaulWay: OK, sorry, kept typing the same style as the spec :-)

insights/parsers/getcert_list.py
PaulWay: Oops!  Thanks, Bob!

insights/parsers/getcert_list.py
PaulWay: Righto, thanks for that.

PaulWay:
OK, I've updated with the above improvements, let me know what you think.

PR #330, opened 2017-08-14
New parser to get info whether there are numeric user or group names.

jsvob:


insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,62 @@
+"""
+NumericUserGroupName - command ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
+=========================================================================================
+
+This parser reads the output of ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
+and detects whether it is 0 or nonzero.
+
+The grep command returns the number of matches. It matches for user and group names that
```
skontar: It actually does not check return value, I am not sure if it is obvious, that the command prints the statistics and we have no way how to check return values.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,62 @@
...
+begin with a number. If it returns 0, then there are no such names, otherwise it returns
+the number of matches (names). The parser only detects whether it is 0 or nonzero so as
+to minimize the amount of collected data.
+
+The boolean is available in the ``data`` attribute but it should not be accessed directly
+(encapsulation). The parser provides methods for comparisons with bool and casting to bool.
+
+Sample input::
+
+    3
```
skontar: I think this is wrong. Sample output of the command is:

```
/etc/passwd:1
/etc/group:1
```

Also you can store how many users and/or groups has name starting with number.

insights/config/specs.py
```diff
@@ -235,6 +235,7 @@
     "ntpq_leap"                 : CommandSpec("/usr/sbin/ntpq -c 'rv 0 leap'"),
     "ntpq_pn"                   : CommandSpec("/usr/sbin/ntpq -pn"),
     "ntptime"                   : CommandSpec("/usr/sbin/ntptime"),
+    "numeric_user_group_name"   : CommandSpec("/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group"),
```
bfahr: @jsvob you'll need to use `/bin/grep` to be able to support RHEL 6.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,62 @@
...
+from .. import Parser, parser
+
+
+@parser('numeric_user_group_name')
+class NumericUserGroupName(Parser):
+    """
+    Reports whether there is a user or group with a name that starts with a digit.
+    """
+    def parse_content(self, content):
+        self.data = not ("0" == ''.join(content).strip())
```
bfahr: It might be simpler for users of this class, if you renamed this attribute to something like `has_numeric_user_group` so that they could just access the attribute directly.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,62 @@
+"""
+NumericUserGroupName - command ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
+=========================================================================================
+
+This parser reads the output of ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
+and detects whether it is 0 or nonzero.
+
```
matysek: Explaining possible use of this parser would be useful. Please add here anything from the related bug report.

insights/parsers/numeric_user_group_name.py
jsvob: I got it wrong and will fix it.

insights/parsers/numeric_user_group_name.py
jsvob: I got it wrong and will fix it.

insights/config/specs.py
jsvob: Will fix it.

insights/parsers/numeric_user_group_name.py
jsvob: Will fix it.

insights/parsers/numeric_user_group_name.py
jsvob: Will fix it.

insights/parsers/numeric_user_group_name.py
```diff
@@ -3,26 +3,38 @@
...
-the number of matches (names). The parser only detects whether it is 0 or nonzero so as
-to minimize the amount of collected data.
+The grep command returns the number of matches for each file. It matches for user and
+group names that begin with a number. The parser only detects whether the sum is 0 or
+nonzero so as to minimize the amount of collected data.
 
-The boolean is available in the ``data`` attribute but it should not be accessed directly
-(encapsulation). The parser provides methods for comparisons with bool and casting to bool.
+The boolean answers are available through the getters ``has_numeric_user_group()``,
+``has_numeric_user()``, ``has_numeric_group()`` and ``bad_data()``.
```
skontar: Should not bad data return an Exception?

insights/parsers/numeric_user_group_name.py
```diff
@@ -3,26 +3,38 @@
...
 
-    >>> bool(shared[NumericUserGroupName])
+    >>> shared[NumericUserGroupName].bad_data()
+    False
+    >>> shared[NumericUserGroupName].has_numeric_user_or_group()
     True
-    >>> True == shared[NumericUserGroupName]
+    >>> shared[NumericUserGroupName].has_numeric_user()
     True
+    >>> shared[NumericUserGroupName].has_numeric_group()
```
skontar: Would not it be more useful to return the actual numbers so they can be displayed in front-end?

insights/parsers/numeric_user_group_name.py
```diff
@@ -3,26 +3,38 @@
...
+
+/etc/passwd:3
+/etc/group:0
 
 Examples:
 
-    >>> bool(shared[NumericUserGroupName])
+    >>> shared[NumericUserGroupName].bad_data()
+    False
+    >>> shared[NumericUserGroupName].has_numeric_user_or_group()
```
skontar: Why not just use attributes? Having getter and setter for everything is considered Java-like.

insights/parsers/numeric_user_group_name.py
jsvob: Where do you suggest to set the line? Throw an exception if at least one file is not available? Or throw an exception only when both files are unavailable?

insights/parsers/numeric_user_group_name.py
jsvob: The reason was to keep it encapsulated 1) because a getter cannot change the attribute, 2) in case we ever want to change the internal implementation. But you're right, this is so simple any refactoring in future will be simple as well.

insights/parsers/numeric_user_group_name.py
jsvob: I think it is not useful. The useful information is whether there is a numeric name. Then the rule & frontend can alert the cutomer to please review it (maybe even suggest to run that grep without the `-c` switch). Other than knowing there is this possible problem, there is little point in having additional data, and this is kind of sensitive.

skontar: OK, it is your call. But I think it may be interesting for statistics anyway.

insights/parsers/numeric_user_group_name.py
skontar: I think you are more concerned that grep command is not available than the files. I think that if you are not able to parse the command output, meaning it does not conform to your expectations, you should rise an exception.

insights/parsers/numeric_user_group_name.py
skontar: Python has mechanics for re-factoring attribute access to transparent setters and getters. See `@property` and `@x.setter` decorators.

If I were you, I would keep it simple and just create two attributes, `nr_numeric_users` and `nr_numeric_groups` (both integers), and then a method `has_numeric_user_or_group()` which would return bool.

jsvob: Right. I will make three attributes like you propose (the last can be an attribute as well, the data won't change during the lifetime of the parser).

insights/parsers/numeric_user_group_name.py
jsvob: On a second thought, I agree with you and will provide the numbers.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,68 @@
...
+=========================================================================================
+
+This parser reads the output of ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
+and detects whether there are any user or group names that begin with a digit.
+
+The grep command returns the number of matches for each file. It matches for user and
+group names that begin with a number. The parser only detects whether the sum is 0 or
+nonzero so as to minimize the amount of collected data.
+
+The boolean answers are available through the attributes ``has_numeric_user_group``,
```
xiangce: `nr_numeric_user` and `nr_numeric_group` are now not boolean variables, we'd better update the documentation correspondingly.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,68 @@
+"""
+NumericUserGroupName - command ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
+=========================================================================================
+
+This parser reads the output of ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
+and detects whether there are any user or group names that begin with a digit.
+
+The grep command returns the number of matches for each file. It matches for user and
+group names that begin with a number. The parser only detects whether the sum is 0 or
+nonzero so as to minimize the amount of collected data.
```
xiangce: It also returns the value of the `sum`.

insights/parsers/numeric_user_group_name.py
jsvob: Thank you for spotting it.

PR #331, opened 2017-08-14
New parser to get info whether there are numeric user or group names.

jsvob:


insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,62 @@
...
+        self.data = not ("0" == ''.join(content).strip())
+
+    def __nonzero__(self):
+        return self.data
+
+    def __bool__(self):
+        # For future compatibility with Python 3
+        return self.data
+
+    def __eq__(self, other):
```
PaulWay: Are you anticipating a lot of inequality comparisons with boolean data?

Your 'not' statement forces self.data to be of type bool.  These inequalities are therefore useless.  I'd advise removing them.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,62 @@
+"""
+NumericUserGroupName - command ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
+=========================================================================================
+
+This parser reads the output of ``/usr/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group``
```
PaulWay: This is interesting, but it never actually explains why this is of any relevance at all.

Why do we even need this?  How would it be used?  What problems does having a user or group name starting with a digit cause?

If there is a reason - and I know there is - then please add it.

In general I'm not hugely keen on having parsers that check for a complicated condition that's only useful to a single rule.  I understand the reason, but they add to the processing time and load on the client.

insights/parsers/tests/test_numeric_user_group_name.py
```diff
@@ -0,0 +1,64 @@
...
+0
+"""
+UNSTRIPPED_NONZERO = """
+
+3
+
+"""
+
+
+# Deliberately testing various ways of comparing the value, some of which are not liked by flake8.
```
PaulWay: I really don't see the point in these inequality tests.

PaulWay: And I also don't see you testing for invalid inputs, like '/bin/grep: command not found'.

skontar: Actually, I think we do not need any of these equality / inequality tests. I think a one method which returns number of users / groups starting with number is the only thing we need there.

insights/parsers/numeric_user_group_name.py
matysek: I agree that explaining possible use of this parser would be useful. Please add here anything from the related bug report.

jsvob: Will fix it.

In my opinion, this is a simple grep, a simple parser (and a simple rule), should be rather fast and lightweight parser & rule.

insights/parsers/numeric_user_group_name.py
jsvob: I'm removing all comparison operators in light of Bob's suggestion.

insights/parsers/tests/test_numeric_user_group_name.py
jsvob: The point was to make sure all conceivable ways of using the class instance as a bool value were covered so that there would be no hidden surprises. In the end, the class will not provide anything like this and will have getter methods.

insights/parsers/tests/test_numeric_user_group_name.py
```diff
@@ -0,0 +1,290 @@
...
+
+/bin/grep: No such file or directory
+
+"""
+
+# Should produce a valid result because it contains valid lines.
+# The last valid lines are used.
+# /etc/passwd:1
+# /etc/group:1
+ERRORS_6 = """
```
PaulWay: I'm really not convinced that we need a test of this kind of impossible output.  But whatever.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,90 @@
...
+used with utilities that accept both user/group IDs and names (like chown, chmod). In
+case such names exist on the system, it is advisable to review whether the used software
+(namely 3rd party software and shell scripts) handles the names correctly.
+
+https://access.redhat.com/solutions/3103631
+
+Sample input:
+
+/etc/passwd:3
+/etc/group:0
```
bfahr: This section is not rendering correctly in the docs, it should be changed to:
```
Sample input::

    /etc/passwd:3
    /etc/group:0
```

insights/parsers/tests/test_numeric_user_group_name.py
jsvob: You are right, I will fix it.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,90 @@
...
+                continue
+            if '/etc/passwd' == line_split[0]:
+                self._has_numeric_user = not zero
+            if '/etc/group' == line_split[0]:
+                self._has_numeric_group = not zero
+
+        # The grep command should report on both files unless a file is missing or something is very
+        # wrong.
+        # This is true if at least one of the files is missing.
+        self._bad_data = None is self._has_numeric_user or None is self._has_numeric_group
```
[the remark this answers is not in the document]
jsvob: You are right and I will change it to raise an exception if the input doesn't contain info for both files. If e.g. only one file is present on the machine, we could theoretically use it but that would be a rare scenario anyway.

insights/parsers/numeric_user_group_name.py
```diff
@@ -0,0 +1,90 @@
...
+    def bad_data(self):
+        return self._bad_data
+
+    def has_numeric_user_or_group(self):
+        return self._has_numeric_user_or_group
+
+    def has_numeric_user(self):
+        return self._has_numeric_user
+
+    def has_numeric_group(self):
```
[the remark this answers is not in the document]
jsvob: Agreed, will fix it.

Issue #300, opened 2017-08-03
CertificatesEnddate fails to parse filenames


miclark:
The `insights/parsers/certificates_enddate.py` contains the `CertificatesEnddate` parser.  This parser appears to have two issues.  First, it assumes the output of the `/usr/bin/openssl x509 -noout -enddate -in path/to/cert/file` command results in, amoung other text, tuplets of the form
```
notAfter=May 25 16:39:40 2019 GMT
FileName= /etc/origin/node/cert.pem  
```

However, it appear that in some customer archives, the `Filename=` prefix is missing:
```
notAfter=Apr 19 13:18:38 2033 GMT
/etc/pki/product/69.pem
notAfter=Mar 12 20:31:09 2036 GMT
/etc/pki/product-default/69.pem
```

Further, a few local tests on Fedora and RHEL 7 don't list a file path at all.

miclark:
Hey @JoySnow, assigning to you since you had put in most of the work on this parser.   Raise questions as necessary, of course.

JoySnow:
@miclark Thanks for the feedback. I'd love to enhance it ASAIC.

PR #334, opened 2017-08-15
Fix broken spec cmd of certificates_enddate.py

JoySnow:
Add "FileName= " as prefix into command.


JoySnow:
https://github.com/RedHatInsights/insights-core/issues/300

JoySnow:
Fix for branch 1.x at https://github.com/RedHatInsights/insights-core/pull/335 .

PR #336, opened 2017-08-15
Add SkipException for dirty parsers

bfahr:
* Add new exception to be thrown by dirty parsers as described in the docs
* Add exception to documentation


insights/parsers/__init__.py
```diff
@@ -15,6 +15,16 @@ class ParseException(Exception):
...
 
+class SkipException(Exception):
+    """
+    Exception that should be thrown from parsers that are explicitly
+    written to look for errors in input data.  If the expected error
+    is not found then the parser should throw this exception to
+    signal to the infrastructure that the parser's output should not be
+    retained.
+    """
+
```
xiangce: You might miss the `pass` here.

bfahr:
@xiangce thanks for catching that, I've also added a test for the exception to ensure that it is functioning properly.

PR #338, opened 2017-08-15
Literal dots 1x

kylape:
This was mentioned a while ago in Jakub's security audit, so I'm just shoring up some regexes to be a bit more specific


kylape:
Fail

PR #341, opened 2017-08-15
add in gpg sig to installation and rotation

brantleyr:


kylape:
Can you rebase against master?  Also, do you think we could factor out the common pieces between `fetch_core` and `fetch_core_sig`?

kylape:
Note that I don't think we have to do the refactoring in this PR, just curious if you think it's feasible.

brantleyr:
@kylape conflicts resolved, will investigate refactoring.

brantleyr:
This PR never happened.

PR #347, opened 2017-08-17
RHNConf documentation for 1.x branch

PaulWay:
Adding documentation and some code coverage test improvements for the RHNConf parser.


PaulWay:
BTW, what's the correct syntax to do a reference in Sphinx to the 'LegacyItemAccess' class definition?

xiangce:
~~~
:class:`insights.core.LegacyItemAccess`
~~~
works

insights/parsers/rhn_conf.py
```diff
@@ -1,20 +1,32 @@
...
 @parser('rhn.conf')
 class RHNConf(LegacyItemAccess, Parser):
-    """Class to parse the configuration file ``rhn.conf``
+    """
+    Class to parse the configuration file ``rhn.conf``.
+
+    The special feature of ``rhn.conf`` is that values can span multiple
+    lines with each intermediate line ending with a comma.
+
+    This parser uses the :py:class::LegacyItemAccess mix-in to provide access to
```
bfahr: You can use the syntax recommended by @xiangce.  Here's the Sphinx reference: http://www.sphinx-doc.org/en/stable/domains.html#cross-referencing-python-objects

PaulWay: OK, great, thanks @xiangce and @bfahr - will use that in the future!  Should be ready to go.

PR #348, opened 2017-08-17
Enhance unsplit_lines to allow keeping continuation character for 1.x branch

PaulWay:
Some continuation formats use the 'line continuation' character as a 'list item separator' character as well.  For these formats, its useful to keep the continuation character on the end of the line.  This update adds a `keep_cont_char` option, defaulting to False.  If set to True, the line continuation character will be left on the end of the line.

Tests included, code coverage is 100%.


insights/parsers/__init__.py
```diff
@@ -165,9 +165,15 @@ def unsplit_lines(lines, cont_char='\\'):
...
+
     Parameters:
         lines (list): List of strings to be evaluated.
         cont_char (char): Char to search for at end of line. Default is ``\\``.
+        keep_cont_char (bool): Whether to keep the continuation on the end of
+            the line.  Defaults to False, which causes the continuation
+            character to be removed.
 
     Yields:
         line (str): Yields unsplit lines
```
xiangce: Better add an example for the new feature.

PaulWay:
OK, I've added documentation for that and `cont_char` usage - should be good to go.

PR #350, opened 2017-08-17
Remove filter in nfs_exports.py

zhangyi733:


xiangce:
It's necessary to remove the `filters` because of some plugins, such as load_nfsmod_failed, need to check if their content is empty.

PR #333, opened 2017-08-15
Enhance config systemd 1x

wushiqinlou:
The former version conflates multi-line strings (those with line continuation characters) and the presence of a key multiple times (like multiple After= lines) in a way that makes it impossible to distinguish the two situations. This PR will return multiple as list.

PR #353, opened 2017-08-18
Enhance config systemd

wushiqinlou:
- The former version conflates multi-line strings (those with line continuation characters) and the presence of a key multiple times (like multiple After= lines) in a way that makes it impossible to distinguish the two situations. This PR will return multiple as list.


insights/parsers/systemd/config.py
```diff
@@ -5,10 +5,11 @@
 ``config``.
 """
 
-from insights.contrib.ConfigParser import RawConfigParser as cp
```
xiangce: Please revert to from "insights.contrib.ConfigParser import RawConfigParser as cp"

xiangce:
See #333

PR #356, opened 2017-08-21
Removing deprecated parsers for 1.x branch

PaulWay:
The `cpuinfo_max_freq` and `OracleConfig` parsers are deprecated and no longer in use.  They should be removed.


PaulWay:
OK - can I deprecate it with #361?

bfahr:
@PaulWay rather than add this to #361 it would be better to revise this PR once #361 has been merged.

PaulWay:
@bfahr Sorry, yes, that's what I meant - once #361 is approved, use the `@deprecated` decorator to mark this as deprecated.

xiangce:
Hey @PaulWay - How is it going? Is there any chance to update this PR as per your newly added "deprecated" feature?

xiangce:
It's fixed in #732 and #733 (#823)

PR #357, opened 2017-08-21
Add specs for pam.conf and oc_get_pvc for 1.x branch

PaulWay:
The `pam.conf` and `oc_get_pvc` specs were missing and have been added.


insights/config/specs.py
```diff
@@ -419,6 +420,7 @@
     "oc_get_project"            : CommandSpec("/usr/bin/oc get project -o yaml --all-namespaces"),
     "oc_get_role"               : CommandSpec("/usr/bin/oc get role -o yaml --all-namespaces"),
     "oc_get_pv"                 : CommandSpec("/usr/bin/oc get pv -o yaml --all-namespaces"),
+    "oc_get_pvc"                : CommandSpec("/usr/bin/oc get pvc -o yaml --all-namespaces"),
```
bfahr: @PaulWay will this result in the `oc_get_pv` spec being deprecated?

PaulWay: Not AFAIK - they're different commands.

PR #358, opened 2017-08-21
Added keyword search to PAM configuration parser

PaulWay:
* Add a keyword search method to PAM configuration objects.  This maps each object's properties over to dictionary keys, with the 'control_flags' key being a key/value dictionary instead of a list.  This makes it possible to search by keyword/value arguments using the regular `keyword_search` helper function, but it returns the original PAMConfEntry objects rather than the dictionary.
* Improved documentation, including documentation for search method.
* Full test coverage of new functionality.


insights/parsers/pam.py
```diff
@@ -288,6 +312,39 @@ def __len__(self):
...
+        the :func:`insights.parsers.keyword_search` function - see its
+        documentation for more information.
+
+        Searching on the list of PAM configuration entries is exactly like
+        they were dictionaries instead of objects with properties.  In
+        addition, the 'control_flags' property becomes a dictionary of
+        keywords and values, so that 'control_flags__contains' allows
+        searching for a particular control flag.
+
+        Yields:
```
bfahr: @PaulWay I believe this should be `Returns` instead of `Yields` shouldn't it?

insights/parsers/pam.py
```diff
@@ -108,8 +108,8 @@ class PamSSHD(PamDConf):
 """
 
 from collections import namedtuple
-from .. import Parser, get_active_lines, parser
-from ..parsers import unsplit_lines
+from insights import Parser, get_active_lines, parser
+from insights.parsers import unsplit_lines, keyword_search
```
bfahr: These links should be left as relative to be consistent with the code base.  However, the second line could be simplified by:
```
from . import unsplit_lines, keyword_search
```

PaulWay: It seems to me that that's not "consistent" at all - almost all parser tests, for example, use the absolute import path, and several parsers use absolute paths as well.  Since we're no longer in doubt about the project name, it seems to me that there's no need for relative paths, and if we're going to 'tidy up' to be consistent I'd personally rather everyone use absolute paths.

Where do we stand on this generally?

bfahr: You are correct that the tests are not consistent with use of relative paths (although they should be), but the majority if not all parsers and combiners do use relative paths.

jhjaggars: Hey folks,

I don't have a *super* strong opinion here, but I generally think that absolute imports are good when you are importing something outside *your* package and relative imports are good when importing things from within your package.  This allows you to rename or relocate a package with the least number of mindless renaming tasks.  The downside is that if you want to extract a subset of a package to another package you may have to fiddle with the imports.

PaulWay: OK, personally I'd prefer absolute imports but I'll go with relative as that does seem to be the standard for parsers.

PR #359, opened 2017-08-21
add image/container flag and remove timestamp for hosts

brantleyr:


insights/client/__init__.py
```diff
@@ -339,10 +337,9 @@ def collect(self, **kwargs):
             config['container_mode'] = True
             config['mountpoint'] = kwargs.get('mountpoint')
 
-        # If check_timestamp is not flagged, then skip this check AND
-        # we are also scanning a host
+        # If check_timestamp is not flagged
         # bypass timestamp checks for other cases
-        if bool(kwargs.get('check_timestamp', True)) and scanning_host:
+        if bool(kwargs.get('check_timestamp', True)):
```
kylape: You can remove the cast to `bool` since it's implicit in the `if` statement.

insights/client/config.py
```diff
@@ -24,6 +24,7 @@
 ]
 
 CONFIG = {
+    'analyze_image': None,
```
kylape: If this is a boolean, should it default to `False`?

insights/client/utilities.py
```diff
@@ -235,7 +242,10 @@ def magic_plan_b(filename):
 
 def generate_container_id(container_name):
     # container id is a uuid in the namespace of the machine
-    return str(uuid.uuid5(uuid.UUID(generate_machine_id()), container_name.encode('utf8')))
```
kylape: If this function is only called when `analyze_image` or `container_mode` are `True`, then why not always return `container_name.encode('utf8')`?  It seems like `generate_machine_id` would always return `None` in this case, so no need to do any of that.

insights/client/config.py
brantleyr: Yes, revising now.

insights/client/__init__.py
brantleyr: Can do coach.

insights/client/utilities.py
brantleyr: It used to be that container IDs were based on their host machine, so if we don't need to see the Container IDs then the call to "generate_machine_id()" can be removed. So either we need to re-think the "generate_machine_id()" functionality or just remove the call to that function here. Note that "generate_machine_id()" did not always return "None" when "container_mode" or "analyze_image" was True.

kylape: IIRC per our discussion earlier this week, we do want containers specifically to take the host into consideration when generating IDs but not images.  Does the `analyze_image` option get used when scanning containers?

kylape:
Merged manually

PR #361, opened 2017-08-23
Add 'deprecated' decorator to log use of deprecated parsers

PaulWay:
This decorator can be used on any parser or combiner, whether a function or a class.  It produces a warning in the logs - we all check the logs, right? - on the use or instantiation of this parser.

One option here for discussion would be to detect if this was being run from the insights-cli or py.test, and warn to stderr.


gavin-romig-koch:
While checking for the use of deprecated code is a good idea, the warning/error messages about them should be signaled during development before merge to master, not at run time in the log file.  At run time there is nothing most users can do about the message except try to ignore it, and so for most users, the message is just additional clutter in the log file.  For me, this issues isn't a bad enough problem that i want to clutter up the log file to solve it.

But i assume it wouldn't be too hard to write a flake8 plugin that checks this same thing which would get the message to the right people at the right time.

csams:
I think we should leave flake8 to style enforcement, but I agree that this could clutter the log file with messages a customer mostly could do nothing about, unless their own plugins are the culprits. Even then, this seems like the kind of warning that should be pretty visible during development/testing but dropped to debug logging at runtime.

bfahr:
@PaulWay have you considered implementing [Python warnings](https://docs.python.org/2/library/warnings.html) to trigger the notice in development/test?

gavin-romig-koch:
@csams why limit flake8 to just style enforcement.  Why not use it to check for any issue we all agree can and should be checked before commit to master.

@bfahr are Python Warnings available in 2.6?

PaulWay:
@gavin-romig-koch the doco for Warnings at https://docs.python.org/2/library/warnings.html says 'new in version 2.1', so they should be supported unless that's a completely different version number.

@bfahr Warnings are a good way to bring this up, and I think they address @csams' point about not disturbing a customer that's using 1.x with deprecated parsers.  Let me have a go at an update for this.

gavin-romig-koch:
Warnings, might resolve my concerns, and may be easier to implement than a flake8 plugin.   As I say above, for users of insights, this message is at best useless clutter that obscures the more important messages in the log.  Imagine "E231 missing whitespace after comma'" appeared while running yum.  This message needs to be delivered to developer's  who can do something about it

I suggested a flake8 plugin because insights-core already uses flake8 as a gating test before check in.   If you choose to implement this with Warning, the following is moot.  The flake8 plugin would not contain a list of deprecated functions. Instead it would look for calls to functions with a docstring containing the warning/deprecated markup, or alternately calls to functions with the 'deprecated' decorator, or some other way of marking such functions.  It would be "part" of insights-core, either directly or as a dependency, pip installing it when we install flake8.

jhjaggars:
Hey Folks,

I think it'd be good enough if we made it obvious to rule developers that they were using deprecated code at unit test time (via warnings in Exception mode perhaps).  I think we probably want to mute warnings when we run stuff in production because of the reasons that @gavin-romig-koch has outlined, so maybe an update to the docs on how to take advantage of the warnings at development time to help discover use of deprecated code is all that is needed here.

PaulWay:
@jhjaggars Is it possible to set that 'python -We' mode in py.test?  That'd probably be ideal...

PaulWay:
Actually, we can set warnings-to-error mode on DeprecationWarnings using `warning.simplefilter`.  I've also added a 'pending' flag as an idea, for using the PendingDeprecationWarning.  It's exactly the same result, so I'm not sure if that's at all useful - what do you think?

insights/core/plugins.py
```diff
@@ -190,6 +192,28 @@ def register_cluster_reducer(func):
     p["module"] = func.__module__
 
 
+def deprecated(reason='', pending=True):
```
jhjaggars: My gut tells me that `pending=False` will be more common.

insights/core/plugins.py
```diff
@@ -190,6 +192,28 @@ def register_cluster_reducer(func):
...
+    """
+    Mark a parser or combiner as deprecated.  This will emit a warning in
+    the logs when the function is used.
+    """
+    warning_exception = PendingDeprecationWarning if pending else DeprecationWarning
+    # If running under py.test, DeprecationWarnings become errors.
+    if 'py.test' in sys.argv[0]:
```
jhjaggars: I think using conftest.py to modify state would be preferable.  Won't the `"Warnings enabled"` message print once per usage this way?

insights/core/plugins.py
PaulWay: Yeah, let's just use DeprecationException throughout.  It was a silly idea.

jhjaggars: We should probably require a `reason` to be set with some information on what the user should use instead.  Requiring that parameter will simplify the message later in this function as well.

insights/core/plugins.py
jhjaggars: To be more clear, if in conftest.py you simply `import warnings` and call `warnings.simplefilter()` that will flip the bits necessary when py.test runs.

PaulWay: Got it, perfect.

Mind you, we now have this to deal with:
```
=========================== short test summary info ============================
FAIL insights/parsers/tests/test_grub_conf.py::test_grub_conf_raise
FAIL insights/parsers/tests/test_installed_rpms.py::test_release_compare

=================================== FAILURES ===================================
_____________________________ test_grub_conf_raise _____________________________

    def test_grub_conf_raise():
        with pytest.raises(Exception) as e:
            Grub2Config(context_wrap(GRUB2_CFG_4))
>       assert e.value.message == "Cannot parse menuentry line: menuentry {"
E       DeprecationWarning: BaseException.message has been deprecated as of Python 2.6

insights/parsers/tests/test_grub_conf.py:204: DeprecationWarning
_____________________________ test_release_compare _____________________________

    def test_release_compare():
        rpm1 = InstalledRpm.from_package('kernel-rt-debug-3.10.0-327.rt56.204.el7_2.1')
        rpm2 = InstalledRpm.from_package('kernel-rt-debug-3.10.0-327.rt56.204.el7_2.2')
        rpm3 = InstalledRpm.from_package('kernel-3.10.0-327.10.1.el7')
        rpm4 = InstalledRpm.from_package('kernel-3.10.0-327.el7')
        rpm5 = InstalledRpm.from_package('kernel-3.10.0-327.el7_1')
        rpm6 = InstalledRpm.from_package('kernel-3.10.0-327')
        rpm7 = InstalledRpm.from_package('kernel-3.10.0-327.1')
        rpm8 = InstalledRpm.from_package('kernel-3.10.0-327.x86_64')
        rpm9 = InstalledRpm.from_package('kernel-3.10.0-327.1.x86_64')
        with pytest.raises(ValueError) as ve:
            rpm4 < rpm6
>       assert "the other does not" in ve.value.message
E       DeprecationWarning: BaseException.message has been deprecated as of Python 2.6

insights/parsers/tests/test_installed_rpms.py:196: DeprecationWarning
===================== 2 failed, 565 passed in 5.75 seconds =====================
```

insights/core/plugins.py
PaulWay: Good point - I'll call it 'solution' instead of 'reason'.

insights/core/plugins.py
PaulWay: So I've also fixed those two deprecations.  Hope that's OK :-)

insights/core/plugins.py
```diff
@@ -190,6 +191,29 @@ def register_cluster_reducer(func):
     p["module"] = func.__module__
 
 
+def deprecated(solution):
```
bfahr: @PaulWay how have you tested this?  I tried it with a class method and it appears to create an error for the entire class regardless of whether the method is used.

PaulWay: @bfahr I tested this initially using the deprecated oracle parser functions from #356.  Do you think I need to put in tests for this?  That'd be a good idea, I suppose...

If we deprecated a class, then I think using the entire class should generate a deprecation error.  What do you think should happen?

bfahr: @PaulWay sorry I wasn't clear, I wasn't necessarily advocating adding tests, I was just wondering how you tested.  If a method in a class is deprecated then it seems that whether or not the method is used, if the class is used then it generates the error.  Is there a way to deprecate a class method and not have an error raised unless the class method is used?

PaulWay: Hi @bfahr - ah, so you did something like this:

```python
class FlangParser(Parser):
    @deprecated("use new `search` method with key=value parameters")
    def old_search(self, name, mode, cpus, sequence, escalation, paradigm):
        ...
```
And the whole class got flagged?  I didn't test that.  Let me see what I can do.

PaulWay: OK, one problem here is that the method I used above causes the mere existence of a deprecated function to cause an error.  Even if that function doesn't get used, the function's existence in the code causes the error.  What I want is for a warning to be generated only when the function gets used.  So I've rewritten the decorator as:

```python
    def _log_deprecated(func):
        message = "Warning: {mod}.{fn} has been deprecated: {sol}".format(
            mod=func.__module__, fn=func.__name__, sol=solution
        )

        def function_with_deprecation_warning(*args, **kwargs):
            warnings.warn(message, DeprecationWarning)
            return func(*args, **kwargs)

        return function_with_deprecation_warning

    return _log_deprecated
```

I'm fine-tuning its output for functions, classes and methods, but this AFAICS works better than deprecating the existence of the function.  What do you think, @bfahr and @jhjaggars ?

PaulWay:
I'm having problems with getting it to recognise a class object method when deciding on the deprecation message, but otherwise this seems to work.  What do you think so far?

insights/core/plugins.py
bfahr: @PaulWay this now seems to trigger on the method, but it still triggers when the method is not being used.  I just added a dummy method to test that.  Did I miss something?

PaulWay:
@bfahr can you share your code?  I'm using:
```python
class ClassWithDeprecatedMethod(Parser):
        pass

    @plugins.deprecated("use new_search method instead")
    def old_search(self, mode, function, execute):
        return "nothing found"

    def new_search(self, **kwargs):
        return "found something"


def test_deprecated_class_method():
    # Make sure class can be instantiated without warning...
    c = ClassWithDeprecatedMethod(context_wrap("CONTENT"))
    assert c
    # And that its new search method doesn't trigger the warning
    assert c.new_search(mode='parametric') == "found something"

    with warnings.catch_warnings(record=True) as w:
        # Make sure old function can be called...
        assert c.old_search("parametric", "system", "libration") == "nothing found"

        # But use is logged as a warning:
        assert len(w) == 1
        assert issubclass(w[-1].category, DeprecationWarning)
        assert "use new_search method instead" in str(w[-1].message)
        # Note that (AFAICS) due to the deprecator decorator, the function
        # actually changes here, but the result is what we want:
        assert "Warning: test_plugins ClassWithDeprecatedMethod.old_search has been deprecated" in str(w[-1].message)
```
And the only thing that's going wrong is that the code seems to think `old_search` is a function, not a class instance method.

csams:
@PaulWay has the updated deprecation decorator been pushed to the branch for the PR?

PaulWay:
@csams Sorry, have just done that, though the tests will fail.

It looks like I can't achieve what I wanted to do, which was to say whether the thing being deprecated was a function, an entire class, or the method from a class.  The former two work, but the latter doesn't - for some reason Python's decorator syntax turns the method into a function before handing it to the decorating code, so `inspect.ismethod()` is True outside the decorator and False inside.

I can live with just saying "here's the thing being deprecated and it was in this file", if that's the only solution.  Any suggestions on how to proceed here?

csams:
@PaulWay We can't decorate the class definitions themselves because the decorator will substitute a function for the class, and that'll cause invocations of `@staticmethod` and `@classmethod` methods to fail. We can decorate `__init__` so we get a warning for the entire class, though this will force all deprecated parsers to have an `__init__` that delegates to a superclass. I don't think that's terrible, but it's also not very useful because defined Parsers will always be constructed if their dependencies are met regardless of what depends on them.

The `inspect.ismethod` check fails because at the point the function and decorator are evaluated in the class body, they're just simple functions. In fact, the thing at the bottom is _always_ a simple function, even when it's attached to a class:
```python
In [1]: class A(object):
   ...:     def my_func(self):
   ...:         pass
   ...:     

In [2]: type(A.__dict__["my_func"])
Out[2]: function

In [3]: type(A.my_func)
Out[3]: instancemethod
```
It's only by accessing it through the dot operator of the class or object instance that you get the special sauce of `instancemethod` (or `classmethod` or `staticmethod`). You can really go down a [rabbit hole](https://docs.python.org/2.7/howto/descriptor.html#functions-and-methods) with that stuff. :smiley:

Maybe try something like this?
```python
def get_name_line(src):                                                         
    for line in src:                                                            
        if "@" not in line:                                                     
            return line.strip()                                                 
                                                                                
def deprecated(solution):                                                       
    def decorator(component):                                                   
        done = [False]                                                          
        msg = "<{c}> at {p}:{l} is deprecated: {s}"                             
                                                                                
        @six.wraps(component)                                                   
        def inner(*args, **kwargs):                                             
            # guard so we don't do all this work every time the thing is called 
            if not done[0]:                                                     
                path = inspect.getsourcefile(component)                         
                src, line_no = inspect.getsourcelines(component)                
                name = get_name_line(src) or "Unknown"                          
                the_msg = msg.format(c=name, p=path, l=line_no, s=solution)        
                warnings.warn(the_msg, DeprecationWarning)                      
                done[0] = True                                                  
            return component(*args, **kwargs)                                   
        return inner                                                            
    return decorator
```

PaulWay:
OK, yeah, I hadn't thought through class decoration.  Can we make a subclass of the original class in a closure with just its init method generating the deprecation warning?  Let me play with that idea.

Also, I don't understand why `done` is a list there, and not just a variable.  Or why all those variables aren't just declared inside `decorator` rather than being evaluated inside the wrapper - `decorator` can see `component` there...  Let me play around with variations to learn what I don't understand there :-)

csams:
You can't assign to variables in an enclosing scope from within a closure, at least not in 2.7. The list is common a trick to get around that.

You could move the other variables used to create the final msg into the enclosing scope. It's probably cleaner that way since msg gets fully constructed all in one place.

PaulWay:
@csams Let me know what you think of it now :-)

jhjaggars:
I'd like to take a step back and ask a few fundamental questions.

1. What's the expected workflow for a developer with this feature?
2. Is the complexity of this change worth the benefit? Would it be so much worse to just add code to deprecated functions and classes that just logs that they are being used?

PaulWay:
@jhjaggars good points!

IMO the workflow is:

1) a parser developer adds something that deprecates a previous function, class or method.  They update the deprecated thing with the `@deprecated` decorator and make sure all tests are now using the new functionality.  The parser developer liaises with the rule developers about the impact of this deprecation.
2) when rule developer update insights-core, rules using the deprecated functionality now cause the test framework to generate a fatal error.  A rule developer is assigned to update the affected rules to not use the deprecated functionality.
3) at some point, the deprecated functionality is then removed.

Step 2 is where I see the main problem - if rule developers update the core and then suddenly they can't do any more development, that'll annoy a few people.

I suppose that was the step at which I thought it'd be useful to have the pending deprecation warning, which would turn into an 'always warn' or 'warn once' message when running `py.test` but would otherwise allow the tests to complete.  The process in that case would be:

1) Parser developer puts something into pending deprecation (e.g. `@deprecated(pending=True)`).
2) Rule developers have time to fix rules without being blocked in new rule development.
3) Some time later, parser developer removes pending status, making the thing deprecated.
4) Rule developers now have to fix rules, but back end functionality still works.
5) Parser developer removes deprecated functionality.
6) Much rejoicing.

How do you see the process going?

PaulWay:
@jhjaggars Re your second point about benefits and just logging, IMO we have two problems:

A) We want to deprecate functionality over time as we improve our methods
B) We have no control over the people who use the functionality we've supplied.

We do have our main internal body of rules, but we're already looking at having multiple repositories so it will get increasingly difficult to inform everyone of the deprecation.  And as we progressively get more people outside Red Hat using rules, any chance of checking who's using a feature we want to deprecate becomes increasingly slim.

For our own purposes, I think we need a formalised way of deprecating things, especially in parsers but even more generally.  Providing something that gives the programmer a clear message of what is deprecated and what they should use to replace it is better than just chopping something out and irritating the developer.  We know that people tend to ignore warning messages, especially if they start seeing them too often, so we want at some point for the warnings to turn into fatal errors that force the developer to address the issue - but still give them a clear warning message rather than just a 'symbol not found' message.

How do we know when we can safely remove the deprecated feature?  Well, we never know - and if Red Hat's experience with RHEL is anything to go by, it doesn't matter that we've told customers that RHEL 4 is end of life, some are still sticking there using it.  Faster moving projects like Django put pressure on the developers to keep their code up to date, but some (like myself, to my shame) avoid this problem by staying on old versions of Django.  The best we can do IMO is to deprecate things, make it obvious that we're doing so, inform the rule developer on how to make their code better, and mainly look at our own code base to see that we're doing it right.

HTH,

Paul

csams:
@PaulWay @jhjaggars Instead of
```python
@deprecated("Thing is Busted!")
def thing(foo):
    pass
```
we could do
```python
def thing(foo):
    deprecated(thing, "Thing is busted.")
    pass

# or for class level:

class A(object):
    def __init__(self):
        deprecated(A, "A is busted.")
        pass
```

That'll remove the need for special handling of classes, etc. in `deprecated`.

PaulWay:
@csams yeah... I guess... :-)

The whole purpose of a decorator is that it gets the function as an argument, removing the need for the caller to supply the function or class object inside the function or method.  But honestly I don't mind - if you and @jhjaggars think that's a better approach, let's do it that way.

PaulWay:
@csams Could we maybe just try this out and see how it goes?  Then modify it if we find problems?

jhjaggars:
@PaulWay sorry for neglecting this PR for so long.  I am definitely a fan of what @csams proposed despite having to pass the object in to the deprecated function.  It simplifies the library code and it's not too bad to have to pass yourself into a function at deprecation time.

PaulWay:
OK, let's do it that way.  One moment...

PaulWay:
@jhjaggars So what do you think of the 'pending' idea?  As I outlined above, I think it may be a good idea to mark things as pending deprecation so they simply alert the user when running testing, before the pending flag is turned off and the warning becomes a fatal error.  This is important so that people using deprecated code still see the message about how to solve the deprecation, rather than just finding a missing piece of code.

Please like this code! :-)

Have fun,

Paul

PaulWay:
Hi @bfahr - I've added some documentation, hopefully in the right place.  Let me know where you'd like me to move it to if not.

docs/api.rst
```diff
@@ -227,3 +227,87 @@ Run a single single unit test one can:
...
+   because of the tests failing.  Anyone receiving a warning about pending
+   deprecation MUST fix their code so that it no longer warns of deprecation.
+3. Once sufficient time has elapsed, the outdated feature is removed from the
+   code.  Any code still using this feature will fail its tests because the
+   feature it requires is absent.
+
+The usual time between each step should be two minor versions of the Insights
+core.
+
+To deprecate code, call the :py:func:``insights.util.deprecated`` function from
```
bfahr: @PaulWay I believe you need to use single back ticks here to get Sphinx to interpret it as a link.

docs/api.rst
```diff
@@ -227,3 +227,87 @@ Run a single single unit test one can:
...
+
+    from insights.util import deprecated
+
+    class ThingParser(Parser):
+        def __init__(self, *args, **kwargs):
+            deprecated(ThingParser, "Use the new_feature() function instead", pending=True)
+            super(ThingParser, self).__init__(*args, **kwargs)
+        ...
+
+The :py:func:``insights.util.deprecated`` function takes three arguments:
```
bfahr: Same comment as above, use single back ticks instead of double.

xiangce:
@PaulWay - Would you please re-submit it to the master branch? Thanks.

PaulWay:
Hi @xiangce - #726 is now submitted (with its own problems, which we can resolve separately).

PR #366, opened 2017-08-24
Added keyword search to PAM configuration parser for master branch

PaulWay:
* Add a keyword search method to PAM configuration objects.  This maps each object's properties over to dictionary keys, with the 'control_flags' key being a key/value dictionary instead of a list.  This makes it possible to search by keyword/value arguments using the regular `keyword_search` helper function, but it returns the original PAMConfEntry objects rather than the dictionary.
* Improved documentation, including documentation for search method.
* Full test coverage of new functionality.

This is the master branch version of #358


insights/parsers/pam.py
```diff
@@ -288,6 +312,39 @@ def __len__(self):
...
+        prop_keys = ('service', 'interface', 'module_name', 'module_args')
+        search = []
+        # Can't find a neat way to do this as a comprehension, so it's back
+        # to loops.
+        for obj in self.data:
+            row = {'entry_obj': obj}
+            for key in prop_keys:
+                row[key] = getattr(obj, key)
+            # Convert control_flags down to key/value dictionary
+            row['control_flags'] = {cf.flag: cf.value for cf in obj.control_flags}
```
bfahr: @PaulWay this dict comprehension will need to be changed to be compatible with Python 2.6.  One solution would be to instead use:
```
row['control_flags'] = dict((cf.flag, cf.value) for cf in obj.control_flags)
```

PaulWay: OK, will do.

PR #369, opened 2017-08-25
make --to-json work

brantleyr:


insights/client/__init__.py
```diff
@@ -523,4 +523,10 @@ def collect():
...
 def upload():
     egg_path = sys.stdin.read().strip()
-    run("upload", egg_path)
+    if not config["to_json"]:
+        logger.debug("Flag --to-json not used. Running upload.")
+        run("upload", egg_path)
+    else:
+        logger.debug("Flag --to-json used. Printing upload results.")
+        print "INIT: %s" % (str(run("upload", egg_path)))
+        return
```
kylape: do we need to return at the end of the function?

insights/client/__init__.py
```diff
@@ -523,4 +523,10 @@ def collect():
 
 def upload():
     egg_path = sys.stdin.read().strip()
-    run("upload", egg_path)
+    if not config["to_json"]:
```
kylape: can we invert the `if` to avoid negative boolean logic?

insights/client/__init__.py
```diff
@@ -344,11 +344,11 @@ def collect(self, **kwargs):
             logger.debug('Check timestamp is True.')
             cached_results = self._cached_results()
             if cached_results:
-                logger.info("Using cached collection: %s", cached_results)
+                logger.debug("Using cached collection: %s", cached_results)
```
kylape: I think we were wanting the fact that a cached collection was being used to be info level because customers might care more about that.

insights/client/__init__.py
```diff
@@ -344,11 +344,11 @@ def collect(self, **kwargs):
...
                 # if to_stdout is flagged then do sys.stdout
                 if config["to_stdout"]:
-                    logger.info("Flag --to-stdout used, using system out.")
-                    logger.info(sys.stdout)
+                    logger.debug("Flag --to-stdout used, using system out.")
+                    logger.debug(sys.stdout)
```
kylape: Shouldn't this use the `INFO: ` thing rather than `logger.debug`?

insights/client/__init__.py
brantleyr: Correct. This is related to changing the way the logging output works for debugs vs infos (when in --verbose mode). This was changed this way so that --to-json and --to-stdout would work as intended without additional random logging occurring. Logging will have to be fixed. I will revert these to .infos()

insights/client/__init__.py
brantleyr: No, this was just added in to test how the wrapper RPM was passing 'sys' to the Core to ensure we still had a correct file pointer for shutil to output the stdout for the .tar.gz file (when using --to-stdout).

insights/client/__init__.py
brantleyr: Can do coach.

insights/client/__init__.py
brantleyr: Maybe? I was just replicating other instances where we simply 'return' back to the wrapper RPM.

insights/client/__init__.py
kylape: So should it be removed now?

brantleyr: It hath been removed.

PR #374, opened 2017-08-27
Added `strip_quotes` option to optlist_to_dict for 1.x branch

PaulWay:
Added a `strip_quotes` option to the `optlist_to_dict()` function, to turn:
```
key='value with spaces',source=yes
```

into

```python
{'key': 'value with spaces', 'source': 'yes'}
```

Rather than including the quotes in the first value.  Quotes are not removed from the inside of the string and mismatched quotes are not removed.


xiangce:
@PaulWay - Your update looks great to me, but it causes the unit test of `storage_netfs_without_netdev.py` failed (caused by the different sort result between `{}` and `dict()`).

PaulWay:
Well, dictionaries are by their nature unordered.  Python does not guarantee the order of items in a dictionary.  Python may change its ordering based on Python version or other factors.  The tests still fail if I change it from a dict constructor on a list of tuples to a dict comprehension on a tuple generator function.

It seems to me here that the mistake is in the `storage_netfs_without_netdev` rule in trying to reconstruct the options list from the dictionary.  I've tested this fix here:

```diff
diff --git a/telemetry/rules/plugins/storage/storage_netfs_without_netdev.py b/telemetry/rules/plugins/storage/storage_netfs_without_netdev.py
index 14856ed..d5247d8 100644
--- a/telemetry/rules/plugins/storage/storage_netfs_without_netdev.py
+++ b/telemetry/rules/plugins/storage/storage_netfs_without_netdev.py
@@ -28,13 +28,7 @@ def _get_vg_lv_name(name):
 
 
 def _add_fstab_entry(fstab_entries, fstab_return_clauses, fstab_mount):
-    fs_mntops_list = []
-    for k, v in fstab_mount.fs_mntops:
-        if v is True:
-            fs_mntops_list.append(k)
-        else:
-            fs_mntops_list.append("{0}={1}".format(k, v))
-    fs_mntops = ",".join(fs_mntops_list) + ",_netdev"
+    fs_mntops = fstab_mount.raw.split(None)[3] + ",_netdev"
     fstab_entries.append(fstab_mount.fs_spec)
     fstab_return_clauses.append(" ".join(
         [fstab_mount.fs_spec, fstab_mount.fs_file, fstab_mount.fs_vfstype, fs_mntops, str(fstab_mount.fs_freq),
```

But the tests then need to be modified to list the options in the order originally given, not the order the dictionary returns them.  I'll submit an update for that rule shortly for your review.

Hope this helps,

Paul

xiangce:
@PaulWay - Great, you made the best solution I think :)

xiangce:
@PaulWay - will merge it after it's sync to the master branch.

PR #375, opened 2017-08-28
More debugging on run_parsers for 1.x branch

PaulWay:
Add more debugging information to show when parsers are being run, and what
they have returned.

Also test that response is not None - some responses (e.g. a zero-length list) may
still be valid outputs and need to be acted on.


PaulWay:
(BTW, there's no master branch version of this MR because the parser subsystem is being substantially reworked).

PR #377, opened 2017-08-28
Enhance specs docker inspect

wushiqinlou:
- The {DOCKER_CONTAINER_NAME} expansion isn't in the 1.x client. So it doesn't look like any of those docker rules have ever worked correctly on client systems. Use a pre_commands instead since they work with the 1.x client.


insights/config/specs.py
```diff
@@ -313,7 +314,9 @@
     "rhsm.conf"                 : SimpleFileSpec("etc/rhsm/rhsm.conf"),
     "rhsm.log"                  : SimpleFileSpec("var/log/rhsm/rhsm.log", large_content=True),
     "root_crontab"              : First([CommandSpec("/usr/bin/crontab -l -u root"),
-                                    SimpleFileSpec("sos_commands/crontab")]),
+                                    SimpleFileSpec("sos_commands/crontab"),
+                                    SimpleFileSpec("sos_commands/cron/root_crontab"),
+                                    SimpleFileSpec("sos_commands/crontab/root_crontab")]),
```
xiangce: The modification to line-234, 317, 318 and 319 should not be there. Please ensure your branch is based on the latest master branch.

xiangce:
See #368

PR #380, opened 2017-08-29
Merge the 'parse_table' helper into 'parse_fixed_table'

PaulWay:
`insights.util.parse_table` and `insights.parsers.parse_fixed_table` both do more or less the same thing: parse a table from a heading string and lines with fields into a list of dictionaries, keyed on the headings.  In both cases column names cannot include spaces and fields are stripped of surrounding spaces.

This update makes `parse_fixed_table` do the things that `parse_table` could do - namely, split the line on delimiters rather than by guessing fixed column widths from the positions of the table headers.

This update also changes the following parser modules to use the updated `parse_fixed_table` function:

* fstab
* ipcs_sem
* lvm
* netstat
* ps
* rhn_schema_stats
* route

The tests in ps were altered only to move the data closer to the tests in the file to aid comprehension; no tests were modified.  All tests pass and there is no loss of test coverage.  Tests in the Red Hat rule repositories all pass.

The `parse_table` function has not been removed in this update for backward compatibility.  The `parse_table` function will be removed in the master branch.


PaulWay:
OK, how about we turn 'parse_table' into 'parse_csv_table', behaving basically the same as it currently does, and put it in insights.parsers rather than insights.util?

jhjaggars:
Hey Paul,

The combination of functionality seems to muddy the interface.  It makes new function more intimidating to use (due to the sheer number of options) and it's not obvious what options are ok to use together or not.  It seems weird to have a `delim` option on a function called `parse_fixed_table`. 

I think it'd make sense to move the current `parse_fixed_table` to insights.util since it's not a `@parser` which the `insights.parsers` package is meant to contain.

Burying the `calc_offset` functionality in the table parsing function contributes to the overall complexity in a way that makes me wish for it to exist outside.  The table parsing functions would be much more concise and easy to understand and use if you fed them more prepared input.  I realize that this is a problem in the existing code, but if we are aiming to refactor I'd much prefer a change that makes these functions simpler and more obvious to use.

PaulWay:
Hi Jesse,

Well, I don't mind either way, but I think there are kind of two paths we can take:

1. Move all the parser 'helper' functions like `parse_fixed_table`, `optlist_to_dict`, `keyword_search` and so forth out of `insights/parsers/__init__.py` into `insights/util/__init__.py` (or even their own module files).  Then tidy up all the parsers that use them to use the `util` path.
2. Move all the parser 'helper' functions like `keys_in`, `parse_table`, and `parse_keypair_lines` into  out of `insights/util/__init__.py` into `insights/parsers/__init__.py`, then tidy up all the parsers using them to use the `parsers` path.

We've got some things in util, and some things in parsers.  We've got all the parser class definitions in core, except for the ones in their own files (e.g. AlternativeOutput) in parsers.  I'm all for consistency, and I'm happy to make the merge requests.  Just let me know how you want things organised and I'll do it.

IMO we should move all the utility functions particularly related to parsing, and all the parser base classes, into `insights/parsers/__init__.py'.

I'm assuming all of this will just be for the 2.0 branch.

Let me know how you want me to proceed here.

Have fun,

Paul

jhjaggars:
Hey Paul,

Great that we agree here on keeping things separate!  I think I agree with moving the parser utilities into the parsers package as well, and even if we change our minds about that later, it certainly seems better to consolidate them into one package rather than having them cross two packages.

I'm not sold on removing the imports from `insights/__init__.py` just yet.  Those are re-exported on purpose to hide packaging details from consumers for commonly imported items.  As such, those re-exports are another part of the interface.

PaulWay:
OK, I've redone this as #445 - will close this one.

PR #385, opened 2017-08-30
add to_json to console handler

brantleyr:


insights/client/client.py
```diff
@@ -76,7 +76,9 @@ def set_up_logging():
         config['to_stdout'] = (config['to_stdout'] or config['from_stdin'] or config['from_file'])
 
         verbose = config['verbose'] or config['verbose']
-        logging.root.addHandler(get_console_handler(config['silent'], verbose, config['to_stdout']))
+        logging.root.addHandler(get_console_handler(config['silent'],
+                                verbose,
```
kylape: Please indent to the inner parens.  It looks like these last two arguments are passed to `addHandler`.

brantleyr:
Correct. We are trying to solve the case where to_stdout should be piped to
an archive and to_json should be consumable, so we can't have any cruft
such as print statements or logging.

On Aug 30, 2017 3:59 PM, "Kyle Lape" <notifications@github.com> wrote:

*@kylape* requested changes on this pull request.

These logging functions need to be refactored because the console handler
actually prints to stderr, not stdout. Thus to_stdout and to_json should
have no bearing on the logging level of the console logger.

Are more concerned with the user experience seeing a ton of logging
alongside your JSON output, or are you having trouble piping a JSON
document to a downstream process?
------------------------------

In insights/client/client.py
<https://github.com/RedHatInsights/insights-core/pull/385#discussion_r136173158>
:

> @@ -76,7 +76,9 @@ def set_up_logging():
         config['to_stdout'] = (config['to_stdout'] or
config['from_stdin'] or config['from_file'])

         verbose = config['verbose'] or config['verbose']
-        logging.root.addHandler(get_console_handler(config['silent'],
verbose, config['to_stdout']))
+        logging.root.addHandler(get_console_handler(config['silent'],
+                                verbose,

Please indent to the inner parens. It looks like these last two arguments
are passed to addHandler.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/RedHatInsights/insights-core/pull/385#pullrequestreview-59675650>,
or mute the thread

kylape:
Pipes should work as-is because console logging goes to stderr.

brantleyr:
There were some log statements that were still getting output to the
console.

kylape:
So with the below changes, I think it still works as expected.  Basically don't take `to_stdout` or `to_json` into account when determining the log level of the console handler.  It will still print its messages to the console, but the data piped to downstream processes will not have log messages in it.

```
$ sudo bin/python run_client.py --to-stdout > archive.tar.gz
Starting to collect Insights data for server.local.kjl.me
Removing archive /var/lib/insights/archive_tmp_dir/insights-server.local.kjl.me-20170905135121.tar.gz
Detected parent temporary directory /var/lib/insights/archive_tmp_dir
Removing /var/lib/insights/archive_tmp_dir

$ file archive.tar.gz
archive.tar.gz: gzip compressed data, last modified: Tue Sep  5 18:51:31 2017, from Unix

$ git --no-pager diff
diff --git a/insights/client/client.py b/insights/client/client.py
index 878d8acf..4d47b219 100644
--- a/insights/client/client.py
+++ b/insights/client/client.py
@@ -35,13 +35,11 @@ def get_file_handler():
     return logging.handlers.RotatingFileHandler(log_file, backupCount=3)


-def get_console_handler(silent, verbose, to_stdout):
+def get_console_handler(silent, verbose):
     if silent:
         target_level = logging.NOTSET
     elif verbose:
         target_level = logging.DEBUG
-    elif to_stdout:
-        target_level = logging.ERROR
     else:
         target_level = logging.INFO

@@ -75,10 +73,7 @@ def set_up_logging():
         # from_stdin mode implies to_stdout
         config['to_stdout'] = (config['to_stdout'] or config['from_stdin'] or config['from_file'])

-        verbose = config['verbose'] or config['verbose']
-        logging.root.addHandler(get_console_handler(config['silent'],
-                                                    verbose,
-                                                    (config['to_stdout'] or config['to_json'])))
+        logging.root.addHandler(get_console_handler(config['silent'], config['verbose']))
         logging.root.addHandler(get_file_handler())
         configure_level()
         logger.debug("Logging initialized")
```

kylape:
I *think* this is resolved with the following commits:

https://github.com/RedHatInsights/insights-core/commit/ff64dbf06fd9788b334dd2ff881301b982867a89
https://github.com/RedHatInsights/insights-core/commit/46344ef4cfdd63f3b4c1e252a50642e412f90bad
https://github.com/RedHatInsights/insights-client/commit/063cb73515354cb41cd2c8cafcf4257c6d03f251

If not, then we can work through it on another PR.

PR #388, opened 2017-08-31
Add parser krb5 1.x

wushiqinlou:
- krb5 Configuration is /etc/krb5.conf and /etc/krb5.conf.d/*, and its content format is
``INI config``.


insights/parsers/krb5.py
```diff
@@ -0,0 +1,110 @@
...
+                key_value = line.strip().split("=")
+                section_value[key_value[0].strip()] = key_value[1].strip()
+            if "=" in line and is_squ:
+                squ_value.append(line)
+            if "=" in line and "{" in line and "}" not in line:
+                is_squ = True
+                squ_value.append(line)
+            if "}" in line and "=" not in line and is_squ:
+                squ_value.append(line)
+                squ_dict = parser_squ().parseString("\n".join(squ_value))[0].asDict()
```
xiangce: I don't think it's a good/efficient method to use the "pyparsing" here. You have already stored the lines in the "{}" pair. And these lines are common "A=B" (if I'm correct), they can be analyzed manually.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,110 @@
...
+        dict_all = {}
+        is_squ = False
+        section_name = ""
+        squ_value = []
+        section_value = {}
+        content_mid = []
+        for line in content:
+            if "includedir" in line or line.startswith("#") or not line:
+                continue
+            content_mid.append(line)
```
xiangce: The first `for` is redundant and can be merged into the next one.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,110 @@
...
+        for line in content_mid:
+            if "[" in line and "]" in line:
+                if not section_name:
+                    section_name = line.strip().strip("[]")
+                    section_value = {}
+                else:
+                    dict_all[section_name] = section_value
+                    section_name = line.strip().strip("[]")
+                    section_value = {}
+            if "=" in line and "{" not in line and "}" not in line and not is_squ:
```
xiangce: Better use `elif` or add a `continue` to each `if`.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,110 @@
...
+                key_value = line.strip().split("=")
+                section_value[key_value[0].strip()] = key_value[1].strip()
+            if "=" in line and "{" in line and "}" in line and not is_squ:
+                key_value = line.strip().split("=")
+                section_value[key_value[0].strip()] = key_value[1].strip()
+            if "=" in line and is_squ:
+                squ_value.append(line)
+            if "=" in line and "{" in line and "}" not in line:
```
xiangce: There are so many `if` and some of their conditions are duplicate, better optimize the code base on current logic.

PaulWay:
See my comments on #389 - I think this needs improvement and should try to use IniConfigFile.

insights/config/specs.py
```diff
@@ -174,6 +174,8 @@
     "keystone.conf"             : SimpleFileSpec("etc/keystone/keystone.conf"),
     "keystone.log"              : SimpleFileSpec("var/log/keystone/keystone.log"),
     "keystone_crontab"          : CommandSpec("/usr/bin/crontab -l -u keystone"),
+    "krb5"                      : SimpleFileSpec("etc/krb5.conf"),
+    "krb5.conf.d"              :  PatternSpec(r"etc/krb5.conf\.d/.*\.conf"),
```
xiangce:
~~~
PatternSpec(r"etc/krb5\.conf\.d/.*\.conf"),
~~~

insights/config/specs.py
```diff
@@ -238,7 +240,6 @@
     "ntpq_leap"                 : CommandSpec("/usr/sbin/ntpq -c 'rv 0 leap'"),
     "ntpq_pn"                   : CommandSpec("/usr/sbin/ntpq -pn"),
     "ntptime"                   : CommandSpec("/usr/sbin/ntptime"),
-    "numeric_user_group_name"   : CommandSpec("/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group"),
```
xiangce: This spec should not be deleted.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,108 @@
...
+"""
+
+
+from insights.core import Parser
+from insights.core.plugins import parser
+
+
+@parser('krb5')
+@parser('krb5.conf.d')
+class Krb5Configuration(Parser):
```
PaulWay: I would recommend, if you're implementing a structure similar to an IniConfigFile parser, that you derive from that and then override the parse_content method as you've done.  That way you get all the methods of IniConfigFile for free.

insights/parsers/tests/test_krb5.py
```diff
@@ -0,0 +1,71 @@
...
+[logging]
+ default = FILE:/var/log/krb5libs.log
+ kdc = FILE:/var/log/krb5kdc.log
+ admin_server = FILE:/var/log/kadmind.log
+
+[realms]
+ dns_lookup_realm = false
+ ticket_lifetime = 24h
+ default_ccache_name = KEYRING:persistent:%{uid}
+ EXAMPLE.COM = {
```
PaulWay: I notice you've only got one kdc here.  The krb5.conf file can define multiple, e.g.:
```ini
[realms]
  REDHAT.COM = {
   kdc = kerberos01.core.prod.int.phx2.redhat.com.:88
   kdc = kerberos.rdu.redhat.com.:88
   kdc = kerberos.bos.redhat.com.:88
   kdc = kerberos.brq.redhat.com.:88
   admin_server = kerberos.corp.redhat.com.:749
   default_domain = redhat.com
  }
```
How does the code handle that?

insights/parsers/krb5.py
```diff
@@ -0,0 +1,108 @@
...
+            """The {} sub_section should end with }, if it is, set the whole value to the sub_section name, and clean the flag"""
+            if "}" in line and "=" not in line and is_squ:
+                section_value[squ_section_name] = squ_value
+                is_squ = False
+                squ_section_name = ""
+                squ_value = {}
+                continue
+        """Set the value to the last section """
+        dict_all[section_name] = section_value
+        self.data = dict_all
```
PaulWay: It seems you have looked at getting the contents of this file, but you have not considered how a rule writer would use the contents.  How does a rule writer find out what realms are defined?  Can they easily find out the ticket lifetime for any domain, with reference back to the defaults?  How would a rule get the reverse mapping from realm back to its domains?

As well as the usual IniConfigFile mechanisms, I would recommend you think about how best to provide useful information from the file to save future rule writers effort.

insights/parsers/krb5.py
```diff
@@ -47,62 +50,104 @@
...
+        # Return a list of section names
+        return self.data.keys()
+   
+    def has_section(self, section):
+        # Indicate whether the named section is present in the configuration.
+        return section in self.data
+
+    def options(self, section):
+        # Return a list of option names for the given section name.
+        try:
```
xiangce: It's not suggested to use `print` to handle exception. How about just returning `None`?
~~~
if self.has_section(section):
    return self.data[section].keys()
~~~

insights/parsers/krb5.py
```diff
@@ -47,62 +50,104 @@
...
+        return section_value.keys()
+  
+    def has_option(self, section, option):
+        # Check for the existence of a given option in a given section.
+        if section not in self.data:
+            return False
+        return option in self.data[section]
+
+    def get(self, section, option):
+        try:
```
xiangce:
~~~
if self.has_option(section, option):
   return self.data[section][option]
~~~

insights/parsers/krb5.py
```diff
@@ -47,62 +50,104 @@
...
 from insights.core.plugins import parser
+from .. import get_active_lines
 
 
-@parser('krb5')
-@parser('krb5.conf.d')
-class Krb5Configuration(Parser):
-    """Class for krb5.conf and krb5.conf.d configuration."""
-
+class Krb5ConfigParser(Parser):
```
xiangce: Better mixin with LegacyItemAccess

insights/parsers/krb5.py
```diff
@@ -47,62 +50,104 @@
...
                 key_value = line.strip().split("=")
                 section_value[key_value[0].strip()] = key_value[1].strip()
                 continue
-            """If in {} sub_section, get the key_value pair"""
+            # If in {} sub_section, get the key_value pair
             if "=" in line and is_squ:
                 key_value = line.strip().split("=")
-                squ_value[key_value[0].strip()] = key_value[1].strip()
+                key = key_value[0].strip()
+                value = key_value[1].strip()
```
xiangce: key, value = [i.strip() for i in line.strip().split('=',1)]

insights/parsers/krb5.py
```diff
@@ -47,62 +50,104 @@
...
         is_squ = False
         section_name = ""
         squ_value = {}
         squ_section_name = ""
         section_value = {}
-        for line in content:
-            """Get active line """
-            if "includedir" in line or line.startswith("#") or not line:
+        for line in get_active_lines(content):
+            if "includedir" in line:
```
xiangce: Better add one line:
~~~
line = line.strip()
~~~

then use it in below code.

insights/parsers/krb5.py
```diff
@@ -47,62 +50,104 @@
...
                     dict_all[section_name] = section_value
-                    section_name = line.strip().strip("[]")
-                    section_value = {}
+                section_name = line.strip().strip("[]")
+                section_value = {}
                 continue
-            """Common key value format is XXX = YYY"""
+            # Common key value format is XXX = YYY
             if "=" in line and "{" not in line and "}" not in line and not is_squ:
                 key_value = line.strip().split("=")
```
xiangce: key, value = [i.strip() for i in line.strip().split('=',1)]

insights/parsers/krb5.py
```diff
@@ -47,62 +50,104 @@
...
+                key = key_value[0].strip()
+                value = key_value[1].strip()
+                if key in squ_value:
+                    if isinstance(squ_value[key], list):
+                        squ_value[key].append(value)
+                    else:
+                        squ_value[key] = [squ_value[key]]
+                        squ_value[key].append(value)
+                else:
+                    squ_value[key] = value
```
xiangce:
~~~
if key in squ_value:
       if isinstance(squ_value[key], str):
            squ_value[key] = [squ_value[key]]
       squ_value[key].append(value)
~~~

insights/config/specs.py
```diff
@@ -238,7 +240,7 @@
...
+    "numeric_user_group_name"   : CommandSpec("/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group"),
```
xiangce: Please remove the trailing whitespace

insights/parsers/krb5.py
```diff
@@ -0,0 +1,152 @@
...
+            # Common key value format is XXX = YYY
+            if "=" in line and "{" not in line and "}" not in line and not is_squ:
+                key_value = line.strip().split("=")
+                section_value[key_value[0].strip()] = key_value[1].strip()
+                continue
+            # Special key value format is XXX = YYY{%ZZZ}
+            if "=" in line and "{" in line and "}" in line and len(line.split("=")) == 2 and not is_squ:
+                key_value = line.strip().split("=")
+                section_value[key_value[0].strip()] = key_value[1].strip()
+                continue
```
xiangce: Above two `if` can be merged?
How about:
~~~
if not is_squ and "=" in line and not line.endswith("{"):
~~~

insights/parsers/krb5.py
```diff
@@ -0,0 +1,152 @@
...
+                    if isinstance(squ_value[key], list):
+                        squ_value[key].append(value)
+                    else:
+                        squ_value[key] = [squ_value[key]]
+                        squ_value[key].append(value)
+                else:
+                    squ_value[key] = value
+                continue
+            # The {} sub_section should start with format XXX = {
+            if "=" in line and "{" in line and "}" not in line and len(line.split("=")) == 2:
```
xiangce: How about:
if not is_squ and "=" in line and line.endswith('{'):

insights/parsers/krb5.py
```diff
@@ -0,0 +1,152 @@
...
+                is_squ = True
+                squ_section_name = line.strip().split("=")[0].strip()
+                continue
+            # The {} sub_section should end with }, if it is, set the whole value to the sub_section name, and clean the flag
+            if "}" in line and "=" not in line and is_squ:
```
xiangce: How about "line == "}"?

insights/parsers/krb5.py
```diff
@@ -0,0 +1,132 @@
...
+        section_name = ""
+        squ_value = {}
+        squ_section_name = ""
+        section_value = {}
+        for line in get_active_lines(content):
+            line = line.strip()
+            if is_squ:
+                # If in {} sub_section, get the key_value pair
+                if "=" in line:
+                    key, value = [i.strip() for i in line.strip().split('=', 1)]
```
xiangce: This and the following `line.strip()` can be replaced with `line` because of line-65 did it.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,132 @@
...
+                    else:
+                        squ_value[key] = value
+                # The {} sub_section should end with }, if it is, set the whole value to the sub_section name, and clean the flag
+                if line.strip() == "}":
+                    section_value[squ_section_name] = squ_value
+                    is_squ = False
+                    squ_section_name = ""
+                    squ_value = {}
+            else:
+                if "includedir" in line:
```
xiangce: This `if` can be moved out (before the `if is_squ:`)

insights/parsers/krb5.py
```diff
@@ -0,0 +1,132 @@
...
+                if "=" in line:
+                    key, value = [i.strip() for i in line.strip().split('=', 1)]
+                    if key in squ_value:
+                        if isinstance(squ_value[key], str):
+                            squ_value[key] = [squ_value[key]]
+                        squ_value[key].append(value)
+                    else:
+                        squ_value[key] = value
+                # The {} sub_section should end with }, if it is, set the whole value to the sub_section name, and clean the flag
+                if line.strip() == "}":
```
xiangce: better use `elif`

insights/parsers/krb5.py
```diff
@@ -0,0 +1,132 @@
...
+                    continue
+                # [XXX] means a section, get the section name and the value format is dict.
+                if line.startswith("[") and line.endswith("]"):
+                    # If first section, just get the section name, if not, set the value to the former section and get the section name
+                    if section_name:
+                        dict_all[section_name] = section_value
+                    section_name = line.strip().strip("[]")
+                    section_value = {}
+                # key value format is XXX = YYY, store as dict
+                if "=" in line and not line.endswith("{"):
```
xiangce: better use `elif`

insights/parsers/krb5.py
```diff
@@ -0,0 +1,108 @@
...
+    >>> krb5_content = '''
+    [realms]
+     dns_lookup_realm = false
+     ticket_lifetime = 24h
+     default_ccache_name = KEYRING:persistent:%{uid}
+     EXAMPLE.COM = {
+      kdc = kerberos.example.com
+      admin_server = kerberos.example.com
+     }
+     89EXAsdcsdMPLE.COM = {
```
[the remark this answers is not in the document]
wushiqinlou: @PaulWay - I made them, change to a working krb5.conf example.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,108 @@
...
+    # rdns = false
+    '''.strip()
+
+    >>> from insights.tests import context_wrap
+    >>> shared = {Krb5Configuration: Krb5Configuration(context_wrap(krb5_content))}
+    >>> krb5_info = shared[Krb5Configuration]
+    >>> krb5_info.data["libdefaults"]["dnsdsd"]
+    "false"
+    >>> krb5_info.data["realms"]["EXAMPLE.COM"]["kdc"]
+    "kerberos.example.com"
```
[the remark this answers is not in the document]
wushiqinlou: @PaulWay - I updated the example exercise. If you think it is not full, please point out in detail.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,151 @@
...
+    def has_option(self, section, option):
+        """
+        Check for the existence of a given option in a given section.
+        Return True if the given option is present, and False if not present.
+        """
+        if section not in self.data:
+            return False
+        return option in self.data[section]
+
+    def get(self, section, option):
```
bfahr: Since this overloads the `LegacyItemAccess.get()` method would it make sense to include a `default` keyword option?

insights/parsers/krb5.py
```diff
@@ -0,0 +1,151 @@
...
+krb5 Configuration file - File
+==============================
+krb5 Configuration are ``/etc/krb5.conf`` and ``/etc/krb5.conf.d/*``,
+and the content format is similar to ``INI config``, but they include
+values that span multiple lines. Multi-line values start with a '{'
+and end with a '}', and we join them together by setting the `is_squ`
+variable to True while in a multi-line value.
+
+The content of file ``/etc/krb5.conf`` and ``/etc/krb5.conf.d/*`` is
+recorded via INI format.
```
bfahr: You might consider rewording this since you've already described the INI format above, for example:
```
Sample content of ``/etc/krb5.conf`` and ``/etc/krb5.conf.d/*`` files 
is shown in the *Examples* section below.
```

insights/parsers/krb5.py
```diff
@@ -0,0 +1,140 @@
...
+    "false"
+    >>> krb5_info["realms"]["EXAMPLE.COM"]["kdc"]
+    "kerberos.example.com"
+    >>> krb5_info.sections()
+    ["libdefaults","realms"]
+    >>> krb5_info.has_section("realms")
+    True
+    >>> krb5_info.has_option("realms", "nosuchoption")
+    False
+    >>> krb5_info.get("realms", "default_ccache_name")
```
xiangce: This `get` should be updated as you have removed it.

insights/parsers/krb5.py
wushiqinlou: @bfahr - As we can get option value by shared[Krb5Configuration][section][option],  I think `get` function is duplicate and we can delete it according to the discussion with @xiangce.

xiangce:
Looks good to me. If @bfahr and @PaulWay are satisfied with the changes, it's ready to merge.

xiangce:
@PaulWay - How do you think about the new changes, if it's OK, I'm going to merge it.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,138 @@
...
+        return self.data[section].keys() if self.has_section(section) else []
+
+    def has_option(self, section, option):
+        """
+        Check for the existence of a given option in a given section.
+        Return True if the given option is present, and False if not present.
+        """
+        if section not in self.data:
+            return False
+        return option in self.data[section]
```
PaulWay: This still provides no assistance to rule writers to determine information about the configuration as opposed to merely what the file contained.  How does a rule writer find out what realms are defined? Can they easily find out the ticket lifetime for any domain, with reference back to the defaults? How would a rule get the reverse mapping from realm back to its domains? Etc.  Things like these should be provided as properties or methods to assist the rule writer.

insights/parsers/tests/test_krb5.py
```diff
@@ -0,0 +1,92 @@
...
+[realms]
+ dns_lookup_realm = false
+ auth_to_local = RULE:[1:$1@$0](.*@.*EXAMPLE.ORG)s/@.*//
+ default_ccache_name = KEYRING:persistent:%{uid}
+ default_ccache_name2 = KEYRING:%{uid}:persistent
+ EXAMPLE.COM = {
+  kdc = kerberos.example.com
+  admin_server = kerberos.example.com
+ }
+ 89EXAsdcsdMPLE.COM = {
```
PaulWay: I would strongly recommend not making up examples, and instead to collaborate with IdM associates and other people who can give you known configuration files that provide the examples you want to test.

insights/parsers/tests/test_krb5.py
```diff
@@ -0,0 +1,92 @@
...
+""".strip()
+
+KRB5CONFIG2 = """
+# Configuration snippets may be placed in this directory as well
+""".strip()
+
+KRB5DCONFIG = """
+# Configuration snippets may be placed in this directory as well
+
+[loggingbb]
```
PaulWay: Please do not make up examples.  Find a configuration file that has what you want to test by asking the IdM associates.

insights/config/specs.py
```diff
@@ -174,6 +174,8 @@
     "keystone.conf"             : SimpleFileSpec("etc/keystone/keystone.conf"),
     "keystone.log"              : SimpleFileSpec("var/log/keystone/keystone.log"),
     "keystone_crontab"          : CommandSpec("/usr/bin/crontab -l -u keystone"),
+    "krb5"                      : SimpleFileSpec("etc/krb5.conf"),
```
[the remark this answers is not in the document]
PaulWay: It's also worth pointing out that once again we are faced with a program's configuration that is comprised of multiple files.  However, Krb5Configuration files will only ever see one file.  It will be up to the rule writer to make sure they parse all the files, and then they will need to resolve dependencies - e.g. default values from the [libdefaults] in /etc/krb5.conf affect a realm setting in /etc/krb5.conf.d/realm.conf.

The solution, at the moment, is to write a combiner that reads all the files and resolves them into one structure.  See the ``insights/combiners/limits_conf.py`` combiner for an example.

I proposed a MultiFileParser class that would take multiple files and provide a united interface to them but that was voted down.

I would really like to see this combiner provided as part of this commit, so that we do not end up with the situation where some rules are using the parser by itself and assuming that they don't have to deal with multiple files.

insights/parsers/tests/test_krb5.py
```diff
@@ -0,0 +1,92 @@
...
+includedir /etc/krb5.conf.d/
+
+[logging]
+ default = FILE:/var/log/krb5libs.log
+ kdc = FILE:/var/log/krb5kdc.log
+ admin_server = FILE:/var/log/kadmind.log
+
+[realms]
+ dns_lookup_realm = false
+ auth_to_local = RULE:[1:$1@$0](.*@.*EXAMPLE.ORG)s/@.*//
```
PaulWay: Once again, do not make up examples.  auth_to_local does not exist outside a realm.  See `man krb5.conf` for details on how the [realms] section is parsed.  Using bad examples means that we miss implementation details that should be checked.

insights/config/specs.py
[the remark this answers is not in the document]
xiangce: @wushiqinlou - yes, Paul is right, we'd better declare them both "PatternSpec", Please refer to `limit.conf` and `limit.d`

insights/parsers/krb5.py
wushiqinlou: @PaulWay - As we need to merge this PR this week for writing a IDM plugin, I plan to write these handy functions in the future.

insights/parsers/krb5.py
```diff
@@ -0,0 +1,138 @@
...
+    >>> krb5_info.has_section("realms")
+    True
+    >>> krb5_info.has_option("realms", "nosuchoption")
+    False
+    >>> krb5_info.options("libdefaults")
+    ["dns_lookup_realm","ticket_lifetime","EXAMPLE.COM"]
+"""
+
+
+from insights.core import Parser, LegacyItemAccess
```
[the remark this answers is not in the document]
jhjaggars: The policy is relative if what you are importing is inside the same package or a subpackage of the current package. Use absolute if you are importing outside of the current package.

If you are working on A.B and want to import A.B.C you should use a relative import, if you want to import A.D you should use an absolute import.

xiangce:
Since @wushiqinlou has already replied/updated your comments, I will go ahead and merge this PR, since there is a rule depends on the basic feature of this parser.

PR #389, opened 2017-08-31
Add parser krb5

wushiqinlou:
- Add krb5 Configuration is ``/etc/krb5.conf`` and ``/etc/krb5.conf.d/*``, and its content format is
``INI config``.


xiangce:
See #388

PaulWay:
Can I make a suggestion here?  It might be better to pre-parse the config file contents into a form that the IniConfigFile parser can understand, so as to not have to re-invent so much of its functionality.

You also need to add a lot more comments throughout your code to explain how it works.  Especially with complicated parser and capture logic like this, you need to make sure that it's as easy as possible for anyone who comes after you to modify this.

I'd also like to see much more testing and a thorough check of every part of the structure created.  It should check that the structure contains exactly what you expect it to contain, no more and no less.

jhjaggars:
This PR seems quite a bit different from #388, what's the reason for that?

xiangce:
@jhjaggars   - this one should be updated after the #388 finally being frozen.

PR #390, opened 2017-08-31
Make raw options string available for 1.x branch

PaulWay:
The mount options is most useful as a dictionary from which to look up values.  However, because of the unordered nature of dictionaries, the reverse process - constructing the options string from the dictionary of options - results in a string with a different ordering from the original fstab line.  Some rules require this string in order to give the user an example of an amended fstab line, so we provide this text as the `raw_fs_mntops` key/attribute.


xiangce:
Will merge it after the same update applied to the master branch.

PR #364, opened 2017-08-23
Updating pcs_status class to show bad nodes, aka nodes that

wduffeeb:
Updating pcs_status class to show bad nodes, aka nodes that show as online but are not in a clean status when the pcs_status output is dug into.

PR #394, opened 2017-09-01
Updating pcs_status class to show bad nodes, aka nodes that for 1.x

xiangce:


xiangce:
See https://github.com/RedHatInsights/insights-core/pull/364

PR #395, opened 2017-09-01
Refine the satellite_version as per the latest version map - 1.x

xiangce:


insights/combiners/satellite_version.py
```diff
@@ -95,30 +104,33 @@ def satellite_version(local, shared):
...
+        # For Satellite 6.2.x, check the ``satellite`` package directly
         sat62_pkg = rpms.get_max('satellite')
         if sat62_pkg:
             version = sat62_pkg.version
             major, minor = _parse_sat_versoin(version)
             return SatelliteVersion(sat62_pkg.package, version,
                                     sat62_pkg.release, major, minor)
 
-        # For Satellite 6.0.x/6.1.x
+        # For Satellite 6.0.x/6.1.x, a little complex
```
chenlizhong: "a liitel complex" as comment is not good. you should give words to explain the codes.

xiangce:
@chenlizhong - updated the code as per our discussion, please review it again, thanks.

insights/combiners/satellite_version.py
```diff
@@ -95,30 +107,36 @@ def satellite_version(local, shared):
...
+        # For Satellite 6.0.x/6.1.x, check the version of:
+        # - foreman, candlepin and katello
+        fman = rpms.get_max('foreman')
+        canp = rpms.get_max('candlepin')
+        ktlo = rpms.get_max('katello')
+        if fman and canp and ktlo:
+            for sat_ver, map_ver in sat6_ver_map.items():
+                if (fman.version.startswith(map_ver.foreman) and
+                        canp.version.startswith(map_ver.candlepin) and
+                        ktlo.version.startswith(map_ver.katello)):
```
chenlizhong: maybe we can simplify 133 to 135 by below code:
  if all([a.startswith(b) for a,b in zip([fman, canp, ktlo], map_ver.values())]):


In further, if just use tuple as value in sat6_ver_map, We can replace map_ver.values() with map_ver.

xiangce: Cool point! will do it

PR #396, opened 2017-09-01
handle multiple lines of output with the mlx4 parser

jhjaggars:
This change covers corner cases in parsing found via production usage of the mlx4_port parser.


xiangce:
@jhjaggars - would you please re-submit this PR again to the master branch?

xiangce:
See #404

PR #399, opened 2017-09-01
want to bypass verification in no_gpg

brantleyr:


kylape:
Can we please read the `gpg` option instead?

brantleyr:
Change made. Now reading 'gpg.'

PR #403, opened 2017-09-05
Enhance class Ethtool with two more ParseExceptions cases

JoySnow:


insights/parsers/tests/test_ethtool.py
```diff
@@ -565,3 +570,5 @@ def test_ethtool(self):
     def test_ethtool_fail(self):
         with pytest.raises(ParseException):
             ethtool.Ethtool(context_wrap(FAIL_ETHTOOL_A_1, path="ethtool_eth1"))
+        with pytest.raises(ParseException):
```
xiangce: Since there are different ParseExceptions in the parser, better check them correspondingly by check the exception message.
And, would you mind refactor the unit test to pytest style in this commit? (unittest2pytest would help you do that easily)

insights/parsers/ethtool.py
```diff
@@ -760,8 +760,16 @@ def parse_content(self, content):
...
         self.iface = self.file_name.replace("ethtool_", "")
 
+        # One way to get this result is run command '$ ethtool'.
+        if "ethtool: bad command line argument(s)" in content[0]:
+            raise ParseException('ethtool: bad command line argument for ethtool', content)
+
         if "Settings for" in content[0]:
             self.data['ETHNIC'] = content[0].split()[-1].strip(':')
+
+        if "No data available" in content[1]:
```
xiangce: As per my test, the output could be:
~~~
[root@liuxc-rhel6-sat5 ~]# ethtool test
Settings for test:
Cannot get device settings: No such device
Cannot get wake-on-lan settings: No such device
Cannot get message level: No such device
Cannot get link status: No such device
No data available
---
[root@liuxc-rhel7-sat6 ~]# ethtool no
Settings for no:
Cannot get device settings: No such device
Cannot get wake-on-lan settings: No such device
Cannot get message level: No such device
Cannot get link status: No such device
No data available

~~~

JoySnow: AFAIK, these lines:  
> Cannot get device settings: No such device
Cannot get wake-on-lan settings: No such device
Cannot get message level: No such device
Cannot get link status: No such device

are the stderr part of this command's output, will not store in collecting-file.
And these 2 examples are from the analysis tool(I made parser to raise the whole content instead of that specific line) .  
Here :  
https://ceehadoop1.gsslab.rdu2.redhat.com/raui/reports/content/?path=%2Fjobs%2F2017-09-04%2Fxiaoxwan%2F1015%2Fanalysis_output%2Frhev_lro_on.html

insights/parsers/tests/test_ethtool.py
JoySnow: Sure.

insights/parsers/ethtool.py
xiangce: Nice @JoySnow , thanks for investigating.

PR #407, opened 2017-09-05
friendly message for saving tar w no-upload

gravitypriest:
not sure if None response will kill anything that calls this, please close if it needs fixin'


kylape:
@gravitypriest Just to make sure - did you test this using the `insights-client --no-upload` RPM wrapper script?

kylape:
I should note that while adding a check for `no_upload` in the `upload`function is okay for now, it will have to be removed during the Great Configuration Refactor since general intuition would side against having a call to upload simply print the path passed in to it ;)

PR #416, opened 2017-09-07
add swift_conf parser module

chenlizhong:


insights/parsers/swift_conf.py
```diff
@@ -0,0 +1,42 @@
+"""
+SwiftProxyServerConf - file ``/etc/swift/proxy-server.conf``
```
xiangce: AFAIK, in the `/etc/swift/` folder, there are also some configuration files like `swift.conf` and `object-server.conf`, and I think in the future some of them will be included in our framework. So, would you please modify the document more compatible?

insights/parsers/swift_conf.py
```diff
@@ -0,0 +1,44 @@
+"""
+swift_conf - files under /etc/swift/
+====================================
+
+This module provides parsers for swift config files under /etc/swift directory.
+
+Class ``SwiftProxyServerConf`` parses the content of the ``/etc/swift/proxy-server.conf``.
```
xiangce: It would be better to shift line-7~35 into the docstring `SwiftProxyServerConf`, just like the `netstat.py`

insights/parsers/swift_conf.py
```diff
@@ -0,0 +1,50 @@
...
+        set
+        log_name = proxy - server
+        set
+        log_facility = LOG_LOCAL1
+
+        [filter:catch_errors]
+        use = egg:swift  # catch_errors
+
+    Examples:
+    >> > proxy_server_conf = shared[SwiftProxyServerConf]
```
xiangce: One more indent is needed for the lines after the `Examples:` and better remove the space in the `>> >`.

PR #417, opened 2017-09-08
Add combiner krb5 1x

wushiqinlou:


insights/combiners/krb5.py
```diff
@@ -0,0 +1,49 @@
...
+            if krb5_parser.file_path == "/etc/krb5.conf":
+                main_data = krb5_parser.data
+            else:
+                self.data = dict(self.data, **krb5_parser.data)
+        self.data = dict(self.data, **main_data)
+
+    def __getitem__(self, item):
+        return self.data[item]
+
+    def get(self, item, default=None):
```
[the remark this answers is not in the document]
xiangce: And should the list of `module` be accessed via this combiner? (and `include` and `includedir`?)

insights/combiners/krb5.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, local, shared):
+        self.data = {}
+        main_data = {}
+
+        for krb5_parser in shared[Krb5Configuration]:
+            if krb5_parser.file_path == "/etc/krb5.conf":
+                main_data = krb5_parser.data
+            else:
+                self.data = dict(self.data, **krb5_parser.data)
```
xiangce: `self.data.update(...)` could be a better choice than `dict()`

insights/combiners/krb5.py
```diff
@@ -0,0 +1,49 @@
...
+        self.data = dict(self.data, **main_data)
```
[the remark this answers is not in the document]
wushiqinlou: @xiangce - krb5.conf is not configuration file for service domain, it is read by krb5 commands, e.g. kinit. And if there is any error in this file, customer will find it immediately while reading it. This file is just used to check configuration, the possibility to change settings from insights' recommends. So I think there is no need to return file_path and file_name in combiner.

insights/combiners/krb5.py
```diff
@@ -0,0 +1,83 @@
...
+@combiner(requires=[Krb5Configuration])
+class AllKrb5Conf(object):
+    """
+    Combiner for accessing all the krb5 configuration files, the format is dict.
+    There may be multi files for krb5 configuration, and the main config file is
+    krb5.conf. In the situation that same section is both in krb5.conf and other
+    configuration files, section in krb5.conf is the available setting. Data from
+    parser krb5 is list of dict(s), this combiner will parse this list and return
+    a dict which containing all valid data.
+    """
```
bfahr: Please add an `Examples:` section to the documentation.

insights/combiners/krb5.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __init__(self, local, shared):
+        self.data = {}
+        main_data = {}
+        self.includedir = []
+        self.include = []
+        self.module = []
```
bfahr: Please include docstrings for the attributes `includedir`, `include`, and `module`.

insights/combiners/krb5.py
```diff
@@ -0,0 +1,115 @@
...
+        >>> all_krb5 = shared[AllKrb5Conf]
+        >>> all_krb5.include
+        ['/etc/krb5test.conf']
+        >>> all_krb5.sections()
+        ['logging', 'realms']
+        >>> all_krb5.options('logging')
+        ['default', 'kdc']
+        >>> all_krb5['logging']['kdc']
+        'FILE:/var/log/krb5kdc.log'
+        >>> all_krb5.has_option('logging', 'admin_server')
```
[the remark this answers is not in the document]
PaulWay: Unfortunately this behaviour is not clearly defined.  `man krb5.conf` says:
```
The krb5.conf file can include other files using either of the following directives at the beginning of a line:

          include FILENAME
          includedir DIRNAME

FILENAME  or  DIRNAME should be an absolute path. The named file or directory must exist and be readable.  Including a directory includes all files within the directory whose names consist solely of alphanumeric characters, dashes, or underscores.  Included profile files are syntactically independent of their parents, so each included file must begin with a section header.
```
I take the point about 'included profile files [being] syntactically independent of their parents' to mean that the files are read independently and then merged, rather than being read as if the included file was the literal content of the original file (as the `#include` directive in C does, for example).  However, I see no evidence there or in discussion about this online to decide whether the `/etc/krb5.conf` file takes precedence or not, and whether files are read in name order or directory order.

Just as a note, it's worth noting that the `include` and `includedir` directives can occur outside a section heading.

One way of resolving this is simply to test what happens in Kerberos.  Does it allow multiple files to redefine the same section?  The same key?  What happens?  Have we tried this?  Can we?

If we can't verify how Kerberos works in the short term, we may have to make a decision and modify it later if it turns out to be incorrect.  OTOH, I'd suggest that we adopt the reverse approach to that taken in the code - I think we should read the main file first, and then override it with any included files.  This is slightly simpler, but it's also in keeping with the general process of include files overriding default behaviour.

insights/combiners/tests/test_krb5.py
```diff
@@ -0,0 +1,90 @@
...
+  admin_server = kerberos.example2.com
+ }
+ EXAMPLE3.COM = {
+  kdc = kerberos.example3.com
+  admin_server = kerberos.example3.com *
+ }
+""".strip()
+
+
+def test_active_krb5_nest():
```
PaulWay: Does this test how the included file overrides the main `/etc/krb5.conf` file?  I can't see how it does...

insights/tests/__init__.py
```diff
@@ -178,7 +178,7 @@ def integrate(input_data, module):
     This method drives the mapreduce framework with test `input_data`.
 
     :param InputData input_data: InputData object filled with test data
-    :param module module: module containing parsers and reducers to test with
```
PaulWay: Not sure this is relevant to this merge request :-)

xiangce: @PaulWay - There are two `module`s declared in docstring (parser.Krb5Configuration.module) and (combiner.AllKrb5Conf.module), but here none is referenced. And it will break making the document. I think, here, the `module` in line-180 is not a `type` used in the core, and it would point to the `plugin`, so I modified it. And how do you think this modification @bfahr.

insights/combiners/tests/test_krb5.py
xiangce: The `assert not result.has_option("logging", "admin_server")` verified it.

insights/tests/__init__.py
xiangce: As per the discussion with @bfahr, I separate the modification of `test/__init__.py` and `parsers/krb5.py` to another 2 PRs, since they are not the part of the update of `combiners/krb5.py`

See #452 and #453.

xiangce:
@bfahr - Please check if this PR could be accepted when you are free.

PR #418, opened 2017-09-08
make cron stuff work

brantleyr:


insights/client/client.py
```diff
@@ -565,6 +567,30 @@ def handle_startup():
...
         return validate_remove_file()
 
+    # handle cron stuff
+    if config['enable_schedule'] is True and config['disable_schedule'] is True:
+        logger.error('Conflicting options: --enable-schedule and --disable-schedule')
+        return False
+
+    if config['enable_schedule']:
+        # enable automatic scheduling
+        InsightsSchedule()
```
kylape: It seems more intuitive to me to not call `set_daily` in `InisghtsSchedule.__init__` and instead call it here.  What do you think?

kylape:
@brantleyr Looks like there are some tests to update as well.

insights/client/client.py
brantleyr: Done.

PR #420, opened 2017-09-11
some more failsafe checks for gpg

brantleyr:


insights/client/__init__.py
```diff
@@ -185,7 +185,22 @@ def _fetch(self, url, etag_file, target_path, force):
             logger.debug('Please check config, error reaching %s', url)
 
     def update(self):
+        # fetch the new eggs and gpg
         egg_paths = self.fetch()
+
```
kylape: This block of code seems to duplicate the checks done in `verify`.

insights/client/__init__.py
```diff
@@ -209,6 +224,25 @@ def verify(self, egg_path, gpg_key=constants.pub_gpg_path):
                     'stdout': None,
                     'rc': 0}
 
+        # check if the provided files (egg and gpg) actually exist
+        if egg_path and not os.path.isfile(egg_path):
```
kylape: Perhaps move this `if` block above the `gpg` option check to make sure it's done even when gpg verification is disabled.

insights/client/__init__.py
```diff
@@ -253,10 +298,12 @@ def install(self, new_egg, new_egg_gpg_sig):
             raise
 
         # Copy the NEW (/tmp/insights-core.egg) egg to /var/lib/insights/newest.egg
+        # Additionally, copy NEW (/tmp/insights-core.egg.asc) to /var/lib/insights/newest.egg.asc
         try:
             logger.debug("Copying %s to %s." % (new_egg, constants.insights_core_newest))
             shutil.copyfile(new_egg, constants.insights_core_newest)
-            shutil.copyfile(new_egg_gpg_sig, constants.insights_core_gpg_sig_newest)
+            if config["gpg"] and new_egg_gpg_sig:
```
kylape: I actually think it would be better to fetch and install the signature file even if gpg verification is currently disabled.  It might not be disabled next time around, thus invalidating the installed egg.

insights/client/__init__.py
brantleyr: Correct. Its more so failsafe, but I can take it out if you find it too redundant.

kylape: Yeah, I think it's redundant enough to be removed IMHO.

brantleyr: It hath been removed.

PR #422, opened 2017-09-12
Fix bug when parsing parameters which presented in both solo format and key-value format.

lonicerae:
- Inspired by the log `'NoneType' object has no attribute 'append'`.


PaulWay:
This looks to me like it's very similar to the `optlist_to_dict` code, which uses `True` instead of `None` to signify that the option was present but had no value.  Personally I think that's better, because a test like `if cmd['audit']` will return True rather than None, which would evaluate to False.  What do you think?

Is there any chance to use `optlist_to_dict` here?

xiangce:
@PaulWay - I'm afraid that `optlist_to_dict` cannot handle the "duplicated" key like "audit" that @lonicerae tested in the case.

insights/parsers/cmdline.py
```diff
@@ -59,8 +59,15 @@ def parse_content(self, content):
         for el in line.strip().split(None):
             # Thanks to strip and split, el always contains something.
             if "=" not in el:
-                cmdline_properties[el] = None
+                if cmdline_properties.get(el) is None:
+                    cmdline_properties[el] = None
+                if isinstance(cmdline_properties[el], list):
+                    cmdline_properties[el].append(None)
             else:
                 (key, value) = el.split("=", 1)
```
jhjaggars: Maybe we should always make things lists so that the interface is consistent.  They way this works now you have to check if the value is a list or a scalar value in the calling code.

jhjaggars: I realize that this comment was applied to the wrong side of the diff.

xiangce:
@jhjaggars - Since there are many plugins used this parser, how about we modify the interface only in the master branch (See #431) and leave the interface of 1.x branch unchanged to avoid plugins broken?

jhjaggars:
@xiangce yes, we should preserve the current, awkward interface in 1.x and disambiguate the interface in master.

PR #423, opened 2017-09-12
add machine ids as soft requirement

brantleyr:


kylape:
We decided this change is no longer needed.

PR #430, opened 2017-09-14
Modprobe combiner for 1.x branch

PaulWay:
This combiner provides a unified interface for all the Modprobe data in `/etc/modprobe.conf` and `/etc/modprobe.conf.d/*.conf` files.  Each item stays associated with the location it was defined in, so that a rule can easily report on where a line was defined if needed.


insights/combiners/modprobe.py
```diff
@@ -0,0 +1,89 @@
...
+                if section not in self.data:
+                    self.data[section] = {}
+                for name, value in sectdict.iteritems():
+                    if name in self.data[section]:
+                        # append to this module's value - should only
+                        # happen for aliases.
+                        self.data[section][name][0].append(value)
+                    else:
+                        # create new tuple
+                        self.data[section][name] = (value, filename)
```
bfahr: @PaulWay have you considered using a `namedtuple` type to make it a bit clearer regarding the data contents and how to access them?

bfahr:
You also need to add the `.rst` file to include this combiner in the documentation.

insights/combiners/modprobe.py
PaulWay: OK, I've used a named tuple, but the documentation is playing up on me.  I have:
```python
ModProbeValue = namedtuple("ModProbeValue", ['value', 'source'])
"""
A value from a ModProbe source

Attributes:
    value (object): The original value stored in the source file, or a list
        of values if this is an alias.
    source (str): The source file name of the associated value.
"""
```

But it lists the value and source attributes twice.  Is there a better way?

bfahr: When you have an `__init__` method in your class you should document attributes there with docstrings following the attribute initialization instead of in the class docstring.  See these examples in this reference http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html:
```
    #: list of str: Doc comment *before* attribute, with type specified
    self.attr4 = ['attr4']

     self.attr5 = None
     """str: Docstring *after* attribute, with type specified."""
```

bfahr: Sorry, misunderstood what you were asking.  I think you should just not use the Attributes tag.

PaulWay:
@bfahr How are we looking with this now?

PR #431, opened 2017-09-14
Enhance cmdline.py and always store values in list

xiangce:
- Store values into list by default no matter there is only one value or more
- Refine the docstring
- Remove the ".data" access method from example and test


bfahr:
I see in the referenced issue #422 that @jhjaggars recommends just making this change in the master branch so no need for a 1.x PR.

PR #433, opened 2017-09-14
print archive path to stdout when using --keep-archive

SteveHNH:


insights/client/__init__.py
```diff
@@ -402,6 +402,8 @@ def upload(self, path, rotate_eggs=True):
             # delete the archive
             if not config['keep_archive']:
                 client.delete_archive(path)
+            else:
```
jhjaggars: A bit of a nit, but could you reverse the checks so we don't do a negative check first?

`if config['keep_archive']: log else: delete`

SteveHNH:
Reversed the order of if/else.  @jhjaggars

PR #436, opened 2017-09-15
Net stat parser

vishwanathjadhav:
Added parser for kernel exported network states in **_/proc/net/softnet_stat_** file.
This parser will parse the packet_drops per cpu instances, will tell available cpus (cpu_instances)
for  processing of network traffic, and will return column details(packet_drops, time_sqeeze) per cpu.


vishwanathjadhav:
Set a wrong base branch, which resulted into lot of conflicts

PR #437, opened 2017-09-15
Net stat parser

vishwanathjadhav:
Added parser for kernel exported network states in **/proc/net/softnet_stat file**. This parser will parse the packet_drops per cpu instances, will tell available cpus (cpu_instances) for  processing of network traffic, and will return column details(packet_drops, time_sqeeze) per cpu.


insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+
+@parser("softnet_stat")
+class SoftNetStats(Parser):
+    """Parses ``/proc/net/softnet_stat`` file contains"""
+
+    def parse_content(self, contents):
+        cpu = 0
+        data = {}
+        for line in contents:
+            stats = line.split(' ')
```
xiangce: Better use `split(None)`

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+        Get network stats per column for all cpu.
+
+        Arguments:
+            (str): Column name for eg. packet_drops or received_rps.
+
+        Returns:
+            (dict): Column states per cpu.
+        """
+        data = {}
+        for each_cpu in self.data.keys()[1:]:
```
xiangce: Do you want to use the [1:]: to exclude the cpu_instances? The elements' order in dict is hashed, I do not suggest you do so.

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+
+        Arguments:
+            None: No input argument for the function.
+
+        Returns:
+            (int): The number of cpus handling network interrupts.
+        """
+        return self.data['cpu_instances']
+
+    def per_cpu_nstat(self, key):
```
xiangce: Better explain each `key` (line49-line54) in the document.

How about adding an interface that can be used like `nstat_of_cpu[0]` which returns a dict like (just a suggestion)
~~~
{'packet_process': ...,
 packet_drops':...,
 'time_squeeze': ...,
 'cpu_collision': ...,
'received_rps': ...,
 'flow_limit_count': ...}
~~~

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+        0000372f 00000001 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+
+    Examples:
+        >>> SOFTNET_STATS = '''
+        ... 00008e78 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+        ... 000040ee 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+        ... 0001608c 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+        ... 0000372f 00000001 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+        ...'''.strip()
+        >>> softnet_obj = shared[SoftNetStats]
```
PaulWay: Declaring SOFTNET_STATS in the code usually implies that you're going to initialise the parser object yourself, e.g. by `softnet_obj = SoftNetStats(context_wrap(SOFTNET_STATS))`.  If you're pulling the declaration out of the shared data, then you don't need to declare SOFTNET_STATS.  I know that there's both styles in use in the code; I'd recommend using the `context_wrap` method as the examples are written as if you're in an interactive Python shell.

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+from .. import Parser, parser
+
+
+@parser("softnet_stat")
+class SoftNetStats(Parser):
+    """Parses ``/proc/net/softnet_stat`` file contains"""
+
+    def parse_content(self, contents):
+        cpu = 0
+        data = {}
```
PaulWay: Save yourself a declaration here and just use `self.data` throughout.

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+
+            data['cpu' + str(cpu)] = {'packet_process': stats[0],
+                                    'packet_drops': stats[1],
+                                    'time_squeeze': stats[2],
+                                    'cpu_collision': stats[8],
+                                    'received_rps': stats[9],
+                                    'flow_limit_count': stats[10]}
+            cpu = cpu + 1
+        self.data = data
+        self.data['cpu_instances'] = cpu
```
PaulWay: I'm not sure what the purpose of this piece of information.  You can tell the number of CPUs by len(self.data).

If you do want to keep it, I'd also recommend removing that from `self.data` and instead make `cpu_instances` a property - e.g. `self.cpu_instances = cpu`.  Then you don't need a getter method.

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def get_cpu_instaces(self):
```
PaulWay: I'd really prefer to use properties directly rather than getter methods.

insights/parsers/softnet_stat.py
PaulWay: This would also be solved by not keeping the cpu_instances in the data property but making it a property of its own.

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+        data = {}
+        for line in contents:
+            stats = line.split(' ')
+
+            data['cpu' + str(cpu)] = {'packet_process': stats[0],
+                                    'packet_drops': stats[1],
+                                    'time_squeeze': stats[2],
+                                    'cpu_collision': stats[8],
+                                    'received_rps': stats[9],
+                                    'flow_limit_count': stats[10]}
```
PaulWay: Why not capture all the information you can get here?

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+class SoftNetStats(Parser):
+    """Parses ``/proc/net/softnet_stat`` file contains"""
+
+    def parse_content(self, contents):
+        cpu = 0
+        data = {}
+        for line in contents:
+            stats = line.split(' ')
+
+            data['cpu' + str(cpu)] = {'packet_process': stats[0],
```
PaulWay: I'm not sure why this is a dictionary rather than an array based on CPU number, since CPU numbers will always be contiguous.  It also makes it easy to do array comprehensions.

insights/parsers/softnet_stat.py
PaulWay: And I'd really prefer this was a meaningful property called e.g. `stats_for_cpu`.  'Data' is a meaningless term here.  There's no need to use a meaningless name.

insights/parsers/softnet_stat.py
PaulWay: @xiangce - I think that'd be solved if this was simply an array by CPU number.

vishwanathjadhav: Some times cpu instances becomes 2 even though ``nproc`` output it "4" , that depends on scheduling policies used. We will never look for the packet drops for let us say "cpu0" we can not predict such thing. we  will look for if there is over all packet drop (or any other non zero stats) on either of the processor, if there is drop then take corresponding action.

insights/parsers/softnet_stat.py
vishwanathjadhav: Sure thing

insights/parsers/softnet_stat.py
vishwanathjadhav: Made it this way because based on the rows in file it doesn't tell what the row implies .

insights/parsers/softnet_stat.py
```diff
@@ -17,20 +17,27 @@
...
+
     Examples:
         >>> SOFTNET_STATS = '''
         ... 00008e78 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
         ... 000040ee 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
         ... 0001608c 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
         ... 0000372f 00000001 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
         ...'''.strip()
-        >>> softnet_obj = shared[SoftNetStats]
+        >>> softnet_obj = context_wrap(SOFTNET_STATS)
```
xiangce: Better make it more correct as `SoftNetStats(context_wrap(SOFTNET_STATS))` :)

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,99 @@
...
+
+            nstats[cpu] = {'packet_process': stats[0],
+                            'packet_drops': stats[1],
+                            'time_squeeze': stats[2],
+                            'cpu_collision': stats[8],
+                            'received_rps': stats[9],
+                            'flow_limit_count': stats[10]}
+            cpu = cpu + 1
+        self.cpu_instances = cpu
+        self.data = nstats
```
xiangce: I agree with @PaulWay, that to use a meaningful property name here. 

And I think it would be better to store the status of each cpu into a list (which has natural order to indicate the cpu number). So you can simplify the `per_cpu_nstat` method by iterating the list instead of iterating the dict (without order) and `inserting` each cpu-value separately.

E.g:
~~~
         nstats = []
         for line in contents:
             stats = line.split(None)
 
             nstats.append({'packet_process': stats[0],
                            'packet_drops': stats[1],
                            'time_squeeze': stats[2],
                            'cpu_collision': stats[8],
                            'received_rps': stats[9],
                            'flow_limit_count': stats[10]})
         self.cpu_instances = len(nstats)
         self.nstat_per_cpu = nstats
~~~

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        It will check for if there is packet drop occurred on any cpu.
+
+        Arguments:
+            None: No input argument for the function
+
+        Returns:
+            (bool): It will return True if observed packet drops.
+        """
+        for each_cpu in self.data.keys()[1:]:
```
xiangce: There is an unsuggested `[1:]` too.

And by replacing it to a `list`, this can be fixed.

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,99 @@
...
+                            'packet_drops': stats[1],
+                            'time_squeeze': stats[2],
+                            'cpu_collision': stats[8],
+                            'received_rps': stats[9],
+                            'flow_limit_count': stats[10]}
+            cpu = cpu + 1
+        self.cpu_instances = cpu
+        self.data = nstats
+
+    def per_cpu_nstat(self, key):
```
xiangce: In addition to this interface, do you think it would be better to add another one such as `nstat_per_cpu[0]`.

Using `list` instead of `dict` to store the `nstats` with the meaningful name can solve this.

vishwanathjadhav: yes its is important because we are going to return a column let's say packet_drops for all cpu , so "key" will be packet_drops , packet_drops  per cpu

insights/parsers/softnet_stat.py
vishwanathjadhav: sure thing

insights/parsers/softnet_stat.py
vishwanathjadhav: Addressed, Please check

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,100 @@
...
+        >>> softnet_obj.per_cpu_nstat('packet_drops')
+        [0, 0, 0, 1]
+"""
+
+from .. import Parser, parser
+
+
+@parser("softnet_stat")
+class SoftNetStats(Parser):
+    """Parses ``/proc/net/softnet_stat`` file contains"""
```
xiangce: Please include docstrings for the attributes `cpu_instances` and `nstat_per_cpu`

insights/parsers/softnet_stat.py
```diff
@@ -93,7 +93,8 @@ def is_packet_drops(self):
         Returns:
             (bool): It will return True if observed packet drops.
         """
-        for each_cpu in self.data.keys()[1:]:
-            if self.data[each_cpu]['packet_drops']:
+        for each_cpu in self.nstat_per_cpu:
+            value = each_cpu.get('packet_drops', '_')
```
xiangce: Here you should replace the '_' with None, or just leave it as `each_cpu.get('packet_drops')`

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,104 @@
...
+softnet_stat - /proc/net/softnet_stat
+=====================================
+
+This parser parses the stats from network devices. These stats includes events
+per cpu(in row), number of packets processed i.e packet_process (first column),
+number of packet drops packet_drops (second column), time squeeze i.e net_rx_action
+performed time_squeeze(third column), cpu collision i.e collision occur while obtaining
+device lock while transmitting cpu_collision packets (Ninth column), received_rps
+number of times cpu woken up received_rps (tenth column), number of times reached flow
+limit count flow_limit_count (eleventh column).
```
PaulWay: I would strongly recommend you add something here about how these stats are somewhat undocumented and may change - e.g. from https://blog.packagecloud.io/eng/2016/06/22/monitoring-tuning-linux-networking-stack-receiving-data/#procnetsoftnetstat:

> These statistics are output to a file in proc: /proc/net/softnet_stat for which there is, unfortunately, very little documentation. The fields in the file in proc are not labeled and could change between kernel releases.

I agree that it's unlikely that the positions of those values will change in the short term, so I don't think it's worth (at this stage) trying to work out which version of the kernel you're extracting data from and trying to cope with field source changes over time.  But it'd be good to have a caveat there.

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,104 @@
...
+        Arguments:
+            (str): Column name for eg. packet_drops or received_rps.
+
+        Returns:
+            (array): Column states per cpu.
+        """
+        nstats = []
+        for each_cpu in self.nstat_per_cpu:
+            value = each_cpu.get(key, '_')
+            nstats.append(int(value) if value.isdigit() else None)
```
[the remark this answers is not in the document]
xiangce: @PaulWay - `int(value, 16)` good point!

xiangce:
Other than the `int(value, 16)` that Paul mentioned, looks good to me.

lonicerae:
For RH's kernel, the output of `softnet_stat` is the same as for RHEL6 ~ RHEL7, and total columns are 10.

https://access.redhat.com/labs/psb/versions/kernel-3.10.0-693.el7/net/core/net-procfs.c#line146
https://access.redhat.com/labs/psb/versions/kernel-3.10.0-123.el7/net/core/net-procfs.c#line146

```
static int softnet_seq_show(struct seq_file *seq, void *v)
{
        struct softnet_data *sd = v;

        seq_printf(seq, "%08x %08x %08x %08x %08x %08x %08x %08x %08x %08x\n",
                   sd->processed, sd->dropped, sd->time_squeeze, 0,
                   0, 0, 0, 0, /* was fastroute */
                   sd->cpu_collision, sd->received_rps);
        return 0;
}
```

https://access.redhat.com/labs/psb/versions/kernel-2.6.32-696.el6/net/core/dev.c#line4438
https://access.redhat.com/labs/psb/versions/kernel-2.6.32-131.21.1.el6/net/core/dev.c#line3573
```
static int softnet_seq_show(struct seq_file *seq, void *v)
{
        struct netif_rx_stats *s = v;

        seq_printf(seq, "%08x %08x %08x %08x %08x %08x %08x %08x %08x %08x\n",
                   s->total, s->dropped, s->time_squeeze, 0,
                   0, 0, 0, 0, /* was fastroute */
                   s->cpu_collision, s->received_rps);
        return 0;
}
```

although upstrem can extend the usage of unused items or add extra column:

https://github.com/torvalds/linux/blob/v3.13/net/core/net-procfs.c#L161-L165
```
seq_printf(seq,
	   "%08x %08x %08x %08x %08x %08x %08x %08x %08x %08x %08x\n",
	   sd->processed, sd->dropped, sd->time_squeeze, 0,
	   0, 0, 0, 0, /* was fastroute */
	   sd->cpu_collision, sd->received_rps, flow_limit_count);
```
I think currently we can add a comment telling people that this parser is compatible for RHEL6 and RHEL7.

insights/parsers/softnet_stat.py
```diff
@@ -80,13 +81,15 @@ def per_cpu_nstat(self, key):
...
-            value = each_cpu.get(key, '_')
-            nstats.append(int(value) if value.isdigit() else None)
+        if not self.cpu_nstats:
+            return []
+
+        if key not in self.cpu_nstats[0]:
+            return []
 
-        return nstats
+        return [int(cpu[key], 16) for cpu in self.cpu_nstats]
```
xiangce: Just came into my mind, it would be better to shift the calculating/casting to `parse_content`, since each column must be a hex-number-string. This will save the workload and avoid calculating/casting it for every calling of `per_cpu_nstat`.

How do you think?

~~~
     def parse_content(self, contents):
         nstats = []
          for line in contents:
              stats = [int(st,16) for st in line.split(None)]
              self.cpu_nstats.append({'packet_process': stats[0],
                              'packet_drops': stats[1],
                              'time_squeeze': stats[2],
                              'cpu_collision': stats[8],
                              'received_rps': stats[9],
                              'flow_limit_count': stats[10]})
~~~
Or, if you want skip the first column "packet_process" (I do not familiy with it)
~~~
              stats = line.split(None)
              stats = [int(st,16) for st in stats[1:]]
~~~

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,106 @@
...
+        """ Total number of CPUs listed in the stats records"""
+        self.cpu_instances = 0
+
+        """ List of network stats per cpu instace """
+        self.cpu_nstats = []
+        super(SoftNetStats, self).__init__(context)
+
+    def parse_content(self, contents):
+        for line in contents:
+            stats = line.split(None)
```
PaulWay: Don't need this line if you're then using the list comprehension to compile the stats line :-)

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,106 @@
...
+
+        Column-01: packet_process: Packet processed by each CPU.
+        Column-02: packet_drop: Packets dropped.
+        Column-03: time_squeeze: net_rx_action.
+        Column-08: cpu_collision: collision occur while obtaining device lock while transmitting.
+        Column-09: received_rps: number of times cpu woken up received_rps.
+        Column-10: flow_limit_count: number of times reached flow limit count.
+
+    .. Note::
+                There is very less doc about the fields in the file, columns are not labeled and
```
PaulWay: I'd say 'minimal documentation' instead of 'very less doc' :-)

insights/parsers/softnet_stat.py
```diff
@@ -0,0 +1,105 @@
...
+limit count flow_limit_count (eleventh column).
+
+    The typical contents of ``cat /proc/net/softnet_stat`` file is as below::
+
+        00008e78 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+        000040ee 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+        0001608c 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+        0000372f 00000001 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
+
+        Column-01: packet_process: Packet processed by each CPU.
```
lonicerae: The column description listed here conflicts with that in line 6~11:

```
number of packets processed i.e packet_process (first column),
number of packet drops packet_drops (second column), time squeeze i.e net_rx_action
performed time_squeeze(third column), cpu collision i.e collision occur while obtaining
device lock while transmitting cpu_collision packets (Ninth column), received_rps
number of times cpu woken up received_rps (tenth column), number of times reached flow
limit count flow_limit_count (eleventh column).
```

Which is correct?

PR #443, opened 2017-09-15
changed local addr parsing to account for ipv6 addrs

jeudy100:


PaulWay:
Great one @jeudy100 - thanks!  Looks good to me.

PR #445, opened 2017-09-18
Rename parse_table to parse_delimited_table and move to parsers for 1.x branch

PaulWay:
`insights.util.parse_table` and `insights.parsers.parse_fixed_table` both do more or less the same thing: parse a table from a heading string and lines with fields into a list of dictionaries, keyed on the headings.  Though they work on different inputs, but they should both be in the `parsers` utility module.

This update renames `parse_table` to `parse_delimited_table`, gives it some of the same parameters as `parse_fixed_table` for interface homogeneity, and moves it from `insights.util` to `insights.parsers`.  The following parser modules have been updated to use the `parse_delimited_table` function:

* fstab
* ipcs_sem
* lvm
* netstat
* ps
* rhn_schema_stats
* route

All tests pass and there is no loss of test coverage. Tests in the Red Hat rule repositories all pass.

This also includes tidying the arguments and handling of `parse_fixed_table`.


insights/parsers/__init__.py
```diff
@@ -353,6 +350,81 @@ def parse_fixed_table(table_lines,
     return table_data
 
 
+def parse_delimited_table(table_lines,
```
jhjaggars: I think for backward compatibility purposes we should make an alias for this function (and mark it as deprecated when that becomes available)

PaulWay: OK, will do.

PaulWay: Actually, I'm not sure there's any point in that - if I've tidied up everything that used `parse_table`, why provide it again?

jhjaggars: The reason is that there are possibly other callers that you haven't fixed.

PaulWay:
@jhjaggars OK, I think I've put everything aright, and tested it all correctly now.  How does that look?

insights/parsers/__init__.py
bfahr: @PaulWay this method must be explicitly added to the file `docs/api_index.rst` due to the way sphinx builds the parser catalog.

PaulWay:
Hi @bfahr - I've added the parse_delimited_table function into the `docs/api_index.rst` index and improved the documentation a bit - let me know what you think now!

PaulWay:
Updating with minor change from master version and to be compatible with test_util changes.

PR #449, opened 2017-09-18
add cron job creation with --register

SteveHNH:


insights/client/__init__.py
```diff
@@ -596,6 +596,9 @@ def post_update():
 
     if config['register']:
         client.try_register()
+        if os.path.exists('/etc/cron.daily'):
```
jhjaggars: any reason to not combine the conditions?

`if path exists and .set_daily():`

SteveHNH:
No reason not to combine. I've fixed it @jhjaggars

PR #450, opened 2017-09-18
Fix disable schedule when registering

SteveHNH:
The disable schedule flag can now be run alongside `--register` so that the
registration will not automatically enable the cron job.
Also fix `--status` so that it properly exists after running once printing the
status message. Previously only exited if the status was registered.


insights/client/__init__.py
```diff
@@ -574,10 +575,7 @@ def post_update():
         reg_check = registration_check()
         for msg in reg_check['messages']:
             logger.info(msg)
-        if reg_check['status']:
```
kylape: It's not clear to me why these changes are here.  Can you explain?

SteveHNH: @kylape Sure. I found that when I used reg_check['status'], it would exit properly only when the system was registered due to the api check coming back as True. If the api check was False, the die(return_code=1) didn't cause the client to exit. The collect and upload would be attempted even though the box was unregistered. This might all be resolved once the behavior of die() is fixed. 

The die() is in place just to exit regardless of reg_check['status']. In reality it should probably exit regardless of the returned response anyway.

kylape: Yeah that sounds like bug with how phases are handled, not with the logic in this function.  Would you be okay to revert this portion of your change with the assumption that fixing the phases will resolve the issue?

SteveHNH: @kylape fo' sho! I'll do it now

PR #452, opened 2017-09-19
Refine the docstring in test/_init_.py

xiangce:
- `module` will be used as `attributes` by different classes (e.g Krb5Configuration and AllKrb5Conf)
    which will cause documentation making error (conflict of cross-reference).
- `component` is more suitable to indicate the type of the 2nd parameter


xiangce:
This is only for the 1.x branch. The master branch has no this issue.

PR #455, opened 2017-09-19
Add md5 check for configuration file /etc/fonts/fonts.conf

zhangyi733:
* Just add fonts.conf path in pre_command md5chk_files, existing parser md5check.py can meet my requirement


xiangce:
See #451

PR #456, opened 2017-09-19
Add md5 check for /etc/fonts/fonts.conf in insights/specs.py

zhangyi733:


xiangce:
This is a supplement for https://github.com/RedHatInsights/insights-core/pull/451

PR #458, opened 2017-09-19
Add nova crontab parser

chenlizhong:
- add nova_crontab into specs.py
- add NovaCrontab class


xiangce:
See #458

PR #459, opened 2017-09-19
Add parser MongodConf at mongod_conf.py

JoySnow:


insights/config/specs.py
```diff
@@ -220,6 +220,9 @@
     "modinfo"                   : CommandSpec("/usr/sbin/modinfo {module}", module=r"\S+"),
     "modprobe.conf"             : PatternSpec(r"etc/modprobe\.conf"),
     "modprobe.d"                : PatternSpec(r"etc/modprobe.d/.*\.conf"),
+    "mongod.conf"               : SimpleFileSpec("etc/mongod.conf"),
+    "mongod.conf_rh_mongodb26"  : SimpleFileSpec("etc/opt/rh/rh-mongodb26/mongod.conf"),
+    "mongodb.conf"              : SimpleFileSpec("etc/mongodb.conf"),
```
xiangce: 1. As we discussed in today's meeting, better replace the '.' to '_'.
2. You can only declare one `spec` by using the `First` method to achieve your goal.
    Note: If you make sure there might be more specs on one host, please use `All` instead of `First`.
~~~
"mongod_conf": First([SimpleFileSpec("etc/mongod.conf"),SimpleFileSpec("etc/opt/rh/rh-mongodb26/mongod.conf"),SimpleFileSpec("etc/mongodb.conf")]),
~~~

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,93 @@
+"""
+mongodb-server configuration files
+==================================
+
+This module contains the following files:
+    ``/etc/mongod.conf`` and ``/etc/mongodb.conf``.
```
xiangce: Missed the `etc/opt/rh/rh-mongodb26/mongod.conf`?

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,93 @@
...
+from .. import parser, Parser, LegacyItemAccess, get_active_lines
+from ..parsers import ParseException, split_kv_pairs
+
+
+@parser('mongod.conf')
+@parser('mongod.conf_rh_mongodb26')
+@parser('mongodb.conf')
+class MongodbConf(Parser, LegacyItemAccess):
+    """
+    Read the ``/etc/mongod.conf`` YAML file.
```
xiangce: Please update the docstring of the class accordingly. (it can also parse the `key-value` mode, right?)

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def _file_type_is_kv(self, content):
+        """
+        Return True if this file type is key-value. Return False means
+        this file will be handled in YAML format.
+        """
+        content = get_active_lines(content)
+        cnt = sum([1 for line in content if "=" in line])
+        percent = float(cnt) / len(content)
+        return True if percent > 0.9 else False
```
xiangce: Better add more explanation for this logic (e.g. why 90%?).

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,93 @@
...
+    # Enables journaling
+    journal = true
+    # Turn on/off security.  Off is currently the default
+    noauth=true
+
+
+Examples:
+
+    >>> mongod_conf = shared[MongodConf]
+    >>> 'storage' in mongod_conf.data
```
xiangce: Since "LegacyItemAccess" is used, better remove the ".data" from example.

xiangce:
@bfahr - Since the logic is newly updated as per the different modes of this configuration file, please review it again. Thanks.

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,93 @@
...
+
+import yaml
+from .. import parser, Parser, LegacyItemAccess, get_active_lines
+from ..parsers import ParseException, split_kv_pairs
+
+
+@parser('mongod.conf')
+@parser('mongod.conf_rh_mongodb26')
+@parser('mongodb.conf')
+class MongodbConf(Parser, LegacyItemAccess):
```
bfahr: We try to avoid having a single parser handling multiple specs.  See issue #14.  It is better to use a base class and a separate class inheriting from that for each separate spec.

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    def parse_content(self, content):
+        kv_type = self._file_type_is_kv(content)
+        try:
+            if kv_type:
+                self.data = split_kv_pairs(content, use_partition=True)
+            else:
+                self.data = yaml.safe_load('\n'.join(content))
+        except Exception, e:
+            raise ParseException('mongod conf parse faild: %s', e)
```
bfahr: `faild` should be `failed`.

bfahr:
Sorry, clicked approved instead of requested changes.  Please see my comments.

insights/parsers/mongod_conf.py
JoySnow: Yes.

insights/config/specs.py
JoySnow: I will use All(...).

insights/parsers/mongod_conf.py
JoySnow: Sure.

insights/parsers/mongod_conf.py
JoySnow: Done.

insights/parsers/mongod_conf.py
JoySnow: Okay.

insights/parsers/mongod_conf.py
JoySnow: Got it, thanks.

insights/parsers/mongod_conf.py
JoySnow: Yep.

insights/parsers/mongod_conf.py
```diff
@@ -48,9 +50,9 @@
 Examples:
 
     >>> mongod_conf = shared[MongodConf]
-    >>> 'storage' in mongod_conf.data
+    >>> 'storage' in mongod_conf
     True
-    >>> mongod_conf.data['storage']['dbPath']
+    >>> mongod_conf['storage']['dbPath']
```
xiangce: As per my env, it's `dbpath` but not `dbPath`, it's all lower case letter. Better cast them to lower case forcibly.

insights/parsers/mongod_conf.py
```diff
@@ -3,10 +3,12 @@
 ==================================
 
 This module contains the following files:
-    ``/etc/mongod.conf`` and ``/etc/mongodb.conf``.
+    ``/etc/mongod.conf``,
+    ``/etc/mongodb.conf`` ,
+    ``etc/opt/rh/rh-mongodb26/mongod.conf``.
```
xiangce: Why remove `/` from `etc/opt/rh/rh-mongodb26/mongod.conf` but leave it in above two lines?

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,171 @@
...
+            in each line. Use 0.9 rather than 1 here, just in case there're
+            any unexpected lines with wrong settings.
+        """
+        content = get_active_lines(content)
+        cnt = sum([1 for line in content if "=" in line])
+        percent = float(cnt) / len(content)
+        return True if percent < 0.9 else False
+
+    @property
+    def is_yaml(self):
```
xiangce: It's OK to rename `self._is_yaml` to `self.is_yaml` and remove this property interface.

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        Return boolean value to flag if this is a yaml format file.
+        """
+        return self._is_yaml
+
+    @property
+    def bindip(self):
+        if self.is_yaml:
```
xiangce: It's necessary to add corresponding docstring for each attribute.

insights/parsers/mongod_conf.py
```diff
@@ -80,14 +80,15 @@ class MongodbConf(Parser, LegacyItemAccess):
     """
     Parse config file in key-value pair or YAML format.
     Make several frequently used config options as properties.
+    Attribute 'is_yaml' is a boolean value, flags if this is a yaml format file.
```
xiangce: You might want to update the docsting of the attribute of `is_yaml` to list it together with below properties.

You can check it in your local by making the html format document:
~~~
cd docs; make html
google-chrome-stable docs/_build/html/shared_parsers_catalog/mongod_conf.html
~~~

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,197 @@
...
+from .. import parser, Parser, LegacyItemAccess, get_active_lines
+from ..parsers import ParseException, split_kv_pairs
+
+
+@parser('mongod_conf')
+class MongodbConf(Parser, LegacyItemAccess):
+    """
+    Parse config file in key-value pair or YAML format.
+    Make several frequently used config options as properties.
+    Attribute 'is_yaml' is a boolean value, flags if this is a yaml format file.
```
bfahr: I agree with @xiangce if you follow the Google docstring style which is the standard for this project you would instead have:
```
Attributes:
    is_yaml (boolean): Flags is True if this is a yaml format file.
```

insights/parsers/mongod_conf.py
```diff
@@ -0,0 +1,197 @@
...
+        content = get_active_lines(content)
+        cnt = sum([1 for line in content if "=" in line])
+        percent = float(cnt) / len(content)
+        return True if percent < 0.9 else False
+
+    @property
+    def bindip(self):
+        """
+        Return option value of:
+            'net.bindIp' in yaml conf;
```
bfahr: It might be clearer if you use backticks to highlight these values. For instance `net.bindIp`.  Also since this renders as just a single line, you might want to simply use prose to describe it, for instance:
```
Returns option value of `net.bindIp` if a yaml conf and `bind_ip` if a key-value pair conf.
```

JoySnow:
For 3.x: https://github.com/RedHatInsights/insights-core/pull/620

PR #460, opened 2017-09-19
Minimize client egg

kylape:
There are lots of modules that are currently included in the client egg that aren't currently used.  In particular, there are several modules in the `contrib` folder that would incur an additional support cost were we to ship them because they are maintained by a third party.  Thus we are removing as much as possible before we ship.  We can always add modules back as we begin to integrate features of the core into the client.


kylape:
Unfortunately it seems like Python 2.7+ ConfigParser features are used more than we thought.

PR #463, opened 2017-09-19
refactor phases, report names

gravitypriest:


kylape:
Corresponding client changes https://github.com/RedHatInsights/insights-client/pull/17

insights/client/__init__.py
```diff
@@ -503,29 +503,15 @@ def _f():
...
-                potentially to be used as input other phases.
-    """
-    print json.dumps({
-        "message": msg,
-        "rc": rc,
-        "retry": retry,
-        "response": response
-    })
-    sys.exit(0)
+def die(msg=None, rc=0, retry=False, response=None, raw=False):
```
kylape: I figure you aren't done here, but I don't see the point of having this function if it just calls sys.exit

insights/client/__init__.py
```diff
@@ -19,7 +19,7 @@
 
 LOG_FORMAT = ("%(asctime)s %(levelname)s %(message)s")
 APP_NAME = constants.app_name
-logger = logging.getLogger(__name__)
+logger = logging.getLogger(APP_NAME)
```
kylape: I don't see how having each module get the same logger is any better than getting a module-specific one.  You may as well just do `logger = logging.root`.

insights/client/client.py
```diff
@@ -52,14 +53,15 @@ def get_console_handler():
     else:
         target_level = logging.INFO
 
-    stdout_handler = logging.StreamHandler(sys.stderr)
+    stdout_handler = logging.StreamHandler(sys.stdout)
```
kylape: Why not put all logging to stderr and only stderr?  This actually breaks `--to-stdout` since stdout will have log messages mixed in.  Also, under normal operation, error messages get printed to the console twice.

insights/client/__init__.py
```diff
@@ -603,7 +591,8 @@ def post_update():
     # put this first to avoid conflicts with register
     if config['unregister']:
         pconn = client.get_connection()
-        die(rc=0 if pconn.unregister() else 1)
+        pconn.unregister()
+        sys.exit(42)
```
kylape: Seems like there should be a difference between "exit now with rc=0" and "exit now with rc=1", but I don't see one with what you've pushed so far.

insights/client/__init__.py
```diff
@@ -599,12 +580,13 @@ def post_update():
...
             logger.info(msg)
-        die(rc=0 if reg_check['status'] else 1)
+        sys.exit(100 + int(validate_remove_file()))
 
     # put this first to avoid conflicts with register
     if config['unregister']:
         pconn = client.get_connection()
-        die(rc=0 if pconn.unregister() else 1)
+        pconn.unregister()
+        sys.exit(100)
```
kylape: are we not wanting to indicate whether unregistration failed?

insights/client/__init__.py
```diff
@@ -500,49 +499,31 @@ def _f():
...
     # validate the remove file
     if config['validate']:
-        die(rc=0 if validate_remove_file() else 1)
+        sys.exit(100 + int(validate_remove_file()))
 
     # handle cron stuff
     if config['enable_schedule'] and config['disable_schedule']:
         logger.error('Conflicting options: --enable-schedule and --disable-schedule')
-        die(rc=1)
+        sys.exit(101)
```
kylape: Would probably be best to have the return codes set as constants

insights/client/client.py
kylape: Thanks for the changes here.  Shouldn't logging go to stderr, though?

kylape:
Oh and I guess there's a merge conflict now

PR #466, opened 2017-09-20
Dont do rotation in core

brantleyr:


gravitypriest:
squash pls

kylape:
I'm not sure if initializing logging and/or config in InsightsClient is the direction we want to go long term, but for now it's okay.  We can talk about it more when we get to focus more on the InsightsClient API.

PR #471, opened 2017-09-20
remove 'only' flag in favor of 'analyze_image'

brantleyr:


kylape:
Looks like you're using `image_id` instead of `analyze_image` to store the image ID.  I agree that `image_id` is more intuitive than `analyze_image`, but it seems inconsistent.  For example, a user could set `image_id` in their `insights-client.conf`.

Maaybe instead of using `analyze_image` as the destination key, perhaps keep it `--analyze-image` but change the destination to `image_id`?

brantleyr:
So there are kind of two entry points for doing image/container scanning. One, using image_id will do the registration/machine-id stuff and upload to the /uploads endpoint. Two, using analyze_image was our new use-case that bypasses that stuff and goes to /uploads/image.

PR #472, opened 2017-09-20
Fix traceback when --offline JSON fails

SteveHNH:
Add a `try` around the json.loads so that the client fails cleanly if it gets a bad json response from the API on upload.


kylape:
Shouldn't the `--offline` option prevent `upload()` from being called at all?

insights/client/connection.py
```diff
@@ -692,7 +692,8 @@ def upload_archive(self, data_collected, duration, cluster=None):
 
         logger.debug("Upload status: %s %s %s",
                      upload.status_code, upload.reason, upload.text)
-        the_json = json.loads(upload.text)
+        if upload.status_code == 201:
+           the_json = json.loads(upload.text)
         try:
```
kylape: This try/except block should go under the new `if` block as well.  We should probably better handle the case where the `status_code` is not 201.  Not sure if that is done in the calling method or not, but we should somehow indicate that the upload has failed.

insights/client/connection.py
```diff
@@ -692,7 +692,10 @@ def upload_archive(self, data_collected, duration, cluster=None):
 
         logger.debug("Upload status: %s %s %s",
                      upload.status_code, upload.reason, upload.text)
-        the_json = json.loads(upload.text)
+        if upload.status_code == 201:
+            the_json = json.loads(upload.text)
+        else:
+            logger.debug("Upload archive failed with status code  %s" % upload.status)
```
jhjaggars: Should we bail out here?  If `the_json` isn't set then the rest of the function is in a bad state.

kylape:
I believe @brantleyr's change to do `if status_code in (200, 201)` conflicts with your change.  Can you resolve the conflict?

kylape:
Uhh I think I just resolved it in the github UI.  We'll see how this goes...

kylape:
I went ahead and manually pushed this change.

PR #473, opened 2017-09-21
Add parser fc-match

zhangyi733:
* Add Class FCMatch
* Add property fonts_info
* Add testcase
* Add corresponding item in insights/config/specs.py and insights/specs.py


insights/config/specs.py
```diff
@@ -94,6 +94,7 @@
     "exim.conf"                 : SimpleFileSpec("etc/exim.conf"),
     "facter"                    : First([CommandSpec("/usr/bin/facter"),
                                     SimpleFileSpec("facts")]),
+    "fc-match"                  : CommandSpec("/usr/bin/fc-match -f 'sans:regular:roman' -sv | grep -e 'family:' -e 'fontformat'"),
```
xiangce: I don't think it's a good idea to add the pipe "|" into the CommandSpec. Better mark it as `large_content` and filter what you want via `add_filter` in the parser file.

insights/parsers/fc_match.py
```diff
@@ -0,0 +1,54 @@
...
+from .. import Parser, parser
+
+
+@parser("fc-match")
+class FCMatch(Parser):
+    """Class to parse command `/usr/bin/fc-match -f 'sans:regular:roman' -sv | grep -e 'family:' -e 'fontformat'`."""
+
+    def parse_content(self, content):
+        self.data = []
+        font = {}
```
xiangce: Shift this initialization to the begging of the loop, you can save the line-47.

insights/parsers/fc_match.py
```diff
@@ -0,0 +1,54 @@
...
+            k, v = l.strip().split(":")
+            if k == "family":
+                font["family"] = v.strip()
+            else:
+                font["fontformat"] = v.strip()
+                if font not in self.data:
+                    self.data.append(font)
+                    font = {}
+
+    @property
```
xiangce: Instead of this one, better add a `__iter__`  interface to `yield` each item in the list and a `__getitem__` to support `[]` index.

insights/config/specs.py
bfahr: Please check the command syntax, the `-f` switch requires a format argument and you should be able to specify the elements you want after the search pattern.  For instance here's an example on RHEL7:
```
# fc-match -sv "sans:regular:roman" family fontformat
Pattern has 2 elts (size 16)
	family: "DejaVu Sans"(s)
	fontformat: "TrueType"(w)

Pattern has 2 elts (size 16)
	family: "DejaVu Sans"(s)
	fontformat: "TrueType"(w)

Pattern has 2 elts (size 16)
	family: "DejaVu Sans"(s)
	fontformat: "TrueType"(w)
...
``

xiangce:
@bfahr - @zhangyi733 has updated the spec as per your suggestion and it works on both RHEL6&7. I will go ahead and merge it today. Thanks.

zhangyi733:
Thank you @xiangce and @bfahr

PR #477, opened 2017-09-21
Rename with_args_from to foreach in spec_factory

csams:
- Rename `with_args_from` to `foreach` in `SpecFactory`.
- Remove `stored_command` from `SpecFactory`.
- Remove unused `set()` in `generate_api_config.py` that caused incorrect log messages saying no specs were added to the config.
- Remove `package_name` from `APIConfigGenerator` constructor. Plugins are loaded by passing arguments when the script is run.
- Remove unused parameters from `APIConfigGenerator` constructor.
- Refactor filters to remove circular dependencies.


insights/tools/generate_api_config.py
```diff
@@ -15,7 +15,7 @@
 import logging
 import os
 import sys
-from optparse import OptionParser
+import argparse
```
jhjaggars: Argparse doesn't exist in 2.6.

csams:
We decided to resolve this by adding python-argparse to the client rpm spec file and adding a dependency in `setup.py` for virtual environments since argparse is in pypi for python >= 2.3. I've pushed the `setup.py` change, and @kylape is changing the rpm spec for the client.

insights/tools/generate_api_config.py
csams: We decided to resolve this by adding python-argparse to the client rpm spec file and adding a dependency in `setup.py` for virtual environments since argparse is in pypi for python >= 2.3. I've pushed the `setup.py` change, and @kylape is changing the rpm spec for the client.

PR #478, opened 2017-09-21
Fix authentication lookuperror and fix --force-register typo

SteveHNH:
Change the authentication failure to a logger.error instead of a LookupErrorso it looks better in the console.
Fix a typo referring customer to a non-existent `--force-register` switch.


insights/client/connection.py
```diff
@@ -459,7 +459,8 @@ def branch_info(self):
         response = self.session.get(self.branch_info_url)
         logger.debug("GET branch_info status: %s", response.status_code)
         if response.status_code != 200:
-            raise LookupError("Bad status from server: %s" % response.status_code)
+            logger.error("Bad status from server: %s" % response.status_code)
```
jhjaggars: Loggers use varargs to defer interpolation of values so please do the following:

`logger.error("Bad status from server: %s", response.status_code)`

PR #481, opened 2017-09-21
remove update collection rules flag

brantleyr:


kylape:
Can you still disable the updates in the config file?

brantleyr:
Now adheres to auto_update on config for updating rules. "Offline" flag now implies auto_update=False.

PR #482, opened 2017-09-21
unwrap the api response

brantleyr:


kylape:
Looks like `insights/client/__init__.py` line 397 needs to be updated because it assumes the old data structure is in place.

brantleyr:
The call to upload() has been reworked to expect the correct response now.

PR #492, opened 2017-09-26
Add `get_dir` method to find mount point for given directory

PaulWay:
The `get_dir` method was recently added to the DiskFree parsers and is also needed for Mount.

Added code is fully tested, and input data corrected to have a valid root mount.


PaulWay:
I agree about sharing the get_dir code; do you mind if I do that refactoring in a separate merge request?

Regarding the idea of basing Mount and DiskFree on common code; that's difficult.  They provide their data in somewhat similar ways, but the data is a bit different.  I think it'd be better to do any tidy-up like that as a separate job.

xiangce:
@PaulWay - OK, it's ready to merge after applying the same patch to the master branch.

PR #493, opened 2017-09-26
Mount keyword search for 1.x branch

PaulWay:
The list of mount points needs a `search` method that can take keywords, e.g.:

```python
mounts.search(mount_type='ext4')
```

This merge also adds throwing an IndexError exception if the index given is not a string or integer.

Tests now at 100% code coverage.


insights/parsers/mount.py
```diff
@@ -118,3 +120,25 @@ def parse_content(self, content):
             self.rows.append(entry)
             if match:
                 self.mounts[mount['mount_point']] = entry
+
+    def search(self, **kwargs):
+        """
+        Returns a list of the mounts (in order) matching the given criteria.
+        Keys are searched for directly - see the ``keyword_search`` utility
```
bfahr: @PaulWay it would be helpful to add a reference to this function instead of just highlighting the name, for instance:
```
:py:func:`insights.parsers.keyword_search`
```

insights/parsers/mount.py
```diff
@@ -89,6 +89,8 @@ def __getitem__(self, idx):
             return self.rows[idx]
         elif isinstance(idx, str):
             return self.mounts[idx]
+        else:
+            raise IndexError("Mounts can only be indexed by mount string or line number")
```
bfahr: Would it make more sense to use a `TypeError` here? https://docs.python.org/2.7/library/exceptions.html#exceptions.TypeError

PaulWay: That's a judgement call - I think you're probably right though.  I'll do that - thanks Bob.

insights/parsers/mount.py
PaulWay: OK, done!

insights/parsers/mount.py
PaulWay: Yep, I think that's a better choice.  Thanks @bfahr !

xiangce:
Paul has modified the patch as per Bob's suggestion, I go ahead and merge it before @bfahr approving these changes.

PR #497, opened 2017-09-26
Removing redundant config logging

kylape:


brantleyr:
Do we dump the config in another location? :eyes: for it now but can't find it.

brantleyr:
If so, or not, we should consolidate all config dumping in compile_config().

kylape:
@brantleyr https://github.com/RedHatInsights/insights-core/blob/2e4c9ba4cc2c173e806b6d16b2f976daa1eb3bd2/insights/client/__init__.py#L477

brantleyr:
@kylape do we call that anywhere?

kylape:
@brantleyr https://github.com/RedHatInsights/insights-core/blob/2e4c9ba4cc2c173e806b6d16b2f976daa1eb3bd2/insights/client/__init__.py#L585

brantleyr:
@kylape but would we want to call this during every phase instead of just post_update? Would we ever run into a funky case where phases might get a different config for some crazy reason?

kylape:
@brantleyr the main case where the config might be different between phases is if something about the config changed when updating the egg.  So the first two phases (pre_update and update) might be different the the ones after them.

But I don't think printing the config in every phase is worth being able to see potential differences between phases.  It really gunks up the logs, and the config being different between phases should be a very edgy edge case.

PR #498, opened 2017-09-27
Improvements to the LSBlock and LSBlockPairs parsers for 1.x branch

PaulWay:
* Minor improvements to documentation.
* BlockDevice objects now have an `iteritems` method.
* BlockDevice objects use get on their explicit data dictionary, not the implicit one.
* BlockDevice objects will raise a ParseException if TYPE is not found in the LSBlockPairs data, rather than `assert`ing.
* BlockDevices now have a `search` method that uses `keyword_search` to search the device list for devices matching a given set of criteria.
* LSBlockPairs objects now have the same `device_data` attribute that LSBlock objects have.
* Test data is now supplied as strings rather than lists.
* Tests of all the above functionality.
* Tests now at 100% code coverage.


insights/parsers/lsblk.py
```diff
@@ -138,8 +145,10 @@ def __repr__(self):
             )
         else:
             # As long as the regular expression in LsBlock works, we must end
-            # up with NAME and TYPE records here.
-            assert 'TYPE' in self.data
+            # up with NAME and TYPE records here.  However, the LSBlockPairs
+            # parsing isn't able to be that strict.
+            if 'TYPE' not in self.data:
+                raise ParseException("TYPE not found in LsBlockPairs line")
```
bfahr: @PaulWay its not clear why `__repr__` should raise an exception?  If the data is not parsed correctly shouldn't the exception be thrown before the parser data is returned from `parse_content`?

insights/parsers/lsblk.py
```diff
@@ -270,3 +302,4 @@ def str2bool(s):
                     d[replace] = transform(d[original]) if transform else d[original]
                     del d[original]
             self.rows.append(BlockDevice(d))
+        self.device_data = {dev.name: dev for dev in self.rows}
```
bfahr: Please add this attribute to the docs.

insights/parsers/lsblk.py
PaulWay: Yeah, that's probably better - I think this may be the first parser that I know of that checks the correctness of the data after all the data has been read (rather than raising a ParseException during the read process).

insights/parsers/lsblk.py
PaulWay: I've moved this documentation up into the BlockDevices as it now applies to both LsBlock and LsBlockPairs.

bfahr:
Approved pending submission of PR for 2.x branch.

PR #499, opened 2017-09-27
Add `get_dir` method to find mount point for given directory for master branch

PaulWay:
The `get_dir` method was recently added to the DiskFree parsers and is also needed for Mount.

Added code is fully tested, and input data corrected to have a valid root mount.


xiangce:
See #492

PR #511, opened 2017-09-27
Add net_debug for registration check

kylape:


insights/client/connection.py
```diff
@@ -562,7 +562,9 @@ def api_registration_check(self):
         logger.debug('Checking registration status...')
         machine_id = generate_machine_id()
         try:
-            res = self.session.get(self.api_url + '/v1/systems/' + machine_id, timeout=10)
+            url = self.api_url + '/v1/systems/' + machine_id
+            net_logger.info("GET %s", url)
+            res = self.session.get(url, timeout=10)
```
brantleyr: Just being nitpicky. Should we log the `res` response as well?

kylape: Not a bad idea, but we aren't doing it elsewhere.  Maybe a future enhancement to the --net-debug option?

brantleyr: :+1:

PR #512, opened 2017-09-27
to stdout flag should not imply quiet

brantleyr:


kylape:
I think you should just remove that whole line because it looks like `quiet` will be set when it shouldn't.

brantleyr:
Removed. o7

PR #515, opened 2017-09-27
expand phases to have run_as_root attribute

gravitypriest:


insights/client/phase/v1.py
```diff
@@ -34,10 +34,16 @@ def _f():
 
 
 def get_phases():
-    return ['pre_update',
-            'update',
-            'post_update',
-            'collect_and_output']
+    return [{
+        'name': 'pre_update'
```
kylape: What do you think about adding `run_as_root` to all phase dicts?

PR #519, opened 2017-09-28
Add QemuConf class to parse /etc/libvirt/qemu.conf

chenlizhong:


insights/parsers/qemu_conf.py
```diff
@@ -0,0 +1,54 @@
...
+    "/dev/random", "/dev/urandom",
+    "/dev/ptmx", "/dev/kvm", "/dev/kqemu",
+    "/dev/rtc","/dev/hpet", "/dev/vfio/vfio"
+    ]
+
+Example:
+    >>> config = shared[QemuConf]
+    >>> config.get('vnc_listen')
+    '0.0.0.0'
+    >>> config.getint('vnc_tls')
```
xiangce: `getint`?

insights/parsers/qemu_conf.py
```diff
@@ -0,0 +1,54 @@
+"""
+QemuConf - file ``/etc/libvirt/qemu.conf``
+=============================================
```
xiangce: the underlying '===' is too long

insights/parsers/qemu_conf.py
```diff
@@ -0,0 +1,54 @@
...
+        _lines = get_active_lines(content, comment_char="#")
+        self.data = {}
+        current_key = ""
+        for line in _lines:
+            if "=" in line:
+                k, v = line.split('=')
+                current_key = k.strip()
+                self.data[current_key] = v.strip("\"' ")
+            else:
+                self.data[current_key] = self.data[current_key] + line
```
xiangce: This will store it into a long string like "['/dev/', 'sxxx/']", how about refactor it to a list?

PR #520, opened 2017-09-28
New parser httpd_limits

xiangce:


insights/config/specs.py
```diff
@@ -410,6 +411,7 @@
     "docker_images"             : "/usr/bin/docker images -q",
     "dumpdev"                   : "/bin/awk '/ext[234]/ { print $1; }' /proc/mounts",
     "getblockschedulers"        : "for device in $(ls /sys/block); do echo /sys/block/$device/queue/scheduler; done",
+    "httpd_pid"                 : "/bin/cat /etc/httpd/run/httpd.pid",
```
chenlizhong: we just check the pid under the default path  "/etc/httpd/run/httpd.pid", if customer change the "server root" or log and pid location, we can not get the pid file. 

Can we use ps to get pid for httpd? is there a robust way?

PR #526, opened 2017-09-29
Enhance the `limits` parser for `httpd`

xiangce:
- Make a base class `ProcLimits`
- Modify the pre-command of `httpd_pid` more accurate
- Provide attribute for each resource and the `Hard Limit` as its value


xiangce:
Hi, @bfahr - Thank you so much for your suggestion for this parser. I improved it after a discussion with @koalakangaroo. Would you please review one more time.

Btw, since we would like to finish the code of a rule which depends on this parser before our holiday (Oct.1 to Oct.8), would you mind to finish your review on your Friday? Thanks a lot.

insights/config/specs.py
```diff
@@ -414,7 +414,7 @@
     "docker_images"             : "/usr/bin/docker images -q",
     "dumpdev"                   : "/bin/awk '/ext[234]/ { print $1; }' /proc/mounts",
     "getblockschedulers"        : "for device in $(ls /sys/block); do echo /sys/block/$device/queue/scheduler; done",
-    "httpd_pid"                 : "ps aux | grep httpd | grep -v grep | head -1 | awk '{print $2}'",
+    "httpd_pid"                 : "ps aux | grep /usr/sbin/httpd | grep -v grep | head -1 | awk '{print $2}'",
```
PaulWay: I was going to suggest that this wasn't as accurate as the suggestion from https://access.redhat.com/solutions/1346533 to use:

```bash
systemctl show httpd -p MainPID | cut -d = -f 2
```
But that won't work on RHEL 6 systems.  I'm not keen on running long pipelines of shell commands to get data, but at the moment we don't seem to have a good way to get this kind of information in any other way, so I think that's OK.

insights/parsers/proc_limits.py
```diff
@@ -0,0 +1,120 @@
...
+        >>> proc_limits.max_processes
+        '99'
+        >>> 'max_cpu_time' in proc_limits
+        True
+        >>> proc_limits.max_cpu_time
+        'unlimited'
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.max_cpu_time = ''
```
PaulWay: This declaration stuff seems completely unnecessary, as I think the meanings of the particular limits aren't particularly relevant here.  I really applaud you setting properties on the object rather than relying simply on the data property.  To me it seems best to remove all the declarations from `__init__`, and just populate the attributes directly exactly as you've done in `parse_content`.

PaulWay: I can also understand that you want the properties to be set even if the file is incomplete - but I'm not sure that's a likely scenario with files in `/proc`.  And if a new limit is introduced, this will be inaccurate as a listing of properties anyway.

insights/config/specs.py
xiangce: Just as you said, the way I used, is to make it compatible for both RHEL6 and RHEL7, Thanks @PaulWay

insights/parsers/proc_limits.py
xiangce: I would like to keep them `__init__` :). And will update the docsting in case there will be some new resources on different system/machine.

xiangce: OK, makes sense to me. will do it Thanks Paul

insights/parsers/proc_limits.py
```diff
@@ -62,9 +76,16 @@ def __iter__(self):
 
     def parse_content(self, content):
         self.data = []
-        self.max_processes = ''
         if 'No such file or directory' not in content[0]:
             self.data = parse_fixed_table(content, header_substitute=HEADER_SUBSTITUTE)
```
bfahr: @xiangce  should you throw a `ParseException` here to indicate that no data is available?

insights/parsers/proc_limits.py
```diff
@@ -62,9 +76,16 @@ def __iter__(self):
...
         for row in self.data:
-            if 'Max processes' == row['Limit']:
-                self.max_processes = row['Hard_Limit']
+            row['Limit'] = row['Limit'].lower().replace(' ', '_')
+            setattr(self, row['Limit'], row['Hard_Limit'])
```
bfahr: Is the *Units* column important here, for example if *Hard Limit* is 4096, would a user of the value need to know the units?

insights/config/specs.py
```diff
@@ -414,7 +414,7 @@
     "docker_images"             : "/usr/bin/docker images -q",
     "dumpdev"                   : "/bin/awk '/ext[234]/ { print $1; }' /proc/mounts",
     "getblockschedulers"        : "for device in $(ls /sys/block); do echo /sys/block/$device/queue/scheduler; done",
-    "httpd_pid"                 : "ps aux | grep httpd | grep -v grep | head -1 | awk '{print $2}'",
+    "httpd_pid"                 : "/bin/ps aux | grep /usr/sbin/httpd | grep -v grep | head -1 | awk '{print $2}'",
```
bfahr: Its not clear how this change is related to the other changes in this PR? Should this be a separate PR?

insights/parsers/proc_limits.py
xiangce: @bfahr - Thanks.
I think it's OK to throw a `ParseException`.
However, for the hosts which are not running `httpd`, the pre_command `httpd_pid` will return empty. I'm not sure in that case if the CommandSpec will be executed by our client. (In that case the CommandSpec will be `cat /proc//limits` which always results in "cat: /proc//limits: No such file or directory".) If the CommandSpec continues to be executed, there will be a lot of `ParseException` raised. Does this make sense?

insights/parsers/proc_limits.py
xiangce: It depends. I will add it together with the soft_limit. Thanks.

insights/config/specs.py
xiangce: @bfahr - This `httpd_pid` is the pre_command to get the PID of httpd for the `CommandSpec` of this parser.

insights/parsers/proc_limits.py
bfahr: If there will be no data in the output then either an exception should be thrown to let the core know that data should not be stored.  We are trying to avoid the situation where rules have to get the parser data and then have to check for `None`.

insights/config/specs.py
bfahr: Got it, I should have paid attention to the line number.

insights/parsers/proc_limits.py
xiangce: Got it. Will do it.

PR #527, opened 2017-09-29
Add nova_compute_log parser

chenlizhong:


xiangce:
@chenlizhong - Since there is another parser for `nova` which also bases on `LogFileOupout`, would you please to shift them to one python file `nova_log.py` (just like, `httpd_log` and `foreman_log`)?
And since there was some rule used the old `nova_api_log` already, we can only do the `shifting` in the master branch.

insights/parsers/nova_log.py
```diff
@@ -0,0 +1,58 @@
...
+            elif type(keywords) == str and keywords in l:
+                r.append(l)
+        return r
+
+
+@parser('nova-compute.log')
+class NovaComputeLog(LogFileOutput):
+    """Class for parsing the ``/var/log/nova/nova-compute.log`` file.
+
+    Note:
```
xiangce: Please modify the `Note` to sphinx style:
http://www.sphinx-doc.org/en/stable/markup/para.html#directive-note

insights/parsers/nova_log.py
```diff
@@ -0,0 +1,58 @@
...
+This module contains classes to parse logs under ``/var/log/nova/``
+"""
+from .. import LogFileOutput, parser
+
+
+@parser('nova-api_log')
+class NovaApiLog(LogFileOutput):
+    """Class for parsing the ``/var/log/nova/nova-api.log`` file.
+
+    Note:
```
xiangce: Please modify the Note to sphinx style:
http://www.sphinx-doc.org/en/stable/markup/para.html#directive-note

PR #530, opened 2017-09-29
Add parser package_provides_java and combiner running_java_package

zhangyi733:
* Add parser package_provides_java
* Add combiner running_java_package
* Add items in insights/config/specs.py and insights/specs.py
* Add rst files


xiangce:
In spite of my comments inside the `File changed` page,  as per our discussion, please try the new `pre_command`

~~~
for jp in `/bin/ps auxwww | grep java| grep '\-Dcatalina.base'|   awk -F' ' '{print $11}' | sort -u`; do readlink -f $jp; done | sort -u
~~~

zhangyi733:
Thank you @xiangce, as we discussed, I will use the new pre_command to implement parser again.

insights/parsers/package_provides_java.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+
+from insights import parser, Parser
+
+
+@parser('package_provides_java')
+class PackageProvidesJava(Parser):
+    def parse_content(self, content):
+        """
+        Parse the output of the ``rpm -qf package`` command.
```
bfahr: Same comment at above, `package` is something very specific here.

insights/parsers/package_provides_java.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        self.data = None
+        for line in content:
+            if 'not owned by any package' not in line:
```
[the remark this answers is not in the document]
bfahr: If you are unable to determine the results here, then an exception should be raised instead of an empty string or `None`.  See the documentation on exceptions here: http://insights-core.readthedocs.io/en/latest/exception_model.html#parsing-and-exceptions

insights/parsers/package_provides_java.py
```diff
@@ -0,0 +1,36 @@
...
+PackageProvidesJava - Command ``rpm -qf``
+=========================================
+
+This parser handles ``rpm -qf`` output.
+
+Typical contents of the ``rpm -qf package`` command::
+
+    java-1.8.0-openjdk-1.8.0.144-0.b01.el7_4.x86_64
+
+Examples:
```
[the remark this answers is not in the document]
xiangce: Java has many different versions and is so flexible that some of them can be installed at the same time on the same system. Some customers might not use the default java (/usr/bin/java) for some special application, we have to get the full path of each specific java process and use `rpm -qf` to check which package it belongs to.

 And since `readlink` can only check one symbol line once, so we have to put it into a loop.

Currently, we'd like to use the following pre_command to get the 
~~~
for jp in `/bin/ps auxwww | grep java | grep '\-Dcatalina.base'| awk '{print $11}' | sort -u`; do readlink -e $jp; done | sort -u
~~~

I totally agree with you that we'd better add more detail documentation to explain this. 

@zhangyi733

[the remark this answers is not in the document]
bfahr: That was the source of my confusion, if the doc is clearer as to the purpose of this parser then that would help significantly.

zhangyi733:
Thank you @xiangce and @bfahr for your review and comments. After a further discussion with Xiangce, we decide to and use pre_command "for jp in `/bin/ps auxwww | grep java | grep -v grep| awk '{print $11}' | sort -u`; do echo $jp `readlink -e $jp | xargs rpm -qf`; done".
The reason why using above pre_command is that we need to record the links between running_java_command and package which provides the java command. In ``ps aux`` output, we can only get what java command starts a java application, instead of java package.
Through this way, when there is jdk bug, we can detect whether a running java application will be affected.

insights/config/specs.py
```diff
@@ -410,6 +411,7 @@
     "iface"                     : "/sbin/ip -o link | awk -F ': ' '/.*link\\/ether/ {print $2}'",
     "md5chk_files"              : "/bin/ls -H /usr/lib*/{libfreeblpriv3.so,libsoftokn3.so} /etc/pki/product*/69.pem /etc/fonts/fonts.conf /dev/null 2>/dev/null",
     "module"                    : "/bin/ls /sys/module",
+    "java_command_package"      : "for jp in `/bin/ps auxwww | grep java | grep -v grep| awk '{print $11}' | sort -u`; do echo $jp `readlink -e $jp | xargs rpm -qf`; done",
```
xiangce: Since there are two parts for the `echo` in the pre_command, better add quotes to surround the two parts.

``echo "$jp `readlink -e $jp | xargs rpm -qf`"``

insights/parsers/package_provides_java.py
```diff
@@ -0,0 +1,49 @@
...
+    /usr/lib/jvm/jre/bin/java java-1.8.0-openjdk-headless-1.8.0.141-3.b16.el6_9.x86_64
+
+Parsed result::
+
+    {"command": "/usr/lib/jvm/jre/bin/java", "package": "java-1.8.0-openjdk-headless-1.8.0.141-3.b16.el6_9.x86_64"}
+
+Examples:
+
+    >>> command_package = shared[PackageProvidesJava]
+    >>> command_package.data
```
xiangce: Since it's already mixin with the `LegacyItemAccess`, better avoid using the `.data` in the example.

insights/parsers/package_provides_java.py
```diff
@@ -0,0 +1,49 @@
...
+    do echo $jp `readlink -e $jp | xargs rpm -qf`; done``.
+    """
+
+    def parse_content(self, content):
+        comamnd_package_dict = {}
+        for line in content:
+            l = line.split()
+            if len(l) == 2:
+                comamnd_package_dict = ({"command": l[0], "package": l[1]})
+        self.data = comamnd_package_dict
```
xiangce: Better add two `Attributes` directly.

self.command = l[0]
self.package = l[1]

Then you do not need to mixin LegacyItemAccess

insights/config/specs.py
zhangyi733: As we discussed, will not change it.

insights/parsers/package_provides_java.py
zhangyi733: Have updated according to this comment.

insights/parsers/package_provides_java.py
zhangyi733: Have updated according to this comment.

insights/parsers/package_provides_java.py
```diff
@@ -2,48 +2,52 @@
...
     """
 
     def parse_content(self, content):
-        comamnd_package_dict = {}
-        for line in content:
-            l = line.split()
-            if len(l) == 2:
-                comamnd_package_dict = ({"command": l[0], "package": l[1]})
-        self.data = comamnd_package_dict
+        l = content[0].split()
```
xiangce: I'm sorry for the late comment, you should process the situation when there is not Java is running on some hosts. In this case, the pre_command will return nothing.

Better raise a ParseException for it. for the docstring, pls refer to`insights/parsers/proc_limits.py`

insights/parsers/package_provides_java.py
```diff
@@ -0,0 +1,53 @@
+"""
+PackageProvidesJava - Command
+=============================
+
+    This parser reads the output from pre_command::
+
+        ``for jp in `/bin/ps auxwww | grep java | grep -v grep| awk '{print $11}' | sort -u`; do echo $jp `readlink -e $jp | xargs rpm -qf`; done``
+
+    and coverts it into a dict that links running_java_command with its package name.
```
xiangce: And currently, it's not a dict any more.

insights/combiners/package_provides_java.py
```diff
@@ -0,0 +1,48 @@
...
+        data (list): List of objects containing package name and its providing java command
+    """
+
+    def __init__(self, package_provides_java):
+        packages_list = []
+        for package in package_provides_java:
+            packages_list.append(package)
+        self.data = packages_list
+
+    def __getitem__(self, index):
```
xiangce: How about adding an interface like `get_package(command)`, then you can call it directly after getting the command from `PS`, e.g.: get_package(ps_obj.row["COMMAND"])

xiangce: And you can mixin it with LegacyItemAccess to save some duplicated codes.

zhangyi733: @xiangce Good suggestion. I add a get_package() interface and mixin LefacyItemAccess as well.

insights/combiners/package_provides_java.py
```diff
@@ -0,0 +1,55 @@
...
+            packages_list.append(package)
+        self.data = packages_list
+        super(PackageProvidesJavaAll, self).__init__()
+
+    def __iter__(self):
+        for item in self.data:
+            yield item
+
+    def get_package(self, ps_command):
+        for item in self.data:
```
xiangce: It's necessary to add a corresponding docstring for the new `get_package` interface.

zhangyi733:
@xiangce Your modification looks good to me. I think you we can merge this PR.

PR #534, opened 2017-10-02
Add functionality to parse_table

SteveHNH:
parse_table assumes the same delimiter is used for headers as well as the content of the table. This feature makes it so the delimiter can be different between headers and content. It's set to default to
the header being the same.

This is needed for parsing the lssap spec content properly.


bfahr:
@SteveHNH please submit a PR for the 1.x branch for this change.

SteveHNH:
@bfahr will do

PR #537, opened 2017-10-03
Add spec and parser for lssap for SAP rules

SteveHNH:
A new spec and mapper for the `lssap` SAP command. This provides us with better data to create rules for SAP systems.


insights/parsers/lssap.py
```diff
@@ -0,0 +1,118 @@
+
+"""
+lssap - Command
+============
```
bfahr: @SteveHNH please include a `lssap.rst` file to include this module in the documentation build.

insights/parsers/lssap.py
```diff
@@ -0,0 +1,118 @@
...
+        """list: Returns a list of instances from the Instance column."""
+        return [row["Instance"] for row in self.data if "Instance" in row]
+
+    @property
+    def sid(self):
+        """list: Returns a list of the SIDs from the SID column."""
+        return list(set(row["SID"] for row in self.data if "SID" in row))
+
+    def instance_types(self):
+        types = []
```
bfahr: Please include docstrings for this method. and the other public methods below so that they will be included in the documentation.

insights/parsers/lssap.py
```diff
@@ -0,0 +1,118 @@
...
+        hana = re.compile(r"HDB\d+")
+        ascs = re.compile(r"ASCS\d+")
+        if re.search(netweaver, instances):
+            types.append('netweaver')
+        if re.search(hana, instances):
+            types.append('hana')
+        if re.search(ascs, instances):
+            types.append('ascs')
+        else:
+            ParseException("No valid instance types found")
```
bfahr: Is it possible to determine if no valid instance types exist in `parse_content`, if so it would be better to raise the exception there, assuming no other useful data is available in that situation.  If all other data is valid then you need to document this exception so that a rule developer will understand that it will have to be caught in the rule.  We typically try to avoid this.

bfahr: Also, it doesn't look like you are testing this method directly.

insights/parsers/lssap.py
PaulWay: Also, the preferred format for the heading is e.g. 'lssap - Command ``/usr/sap/hostctrl/exe/lssap`` with a matching length of double underline :-)

insights/parsers/lssap.py
PaulWay: I agree.  I'd normally suggest computing this kind of stuff in parse_content, but you seem to do that in the LSSap parser class rather than in the SAPList class.  If it was all in one thing I'd do:
```python
        # remove lssap version and bar text from content
        clean_content = content[2:]  # There's a better way of doing this coming in parse_delimited_table...
        if len(clean_content) > 0 and "SID" in clean_content[0]:
            self.data = parse_table(clean_content, delim='|', header_split=False)
        else:
            raise ParseException("Lssap: Unable to parse {0} line(s) of content: ({1})".format(len(content), content))

        # Set attributes
        self.instances = [row["Instance"] for row in self.data if "Instance" in row]
        self.sids = list(set(row["SID"] for row in self.data if "SID" in row))
        self.instance_types = []
        if any(i.startswith('HDB') and i[3:].isdigit() for i in self.instances):
            self.instance_types.append('hana')
        if any(i.startswith('ASCS') and i[4:].isdigit() for i in self.instances):
            self.instance_types.append('ascs')
        if any(i.startswith('D') and i[1:].isdigit() for i in self.instances):
            self.instance_types.append('netweaver')
        if self.instance_types == []:
            raise ParseException("No valid instance types found")
```

But you probably  know this content and how it's going to change better than I do.

insights/parsers/tests/test_lssap.py
```diff
@@ -0,0 +1,133 @@
...
+==========================================
+  SID   Nr   Instance    SAPLOCALHOST                        Version                 DIR_EXECUTABLE
+  HA2|  16|       ASCS16|         lu0417|749, patch 10, changelist 1698137|          /usr/sap/HA2/ASCS16/exe
+""".strip()
+
+Lssap_runAll_Test = """
+ - lssap version 1.0 -
+==========================================
+  SID   Nr   Instance    SAPLOCALHOST                        Version                 DIR_EXECUTABLE
+  HA2|  16|       D16|         lu0417|749, patch 10, changelist 1698137|          /usr/sap/HA2/D16/exe
```
PaulWay: Is this really the way this output would be formatted?  Or have you constructed it by hand?  It'd be really good if we could get real data.

insights/parsers/tests/test_lssap.py
```diff
@@ -0,0 +1,133 @@
...
+                       "Version": "749, patch 10, changelist 1698137",
+                       "DIR_EXECUTABLE": "/usr/sap/HA2/D22/exe"}
+    assert sap.is_netweaver() is True
+    assert sap.is_hana() is True
+    assert sap.is_ascs() is True
+
+
+def test_fail():
+    with pytest.raises(ParseException) as excinfo:
+        lssap.Lssap(context_wrap(Lssap_BAD))
```
PaulWay: No test of which parse exception this raises here?

insights/parsers/tests/test_lssap.py
```diff
@@ -0,0 +1,133 @@
...
+        lssap.Lssap(context_wrap('test'))
+    assert "Lssap: Unable to parse 1 line(s) of content: (['test'])" in str(excinfo)
+
+    with pytest.raises(ParseException) as excinfo:
+        lssap.Lssap(context_wrap(Lssap_BAD_2))
+    assert "Lssap: Unable to parse 1 line(s) of content: (['HA2 | N16|  D16'])" in str(excinfo)
+
+    with pytest.raises(ParseException) as excinfo:
+        lssap.Lssap(context_wrap(Lssap_BAD_3))
+    assert "Lssap: Unable to parse 2 line(s) of content: (['HA2 | N16 | D16', 'HB2 | foo | bar'])" in str(excinfo)
```
PaulWay: I agree with Bob - it'd be good to have a test that parsing data which does not contain any valid instance types raises the relevant parse exception here.

insights/parsers/tests/test_lssap.py
SteveHNH: @PaulWay This is exactly how the output looks. The SAP experts I'm working with sent me doc with the data in it and expressed that this is what the output looks like. I don't have a machine for myself to test with yet.

PaulWay: I was more remarking on the odd semi-justification and names being longer and shorter - sorry to have doubted you @SteveHNH :-)

insights/parsers/lssap.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, *args, **kwargs):
+        self.data = {}
+        super(Lssap, self).__init__(*args, **kwargs)
+
+    def parse_content(self, content):
+        # remove lssap version and bar text from content
+        clean_content = content[2:-1]
+        invalid_inst = []
+        if len(clean_content) > 0 and "SID" in clean_content[0]:
+            self.data = parse_table(clean_content, delim='|', header_delim=None)
```
xiangce: Since the `parse_table` is deprecated and remove by #557, better use the `parse_delimited_table` here instead.

insights/config/specs.py
```diff
@@ -198,6 +198,7 @@
     "lsmod"                     : CommandSpec("/sbin/lsmod"),
     "lspci"                     : CommandSpec("/sbin/lspci"),
     "lsof"                      : CommandSpec("/usr/sbin/lsof", large_content=True),
+    "lssap"                     : CommandSpec("/usr/sap/hostctrl/exe/lssap"),
```
xiangce: For the master branch, better add the spec to `insights/specs.py` too.

insights/parsers/lssap.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Raises:
+        ParseException: Raised if any error occurs parsing the content.
+    """
+    instance_dict = {'D': 'netweaver',
+                     'HDB': 'hana',
+                     'ASCS': 'ascs'}
+
+    def __init__(self, *args, **kwargs):
+        self.data = {}
```
xiangce: `parse_table` (`parse_delimited_table`) returns a `list`, better initialize the `self.data` as a `list` correspondingly.

insights/parsers/lssap.py
```diff
@@ -0,0 +1,106 @@
...
+        clean_content = content[2:-1]
+        invalid_inst = []
+        if len(clean_content) > 0 and "SID" in clean_content[0]:
+            self.data = parse_table(clean_content, delim='|', header_delim=None)
+        else:
+            raise ParseException("Lssap: Unable to parse {0} line(s) of content: ({1})".format(len(content), content))
+
+        self.running_inst_types = [row["Instance"].rstrip('1234567890') for row in self.data if "Instance" in row]
+
+        for i in self.running_inst_types:
```
xiangce: Just like the list-comprehension used in line-69, the `for` loop at line-71 can be simplified as:

~~~
        invalid_inst = [i for i in self.running_inst_types if i not in self.instance_dict]

~~~
btw. the `.keys()` an be ignored.

insights/parsers/tests/test_lssap.py
```diff
@@ -0,0 +1,153 @@
...
+
+    with pytest.raises(ParseException) as excinfo:
+        lssap.Lssap(context_wrap(Lssap_bad_instance_TEST))
+    assert "Lssap: Invalid instance parsed in content: (['B', 'QR', 'DS']"
```
xiangce: You might miss a piece of `in str(excinfo)` in the last assert line.

insights/parsers/tests/test_lssap.py
SteveHNH: @PaulWay No worries! I got some test machines finally and the output was accruate. I did have to add a `=======` at the bottom as well, but that was the only modification .

PR #539, opened 2017-10-03
add parse_table functionality to 1.x

SteveHNH:
Add ability to use the split delimiter only on the table content rather than assuming the header users the same delimiter


insights/util/__init__.py
```diff
@@ -93,17 +93,15 @@ def __init__(self, dict_):
         self.__dict__ = dict_
 
 
-def parse_table(content, delim=None, max_splits=-1, strip=True):
+def parse_table(content, delim=None, max_splits=-1, strip=True, header_split=True):
```
bfahr: Based on this information from #534, should the attribute be `header_delim` instead so that you can specify a different header delimiter?  The code at 103 will always split the header, but use either `delim` or ` ` (space).

> parse_table assumes the same delimiter is used for headers as well as the content of the table. This feature makes it so the delimiter can be different between headers and content. It's set to default to the header being the same.
> 
> This is needed for parsing the lssap spec content properly.
```

File: insights/util/__init__.py [hunk 2]
```diff
@@ -93,17 +93,15 @@ def __init__(self, dict_):
...
     Parses table-like text. Assumes the first row contains column names.
     Column names cannot contain spaces.  Fields cannot be blank.
     """
     if not content:
         return []
-    if strip:
-        cols = [c.strip() for c in content[0].split(delim)]
-    else:
-        cols = content[0].split(delim)
+    cols = [c.strip() if strip else c for c in content[0].split(delim if header_split else None)]
```
bfahr: @SteveHNH I went looking for the tests for this method and could not find any.  Would you mind to add tests?

SteveHNH: Sure! I'll get all the above changes put in and add a test as well.
Thanks, @bfahr

insights/util/__init__.py
SteveHNH: @bfahr Ended up implementing a header_delim and a delim separately. There is actually only one parser using the `delim` argument, so I updated that to use the `header_delim` as well in order to make everything work nicely. All the other parsers that use the `parse_table` function simply use None as the delimiter so everything works.

insights/tests/test_util.py
```diff
@@ -114,3 +115,44 @@ def test_deep_nest(self):
         with pytest.raises(AssertionError):
             b["stuff"]["abba"][0]["foo"] = "cake"
             unordered_compare(a, b)
+
+
+class TestParseTable(unittest.TestCase):
```
bfahr: @SteveHNH would you mind to remove this class declaration and just have the `test_parse_table` function?  All new tests should use `pytest` instead of `unittest`.  I believe all of your tests will work as a function instead of a class method and you are already using the `pytest` assert.

SteveHNH:
@bfahr Took care fo the class to function deal. Should be good to go once tests complete. Thanks!

bfahr:
@SteveHNH just one minor flake8 error.

SteveHNH:
@bfahr yep! just realized that. Should be fixed in a minute

PR #544, opened 2017-10-05
Handle missing option in /etc/sysconfig/selinux

skontar:
This fixes occasional exceptions thrown by SELinux combiner.


skontar:
1.x https://github.com/RedHatInsights/insights-core/pull/547

PR #553, opened 2017-10-09
Enhance combiner krb5 1x

wushiqinlou:
- Same section in main krb5 will cover the same option from other confiuration, and the different ones will be kept.


insights/combiners/krb5.py
```diff
@@ -82,7 +82,11 @@ def __init__(self, local, shared):
                 self.module = krb5_parser.module
             else:
                 self.data.update(krb5_parser.data)
-        self.data.update(main_data)
+        for key, value in main_data.items():
```
PaulWay: Could you explain, preferably in a comment in the code as well, why we don't use update here?  What is it about `self.data.update(main_data)` that works the wrong way?

PR #555, opened 2017-10-09
Add TomcatVirtualDirContext

skontar:
Search for VirtualDirContext in any TomCat configuration files.


insights/parsers/tomcat_virtual_dir_context.py
```diff
@@ -0,0 +1,46 @@
...
+
+    /usr/share/tomcat/conf/server.xml:    <Resources className="org.apache.naming.resources.VirtualDirContext"
+
+Examples::
+
+    >>> shared[TomcatVirtualDirContext].has_virtual_dir_context
+    True
+    >>> shared[TomcatVirtualDirContext].lines
+    ('/usr/share/tomcat/conf/server.xml',
+    '    <Resources className="org.apache.naming.resources.VirtualDirContext"')
```
xiangce: Do you keep the whitespace on any purpose? If not, better `strip` them.

insights/parsers/tests/test_tomcat_virtual_dir_context.py
```diff
@@ -0,0 +1,57 @@
...
+    tomcat_virtual_dir_context = TomcatVirtualDirContext(context_wrap(FOUND_2))
+    assert tomcat_virtual_dir_context.has_virtual_dir_context is True
+    assert len(tomcat_virtual_dir_context.lines) == 2
+    assert tomcat_virtual_dir_context.lines[0] == ('/usr/share/tomcat/conf/server.xml',
+                                                   '    <Resources className="org.apache.naming.resources.VirtualDirContext"')
+    assert tomcat_virtual_dir_context.lines[1] == ('/usr/share/tomcat6/webapps/whatever/META-INF/context.xml',
+                                                   'className="org.apache.naming.resources.VirtualDirContext"')
+
+
+def test_tomcat_virtual_dir_context_not_found():
```
xiangce: How about raising a "ParseException" or "SkipExecption" for empty input content (NOT_FOUND)? In the case of raising exception, the `.has_virtual_dir_context` attribute is not needed since this parser will not be triggered. 

That is to say, for the cases, this parser is triggered the `.has_virtual_dir_context` must be `True`.

And by doing that, the plugin's code would be simpler.

insights/parsers/tomcat_virtual_dir_context.py
```diff
@@ -0,0 +1,46 @@
...
+
+@parser('tomcat_virtual_dir_context')
+class TomcatVirtualDirContext(Parser):
+    """
+    Reports whether there is a VirtualDirContext option used in TomCat configuration file.
+    """
+    def __init__(self, *args, **kwargs):
+        self.has_virtual_dir_context = False
+        """bool: True if VirtualDirContext is used."""
+        self.lines = []
```
xiangce: It would be convenient to store/parse the `file_name` and `file_line` as a `dict` than `list`.
~~~
self.data = {'/usr/share/tomcat/conf/server.xml': 
                   ['<Resources className="org.apache.naming.resources.VirtualDirContext'],  <-- if line is unique in one .xml file, `str` is more suitable here than `list`
}
~~~

And add several useful interfaces like `files_has_virtual_dir_context`:
~~~
def files_has_virtual_dir_context(self):
    return self.data.keys()
~~~

insights/parsers/tomcat_virtual_dir_context.py
skontar: Yes, I want to show the exact offending line to the user.

insights/parsers/tests/test_tomcat_virtual_dir_context.py
skontar: `SkipExecption` seems like a good idea. It is a bit outside my usual thinking about parsers, but makes sense. Will be updated.

insights/parsers/tomcat_virtual_dir_context.py
skontar: Storing in a dictionary in case that there are more lines with VirtualDirContext in one file makes sense. Will be updated. 

However, I do not need any extra interfaces for the rule, as I will just iterate through the dictionary and display offending files / lines. Also, there is a very high probability that this parser will never be re-used.

insights/config/specs.py
```diff
@@ -373,6 +373,7 @@
     "teamdctl_state_dump"       : CommandSpec("/usr/bin/teamdctl {iface} state dump", iface=r"team\S+"),
     "tomcat_web.xml"            : First([PatternSpec(r"etc/tomcat.*/web\.xml"),
                                     PatternSpec(r"conf/tomcat/tomcat.*/web\.xml")]),
+    "tomcat_virtual_dir_context": CommandSpec("/bin/grep -R --include '*.xml' 'VirtualDirContext' /usr/share/tomcat*"),
```
xiangce: For the `master` branch, we'd better add the specs to both the `insights/config/specs.py` and `insights/specs.py`.

Sorry for missing this in the first glance.

skontar:
1.x https://github.com/RedHatInsights/insights-core/pull/562

PR #557, opened 2017-10-10
Rename parse_table to parse_delimited_table and move to parsers for master branch

PaulWay:
`insights.util.parse_table` and `insights.parsers.parse_fixed_table` both do more or less the same thing: parse a table from a heading string and lines with fields into a list of dictionaries, keyed on the headings.  Though they work on different inputs, but they should both be in the `parsers` utility module.

This update renames `parse_table` to `parse_delimited_table`, gives it some of the same parameters as `parse_fixed_table` for interface homogeneity, and moves it from `insights.util` to `insights.parsers`.  The following parser modules have been updated to use the `parse_delimited_table` function:

* fstab
* ipcs_sem
* lvm
* netstat
* ps
* rhn_schema_stats
* route

All tests pass and there is no loss of test coverage. Tests in the Red Hat rule repositories all pass.

This also includes tidying the arguments and handling of `parse_fixed_table`.

This is the master branch version of #445


insights/__init__.py
```diff
@@ -14,7 +14,8 @@
 from .core.plugins import make_response, make_metadata  # noqa: F401
 from .core.filters import add_filter, apply_filters, get_filters  # noqa: F401
 from .parsers import get_active_lines  # noqa: F401
-from .util import defaults, parse_table  # noqa: F401
+from .util import defaults  # noqa: F401
+from .parsers import parse_delimited_table as parse_table  # noqa: F401
```
xiangce: In the `master` branch, do you think it's still necessary to keep the old `parse_table` here? how about use the `parse_delimited_table` directly?
And would you mind put the `parse_fixed_table` here too?

PaulWay: Well, we export neither parse_delimited_table nor parse_fixed_table, so IMO that's the way it should be.  The purpose of this move and rename was to move the `parse_delimited_table` into the `parsers` module where it is commonly used.  IMO we'd like to avoid other things using these internal functions to parse text, and would instead like them to use the parsers :-)

So thanks for reminding me to not export the symbol I didn't want to export :-)

xiangce: It's OK to me that do not export them there.

PR #561, opened 2017-10-10
add distinct versions

brantleyr:


jharting:
from the insights playbook POV (where we'll need to determine what version we're dealing with) it would be nice for `--version` to behave in a way compatible with the old client

brantleyr:
@jharting yea thats why i added new flags, and simply left the old one alone.

kylape:
I'm leaning towards having --version print only the RPM version mainly due to backwards compatibility concerns.  Also, --version handling should be done in the RPM script because it's not really valid to import insights_client package from within the egg.  I think it's valid to have --core-version, but --client-version seems to just duplicate --version.

brantleyr:
@kylape so then if we want to put --version handling in the RPM we need to do another release.

gravitypriest:
WRT backward compat concerns, I think I'd want --version to print the RPM version, and --core-version to print the egg version, in that case.

gravitypriest:
Yeh understandable.  Should --version then print the core version and we have a separate one for the wrapper?

brantleyr:
@gravitypriest you mean like --client-version?

gravitypriest:
Yeah

brantleyr:
@kylape are we closing out this PR in favor of yours (I assume)?

PR #565, opened 2017-10-10
InsightsClient.version() should just return the version + release string

kylape:


kylape:
Whelp, didn't update the test.  I can update it, but you guys can still review the implementation to see if you approve of it.

insights/client/__init__.py
```diff
@@ -47,14 +47,7 @@ def set_up_logging(self):
...
     def version(self):
-        """
-            returns (dict): {'core': str,
-                            'client_api': str}
-        """
-        core_version = get_nvr()
-        client_api_version = constants.version
-
-        return {'core': core_version, 'client_api': client_api_version}
+        return "%s-%s" % (package_info["VERSION"], package_info["RELEASE"])
```
brantleyr: So we are no longer using the versioning in constants.py?

brantleyr:
So it looks like we are no longer using the versioning in constants.py? In addition, are we wanting to put the core and client versions in the user agent headers (as was in my PR)?

kylape:
@brantleyr I think adding the client version to the HTTP requests is a good idea, so I'll add that back in.  

@jhjaggars Just to confirm, we can use the version of the core now because we're gonna move insights-core version to 3.x, right?

brantleyr:
This will need to be merged in sync with https://github.com/RedHatInsights/insights-client/pull/36

PR #572, opened 2017-10-11
Enhance parser and package_provides_java.py

zhangyi733:
* Update pre_command java_command_package to handle process using java
  directly instead of /usr/bin/java
* Raise exception when java command is not provided by package installed
  through yum or rpm
* Update docstring in parser
* Enhance testcase to cover new added code logic


zhangyi733:
See https://github.com/RedHatInsights/insights-core/pull/570

PR #575, opened 2017-10-11
Xfs_info parser doc and code enhancements for 1.x branch

PaulWay:
Documented XFSInfo parser and got to 100% code coverage in tests.


insights/parsers/xfs_info.py
```diff
@@ -47,18 +105,20 @@ def parse_content(self, content):
...
                         value = int(value)
                     sect_info[key] = value
 
         self.xfs_info = xfs_info
 
         # Calculate a few things
-        self.data_size = xfs_info['data']['blocks'] * xfs_info['data']['bsize']
-        self.log_size = xfs_info['log']['blocks'] * xfs_info['log']['bsize']
+        if 'data' in xfs_info:
+            self.data_size = xfs_info['data']['blocks'] * xfs_info['data']['bsize']
```
xiangce: I think it would be safer to initialize them before the condition to avoid "AttributeError:" when the condition is not satisfied.

insights/parsers/tests/test_xfs_info.py
```diff
@@ -149,3 +149,10 @@ def test_ext_log_xfs_info(self):
         # Calculated information checks
         assert xfs_obj.data_size == 262144 * 4096
         assert xfs_obj.log_size == 25600 * 4096
+
+    def test_bad_xfs_info(self):
+        xfs_obj = xfs_info.XFSInfo(context_wrap("xfs_info: bad command or file name"))
+        assert hasattr(xfs_obj, 'xfs_info')
+        assert xfs_obj.xfs_info == {}
+        assert not hasattr(xfs_obj, 'data_size')
```
xiangce: For these XFSInfo which does not contain any `data` or `log` section, how about leave these `_size` as zero?
Otherwise, we'd better explain in the `Attributes` docstring that these attributes might do not exist.

insights/parsers/xfs_info.py
PaulWay: Good point, will do.

insights/parsers/tests/test_xfs_info.py
PaulWay: Yep, done.  Thanks @xiangce

PR #576, opened 2017-10-11
Add parser ipaupgrade 1x

wushiqinlou:


insights/parsers/ipaupgrade_log.py
```diff
@@ -0,0 +1,33 @@
+"""
+ipaupgrade - "/var/log/ipaupgrade.log" File
```
xiangce: To get a fine look for the documentation, better use '``' to surround the `/var/log/ipaupgrade.log`

~~~
IpaupgradeLog - File ``/var/log/ipaupgrade.log``
~~~

And better add a brief description to the file header.

xiangce:
See #577

PR #577, opened 2017-10-11
Add parser ipaupgrade

wushiqinlou:


insights/config/specs.py
```diff
@@ -160,6 +160,7 @@
                                     SimpleFileSpec("sos_commands/networking/ip_address"),
                                     SimpleFileSpec("ip_a")]),
     "ip_route_show_table_all"   : CommandSpec("/sbin/ip route show table all"),
+    "ipaupgrade.log"            : SimpleFileSpec("var/log/ipaupgrade.log"),
```
xiangce: Should add it to `insights/specs.py` for master.

PR #579, opened 2017-10-11
InitScript parser and unit tests

rmetrich:
This parser read files in etc/rc.d/init.d/ and creates InitScript instances, skipping non-initscript files.

To determine whether a file in etc/rc.d/init.d/ is an initscript,

the parser checks for "# chkconfig: " or "# Provides: " strings in the script. If that matches, then it assumes it is an initscript.

Otherwise, it tries to find out if it is by searching for:
- shebang (e.g. #!/bin/bash) on first line
- start/stop/status tokens in script
If 3 or more items are found (called "confidence" in the code), (e.g. shebang + start + stop), then we assume it is an initscript
Otherwise the parser raises a SkipException.


insights/config/specs.py
```diff
@@ -148,6 +148,7 @@
                                     PatternSpec(r"usr/lib64/ImageMagick-6.5.4/config/policy\.xml"),  # ImageMagick 6.5 on rhel-6 64bit.
                                     PatternSpec(r"usr/lib/ImageMagick-6.5.4/config/policy\.xml")]),  # ImageMagick 6.5 on rhel-6 32bit.
     "init.ora"                  : SimpleFileSpec("{ORACLE_HOME}/dbs/init.ora"),
+    "initscript"                : PatternSpec(r"etc/rc\.d/init\.d/.*"),
```
jhjaggars: `PatternSpec` isn't exactly a normal regular expression.  The dirname and basename are split and only the basename may be a regular expression.  This is an unfortunate implementation detail of the current client that we're still dealing with.  You should change this to:

```
PatternSpec(r"etc/rc.d/init.d/.*"),
```

insights/parsers/initscript.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class NotInitscriptException(SkipException):
+    pass
+
+
+@parser("initscript")
+class InitScript(Parser):
+    """
+    Parse initscript files. Each item is a dictionary with following fields:
```
bfahr: @rmetrich would you mind to include your description of the algorithm you are implementing here, I think it would be helpful for users of your parser in understanding how it works, and the outputs.  You could start with your PR comments and add if you feel like anything needs further explanation.

insights/parsers/initscript.py
```diff
@@ -0,0 +1,84 @@
...
+
+    * ``file_name`` - initscript name
+    * ``file_path`` - initscript path (without leading '/')
+    * ``file_content`` - initscript content (list of lines)
```
bfahr: If you wanted, you could convert this to an attributes section in the docstring, for example:
```
Attributes:
    file_name (str): initscript name
    file_path (str): initscript path (without leading '/')
    file_content (list): initscript content (list of lines)
```

insights/parsers/tests/test_initscript.py
```diff
@@ -0,0 +1,152 @@
...
+
+class TestInitScript(unittest.TestCase):
+    def test_initscript1(self):
+        context = context_wrap(NOTINITSCRIPT_CONTENT)
+        context.path = NOTINITSCRIPT_SCRIPT
+
+        with pytest.raises(NotInitscriptException) as e_info:
+            r = InitScript(context)
+            self.assertIsNone(r)
+            assert context.path in str(e_info.value)
```
bfahr: Once the exception is thrown on line 113, lines 114 and 115 will not execute.  If the exception is not thrown then an assertion will be raised.  You could probably just remove line 114 as it is implied by the exception being raised, and move line 115 out one level to be checked after the exception is thrown.

insights/parsers/tests/test_initscript.py
```diff
@@ -0,0 +1,152 @@
...
+        self.assertEqual(r.file_name, "script_hintsonly")
+
+    def test_initscript6(self):
+        context = context_wrap(EMPTY_CONTENT)
+        context.path = EMPTY_SCRIPT
+
+        with pytest.raises(EmptyFileException) as e_info:
+            r = InitScript(context)
+            self.assertIsNone(r)
+            assert context.path in str(e_info.value)
```
bfahr: Same comment as above regarding the exception.

insights/parsers/tests/test_initscript.py
```diff
@@ -0,0 +1,152 @@
...
+case "$1" in
+start)
+        ;;
+stop)
+        ;;
+esac
+""".strip()
+
+
+class TestInitScript(unittest.TestCase):
```
bfahr: Please use pytest instead of unittest for your test cases.  All that is necessary to do that is to make all of the class methods functions, remove the class, and make all of the `self.assert...` statements `assert`.  For instance `test_initscript2` becomes:
```
def test_initscript2():
    context = context_wrap(CHKCONFIG_CONTENT)
    context.path = CHKCONFIG_SCRIPT

    r = InitScript(context)
    assert r.file_name == "script_chkconfig"
```

bfahr:
PR for 1.x branch has also been submitted #580.

rmetrich:
I'm surprised about the regex comment, because I already implemented a
plugin in DiagInsights and initscripts get caught by the parser, even
though etc/rc\.d/init\.d/.* is used.

Renaud.

Sent from my phone.

Le 11 oct. 2017 4:00 PM, "Jesse Jaggars" <notifications@github.com> a
écrit :

rmetrich:
I'll do so, thanks for pointing that out.

Renaud.

insights/config/specs.py
rmetrich: I can confirm, with branch 1.x at least that PatternSpec is a regular expression:
- if I use PatternSpec(r"etc/rc.d/init.d/.*"), then etc/rcXd/initXd/something also matchs, which I don't want to
- if I use PatternSpec(r"etc/rc\.d/init\.d/.*"), it perfectly works: only etc/rc.d/init.d/* matches

rmetrich:
@jhjaggars Please confirm for the PatternSpec regex, for now I didn't change the code because my spec works well whereas removing "\." doesn't (catches more directories).
@bfahr I integrated your suggestions/requests. Sorry for the wreck, I just inspired myself from existing unit tests in 1.x branch ;-)

insights/parsers/initscript.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def parse_content(self, content):
+        """
+        Raises:
+            EmptyFileException: Raised if file is empty.
+            NotInitscriptException: Raised if likely not an initscript.
+        """
+        self.file_content = content
+
+        # If we find 'chkconfig: XYZ' or 'Provides: NAME', assume we have an
```
xiangce: Hi @rmetrich, how about adjusting the logic a little to save some workloads?
1. Check the `shebang` in content[0] at the very first
2. The check the ``# chkconfig: <values>`` or ``# Provides: <names>`` together with the ``start``/``stop``/``status`` token in the same loop

E.g.

~~~
        # Check for hints (shebang + presence of start/stop/status keywords)
        # Allocate 1 point for each item found. And declare as valid if
        # confidence >= 3 (half of the number of items + 1).
        confidence = 0
        # Check 'shebang' at first
        try:
            line = content[0]
            if SHEBANG_REGEX.match(line):
                confidence += 1
        except IndexError:
            raise EmptyFileException(self.file_path)

        key = {'start': 0, 'stop': 0, 'status': 0}
        for line in content:
            # If we find 'chkconfig: XYZ' or 'Provides: NAME', assume we have an
            # initscript
            if CHKCONFIG_REGEX.match(line) or LSB_REGEX.match(line):
                return
            # Check presence of 'start/stop/status' keywords
            m = STARTSTOPSTATUS_REGEX.search(line)
            if m:
                key[m.group(1)] += 1
        for v in key.itervalues():
            if v != 0:
                confidence += 1

        if confidence < 3:
            raise NotInitscriptException("path: %s, confidence: %d" % (self.file_path, confidence))
~~~

rmetrich: Well doing so will cost more.
With the current code, for nicely written initscripts, 5 to 10 lines will be matched against before finding "chkconfig" (regex cost is 2xN, with N ~= 7) 
Whereas with your code proposal, there will be 1 (shebang) + 5xN with N ~= 7

For badly written initscripts, the cost will be 5 x number of lines of the script in all cases.

insights/config/specs.py
jhjaggars: Sorry I managed to mis-explain it.  When the client collects the spec it doesn't treat it as a regular expression.  When we try to extract it from the archive we do treat it as a regular expression.

insights/parsers/initscript.py
```diff
@@ -0,0 +1,104 @@
...
+
+SHEBANG_REGEX = re.compile(r"^#!\s*/.*$")
+STARTSTOPSTATUS_REGEX = re.compile(r"\b(start|stop|status)\b")
+
+
+class EmptyFileException(SkipException):
+    pass
+
+
+class NotInitscriptException(SkipException):
```
bfahr: @rmetrich I apologize for not noticing this in the first pass, but you should be using `ParseException` instead of `SkipException` here.  See this document for the difference http://insights-core.readthedocs.io/en/1.x/exception_model.html#parsing-and-exceptions

insights/config/specs.py
rmetrich: So what should I do, change to r"etc/rc.d/init.d/.*" anyway?

insights/parsers/initscript.py
xiangce: My main target is to merge the two iterators of the content into one loop.
Even the `shebang` check can also be moved into the loop:
~~~
        confidence = 0
        key = {'start': 0, 'stop': 0, 'status': 0}
        for i, line in enumerate(content):
            # If we find 'chkconfig: XYZ' or 'Provides: NAME', assume we have an
            # initscript
            if CHKCONFIG_REGEX.match(line) or LSB_REGEX.match(line):
                return
            if i == 0 and SHEBANG_REGEX.match(line):
                confidence += 1
            # Check presence of 'start/stop/status' keywords
            m = STARTSTOPSTATUS_REGEX.search(line)
            if m:
                key[m.group(1)] += 1
~~~
It basicaly has the same executing order with yours.

insights/parsers/initscript.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # Otherwise, check for hints (shebang + presence of start/stop/status
+        # keywords)
+        #
+        # Allocate 1 point for each item found. And declare as valid if
+        # confidence >= 3 (half of the number of items + 1).
+        confidence = 0
+        try:
+            line = content[0]
+            if SHEBANG_REGEX.match(line):
```
PaulWay: I love regexes as much as anyone but you REALLY don't need a regex here.  You can do:
```python
if line.startswith('#!/'):
```

PaulWay:
I'm trying to work out the raison d'être of this parser.  What rules are going to use it?  What is it getting out of the script files?

If all you're trying to do is to find whether the script contains 'start', 'stop' or 'status'... why?  Can 'grep' tell you this information more easily?  What does this give the user that the existing 'chkconfig' parser does not?

If there's more to it, then please let us know.  I'm really not sure I see the point of scanning all these files in this way.

insights/parsers/initscript.py
```diff
@@ -0,0 +1,104 @@
...
+                confidence += 1
+        except IndexError:
+            raise EmptyFileException(self.file_path)
+
+        key = {'start': 0, 'stop': 0, 'status': 0}
+        for line in content:
+            m = STARTSTOPSTATUS_REGEX.search(line)
+            if m:
+                key[m.group(1)] += 1
+        for v in key.itervalues():
```
PaulWay: or:
```python
confidence = len([v for v in key.values() if v != 0])
```

insights/parsers/initscript.py
rmetrich: The parser will be used with a (for now) DiagInsights rule that will search for calls for "su" which have to be avoided on RHEL7 and later (i.e. with systemd).
Regarding the shebang, "#!/" is not enough, because some people put a space between ! and /.

insights/parsers/initscript.py
rmetrich: We can merge the iterations for sure, but it is more CPU costly.

insights/config/specs.py
jhjaggars: Since this spec is designed to work with sosreports explicitly, leaving it as is will work.  We should wrap the spec with a NoneGroup so that we don't attempt to collect it from customers.  There is an open issue #594 to track the differences between sosreport collection and insights archive collection.

insights/parsers/initscript.py
rmetrich: @bfahr Done

rmetrich:
@jhjaggars I wrapped the PatternSpec into a NoneGroup as requested

insights/parsers/initscript.py
PaulWay: Sure.  Search for '#!'.  Works for me.

PaulWay:
So, if your purpose is to search for use of 'su' in init scripts, then:
```bash
/usr/bin/grep -E "[:space:]+su[:space:]" /etc/rc.d/init.d/*.sh
```
would do it just fine.  That will give you the name of all scripts using `su` as a command.  Remove the 'l' option if you want the offending line(s) as well.

To read every single init script into Insights is overkill, in my opinion.

insights/parsers/initscript.py
rmetrich: Still, different because shebang has to be an absolute path, so executable path should start with "/". My idea was to provide something clean, still I won't argue anymore, I have other higher priority tasks, including customer satisfaction.

Regarding your "grep" proposal, I want to parse sosreports from customers.
If there is a way to use system commands on sosreport files, then please explain how to do that.
I can then rewrite most of the current DiagInsights rules into basic greps, which will be faster than using Insights code.

insights/parsers/initscript.py
xiangce: :) It's just my suggestion. The logic of your current code works, and it's OK to me that you do not change them.

PR #580, opened 2017-10-11
InitScript parser and unit tests branch 1.x

rmetrich:
This parser read files in etc/rc.d/init.d/ and creates InitScript instances, skipping non-initscript files.

To determine whether a file in etc/rc.d/init.d/ is an initscript,

the parser checks for "# chkconfig: " or "# Provides: " strings in the script. If that matches, then it assumes it is an initscript.

Otherwise, it tries to find out if it is by searching for:

- shebang (e.g. #!/bin/bash) on first line
- start/stop/status tokens in script

If 3 or more items are found (called "confidence" in the code), (e.g. shebang + start + stop), then we assume it is an initscript
Otherwise the parser raises a SkipException.


rmetrich:
The Exception Handling in first commit (https://github.com/RedHatInsights/insights-core/pull/580/commits/7699c28e1eb711796470b3c6b4a5c92e7e2b4e7f) works well on branch 1.x but not master. I hence added the second commit so that code on master and 1.x remains the same.

insights/config/specs.py
```diff
@@ -148,6 +148,7 @@
                                     PatternSpec(r"usr/lib64/ImageMagick-6.5.4/config/policy\.xml"),  # ImageMagick 6.5 on rhel-6 64bit.
                                     PatternSpec(r"usr/lib/ImageMagick-6.5.4/config/policy\.xml")]),  # ImageMagick 6.5 on rhel-6 32bit.
     "init.ora"                  : SimpleFileSpec("{ORACLE_HOME}/dbs/init.ora"),
+    "initscript"                : PatternSpec(r"etc/rc\.d/init\.d/.*"),
```
jhjaggars: As mentioned on my other review this should be changed to:

```
PatternSpec(r"etc/rc.d/init.d/.*"),
```

bfahr:
See my comments in #579

PR #581, opened 2017-10-11
exit when no docker-py

brantleyr:


kylape:
Is there a reason we don't add this as an optional dependency in our RPM spec?  It seems like that would be a more seamless user experience.

brantleyr:
@kylape are you suggesting adding it into the RPM as a weak dependency?

brantleyr:
@kylape even if we do add it in as an optional dependency in the RPM, I think we should still have this failsafe check.

kylape:
Yeah, good point.  Doesn't hurt to have the check.

PR #585, opened 2017-10-11
remove --docker-image-name

gravitypriest:


brantleyr:
Are we wanting to simply remove the flag invocation of docker image name? Or do we want to remove its functionality entirely in other places? Is this something we still want to keep in the config file or completely remove? I saw "docker_image_name" referenced other places.

gravitypriest:
Oh I'm a dumbass, I forgot the hypens go to underscores and just removed it where I thought I saw it....  gonna close this out of shame.

PR #617, opened 2017-10-18
Add Parser for sysconfig_mongod in branch 3.x

JoySnow:


insights/parsers/sysconfig.py
```diff
@@ -221,3 +221,26 @@ class VirtWhoSysconfig(SysconfigOptions):
...
+    Examples:
+        >>> mongod_syscfg = shared[MongodWhoSysconfig]
+        >>> mongod_syscfg['OPTIONS']
+        '--quiet -f /etc/mongod.conf'
+        >>> mongod_syscfg.get('NO_SUCH_OPTION')
+        None
+        >>> 'NOSUCHOPTION' in mongod_syscfg
+        False
+    """
+    pass
```
PaulWay: I'd suggest making an `options` property here to contain the 'OPTIONS' keyword data or the empty string.  That'll make it a bit easier to get at this regularly-used piece of data.

JoySnow: @PaulWay    
Currently,  `mongod_syscfg.get('OPTIONS')` works well.
Or with this `options` property, users can use `mongod_syscfg.options` instead.
I can't see any necessity there for this improvement.

And I think it's always good for user to know if an option's value is `an un-empty string` or `a empty string` or None.

PR #625, opened 2017-10-20
Keyword search in Netstat parser for 1.x branch

PaulWay:
This method searches both the active internet connections and active UNIX domain sockets.  If you only want to search one, specify the name via the `search_list` keyword, e.g.:
```python
    from insights.parsers import Netstat, ACTIVE_UNIX_DOMAIN_SOCKETS
    conns.search(search_list=[ACTIVE_UNIX_DOMAIN_SOCKETS], State='LISTEN')
```
If `search_list` is not given as a parameter, both the active internet connections and active UNIX domain sockets are searched, in that order.  Results are compiled into one list.  This allows you to search for all listening processes, whether for internet connections or UNIX sockets, by e.g.:
```python
    conns.search(State__contains='LISTEN')
```
This method uses the `insights.parsers.keyword_search` function.


insights/parsers/netstat.py
```diff
@@ -292,17 +297,14 @@ def add_data(self, line):
...
         i = 0
         while i < len(self.indexes):
             index = self.indexes[i]
             m = self.meta[index]
             merged_data[m] = self.data[i]
-            component_len[m] = self.indexes[i + 1] - self.indexes[i] if i != len(self.indexes) - 1 else None
             i += 1
 
         self.data = merged_data
-        self.data[COMPONENT_LEN] = component_len
```
bfahr: @PaulWay since this is in the existing api, even thought it does not appear to have been tested, we should leave this in unless there's a problem associated with it.

PaulWay: That seems to be unnecessary work to me, but OK.

xiangce:
@PaulWay - Would you please re-submit it to the master branch? Thanks.

PR #635, opened 2017-10-24
fix analyze_mountpoint

brantleyr:


rexwhite:
Looks harmless enough... ;-)

PR #636, opened 2017-10-25
avoid status_code in type bool bug by returning the upload obj

gravitypriest:


insights/client/phase/v1.py
```diff
@@ -189,6 +189,6 @@ def collect_and_output():
             resp = c.upload(tar_file)
         else:
             logger.info('Archive saved at %s', tar_file)
-        if config["to_json"]:
+        if resp and config["to_json"]:
```
brantleyr: I think this is going to throw a "resp" is not defined error in the case it falls into the "else" above.

gravitypriest: Will do to make this more robust.  I added this check because when the upload fails, `c.upload` returns None, causing the json dump to print `null`

brantleyr: `resp` is never defined unless you fall into the logic above, so this could fail in some instances.

PR #638, opened 2017-10-26
xinetd.conf spec needs to match PatternSpec with xinetd.d for 1.x branch

PaulWay:
Otherwise this will hit the spec mismatch handling problem.

All plugin tests pass.


bfahr:
@PaulWay I assume this is related to #14 and the fact that these two specs are stacked for the `XinetdConf` parser?

PaulWay:
@bfahr Yep, correct.

PaulWay:
@xiangce That is a much better idea - will do!

kylape:
I'm pushing a fix directly to master (due to pending release) to fix this spec:

    PatternSpec(r"etc/xinetd\.d/.*")

The backslash in `xinetd\.d` is not interpreted as a regular expression and thus will not match anything.

PR #639, opened 2017-10-26
LVM: Improved tests and code for near 100% test coverage for 1.x branch

PaulWay:
This patch:

* Removes no longer used functionality that was only used internally.
* Simplifies some parsing to remove unnecessary branches and unused clauses.
* Tests all methods provided by all classes.

The only thing that isn't tested is the `if __name__ == '__main__':` code at the bottom which generates the key lists.


bfahr:
@PaulWay could you check these changes against the plugins, I see errors when I run plugin tests.

xiangce:
@PaulWay - It breaks the unit test of `multipath_underlying_devices_are_not_filtered_out.py`, would you please check it? We cannot merge it due to this break.

PaulWay:
Yeah, one of the internal plugins was supplying data with no headers, which only worked because the plugin was only looking at the errors.  I've submitted a bug fix for that.

PaulWay:
@xiangce Should be ready to go now :-)

xiangce:
@PaulWay - Sorry `multipath_underlying_devices_are_not_filtered_out.py` is a public rule in `insights-plugins` repo under the `storage` folder. Could you check it again?

xiangce:
@PaulWay - Sorry for missing the https://github.com/RedHatInsights/insights-plugins/pull/431.

xiangce:
Would you please re-submit it to the master branch? Thanks.

PR #642, opened 2017-10-27
New specs and parsers for libkeyutils

skontar:
https://bugzilla.redhat.com/show_bug.cgi?id=1507004


insights/config/specs.py
```diff
@@ -186,6 +186,8 @@
     "ksmstate"                  : SimpleFileSpec("sys/kernel/mm/ksm/run"),
     "lastupload"                : All([SimpleFileSpec("etc/redhat-access-insights/.lastupload"),
                                     SimpleFileSpec("etc/insights-client/.lastupload")]),
+    "libkeyutils"               : CommandSpec("/usr/bin/find -L /lib /lib64 -name 'libkeyutils.so*'"),
+    "libkeyutils_objdumps"      : CommandSpec('/usr/bin/find -L /lib /lib64 -name libkeyutils.so.1 -exec objdump -x "{}" \;'),
```
bfahr: @skontar looks like you may be missing a trailing double quote.

skontar: @bfahr Used ' for string instead of ", to include " in the string.

PR #646, opened 2017-10-30
Add parser for file /var/log/audit/audit.log

JoySnow:


insights/parsers/audit_log.py
```diff
@@ -0,0 +1,137 @@
...
+            info['type'] = linesp[0].lstrip('type=')
+
+            if not (linesp[1] and linesp[1].startswith('msg=audit(')):
+                info = None
+                continue
+            timestamp_id = linesp[1].lstrip('msg=audit(').rstrip('):')
+            timestamp_id_sp = timestamp_id.split(':')
+            if len(timestamp_id_sp) != 2:
+                info = None
+                continue
```
bfahr: I think you might be able to simplify this  **if / continue** logic since you've already checked if there are two items in the list and you haven't set `info` to any value?

insights/config/specs.py
```diff
@@ -14,6 +14,7 @@
 
 static_specs = {
     "autofs.conf"               : SimpleFileSpec("etc/autofs.conf"),
+    "audit_log"                 : SimpleFileSpec("var/log/audit/audit.log", large_content=True),
     "auditd.conf"               : SimpleFileSpec("etc/audit/auditd.conf"),
```
bfahr: Would you mind to move `autofs.conf` after the two `audit...` specs to keep them in order.  Thanks!

JoySnow: Sure.

insights/parsers/audit_log.py
```diff
@@ -0,0 +1,137 @@
...
+
+        Yields:
+            (dict): the parsed data of lines with timestamps after this date in the
+            same format they were supplied.
+        """
+        if lines is None:
+            lines = self.lines
+        items = self.parse_lines(lines)
+
+        from datetime import date
```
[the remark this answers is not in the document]
JoySnow: Okay.

insights/parsers/audit_log.py
```diff
@@ -0,0 +1,137 @@
...
+                continue
+            timestamp_id = linesp[1].lstrip('msg=audit(').rstrip('):')
+            timestamp_id_sp = timestamp_id.split(':')
+            if len(timestamp_id_sp) != 2:
+                info = None
+                continue
+            info['timestamp'] = timestamp_id_sp[0]
+            info['msg_ID'] = timestamp_id_sp[1]
+
+            for index in xrange(len(linesp) - 1, 1, -1):
```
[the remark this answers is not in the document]
JoySnow: For a log like this(I made it up ),   
`type=AVC msg=audit(1506487181.009:32794): what ev=er string  pid= 27960`  
We want unparsed part to be like `what ev=er string` instead of `what`.
So I'd like to keep my reversed iteration.

insights/parsers/audit_log.py
JoySnow: Yep, I can remove several `info = None`  lines.

(If I understand u correctly)   
Checking of    
`if len(linesp) < 2:`   
and    
`if len(timestamp_id_sp) != 2`   
are for different parts.

insights/parsers/audit_log.py
```diff
@@ -0,0 +1,133 @@
...
+            'comm': 'mongod',
+            'scontext': 'system_u:system_r:mongod_t:s0',
+            'tclass': 'unix_dgram_socket',
+            'type': 'AVC',
+            'tcontext': 'system_u:system_r:mongod_t:s0'
+        }
+        >>> assert avc_log_info == list(log.get_after(timestamp=date.fromtimestamp(1506047401.407)))[0]
+    """
+
+    def parse_lines(self, lines):
```
xiangce: As we discussed, this `parse_lines` should not be a public interface of this class, otherwise it could accept any kinds of `lines` and parse them as the log of `audit.log`

JoySnow: Changed.

insights/parsers/audit_log.py
```diff
@@ -0,0 +1,162 @@
...
+             {'timestamp': '1506487181.009',
+              'msg_ID': '32794',
+              'type': 'AVC',
+              'unparsed': 'avc:  denied  { create } for',
+              'raw_message': '...: ...'
+              ...
+             }, ...
+            ]
+        """
+        r = []
```
xiangce: Would you like to let it supports `list`? See `nova_log.py`

JoySnow: Enhanced.

PR #648, opened 2017-10-31
Modify blkid item in specs.py to make it work both in rhel 6 and 7

zhangyi733:


zhangyi733:
See https://github.com/RedHatInsights/insights-core/pull/649

PR #651, opened 2017-10-31

skontar:


skontar:
3.x https://github.com/RedHatInsights/insights-core/pull/652

insights/config/specs.py
```diff
@@ -391,7 +391,7 @@
     "teamdctl_state_dump"       : CommandSpec("/usr/bin/teamdctl {iface} state dump", iface=r"team\S+"),
     "tomcat_web.xml"            : First([PatternSpec(r"etc/tomcat.*/web\.xml"),
                                     PatternSpec(r"conf/tomcat/tomcat.*/web\.xml")]),
-    "tomcat_virtual_dir_context": CommandSpec("/bin/grep -R --include '*.xml' 'VirtualDirContext' /usr/share/tomcat*"),
+    "tomcat_virtual_dir_context": CommandSpec("/usr/bin/find /usr/share -maxdepth 1 -name 'tomcat*' -exec grep -R -s 'VirtualDirContext' --include '*.xml' '{}' +"),
```
xiangce: Please correct me if something goes wrong here.
Since the `catalina.base` (or `catalina.home`) can be specified manually by the CU manually, how about using below pre-command to get the directory accurately?

~~~
pre_command:
catalina_home: "/bin/ps auxwww | grep java | grep '\-Dcatalina.base' | awk -F'-Dcatalina.base=' '{print $2}' | awk '{print $1}' | sort -u"
===

"tomcat_virtual_dir_context": CommandSpec("/bin/grep -R /bin/grep -R --include '*.xml' 'VirtualDirContext'  {catalina_home}/*", catalina_home=r"\S+"),
~~~

skontar: That is a great idea. I think I will use this spec as a fallback for instances where tomcat is not running at the moment, but may want to have another one with a similar pre-command. Just note that `{catalina_home}/*` will not work as shell expansions are  disabled.

skontar: @xiangce also, can pre-commands do pipes? I am asking because CommandSpecs cannot.

xiangce: I'm sure the pipe works in pre-commands (but doesn't in CommandSpecs.) And in the `insights/specs.py` of the master branch, you can implement it more simply :)

skontar:
It is not done yet, please do not merge.

skontar:
@xiangce @bfahr Hi, I have modified the code significantly. Please re-review. Thanks!

skontar:
@xiangce @bfahr I have another question. Will the `find` command with `{}` as a parameter work, when there are pre-commands which are substituted into the strings using the same characters to define "variables"?

insights/config/specs.py
```diff
@@ -425,6 +426,7 @@
 
 pre_commands = {
     "block"                     : "/bin/ls /sys/block | awk '!/^ram|^\\.+$/ {print \"/dev/\" $1 \" unit s print\"}'",
+    "catalina"                  : r"""/bin/ps auxww | awk '/java/ { match($0, "\\-Dcatalina\\.home=([^[:space:]]+)", a); match($0, "\\-Dcatalina\\.base=([^[:space:]]+)", b); if (a[1] != "" || b[1] != "") print a[1] " " b[1] }'""",
```
jsvob: There is not anything obvious that I think would break this, but until someone actually runs it...

insights/parsers/tomcat_virtual_dir_context.py
```diff
@@ -1,46 +1,80 @@
...
+                    self.data[file_name].append(file_line)
+                else:
+                    self.data[file_name] = [file_line]
+
+        if self.data == {}:
+            raise SkipException('VirtualDirContext not used.')
+
+
+@parser('tomcat_vdc_targeted')
+class TomcatVirtualDirContextTargeted(TomcatVirtualDirContextFallback):
```
jsvob: Is it certain that TomcatVirtualDirContextFallback being already decorated with `@parser` is not going to cause problems? I remember there used to be such a problem somewhere sometime and the solution was to have one undecorated class and only the leaf children in the inheritance tree could be decorated but IDK if it has been resolved.

insights/config/specs.py
```diff
@@ -391,7 +391,8 @@
     "teamdctl_state_dump"       : CommandSpec("/usr/bin/teamdctl {iface} state dump", iface=r"team\S+"),
     "tomcat_web.xml"            : First([PatternSpec(r"etc/tomcat.*/web\.xml"),
                                     PatternSpec(r"conf/tomcat/tomcat.*/web\.xml")]),
-    "tomcat_virtual_dir_context": CommandSpec("/bin/grep -R --include '*.xml' 'VirtualDirContext' /usr/share/tomcat*"),
+    "tomcat_vdc_fallback"       : CommandSpec("/usr/bin/find /usr/share -maxdepth 1 -name 'tomcat*' -exec /bin/grep -R -s 'VirtualDirContext' --include '*.xml' '{}' +"),
+    "tomcat_vdc_targeted"       : CommandSpec("/bin/grep -R -s 'VirtualDirContext' --include '*.xml' {catalina}"),
```
bfahr: You'll need to add `, catalina=r"\S+"` or a modified version to the end of this in order for it to work with the precommand.

insights/parsers/tomcat_virtual_dir_context.py
matysek: @skontar I agree with Jakub. There should not be multiple parsers in inheritance. You shoudl add one parent class that inherits from Parser. And two child classes that inherit from the parent class. The child classes should then be decorated with @parser. Look at the iptables parser where I added it like this.

matysek: Or another example is the 'ps.py' parser.

insights/config/specs.py
skontar: @jsvob I ran this in Bash. Could not find how it works in the Insights yet.

insights/config/specs.py
skontar: Does it only check if it is there? I could not find the code which handles that. I am asking because I expect either `/some/path` or `/some/path /some/other/path` as a result.

skontar:
3.x https://github.com/RedHatInsights/insights-core/pull/683

PR #652, opened 2017-10-31

skontar:


skontar:
1.x https://github.com/RedHatInsights/insights-core/pull/651

skontar:
Needs a lot of rework when https://github.com/RedHatInsights/insights-core/pull/651 is approved. Closing for now.

PR #653, opened 2017-10-31
Add spec, parser, and combiner for tmpfiles.d

SteveHNH:
Spec, Parser, and Combiner for tmpfiles.d files.


insights/combiners/tmpfilesd.py
```diff
@@ -0,0 +1,79 @@
...
+        files = set([])
+        file_paths = []
+        self.tmpfiles = tmpfiles
+        rules_found = {}
+        rule_number = 0
+        # Files must be sorted so that /etc/tmpfiles.d overrides /run/tmpfiles.d
+        # which overides /usr/lib/tmpfiles.d. To do this, we have to list file
+        # paths in reverse order.
+        for tmpfiles in sorted(tmpfiles, key=lambda f: f.file_path, reverse=True):
+            for file_ in tmpfiles.rules:
```
xiangce: A little tricky too (to save the `rules_found` and `rule_number`):
Just FYI:
~~~
             for new_rule in tmpfiles.rules:                                                                                                                                                                      
                 _file = (new_rule['path'])                                                                                                                                                                       
                 if _file in files:                                                                                                                                                                               
                      [rules.remove(_rule) for _rule in rules if _rule['path'] == _file]                                                                                                                           
                 rules.append(new_rule)
~~~

insights/parsers/tmpfilesd.py
```diff
@@ -0,0 +1,69 @@
...
+    Examples:
+
+        >>> tmpfiles = shared[TmpFilesd][0] # List is per filename
+        >>> len(tmpfiles.rules)
+        2
+        >>> tmpfiles.files
+        ['/var/cache/dnf/*/*/download_lock.pid', '/var/cache/dnf/']
+        >>> tmpfiles.rules[0]
+        {'path': '/var/cache/dnf/*/*/download_lock.pid', 'type': 'r!',
+        'mode': '-', 'age': '-', 'gid': None, 'uid': None, 'argument': None}
```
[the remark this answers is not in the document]
SteveHNH: This is just an example of what you may see in the file. Sometimes '-' is used in order to get the order of arguments correct, and then any missing arguments are None. I can make this more accurate

insights/parsers/tmpfilesd.py
```diff
@@ -0,0 +1,69 @@
...
+                if key not in d.keys():
+                    d[key] = None
+
+            self.rules.append(d)
+
+        for i in self.rules:
+            self.files.append(i['path'])
+
+    def find_file(self, filename):
+
```
bfahr: @SteveHNH please add a docstring to this public method.

insights/parsers/tmpfilesd.py
```diff
@@ -0,0 +1,69 @@
...
+        {'path': '/var/cache/dnf/*/*/download_lock.pid', 'type': 'r!',
+        'mode': '-', 'age': '-', 'gid': None, 'uid': None, 'argument': None}
+        >>> tmpfiles.find_file('download_lock.pid')
+        [{'path': '/var/cache/dnf/*/*/download_lock.pid', 'type': 'r!',
+          'mode': '-', 'age': '-', 'gid': None, 'uid': None, 'argument': None}]
+    """
+
+    def parse_content(self, content):
+        self.files = []
+        self.rules = []
```
bfahr: Please document these two attributes in your class docstring.

insights/combiners/tmpfilesd.py
```diff
@@ -0,0 +1,79 @@
...
+        rules = []
+        files = set([])
+        file_paths = []
+        self.tmpfiles = tmpfiles
+        rules_found = {}
+        rule_number = 0
+        # Files must be sorted so that /etc/tmpfiles.d overrides /run/tmpfiles.d
+        # which overides /usr/lib/tmpfiles.d. To do this, we have to list file
+        # paths in reverse order.
+        for tmpfiles in sorted(tmpfiles, key=lambda f: f.file_path, reverse=True):
```
bfahr: It would be clearer if your loop variable `tmpfiles` was instead `tmpfile` or some other name so that is doesn't override the input parameter.

insights/combiners/tmpfilesd.py
```diff
@@ -0,0 +1,79 @@
...
+            {'path': '/tmp/sap.conf', 'mode': '644', 'type': 'x', 'age': None,
+             'gid': None, 'uid': None, 'argument': None}
+
+        Parameters:
+            filename(str): File path for the file to be searched in the rules
+
+        Returns:
+            (list): a list of rules where the path matches the file searched
+            for.
+            """
```
bfahr: The trailing quote should be indented out one level.

SteveHNH:
@bfahr @xiangce Cleaned it up. let me know if there is more I should do. Thanks!

insights/combiners/tmpfilesd.py
```diff
@@ -0,0 +1,74 @@
...
+             'gid': None, 'uid': None, 'argument': None}
+
+        Parameters:
+            filename(str): File path for the file to be searched in the rules
+
+        Returns:
+            (list): a list of rules where the path matches the file searched
+            for.
+        """
+        return [match for data in self.tmpfiles for match in data.find_file(filename)]
```
xiangce: Do you intend to return a valid `rule` via this method?
As per current code, it does not return the valid rule (when there are duplicates) but returns all of them. The plugin developers will have to do a further check to get the valid one. (And the `file_paths` is not returned together, it seems difficult to do the check)

insights/combiners/tmpfilesd.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Attributes:
+        files(set): the set of files found in all data files.
+        rules(dict): a dictionary of rules using the config file as the key
+        file_paths(list): a list of the file paths for the configurations.
+    """
+    def __init__(self, tmpfiles):
+        rules = {}
+        files = set([])
+        self.tmpfiles = tmpfiles
```
xiangce: Since the `combiner` will parse all the data passed from each `parser`, better do not store the original data of the `parser` here.

insights/combiners/tmpfilesd.py
```diff
@@ -0,0 +1,74 @@
...
+        # Files must be sorted so that /etc/tmpfiles.d overrides /run/tmpfiles.d
+        # which overides /usr/lib/tmpfiles.d. To do this, we have to list file
+        # paths in reverse order.
+        for tmpfile in sorted(tmpfiles, key=lambda f: f.file_path, reverse=True):
+            file_paths.append(tmpfile.file_path)
+            rules[tmpfile.file_path] = []
+            for new_rule in tmpfile.rules:
+                _file = (new_rule['path'])
+                if _file in files:
+                    [rules[tmpfile.file_path].remove(_rule) for _rule in rules[tmpfile.file_path] if _rule['path'] == _file]
```
xiangce: I don't think this `.remove` will work as expected according to current structure (`file_path` as the key), since at this time in each pass of loop the `rules[tmpfile.file_path]` is still an empty `[]`.

As per current structure, you have to iterate all the data which is already stored into `rules` to remove the duplicate rules.

Would you mind to try to add a new key `file_path` for each `rule`? It's redundant, but could simplify the logic a little. You can take a look at the `combiners/httpd_conf.py` for a reference.

insights/combiners/tmpfilesd.py
xiangce: And better use the parsed data did by this `combiner` to implement this method.

xiangce:
Btw, @SteveHNH, You'd better commit your following commit directly and do not use the `--amend` option. Since the commit with `--amend` option, will cover all former modification points and you have to `push` the commit by `--force`. It's not so convenient for the reviewers to check what have modified in the last commit :(
Thanks.

SteveHNH:
@xiangce Updated to take into account the other changes. Thanks for the feedback and input. I think I handled the remaining cases, but feel free to let me know any other changes I should make. I also made sure to not --amend this commit.

insights/combiners/tmpfilesd.py
xiangce: @SteveHNH - Sorry for my poor description. The `.remove` might still do not work as what you want, if you want to override the rule with the same `file` but defined in different `file_path` (it's not because of the empty `[]`, sorry)

The `tmpfile.file_path` is an iteration of the `outer` loop, it will be always the same one for the `inner` loop. So, it only works for if there are re-defined rules inner the same file, but does not work for different files.

For example, in case of the test `all_data = AllTmpFiles([data1, data3])`. It cannot remove the rule `r! /tmp/.sap*`.

It's OK to me if you just want to remove the duplicated rules inner the same file. But you might need to note that it won't remove between different file.

SteveHNH:
@xiangce I took another shot at it and I think I got it this time. Now the rule path is checked and if it's over-ridden by one of the files that takes precedence, the rule is removed from the dictionary. I believe the behavior is what we want now.

PR #645, opened 2017-10-30
Completely test ethtool parsers for 1.x branch

PaulWay:
* Rewrote the main parsing of ring information for consistency
* Some better exception handling (not all)
* Tests now cover 100% of the code

PR #655, opened 2017-10-31
Completely test ethtool parsers for master branch

PaulWay:
* Rewrote the main parsing of ring information for consistency
* Some better exception handling (not all)
* Tests now cover 100% of the code

This is the master branch version of #645


insights/parsers/ethtool.py
```diff
@@ -815,31 +823,16 @@ def parse_content(self, content):
                 except:
                     raise ParseException('Ethtool unable to parse content', line)
 
-        self._supported_link_modes = []
+        self.supported_link_modes = []
```
xiangce: Please add their description to the class's docsting.

PaulWay: Hi @xiangce - sorry for the delay, I was away.  The attributes in this are badly documented, so I'm going to document them throughout.

PaulWay:
OK, @xiangce, I've updated this - let me know what you think now.

PR #656, opened 2017-11-01
Parser For DU command

Sheersha-jain:
Parser for du command, and parsing the command output (which is written in file) from files.


Sheersha-jain:
@xiangce  @vishwanathjadhav 
Please review parser and test case for du command.

insights/config/specs.py
```diff
@@ -80,6 +81,8 @@
     "docker_storage"            : DockerHostSimpleFileSpec("etc/sysconfig/docker-storage"),
     "docker_storage_setup"      : DockerHostSimpleFileSpec("etc/sysconfig/docker-storage-setup"),
     "docker_sysconfig"          : DockerHostSimpleFileSpec("etc/sysconfig/docker"),
+    "du_-h"                     : CommandSpec("/usr/bin/du -h /"),
+    "du_-sh"                    : CommandSpec("/usr/bin/du -sh /"),
```
vishwanathjadhav: No need of this

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,55 @@
...
+
+    return du_output
+
+
+@parser("du_-h")
+class DiskUsage(LegacyItemAccess, Parser):
+    def __len__(self):
+        return len(self.data)
+
+    def parse_content(self, content):
```
vishwanathjadhav: Please add proper documentation for this. Refer other plugins for the same.

insights/config/specs.py
vishwanathjadhav: or if need specific to mongodb please add it.

Sheersha-jain: ok ,I'll remove `du` entries from config/specs.py file, actually from forman-debug archieve, we need to parse content of files"mongodb_disk_space" and "postgres_disk_space" which contain output of du in it (e.g, du -sh /var/lib/mongodb), i'll replace the du command specs with file specs.

insights/parsers/du_usage.py
Sheersha-jain: ok, i'll add documentation to it.

bfahr: FYI, publicly available documentation is available here: http://insights-core.readthedocs.io/en/1.x/

Sheersha-jain: @bfahr  I'll look into it.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,55 @@
...
+        self.data = parse_du_lines(content)
+
+    def get_size(self, dir_path):
+        return self.data[dir_path]
```
xiangce: You should check if the `dir_path` is in the `self.data` or not to avoid the `KeyError`.
Or use `return self.data.get(dir_path)` which will return None when `dir_path` does not exist.

Btw, seems the `get` of the `LegacyItemAccess` already provides the same functions to this `get_size`

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,55 @@
...
+
+Sample input data for the ``du -sh`` command looks like::
+22G     /var/lib/mongodb
+
+"""
+
+from .. import Parser, parser, LegacyItemAccess
+
+
+def parse_du_lines(du_content):
```
xiangce: If no other modules will use this function, better add a `_` as a prefix to indicate it's a private/local interface.

insights/config/specs.py
```diff
@@ -62,6 +62,7 @@
     "df_-al"                    : First([CommandSpec("/bin/df -al"),
                                     SimpleFileSpec("diskinfo")]),
     "df_-alP"                   : CommandSpec("/bin/df -alP"),
+    "df_-h"                     : CommandSpec("/bin/df -h"),
```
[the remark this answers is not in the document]
PaulWay: This seems redundant, especially as the human sizes lose accuracy compared `df -a`.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,55 @@
...
+    reassembled prior to parsing, and that mount names containing spaces
+    are maintained.
+
+    Parameters:
+        du_content (list): Lines of du output to be parsed.
+
+    """
+    for line in du_content:
+        line_split = line.split()
+        du_output[line_split[1]] = line_split[0]
```
PaulWay: I feel you're missing a perfectly good opportunity to provide this as an integer value so that rule writers can compare values directly rather than having to do the conversion to integer themselves.

But then I think this would work much better reading the `du -k` output to get an integer number of kilobytes so that you retain accuracy across all sizes of file and directory.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,55 @@
...
+    return du_output
+
+
+@parser("du_-h")
+class DiskUsage(LegacyItemAccess, Parser):
+    def __len__(self):
+        return len(self.data)
+
+    def parse_content(self, content):
+        self.data = parse_du_lines(content)
```
PaulWay: Don't bother with having a separate function to do this, just put the function code here.  It needlessly obscures the flow of the code.

insights/parsers/du_usage.py
PaulWay: I see no real point in having a separate parse function when you can just include this code in the `parse_content` method of the class.

insights/parsers/tests/test_du_usage.py
```diff
@@ -0,0 +1,62 @@
...
+8.0K	/var/lib/xkb
+4.0K	/var/lib/gssproxy/clients
+4.0K	/var/lib/gssproxy/rcache
+12K	/var/lib/gssproxy
+4.0K	/var/lib/net-snmp/cert_indexes
+4.0K	/var/lib/net-snmp/mib_indexes
+12K	/var/lib/net-snmp
+56K	/var/lib/upower
+4	/root
+92	/opt/google/chrome/default_apps
```
PaulWay: This looks like you've mixed 'du -h' and 'du' output here.  Are you sure this is accurate?

insights/parsers/tests/test_du_usage.py
```diff
@@ -0,0 +1,62 @@
...
+267032	/opt/google/chrome
+267036	/opt/google
+""".strip()
+
+
+def test_du_disk_usage():
+    context = context_wrap(du_content)
+    du_info = DiskUsage(context)
+
+    assert len(du_content) == 482
```
PaulWay: Why are you asserting something about the input data?

insights/parsers/du_usage.py
Sheersha-jain: Done

insights/parsers/du_usage.py
Sheersha-jain: done.

insights/parsers/tests/test_du_usage.py
Sheersha-jain: I was using that to test something, removed now.

insights/config/specs.py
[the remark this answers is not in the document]
Sheersha-jain: df command parser is out of my development scope, as I have not developed this.

insights/config/specs.py
Sheersha-jain: I have fixed this.

insights/parsers/du_usage.py
Sheersha-jain: done.

insights/config/specs.py
```diff
@@ -216,6 +216,7 @@
     "modinfo"                   : CommandSpec("/usr/sbin/modinfo {module}", module=r"\S+"),
     "modprobe.conf"             : PatternSpec(r"etc/modprobe\.conf"),
     "modprobe.d"                : PatternSpec(r"etc/modprobe.d/.*\.conf"),
+    "mongodb_disk_space"        : SimpleFileSpec("mongodb_disk_space"),
```
xiangce: The simple file `mongodb_disk_space` and `postgres_disk_space` you required must not exist if you do not generate them explicitly. 
Are they generated by any collection tool (sosreport, spacewalk or foreman-debug)?
If there are such simple files, you write them like:
~~~
db_disk_space : All([SimpleFileSpec("mongodb_disk_space"),
                             SimpleFileSpec("postgres_disk_space")]),
~~~

But for `insights-data` (the archive collected by insights-client), I don't suggest you generate these files.
So, at last, you can write them like:

~~~
db_disk_space : All([CommandSpec("/bin/du -sh /var/lib/mongodb /var/lib/pgsql"),
                             SimpleFileSpec("mongodb_disk_space"),
                             SimpleFileSpec("postgres_disk_space")]),
~~~

insights/parsers/du_usage.py
xiangce: I think @PaulWay must mean you'd better remove the `parse_du_lines` and shift its body into the `parse_content` directly.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,91 @@
...
+        return self.data
+
+    def get_size(self, dir_path):
+        return self.data.get(dir_path)
+
+
+@parser("du_-sh")
+class DiskUsage_SH(DiskUsage):
```
xiangce: There is no spec "du_-sh" anymore.
I think this class can be removed.

insights/parsers/tests/test_du_usage.py
xiangce: And this comment pointed out by Paul should be fixed too.

xiangce: As per your current logic, there is not `du -h` at all. Test `du -sh` only is enough.

insights/config/specs.py
Sheersha-jain: done.

insights/parsers/du_usage.py
Sheersha-jain: yup, I fixed it.

insights/parsers/du_usage.py
Sheersha-jain: removed.

Sheersha-jain: Removed now.

insights/parsers/tests/test_du_usage.py
Sheersha-jain: yup, my bad that was mixed output, I have changed it.

insights/parsers/du_usage.py
Sheersha-jain: Done.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,60 @@
+"""
+du - Command
+============
+
+Module for the processing of output from the ``du`` command stored in
+file.  The base class ``DBDiskSpace`` provides all of the functionality.
```
abhaykadam: As we are considering this as a base class, and this may not be database always, should we remove `DB` prefix from `DBDiskSpace`? May be we can name it `DirSpace` or `DirUsage`.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,60 @@
...
+
+This example demonstrates the funcionality of ``DBDiskSpace`` class.
+
+Examples:
+    >>> FILE_CONTENT = '''
+    ... COMMAND> du -sh /var/lib/mongodb
+    ...
+    ... 22G     /var/lib/mongodb
+    ... 635M    /var/lib/pgsql
```
abhaykadam: As `du -sh` command particularly asks for `/var/log/mongodb`, the output should contain usage for only that path.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@parser("db_disk_space")
+class DBDiskSpace(LegacyItemAccess, Parser):
+    """
+    Typical content of `mongodb_disk_space` command looks like:
+        COMMAND> du -sh /var/lib/mongodb
+
+        22G     /var/lib/mongodb
+        635M    /var/lib/pgsql
```
abhaykadam: Please refer my above comment.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+    def __len__(self):
+        return len(self.data)
+
+    def parse_content(self, content):
+        """Parse contents of each line in file which stores ``du`` output.
+
+        Parameters:
+            dbdisk_content : Lines of df output stored in file to be parsed.
```
abhaykadam: it should be `du` instead of `df`.

insights/parsers/tests/test_du_usage.py
```diff
@@ -0,0 +1,50 @@
+from insights.parsers.du_usage import DBDiskSpace
+from insights.tests import context_wrap
+
+dbdisk_content = """
+COMMAND> du -sh /var/lib/mongodb
+
+22G     /var/lib/mongodb
```
abhaykadam: Please refer my above comment regarding showing list of paths.

insights/parsers/du_usage.py
Sheersha-jain: I have changed it as DirUsage.

insights/parsers/du_usage.py
Sheersha-jain: updated this.

insights/parsers/du_usage.py
Sheersha-jain: done.

insights/parsers/du_usage.py
Sheersha-jain: done.

insights/parsers/tests/test_du_usage.py
Sheersha-jain: done

insights/parsers/du_usage.py
Sheersha-jain: I have again modified it as DiskUsage, and changed all names as well which contained db prefix.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,58 @@
...
+
+This example demonstrates the funcionality of ``DiskUsage`` class.
+
+Examples:
+    >>> FILE_CONTENT = '''
+    ... COMMAND> du -sh /var/lib/mongodb
+    ...
+    ... 22G     /var/lib/mongodb
+    ... '''
+    >>> du_info.get_size('/var/lib/mongodb')
```
abhaykadam: we need to initialize `du_info` before using it.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,58 @@
...
+    22G
+    >>> du_info.get_size('/var/lib/pgsql')
+    635M
```
abhaykadam: shouldn't we get exception here rather than some random value?

abhaykadam:
@Sheersha-jain, naming file as `du_usage` seems redundant as `du` itself implies usage.

insights/parsers/du_usage.py
Sheersha-jain: Modified.

insights/parsers/du_usage.py
Sheersha-jain: We will get here "None", I have modified test case and documentation according to this.

Sheersha-jain:
@xiangce After writing in specs file like 
```
db_disk_space : All([SimpleFileSpec("mongodb_disk_space"),
                             SimpleFileSpec("postgres_disk_space")]),
```
rule is not hitting for both files at the same moment, i.e both files are not parsed at the same time.
```
sat6_report_free_space:
    error_key     : "SAT6_REPORT_FREE_SPACE"
    mongodb_space : "22G"
    pgsql_space   : None
```

insights/parsers/du.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+from .. import Parser, parser, LegacyItemAccess
+from .. import get_active_lines
+
+
+@parser("disk_space")
+class DiskUsage(LegacyItemAccess, Parser):
+    """
+    Typical content of `mongodisk_space` command looks like:
```
bfahr: There should be a double colon `::` at the end of this line to properly format the following lines.

docs/shared_parsers_catalog/du_usage.rst
```diff
@@ -0,0 +1,3 @@
+.. automodule:: insights.parsers.du_usage
```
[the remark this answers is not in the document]
Sheersha-jain: Yes, I have actually changed the name of file from du_usage to du.

insights/parsers/du_usage.py
```diff
@@ -0,0 +1,58 @@
+"""
+du - Command
+============
```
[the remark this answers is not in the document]
Sheersha-jain: yes, I have just renamed this module as du.py instead of du_usage.py, as @abhaykadam reviewed.

docs/shared_parsers_catalog/du_usage.rst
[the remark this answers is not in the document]
bfahr: You still need to delete du_usage using `git rm`.

insights/config/specs.py
```diff
@@ -67,6 +67,8 @@
     "dirsrv"                    : SimpleFileSpec("etc/sysconfig/dirsrv"),
     "dirsrv_access"             : PatternSpec("var/log/dirsrv/.*/access"),
     "dirsrv_errors"             : PatternSpec("var/log/dirsrv/.*/errors"),
+    "disk_space"                : All([SimpleFileSpec("mongodb_disk_space"),
+                                    SimpleFileSpec("postgres_disk_space")]),
```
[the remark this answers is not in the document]
bfahr: Currently in 1.x, regardless of whether this is a PatternSpec or a SimpleFileSpec, only the first file found will be collected.  So you can split this into two specs or wait until the problem is corrected in 1.x.

insights/parsers/tests/test_du_usage.py
```diff
@@ -0,0 +1,15 @@
+from insights.parsers.du_usage import DiskUsage
```
bfahr: Remember to `git rm` this file as well.

docs/shared_parsers_catalog/du_usage.rst
bfahr: Remember to `git rm` this file as well.

insights/config/specs.py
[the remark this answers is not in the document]
PaulWay: Aren't these SimpleFileSpecs looking for a file called 'mongodb_disk_space' in the root directory (of the client file system, or the sosreport)?  Shouldn't they be an actual command?  How do these files actually get generated?

insights/parsers/du.py
```diff
@@ -0,0 +1,59 @@
...
+        Parameters:
+            disk_content : Lines of du output stored in file to be parsed.
+
+        Returns:
+            Dict : A dict of key pair values where key is filesystem path
+            and value is the size of filesystem.
+        """
+
+        data = {}
+        for line in get_active_lines(content, comment_char="COMMAND>"):
```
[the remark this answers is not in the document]
PaulWay: @bfahr I think it'd better to allow multiple lines, so that one du command can gather statistics from multiple directories.

insights/parsers/du.py
```diff
@@ -0,0 +1,59 @@
...
+            size, path = line.split()
```
PaulWay: This parsing makes a lot of assumptions:
* The path doesn't contain any spaces.
* The line always has at least two parts (what about blank lines mentioned above)
* The line contains a size and a path - what about "/bin/du: Permission denied" or "/bin/du: No such file or directory"?

insights/parsers/du.py
```diff
@@ -0,0 +1,59 @@
...
+            data[path] = size
+        self.data = data
+        return self.data
+
+    def get_size(self, dir_path):
```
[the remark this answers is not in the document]
PaulWay: I don't see the need for a 'get_size' method if the user can simply use the `data` attribute to look directories up directly - especially as this is also using LegacyItemAccess for direct access to the data.

insights/parsers/du.py
```diff
@@ -0,0 +1,59 @@
...
+Module for the processing of output from the ``du`` command stored in
+file.  The base class ``DiskUsage`` provides all of the functionality.
+
+This example demonstrates the funcionality of ``DiskUsage`` class.
+
+Examples:
+    >>> FILE_CONTENT = '''
+    ... COMMAND> du -sh /var/lib/mongodb
+    ...
+    ... 22G     /var/lib/mongodb
```
PaulWay: I think it's a bad idea to use 'human' units for this command - it should be in kilobytes (`du -sk`).  You retain a lot more accuracy, and it's much easier for rules based on this input to compare values to other things - e.g. free disk space, file sizes, etc.

insights/parsers/du.py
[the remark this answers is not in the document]
PaulWay: I would, however, like to see something like `get_path_in_kb` which converted human readings (should they continue to be used) into kilobytes.

insights/parsers/tests/test_du_usage.py
```diff
@@ -0,0 +1,15 @@
...
+
+22G     /var/lib/mongodb
+""".strip()
+
+
+def test_du_disk_usage():
+    context = context_wrap(disk_content)
+    dbdisk_info = DiskUsage(context)
+
+    assert dbdisk_info.get_size('/var/lib/mongodb') == "22G"
```
PaulWay: You need a lot more tests:
* Tests of input of multiple directories.
* Tests of handling paths with spaces.
* Tests of handling other invalid inputs.
* Tests of what should be returned when retrieving directories that weren't read.
* Tests of other access methods - access via `data` attribute and via the LegacyItemAccess methods.

insights/config/specs.py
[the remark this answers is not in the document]
xiangce: @bfahr - I have made a test for PatternSpec and SimpleFileSpec in `All` and `NoneGroup`, the result is:
1. No matter PatternSpec or SimpleFileSpec, if there are put in an `All` function, all the existed files will be collected by our client_1.x.
2. **However**, for All([SimpleFileSpec()..]), only the first file will be parsed by the framework, all others are just skipped.
3. For `All([PatternSpec()..]) or NoneGroup([PatternSpec()..])`, all the collected files will be parsed, but be careful, the parsed result will be **wrapped into a list**, so it's necessary for rule developer to iterate the list to get each file's result.

@Sheersha-jain, for the `NoneGroup` we talked on IRC yesterday, it's used to indicate the collector do not attempt to collect the files in it.
So, here you can 
1). use PatternSpec instead of SimpleFileSpec, and replace the `All` to `NoneGroup`.
2). continue using SimpleFileSpec, but split them into two `NoneGroup` specs.
I prefer to choose 1), since the two files are already collected by sosreport or foreman-debug and they are basically in the same format and can be parsed by one Parser by using one `spec`
How do you think it, @Bob?

@PaulWay - Just FYI, these two files can be found in sosreport/foreman-debug archive when corresponding files exist in the system.

```
[root@liuxc-rhel7-sat6 ~]# find -name "*_disk_space*"                                          
./sosreport-liuxc-rhel7-sat6-20171106112737/sos_commands/foreman/foreman-debug/postgres_disk_space                                                                                             
./sosreport-liuxc-rhel7-sat6-20171106112737/sos_commands/foreman/foreman-debug/mongodb_disk_space                                                                                              
./foreman-debug-HBR7p/postgres_disk_space      
./foreman-debug-HBR7p/mongodb_disk_space       
[root@liuxc-rhel7-sat6 ~]# cat ./foreman-debug-HBR7p/mongodb_disk_space
COMMAND> du -sh /var/lib/mongodb

5.1G    /var/lib/mongodb
[root@liuxc-rhel7-sat6 ~]# cat ./sosreport-liuxc-rhel7-sat6-20171106112737/sos_commands/foreman/foreman-debug/postgres_disk_space
COMMAND> du -sh /var/lib/pgsql

216M    /var/lib/pgsql
```
IMHO, I think we can add these two commands "du -sh /var/lib/mongodb(or pgsql)" to the `specs.py` when there are some public rules need them in the future.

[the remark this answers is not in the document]
PaulWay: @xiangce Then the correct way AFAIK to do that is with the specs:
```python
    "mongodb_disk_space"                : First([SimpleFileSpec("mongodb_disk_space"), CommandSpec("du -sh /var/lib/mongodb")]),
    "postgres_disk_space"                : First([SimpleFileSpec("postgres_disk_space"), CommandSpec("du -sh /var/lib/pgsql")]),
```

And then have one base parser class - e.g. `DiskUsage` - and two parsers, say `MongoDBDiskUsage` and `PostgresDiskUsage` based off that base class.

A path that's specific to Red Hat's way of gathering customer data in a sosreport is useless for the rest of the world that doesn't use a sosreport.  It's useless for the customers we have that use Insights rather than the CLI.  So I would really strongly encourage @Sheersha-jain to consider all users of this parser, not only those based on reading a sosreport.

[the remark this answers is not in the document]
xiangce: @PaulWay - Yep, splitting them into two also makes sense to me. And in your example, it easily achieved to support the sosreport and insights-data.

**But**, you should put the CommandSpec in the first place, since the insights-client will only collect the first command/simple-file for `First`:

```
    "mongodb_disk_space"                : First([CommandSpec("du -sh /var/lib/mongodb"), SimpleFileSpec("mongodb_disk_space")]),
    "postgres_disk_space"               : First([CommandSpec("du -sh /var/lib/pgsql"), SimpleFileSpec("postgres_disk_space")]),

```

insights/parsers/du.py
csams: Blank lines are filtered in `get_active_lines`.

insights/config/specs.py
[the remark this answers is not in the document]
csams: @xiangce @PaulWay Even if an All group has a PatternSpec, only the first one that produces a result is evaluated by the engine in 1.x. [The code](https://github.com/RedHatInsights/insights-core/blob/1.x/insights/core/specs.py#L106) breaks after the first match. The [intent of the All group](https://github.com/RedHatInsights/insights-core/blob/1.x/insights/config/__init__.py#L309) is to cover cases where the same file(s) might be in different places in different versions of RHEL, so the collector is instructed to collect everything in an All group, but in reality the engine expects only one of the individual entries to be in a given archive.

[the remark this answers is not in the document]
PaulWay: @csams So what's the difference between the 'First' and 'All' groups?
@xiangce Yes, that'd be better to put the command first :-)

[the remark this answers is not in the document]
Sheersha-jain: @PaulWay I think need to specify "/usr/bin" in command specs 
```
    "mongodb_disk_space"        : First([CommandSpec("/usr/bin du -sh /var/lib/mongodb"), SimpleFileSpec("mongodb_disk_space")]),
    "postgres_disk_space"       : First([CommandSpec("/usr/bin du -sh /var/lib/pgsql"), SimpleFileSpec("postgres_disk_space")]),
```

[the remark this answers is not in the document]
xiangce: @csams - Yes, my fault. The `All([PatternSpecs..]) only works in unit test :(. I tested it with the `insights-cli` tool, it indeed only parsed the first file.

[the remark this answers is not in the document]
xiangce: And I have the same question: what's the difference between `First` and `All`?  :((

insights/parsers/tests/test_du_usage.py
Sheersha-jain: Done

xiangce: Not yet, the files are still here.

eduardocerqueira:

bfahr:
ok to test

xiangce:
Since there are many problems are not handled in this PR, I will go ahead and close it. If your rules require this parser, please raise another PR to implement this parser freely. Thanks.

PR #657, opened 2017-11-01
Add ipython to setup.py

matysek:
This adds ipython as develop dependency to setup.py when we started using .ipynb files.


kylape:
@matysek Please specify a version of IPython less than 6.0 in setup.py so that shippable doesn't fail.

matysek:
@kylape done :)

PR #650, opened 2017-10-31
Make `parse_lines` a private interface

xiangce:
- rename `parse_lines` to `_parse_line`, and let it parse  only one line each time
- make it as a private interface of the class, to avoid parsing `lines`
  which are not belongs to the instance of the class object
- add `_parse_line` for the base class
- let `get` and `get_after` return parsed line directly
  then,  no need to call `parse_lines` explicitly, e.g.
  CLS.parse_lines(obj.get_after(xxx))
- uniform the `raw_message` as the key for the raw lines

Signed-off-by: Xiangce Liu <xiangceliu@redhat.com>

PR #660, opened 2017-11-02
Refactor `get_after` and `_parse_line` - 3.x

xiangce:
The re-submission of #650 for 3.x version


xiangce:
Will separate it into small ones.

Issue #643, opened 2017-10-27
Update docs and create PR for master branch for libkeyutils parser


bfahr:
Reminder to add the docs to parser and create a PR for #642.

xiangce:
This issue can be closed. See #661 and #667

PR #661, opened 2017-11-02
Fix documentation and parsing algorithm for libkeyutils parser

skontar:
It is not embargoed anymore, so the information can be
shared in the documentation.

https://github.com/RedHatInsights/insights-core/issues/643


skontar:
@xiangce if it looks good I will add it to the master branch.

insights/parsers/libkeyutils.py
```diff
@@ -1,18 +1,115 @@
...
+    NEED_PATTERN = re.compile(r'NEEDED\s+(.*)\s*$')
+
+    def __init__(self, *args, **kwargs):
+        self.linked_libraries = {}
+        """dict: found libraries and their linked libraries."""
+        super(LibkeyutilsObjdumps, self).__init__(*args, **kwargs)
+
     def parse_content(self, content):
-        self.data = list(content)
+        file_name = None
```
jsvob: I think the `file_name = None` line should be moved one line down - to be at the beginning inside the `for` loop.

With the current code as it is, I *think* that if there is file 1 that matches only FILE_PATTERN, it sets `file_name`, and then if there is another file 2 that **doesn't** match FILE_PATTERN but does match NEED_PATTERN, it will get improperly recorded into `linked_libraries`. I didn't test it though.

jsvob: I'm not so sure but maybe some static binaries can have nothing in NEEDED.

skontar: The parser only runs on object dumps from `libkeyutils.so.1` files. See the `specs.py`, the command parsed is `find -L /lib /lib64 -name libkeyutils.so.1 -exec objdump -x "{}" \;`.

skontar:
3.x https://github.com/RedHatInsights/insights-core/pull/667

insights/parsers/libkeyutils.py
bfahr: @skontar did you resolve @jsvob's question about whether or not `file_name = None` should be moved into the loop?

jsvob: @skontar I missed that - in this case, I think the code is fine as it is. @bfahr I deem it resolved and I approve of the request. Sorry for not responding in time.

xiangce: @jsvob - Thanks for the reply, I'm going to merge it.

PR #665, opened 2017-11-03
Fix the spec of installed_rpm

shzhou12:


xiangce:
@shzhou12 - Would you please to explain the detail reason of this modification? Since this spec relates to many plugins, we should involve more reviewer to make sure it's reasonable. Thanks.

shzhou12:
1. The description of the issue:

In the file "spec.py" we will get the following column for the rpm command.

def _make_rpm_formatter(fmt=None):
    if fmt is None:
        fmt = [
            "%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}",
            "%{INSTALLTIME:date}",
            "%{BUILDTIME}",
            "%{RSAHEADER:pgpsig}",
            "%{DSAHEADER:pgpsig}"
        ]

While in the parser "installed_rpm", we defined extra columns "vendor' and 'build server' which we never collect from the client. 
    SOSREPORT_KEYS = [
        'installtime', 'buildtime', 'vendor', 'buildserver', 'pgpsig', 'pgpsig-short'
    ]

This lead to an reducer error when a plugin calls the attributes  'vendor', 'buildserver', 'pgpsig', 'pgpsig-short' of the paser InstalledRpm as the following example.

Reducer telemetry.rules.plugins.container.docker_redhat_version.report failed
    if (docker_rpm and "199e2f91fd431d51" not in docker_rpm["pgpsig-short"]) or not docker_rpm:
  File "/home/shzhou/insights/insights-core/insights/parsers/installed_rpms.py", line 471, in __getitem__
    return getattr(self, item)
AttributeError: 'InstalledRpm' object has no attribute 'pgpsig-short'

shzhou12:
Add the following two columns to get the value of attributes "vendor" and "buildserver" 
+            "%{VENDOR}",   
+            "%{BUILDHOST}",

RSAHEADER and DSAHEADER are two different internal signature types. I changed the following two columns to get the value of the attributes "pgpsig", "pgpsig-short" . 
The variable "%{SIGPGP}"  return a long format older signature. 
The variable "%|DSAHEADER?{%{DSAHEADER:pgpsig}}:{%|RSAHEADER?{%{RSAHEADER:pgpsig}}:{(none}|}|" returns the short signature information.
-            "%{RSAHEADER:pgpsig}",
-            "%{DSAHEADER:pgpsig}"
+            "%{SIGPGP}",
+            "%|DSAHEADER?{%{DSAHEADER:pgpsig}}:{%|RSAHEADER?{%{RSAHEADER:pgpsig}}:{(none}|}|"

shzhou12:
The following outputs are the testing result in rhel6 and rhel7.

```

[vagrant@rhel6-box ~]$ rpm -q fontconfig --qf   "%{NAME}"-"%{VERSION}"-"%{RELEASE}"."%{ARCH}\t""%{INSTALLTIME:date}\t""%{BUILDTIME}\t""%{VENDOR}\t""%{BUILDHOST}\t""%{SIGPGP}\t""%|DSAHEADER?{%{DSAHEADER:pgpsig}}:{%|RSAHEADER?{%{RSAHEADER:pgpsig}}:{(none}|}|\n"
fontconfig-2.8.0-5.el6.x86_64	Wed 19 Jul 2017 07:06:27 AM UTC	1400676162	Red Hat, Inc.	x86-023.build.eng.bos.redhat.com	890215030500537db83c199e2f91fd431d51010842550ffe375b11dc021c79154ac09101864dbfd0149f47d8b9cf9ee9eac0f1e709cf140cd61468285a1ff9c985bc38cfc46c0a18a4fbf921487ceab0c92e1b71ae7a5428f206cac7bbed16f43d7597645092d3880cebba637eea3aee43dd2a58f3bbd375d608046a916de30451d05ac226205344795dc44ee502d76ab03a0d6ac1cf82158ba248a957918aac88efa554e7f15c799a570718bfaa0a49a212e4d6d6adea56405e805990fad78c4b48422ecde29ae107069d736c5fd70f9885a779d09c078efebff446ce0d5fea949943fbb1fff33525db30ae4e90e1fad8d828384ea00a0e1502b2b9ace48f3308da1214ba2bf2b54643d1813636be705b854890d84b2d556a26d139e4d49e5e9c2302e59543a88ba3482885d73e217e17ef27cd9dfb4bb3755ca96972c1ad324f36ad7c1da5df277fac23315849ca8280266db2297cf3e9a2c96381c2051faa8f1d2dfaf6c0fb8516a13b0adb82fd7ab11658bca841ca2e15f651452e5424b60b3cc72afd6d3f93c95a0f50466f0e3a0ec240fdc0d2f575e7bef4eeb46de251691da9617341370fd9ffedd27efa90e84a21515cc44d1bab160c412b9a1b9327c51569892aad9f254908cf9be8f9f23cbb3dee09341c303be91ff6f28c7923d2befa98c22501568df06b04adfb6b0c185284c55b24b38e0d252d43a047c75965e3a7ff5adb998182986eee607d31cfa85e116cc5f9877525	RSA/8, Thu 22 May 2014 08:41:32 AM UTC, Key ID 199e2f91fd431d51

[vagrant@rhel7-box ~]$  rpm -q docker --qf   "%{NAME}"-"%{VERSION}"-"%{RELEASE}"."%{ARCH}\t""%{INSTALLTIME:date}\t""%{BUILDTIME}\t""%{VENDOR}\t""%{BUILDHOST}\t""%{SIGPGP}\t""%|DSAHEADER?{%{DSAHEADER:pgpsig}}:{%|RSAHEADER?{%{RSAHEADER:pgpsig}}:{(none}|}|\n"
docker-1.12.6-61.git85d7426.el7.x86_64	Mon 06 Nov 2017 02:09:49 AM GMT	1506454338	Red Hat, Inc.	x86-039.build.eng.bos.redhat.com	89021503050059cab972199e2f91fd431d51010856890fff45b36b56cc06d0b765fe23f1a6d2e3412b798c9d217c73fa276856130b3d7364cbae910beea2e16d2cbff205dca7349be0e0e6040493b04ef8310fd48176926c1f9c291e730dc6a5779f24d0f3faff14973d092651278433fc8e85f8e381d6b58eb14503ac38aaaca10f67c03740d648e43b5948d4261ddcaf81a87757a6b4c51cedaeef424a5d28676620e6fb4772ad8f6d853a7edd7293b95597d199b1d8aa175a67bf3399c39531866db8e8df82b0971a3249e0f4d831f9d0d3d0958ac97cf8d9f0aa0f8cf0c05a0c61d08f20c06850b64799745789d6ccb1a6f2148bdf2dc1320bcf42a0835ef572a6766a9803cc794f3dd7b02311fdd8e36ef6b996dd9fa767af015a6775f90b165eaf0e5e1f87c42545863c5737ee9532682bf9d2fd144326970edb15332d07ae8e95fdeb1065e7ecfd6c084582a8b9cebaffd2f4c0fa6a82fb0f78f7a11f952163bf39f5ad4d63cd33c3802fdb42397e5bb3c747f0fc9af2d262e63deb79abd35048652abd085cea190d1e03ddb07ee2cf64f65e43bd9e06f439792823972fe81712c095794f723280261ada26240655b0b3f31fdbc7de5808fef44c03fa8250d6ad624f74471756499a84ac45d640fbf53064c864648bfb50439d3a727f54648854994514df65e4ccf88b132eff23131bf0fd2238d9f3792d249d4ded1cae0883fafe838ff1494ac8ee75515b528ed6a7bbf17075fe	RSA/SHA256, Tue 26 Sep 2017 08:32:50 PM GMT, Key ID 199e2f91fd431d51

```

xiangce:
@kylape - Would you please to help check if the `json_format` need to be updated as well as the `_make_rpm_formatter`? I'm not sure when will the `json_format` result of rpm command will be generated. Thanks.

skontar:
To be honest, I do not have an idea what is going on here in this change. But I ran all `prodsec.rules` tests using this branch of 1.x and it passes.

xiangce:
@matysek - I think so, this patch is trying to fix this issue. But it's difficult to test it as per current framework and client structure by ourselves.

kylape:
I'm taking a look at this now.

kylape:
I tested this by:

- applying the patch
- running `gen_api` to create a new uploader.json
- Copied the uploader.json to `/etc/redhat-access-insights/.cache.json`
- Ran `redhat-access-insights --no-gpg --no-upload` to create an archive using the new spec
- `python -m insights.tools.perf -p insights.plugins /path/to/archive.tar.gz`

I created a very basic rule to dump a bunch of info from the `InstalledRpm` parser:

```python
from insights.core.plugins import rule


@rule(requires=[InstalledRpms])
def report(local, shared):
    for name, p in shared[InstalledRpms].packages.iteritems():
        print p[0].name
        print p[0].epoch
        print p[0].version
        print p[0].release
        print p[0].arch
        print p[0].pgpsig
        print p[0].pgpsig_short
```

With this info, I verified that the new archive works as expected, and @matysek's issue is solved (assuming `SIGPGP` and `SIGPGP:sigpgp`).  Unfortunately any archive with the old format does not work.  We'll need to rework the parser to know how to read old formats.

Another problem is that the archive size increased from 132K to 409K.  Unpacking the archive, the "old" format `rpm -qa` file is 61K, whereas the new one is **555K**.  If we can get away with only getting `SIGPGP:sigpgp` and not `SIGPGP`, that would probably help out a lot.

matysek:
@kylape For my current usecase I do not need `SIGPGP`.  I need only `%{BUILDTIME}` and `%{SIGPGP:pgpsig}`.

xiangce:
@kylape - Did you met any fails when using this patch on old-format archives? Would you like to share these failed cases with me? PS. in my test, it works for both the old-format and new-format archives.

BTW, I don't think the `SIGPGP` is needed by the `docker_redhat_versio` too. @shzhou12 will confirm that in further.

shzhou12:
@kylape Thanks for your review.
For your questions:
1. I do not understand the failure you met when you run archives with the old format.
2. Currently, we do not use  "%{SIGPGP}" in plugins.  The format of rpm query in my first commit is aligning with the sos-report.  As the insights-core parses both sos-report and insights archive, I would like to add a dummy column "DUMMY" to replace "%{SIGPGP}" to avoid the large archive size.
I will send my second commit later.

kylape:
To reproduce the failure, use `redhat-access-insights --no-upoad` with the currently published `uploader.json` to create an archive.  Use that archive as input to an rules run.  I use the `insights.tools.perf` module to run the framework directly from the command line:

    python -m insights.tools.perf -p insights.plugins /path/to/archive

See my above comment for the `print_rpm_spec.py` plugin I stick in the `insights/plugins` package.  This setup should produce output indicating that the `installed-rpm` spec was not read.

The reason for this is that the specs are saved in the archive based on their command line invocation, including all options.  So when you change the format of the RPM command, you change the name of the file in the archive.  Specs are matched with their corresponding file using a regular expression based on the command format.  Thus, a modified command name means old formats will not be matched up to the spec.

Perhaps if we scan for the old format in addition to the new one for a short period of time after we merge this change, we can avoid missing any `installed_rpm` specs.

Also, it's possible I'm overestimating the number of clients that don't immediately fetch a new uploader.json, which would cause clients to avoid this issue altogether.

xiangce:
Okay @kylape, I got your point. And I have met the same problem before on the "lvs_noheading" spec before.

How do you think that we temporary modify the `installed-rpms` from `First()` to `All()` and keep both the old-format and new-format in it could solve this problem? After 24-hours or more we then change back to `First()` and remove the old-format.

I'm going to have a test at first.

xiangce:
Hi @kylape, as per my test, changing `First` to `All` can solve the command match issue you mentioned. The `installed_rpms` parser will search the new-format at first. If new-format is not found, it will continue to search the old-format. I think it could solve this problem temporary.

xiangce:
@shzhou12, @matysek - After a short discussion with @bfahr, we think it's better to change the `pgpsig-short` to `pgpsig_short` , since the latter is a valid attribute name and the `InstalledRpm` will set all the `keys` as an attribute.

xiangce:
@shzhou12 - the renaming of "pgpsig_short" will break the test of `docker_redhat_version.py` you should update this plugin as well.

xiangce:
I changed the `All` back to `First`. Since adding the `old-format` to the `First` list works well for the command name issue.

Thanks for your reminder @shzhou12.

xiangce:
Tested according to @kylape 's suggestion:

```python
    for name, p in shared[InstalledRpms].packages.iteritems():
        print '======================================'
        print 'name=        ', p[0].name
        print 'epoch=       ', p[0].epoch
        print 'version=     ', p[0].version
        print 'release=     ', p[0].release
        print 'arch=        ', p[0].arch
        if hasattr(p[0], 'vendor'):
            print 'vendor=      ', p[0].vendor
        if hasattr(p[0], 'buildtime'):
            print 'buildtime=   ', p[0].buildtime
        if hasattr(p[0], 'buildserver'):
            print 'buildserver= ', p[0].buildserver
        if hasattr(p[0], 'pgpsig'):
            print 'pgpsig=      ', p[0].pgpsig
        if hasattr(p[0], 'pgpsig_short'):
            print 'pgpsig_short=', p[0].pgpsig_short
```

The output of old-archive:
```
======================================
name=         glibc
epoch=        0
version=      2.17
release=      196.el7
arch=         x86_64
vendor=       RSA/SHA256, Fri Jun 16 15:48:42 2017, Key ID 199e2f91fd431d51
buildtime=    1497632365
buildserver=  (none)
====================================== 
```
The output of new archive:
```
======================================
name=         glibc 
epoch=        0
version=      2.17
release=      196.el7
arch=         x86_64
vendor=       Red Hat, Inc.
buildtime=    1497632365
buildserver=  x86-017.build.eng.bos.redhat.com
pgpsig=       DUMMY
pgpsig_short= RSA/SHA256, Fri Jun 16 15:48:41 2017, Key ID 199e2f91fd431d51
======================================
```

It looks OK.

xiangce:
Please refer to #688

PR #666, opened 2017-11-03
Fix the spec of installed_rpm

shzhou12:


xiangce:
See #665

xiangce:
The travis/pr test is already finished and passed but seems the result is still not passed to here. I will go ahead and ignore that it reports `tests haven't completed.`

PR #667, opened 2017-11-03
Fix documentation and parsing algorithm for libkeyutils parser (3.x)

skontar:
It is not embargoed anymore, so the information can be
shared in the documentation.

https://github.com/RedHatInsights/insights-core/issues/643


skontar:
1.x https://github.com/RedHatInsights/insights-core/pull/661

PR #683, opened 2017-11-06

skontar:


skontar:
@bfahr @xiangce Hi. I wanted to add the new-style specs, but it seems that there is a new way to do a command composition (probably using `sf.foreach`). I could not find any documentation or a way how to reasonably test it. Do you have any hints?

skontar:
1.x https://github.com/RedHatInsights/insights-core/pull/651

insights/combiners/tomcat_virtual_dir_context.py
```diff
@@ -0,0 +1,39 @@
...
+    {'/usr/share/tomcat/conf/server.xml':
+     ['    <Resources className="org.apache.naming.resources.VirtualDirContext"'],
+     }
+"""
+from ..core.plugins import combiner
+from ..parsers.tomcat_virtual_dir_context import TomcatVirtualDirContextFallback, \
+    TomcatVirtualDirContextTargeted
+
+
+@combiner(requires=[[TomcatVirtualDirContextFallback, TomcatVirtualDirContextTargeted]])
```
bfahr: @skontar the decorator changes for 3.x, you'll need to change this to:
`@combiner([TomcatVirtualDirContextFallback, TomcatVirtualDirContextTargeted])`

Take a look a some of the other combiners for more examples.  They have all been converted to the 3.x format.

insights/combiners/tomcat_virtual_dir_context.py
```diff
@@ -0,0 +1,39 @@
...
+class TomcatVirtualDirContextCombined(object):
+    """
+    Combiner for VirtualDirContext usage in Tomcat config files.
+    """
+    def __init__(self, local, shared):
```
bfahr: Intead of `local, shared`, the arguments to `__init__` are the instance objects for the classes you specify in the `@combiner` decorator.  So for instance:
```
def __init__(self, tomcat_vdc_fallback, tomcat_vdc_targeted):
```

bfahr:
Adding @csams to the discussion re @skontar's question regarding the 3.x specs.

csams:
Hey @skontar, I don't think you need command composition. You can use `SpecFactory.simple_command` like this:
```python
tomcat_vdc_targeted = sf.simple_command("/bin/grep -R --include '*.xml' 'VirtualDirContext' /usr/share/tomcat*", name="tomcat_vdc_targeted")
```
Specs are notoriously hard to test without a QE environment. Here's how I verified that something like it runs on my local environment:
```python
In [1]: from insights import run

In [2]: from insights.core.spec_factory import SpecFactory

In [3]: sf = SpecFactory()

In [4]: nginx_video_types = sf.simple_command("/bin/grep -R --include '*.types' 'video' /etc/nginx", name="nginx_video_types")

In [5]: results = run(nginx_video_types)

In [6]: results[nginx_video_types].content
Out[6]: 
['/etc/nginx/mime.types:    video/3gpp                            3gpp 3gp;',
 '/etc/nginx/mime.types:    video/mp2t                            ts;',
 '/etc/nginx/mime.types:    video/mp4                             mp4;',
 '/etc/nginx/mime.types:    video/mpeg                            mpeg mpg;',
 '/etc/nginx/mime.types:    video/quicktime                       mov;',
 '/etc/nginx/mime.types:    video/webm                            webm;',
 '/etc/nginx/mime.types:    video/x-flv                           flv;',
 '/etc/nginx/mime.types:    video/x-m4v                           m4v;',
 '/etc/nginx/mime.types:    video/x-mng                           mng;',
 '/etc/nginx/mime.types:    video/x-ms-asf                        asx asf;',
 '/etc/nginx/mime.types:    video/x-ms-wmv                        wmv;',
 '/etc/nginx/mime.types:    video/x-msvideo                       avi;']
```

xiangce:
Hey @csams, here to use the command composition is because of the `/usr/share/tomcat*` is not a fixed position, but can be specified in the command line by the customer. So, we'd better get it from the output of `ps auxwww`.

@skontar - a simple way is to put the whole pre-command `catalina` to an `sf.simple_command` and then define the `tomcat_vdc_targeted` with `sf.foreach`. However, the pre-command is so complex, I'm not sure it will work well.

csams:
The shell command for finding `catalina\.(home|base)` doesn't work with `simple_command`. I need to expose the ability to execute through a shell and also to specify whether stderr gets returned with stdout in the content. Some commands like `smartctl` return info on stderr, stdout, AND with a non-zero return code. Until I sort that out, here's a script showing an alternative.

As a side note, I want to get [psutil](https://github.com/giampaolo/psutil) included as a standard dependency for core.

```python
#!/usr/bin/env python                                                           
import re                                                                       
from insights.core.spec_factory import SpecFactory                              

@datasource(HostContext)                                                        
def tomcat_home(broker):                                                        
    ctx = broker[HostContext]                                                   
    ps = ctx.shell_out("ps auxwww")                                             
    results = []                                                                
    findall = re.compile(r"\-Dcatalina\.(home|base)=(\S+)").findall             
    for p in ps:                                                             
        found = findall(p)                                                      
        if found:                                                               
            results.extend(f[1] for f in found)                                 
    return list(set(results))                                                   
                                                                                
sf = SpecFactory()                                                              
tomcat_vdc_targeted = sf.foreach(tomcat_home, "/bin/grep -R -s 'VirtualDirContext' --include '*.xml' %s", name="tomcat_vdc_targeted")
                                                                                
if __name__ == "__main__":                                                      
    from insights import run                                                    
    run(tomcat_vdc_targeted, print_summary=True)
```

insights/specs.py
```diff
@@ -424,7 +425,23 @@ def block(broker):
...
+    findall = re.compile(r"\-Dcatalina\.(home|base)=(\S+)").findall
+    for p in ps:
+        found = findall(p)
+        if found:
+            results.extend(f[1] for f in found)
+    return list(set(results))
+
+
+tomcat_vdc_targeted = sf.foreach(tomcat_home, "/bin/grep -R -s 'VirtualDirContext' --include '*.xml' %s", name="tomcat_vdc_targeted")
+tomcat_vdc_fallback = sf.simple_command("/bin/grep -R --include '*.xml' 'VirtualDirContext' /usr/share/tomcat*", name="tomcat_vdc_fallback")
```
zhangyi733: In 1.x, `tomcat_vdc_fallback` corresponding command is `"/usr/bin/find /usr/share -maxdepth 1 -name 'tomcat*' -exec /bin/grep -R -s 'VirtualDirContext' --include '*.xml' '{}' +"`, in this PR, it's `/bin/grep -R --include '*.xml' 'VirtualDirContext' /usr/share/tomcat*`, though these commands work the same, do you think it's better to keep the same command format?

skontar: @zhangyi733 Good catch. They are not the same and the one with glob (the old one) does not work.

insights/parsers/tomcat_virtual_dir_context.py
```diff
@@ -1,47 +1,88 @@
...
+                    self.data[file_name] = [file_line]
+
+        if self.data == {}:
+            raise SkipException('VirtualDirContext not used.')
+
+
+@parser('tomcat_vdc_fallback')
+class TomcatVirtualDirContextFallback(TomcatVirtualDirContextBase):
+    """
+    Reports whether there is a VirtualDirContext option used in TomCat configuration file. Looks
```
zhangyi733: In docstring, We'd better use "Tomcat" instead of "TomCat". The same comments to other similar issues.

skontar: @zhangyi733 I will fix the comments later, as I will probably need another changes to the parser / combiner.

PR #684, opened 2017-11-06
add spec, parser, and combiner for tmpfilesd

SteveHNH:
1.x version of PR #653


SteveHNH:
@xiangce the only thing that changed is line 14 in combiners/tmpfilesd.py  because the way the combiner decorator is  used was failing the test when using it the same way I did in master. 
`@combiner(requires=[TmpFilesD])` in 1.x
vs.
`@combiner([TmpFilesD]) in master branch.

I can change master to match 1.x, but if I modify 1.x to match master, it fails.

insights/combiners/tmpfilesd.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Attributes:
+        files(set): the set of files found in all data files.
+        rules(dict): a dictionary of rules using the config file as the key
+        file_paths(list): a list of the file paths for the configurations.
+    """
+    def __init__(self, tmpfiles):
+        rules = {}
+        files = set([])
+        self.tmpfiles = tmpfiles
```
xiangce: I mean, this attribute is removed and a new `active_rules` attribute is added. Would you please to check these details?

insights/combiners/tmpfilesd.py
```diff
@@ -0,0 +1,74 @@
...
+        # Files must be sorted so that /etc/tmpfiles.d overrides /run/tmpfiles.d
+        # which overides /usr/lib/tmpfiles.d. To do this, we have to list file
+        # paths in reverse order.
+        for tmpfile in sorted(tmpfiles, key=lambda f: f.file_path, reverse=True):
+            file_paths.append(tmpfile.file_path)
+            rules[tmpfile.file_path] = []
+            for new_rule in tmpfile.rules:
+                _file = (new_rule['path'])
+                if _file in files:
+                    [rules[tmpfile.file_path].remove(_rule) for _rule in rules[tmpfile.file_path] if _rule['path'] == _file]
```
xiangce: And the `.remove` is also deprecated in the master branch.

insights/combiners/tmpfilesd.py
SteveHNH: @xiangce ah! I'm not even sure how that happened -_-

I'll take care of it. My apologies

xiangce:
@SteveHNH - Thanks for updating. I went ahead and merged it :smile:

PR #686, opened 2017-11-07
Refactor the `get` and `get_after` of LogFileOutput and its sub-classes

xiangce:
- Change the 3rd parameter of `get_after` to keyword parameter
  1. Avoid passing irrelevant lines do not belong to the instance
  2. Support searching keywords while comparing with the time-stamp
  3. The keyword parameter can be either a string or a string list
- `get` and `get_after` will always return a parsed line with at least
  one key: `raw_message`
- `get` also supports searching multiple keywords by passing it a string
  list
- Change the `parse_lines` interface to private `_parse_line` to parse
  each line
- Uniform key's name to `raw_message` for the raw data in each parsed
  return value
- Add `note` docsting to all parsers who used the base class:
  `LogFileOutput`
- Remove useless function from some sub-classes
- Correct some of the tests as per above modification


xiangce:
@PaulWay - I apologize for modified the useful `get_after` function you create a long time ago. 
I changed its 3rd parameter to a `keyword` like parameter,  my main concern is: want to avoid pass irrelevant `lines` to this function. 

And by changing it to a `keyword`, it still can achieve the original goal. E.g. To iterate the lines which contain the string "ERROR in httpd" and is after `2017.11.7`,

The original way is:

~~~
logs = log_obj.get("ERROR in httpd").
for log in log_obj.get_after(datetime(2017,11,7), logs):
   ...
~~~

The current way of this PR is:
~~~
for log in log_obj.get_after(datetime(2017,11,7), "ERROR in httpd"):
   ...
~~~

Please take a look at this PR and help correct it if something is wrong in it when you are free.
Many thanks.

insights/parsers/dmesg.py
```diff
@@ -59,42 +63,57 @@ def has_startswith(self, prefix):
...
+            s(str or list): one or more strings to search for.
                 If not supplied, all available lines are searched.
 
         Yields:
             Log lines with time stamps after the given time.
         """
-        if lines is None:
-            lines = self.lines
+        if not isinstance(timestamp, float):
+            raise ParseException(
```
bfahr: @xiangce should this be a `TypeError` instead since it is called by a rule after the data has been parsed?

insights/parsers/sssd_logs.py
```diff
@@ -42,61 +45,45 @@ class SSSDLog(LogFileOutput):
         (Tue Feb 14 09:45:06 2015) [sssd] [monitor_hup] (0x0020): Received SIGHUP.
         (Tue Feb 14 09:45:07 2015) [sssd] [te_server_hup] (0x0020): Received SIGHUP. Rotating logfiles.
 
+    Each line is parsed into a dictionary with the following keys::
+
+        * **timestamp** - the date of the log line (as a string)
```
bfahr: @xiangce did you mean for this to be a bullet list with formatted terms?  Because of the `::` above the formatting chars are being treated as literals.

insights/parsers/vdsm_log.py
```diff
@@ -25,12 +24,29 @@ class VDSMLog(LogFileOutput):
         Thread-65::DEBUG::2015-05-08 18:01:04,857::blockSD::600::Storage.Misc.excCmd::(getReadDelay) SUCCESS: <err> = '1+0 records in\\n1+0 records out\\n4096 bytes (4.1 kB) copied, 0.000157193 s, 26.1 MB/s\\n'; <rc> = 0
         Thread-4662::DEBUG::2015-05-08 18:01:05,560::task::595::TaskManager.Task::(_updateState) Task=`9a7948f6-b6d9-42c2-b91f-7e0346dfc1d6`::moving from state init -> state preparing
 
-    Usage:
+    Each line is parsed into a dictionary with the following keys::
+
+        * **thread** - the thread or task ID
```
bfahr: Same questions as above, should this be formatted?

insights/parsers/dmesg.py
xiangce: `TypeError` makes more sense, will update it in next commit. Thanks.

insights/parsers/sssd_logs.py
xiangce: Thank you @bfahr , I will uniform these docstrings in next commit.

PaulWay:
This looks fine to me - just make sure the documentation is up to date everywhere :-)

PR #687, opened 2017-11-07
Add manila config parser

pratikgadiya12:


insights/config/specs.py
```diff
@@ -219,6 +219,7 @@
     "lvm.conf"                  : SimpleFileSpec("etc/lvm/lvm.conf"),
     "lvs"                       : NoneGroup([CommandSpec('/sbin/lvs -a -o +lv_tags,devices --config="global{locking_type=0}"')]),
     "lvs_noheadings"            : CommandSpec("/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_name,lv_size,lv_attr,mirror_log,vg_name,devices,region_size,data_percent,metadata_percent --config=\"global{locking_type=0}\""),
+    "manila.conf"               : SimpleFileSpec("etc/cinder/manila.conf"),
```
xiangce: Would you mind to rename the spec to `manila_conf`?

All newly added parsers and the corresponding specs should be added to both the 1.x branch and the master branch (insights v3.x)  to keep them synchronized with each other. In the master branch, a new specs style ( in `insights/specs.py`) will be used in the future, in which all specs are wrapped as (datasource) functions. These special characters, e.g. `.`, are not allowed for function/variable name. It's recommended to replace them to "_" and it would be better to give the spec a unified name for both the `insights/config/specs.py` and `insights/specs.py`.

xiangce:
And I would like to merge them together if you can re-submit it to the master branch. Thanks.

pratikgadiya12:
@xiangce : 
- Renamed manila.conf to manila_conf
- Added a patch to insights 3.x with suggested changes https://github.com/RedHatInsights/insights-core/pull/690

Issue #658, opened 2017-11-01
satellite_version combiner should not return None


csams:
The combiner `insights.combiners.satellite_version` has several conditions under which it might return None. Combiners whose conditions are met should either return a usable value or raise an exception.

xiangce:
I will fix it ASAP.

PaulWay:
Hi @csams - this has come back because I noticed that the rules were getting the Exception being raised by the `satellite_version` parser, and wanted to eliminate this.  I raised #751 to deal with this but @xiangce  has pointed out this issue which I had forgotten.

So my questions are:

* Why is returning None bad?
* What Exception should a combiner raise if it has determined that it is invalid - e.g. in this case, that Satellite is not installed and therefore has no valid version number?  Maybe we should just raise a ParseException here?
* I don't see anything in http://insights-core.readthedocs.io/en/latest/exception_model.html about this, and http://insights-core.readthedocs.io/en/latest/combiner_tutorial.html is blank; is there some documentation about what combiners should and should not return?

PaulWay:
Hi @csams - @xiangce has filled me in on some of the discussion.  So let's change my questions to:

* I'm assuming we should raise a SkipException if it can't end up in a valid state and wants to be skipped.  Is that correct?
* Can I suggest that the documentation of SkipException be modified to include this kind of use case?  See http://insights-core.readthedocs.io/en/latest/api_index.html#insights.parsers.SkipException and http://insights-core.readthedocs.io/en/latest/exception_model.html#case-3b-parsing-data-to-find-errors-dirty-parser

PaulWay:
Hi @csams 

Just a further note - I've tried raising a SkipException (see #751) and the SkipException gets caught in the rule tests.  I can forward you more information about this but I thought the SkipException was going to be absorbed by the parsing system.  Is this feature present in version 1.x?

Thanks in advance,

Paul

csams:
Hey @PaulWay, `SkipComponent` and all of its subclasses (`SkipException` for now for backward compatibility) should be ignored by the core in 1.x and 3.x. #755 included the changes for 1.x and was recently merged. I'll update the docs to reflect the use case.

csams:
Fixes in #755 #751. Combiner in master and 1.x branches are updated.

PR #692, opened 2017-11-10
Refactor satellite_version combiner to class

xiangce:


xiangce:
For #658

xiangce:
Hi @csams, different with the #693, in the master branch, I refactored the original `satellite_version` function combiner to a class combiner.  How do you think?

insights/combiners/satellite_version.py
```diff
@@ -101,43 +42,115 @@ def _parse_sat_versoin(version):
...
-                                    sat5_pkg.release, major, minor)
+    def __init__(self, sat6_ver, rpms):
+        self.full = None
+        self.version = None
+        self.release = None
+        self.major = None
+        self.minor = None
+
+        # For Satellite 6.1.x, if satellite_version/version.rb is available:
+        if sat6_ver and sat6_ver.version:
```
csams: If this condition and the ones below it within "if rpms:" are all false (i.e., this isn't a satellite machine), then the combiner will not be correctly initialized in almost every case since InstalledRpms will almost always be available to satisfy one of its dependencies. If the combiner doesn't initialize correctly, it should raise `SkipComponent` (from `insights.core.dr`), or we should move `SkipException` up out of `insights/parsers/__init__.py` to a higher level and have it raise that.

xiangce: In fact, I have once tried to raise an `Exception` for the case it's not a satellite machine (just like what 1.x did). However, just like you said, the `InstalledRpms` will almost always be available. Then for most situations which are not satellite machine, the `Exception` will be raised (it seems like something is wrong, but it's just not a satellite machine). So I left (initialized) all attributes as `None` when it's not a satellite machine instead of raising an `Exception`.  (And I'm afraid that these `Exception` might be treated as a kind of `Parser Error` and will cause maintainer's unnecessary attention). Does that make sense to you?

csams: Hey @xiangce, `insights.core.dr.SkipComponent` is the general `Exception` class in master that should be raise if a component wants to take itself out of dependency resolution and not have it counted as something going wrong. If a component raises `SkipComponent` or a subclass of it (like `insights.parsers.SkipException in master), the exception is ignored.

I think we might need to create something like `SkipComponent` in the 1.x branch and then have the `satellite_version` combiner raise it in both master and 1.x. I think proper handling of `insights.parsers.SkipException` needs to be added to 1.x anyway.

Let's hold this PR for now, and I'll chat with Kyle about some changes to 1.x core that it can use.

xiangce: @csams - Thanks for your explanation, that sounds good. I will modify this PR at first.

bfahr:
@csams is working on a solution for #720 so we will wait until it is resolved before considering this PR.

xiangce:
@bfahr - Thanks. I'll re-submit this one to the 1.x branch after the #720 is solved.

xiangce:
See #751

PR #693, opened 2017-11-10
Raise Exception instead of return None for satellite_version

xiangce:


xiangce:
For #658

PR #713, opened 2017-11-13
Unpack list of parsers in the TomcatVirtualDirContextCombined combiner

skontar:
TomcatVirtualDirContextTargeted returns list of parsers instead of
a parser, because spec command is different.


skontar:
3.x https://github.com/RedHatInsights/insights-core/pull/714

PR #725, opened 2017-11-17
Netstat parser minor improvements for master branch

PaulWay:
This merge roughly parallels #625 but since the master branch already has a '`rows_by`' method there's no need for the '`search`' method.

Or is there?  Should we replace the `rows_by` function, which doesn't take keyword arguments, with the 1.x version of `search`, which does?

@bfahr @xiangce Opinions please, I'll call this a work in progress until we decide how to resolve this conflict.


insights/parsers/netstat.py
```diff
@@ -182,28 +177,31 @@ def parse_content(self, content):
...
                                 break
-                        else:
-                            # Line contained no number, ignore
-                            continue
-                        data.remove(val)
-                        key = "_".join([k.lower() for k in data])
-                        first_layer[key] = val
+                        if found_val is not None:
+                            key = '_'.join(k.lower() for k in parts if k != found_val)
+                            first_layer[key] = found_val
```
xiangce: Would you raise a "ParseException" here for `found_val` is None?

PaulWay: According to http://insights-core.readthedocs.io/en/latest/exception_model.html#parsing-and-exceptions we should only raise a ParseException if the entire output is unusable.  This change specifically removes the previous ParseException, which was counter to that policy.

PaulWay:
OK, I've replaced the old `rows_by` method with the `search` method identical to that used in the 1.x code, so there should be no feature disparity.  @xiangce - should be ready to go now!

insights/parsers/netstat.py
xiangce: Okay, makes sense to me.

PR #726, opened 2017-11-17
Add 'deprecated' function to log use of deprecated parsers for master branch

PaulWay:
This is the master branch version of #361.

However, you'll note that the tests are failing.  This is because some of the libraries we depend on are using deprecated functionality.  Raising a DeprecationWarning is now considered a fatal error, because for us the 'DeprecatedWarning' means 'you should have already corrected your code to not use this'.

I see two solutions:

1) Work out which libraries we're using that have deprecated functionality, and use updated ones.
2) Make the DeprecationWarning a non-fatal error.

Any preferences?


xiangce:
@PaulWay -
FYI: https://github.com/xiangce/insights-core/pull/1
Maybe you want to backport the `DeprecationWarning` from python2.7 to `insights-core` explicitly to avoid this test issue in python2.6. 

Just like what Kyle did on `ConfigParser`: https://github.com/RedHatInsights/insights-core/pull/325/commits/c15fea1

PaulWay:
@xiangce I think what you mean is to copy the `cryptography` library into `contrib`, since the `warnings` library that provides `DeprecationWarning` is already there.

As far as I can see even if we did that we'd still have to edit the code so it would not warn about deprecation, and I think we really want to avoid going down that path.

One option that we may have to consider is to have a separate fork of the core code that is specifically for Python 2.6.  @jhjaggars @kylape @csams Is that a better solution to this problem of libraries not being backward compatible with old version of Python?

PaulWay:
@xiangce I think what you mean is to copy the `cryptography` library into `contrib`, since the `warnings` library that provides `DeprecationWarning` is already there.

As far as I can see even if we did that we'd still have to edit the code so it would not warn about deprecation, and I think we really want to avoid going down that path.

One option that we may have to consider is to have a separate fork of the core code that is specifically for Python 2.6.  @jhjaggars @kylape @csams Is that a better solution to this problem of libraries not being backward compatible with old version of Python?

kylape:
Creating a py2.6 fork sounds like a lot of maintenance effort.  We need to make one codebase support all target platforms.  `exceptions.DeprecationWarning` looks to be in by py2.6 and py2.7.  

I think my vote is for making `DeprecationWarning` a non-fatal error.  It would be great if deprecation warnings were still visible when running the tests, though.

PaulWay:
@kylape The problem is that we agreed in the initial discussion that making the DeprecationWarning a fatal error was necessary to prevent people just ignoring the warnings and continuing to use a deprecated parser.  A fatal error prevents people merging code that includes a deprecated feature.  That's why we use the `PendingDeprecationWarning` when initially deprecating something as a non-fatal but verbose reminder.  However, this conflicts with the general usage, which seems to be that the DeprecationWarning is the non-fatal but verbose reminder and removal of the deprecated feature causes the fatal error in the dependent code.

I think it's probably better in the long run to conform to community expectations and use of the DeprecationWarning, which is something I guess we should have been looking at when we agreed on this approach.

I'm happy to rework this so there's no 'pending' state, that DeprecationWarning is issued as a non-fatal verbose warning when something shouldn't be used.  I just want to make sure we're all on the same page about how it's supposed to work.

xiangce:
@PaulWay - And I think we'd better keep the 1.x and 3.x in sync, would you please cherry-pick the last commit to the 1.x branch as well?

PaulWay:
@xiangce Sure thing, it's in #819.  I also updated the documentation in both branches.

PR #732, opened 2017-11-20
Marked cpuinfo_max_freq as deprecated, removed tests

PaulWay:


insights/parsers/tests/test_cpuinfo_max_freq.py
```diff
@@ -1,9 +0,0 @@
-from insights.tests import context_wrap
-from insights.parsers.cpuinfo_max_freq import cpu_max_freq
-
-cpuinfo_content = '3200000'
-
-
-def test_cpu_max_freq():
-    result = cpu_max_freq(context_wrap(cpuinfo_content))
-    assert result == 3200000
```
bfahr: @PaulWay whats the motivation for removing the tests before the function is actually removed?  Seems like you would leave the tests and remove all at once, just in case some other code change breaks this function.

PaulWay: The tests trigger a deprecation warning, causing the tests to fail.  This was my lazy way of avoiding fixing the tests to catch and detect the deprecation warning :-)

xiangce:
@PaulWay - Since #819 is merged first, this PR should update correspondingly.  Sorry for that.

PaulWay:
@xiangce This should be good to go after #819

PR #733, opened 2017-11-20
Marked OracleConfig as deprecated, removed tests

PaulWay:


bfahr:
@PaulWay I recommend leaving tests in place until function is removed.

insights/parsers/oracfg.py
```diff
@@ -71,6 +72,9 @@ class OracleConfig(Parser):
             string
 
     """
+    def __init__(cls, *args, **kwargs):
```
bfahr: I expected to see `self` here instead of `cls`.

xiangce: Hi @PaulWay - How do you think about the comments that Bob commented above?

PaulWay:

xiangce:
@PaulWay - I have to revert this PR. Since I have merged the #819 already, the test breaks.
Please update this parse and re-submit it in a new PR. Sorry for that.

PaulWay:
Hi @xiangce OK, try #823 instead.

PR #734, opened 2017-11-21
Marked LimitsConf and NprocConf as deprecated, use limits_conf instead

PaulWay:


bfahr:
@PaulWay I recommend leaving tests in place until the deprecated functionality is removed.

insights/parsers/nproc.py
```diff
@@ -46,12 +47,20 @@ def _parse_content(content):
 @parser("limits.conf")
 class LimitsConf(Parser):
     """Parse ``limits.conf`` file contents."""
+    def __init__(cls, *args, **kwargs):
```
bfahr: I would expect to see `self` here instead of `cls`.

insights/parsers/nproc.py
```diff
@@ -46,12 +47,20 @@ def _parse_content(content):
...
+        super(LimitsConf, cls).__init__(*args, **kwargs)
+
     def parse_content(self, content):
         self.data = _parse_content(content)
 
 
 @parser("nproc.conf")
 class NprocConf(Parser):
     """Parse ``*-nproc.conf`` file contents."""
+    def __init__(cls, *args, **kwargs):
```
bfahr: Same comment as above regarding use of `cls`.

xiangce: @PaulWay - How do you think about the comments that Bob commented?

PaulWay:

xiangce:
@PaulWay - Since #819 is merged first, this PR should update correspondingly.

PR #735, opened 2017-11-21
Deprecated get_limits function, use LimitsConf class instead

PaulWay:
This function is already removed in the master branch, so it is marked as deprecated in the 1.x branch.


xiangce:
No 3.x patch is needed

PR #737, opened 2017-11-21
Deprecating other Sysconfig parsers in favour of those in sysconfig module

PaulWay:
These classes have been moved into the sysconfig module in the master branch, so it is marked as deprecated in the 1.x branch


PaulWay:
Hmmm - I left the tests in.  Should we be removing the tests?

insights/parsers/docker_sysconfig.py
```diff
@@ -50,7 +51,9 @@ class DockerSysconfig(SysconfigOptions):
 
     Parse the ``/etc/sysconfig/docker`` file.
     """
-    pass
+    def __init__(cls, *args, **kwargs):
```
bfahr: I would expect to see `self` instead of `cls` as the first argument here.

bfahr: There is a deprecated function in this module, would it make sense to also add the deprecation functionality to it as well since you are modifying this module?

insights/parsers/irqbalance_conf.py
```diff
@@ -46,5 +47,9 @@ class SysconfigIrqbalance(LegacyItemAccess, Parser):
         False
 
     """
+    def __init__(cls, *args, **kwargs):
```
bfahr: Same comment as above regarding use of `cls` here.

insights/parsers/kdump.py
```diff
@@ -285,6 +286,9 @@ class SysconfigKdump(SysconfigOptions):
     These are set to the value of the named variable in the kdump sysconfig
     file, or '' if not found.
     """
+    def __init__(cls, *args, **kwargs):
```
bfahr: Same comment as above regarding use of `cls`.

xiangce: @PaulWay - Hi, How do you think about the comments Bob commented? And BTW, the shippable test failed, would you mind to fix it at first?

PaulWay:

xiangce:
@PaulWay - Since #819 is merged first, this PR should update correspondingly.

PR #739, opened 2017-11-22
Python bindings to libmagic now come from python-magic RPM

kylape:
This enables us to depend on the `python-magic` RPM in RHEL.


kylape:
CI failed because the python-magic RPM is not installed on the Shippable or Travis builder image.  I'm going to throw some commits on here to see if I can switch CI builder images easily enough.

PR #742, opened 2017-11-22
Adding 'get_param' and 'get_last' methods to NTP_conf for 1.x branch

PaulWay:
Rules need to look up the last given value of a keyword by parameter, or all the values.  These operations are now provided in the `get_param` and `get_last` methods respectively.

Documentation and test coverage are also provided.


insights/parsers/system_time.py
```diff
@@ -103,7 +103,30 @@ class NTP_conf(NTPConfParser):
 
     Uses the ``NTPConfParser`` class defined in this module.
     """
-    pass
+    def get_tinker(self, param):
```
bfahr: What about just adding a second arg so that it could be called as `get_last('tinker', 'step')`?  That would seem to make it more generally useful for all parameters here.

PaulWay: OK, I've redone this as a `get_param` method and renamed my branch.  I think I've pushed it to this merge request - let's see...

PaulWay:
Hi @bfahr - I've generalised this into the `get_param` method - let me know what you think.

insights/parsers/tests/test_system_time.py
```diff
@@ -147,18 +146,51 @@
...
 """
 
+NTP_TINKER_CONF = """
+driftfile /var/lib/ntp/ntp.drift
+restrict default kod nomodify notrap nopeer noquery
+restrict 127.0.0.1
+server 192.168.17.62 iburst
+tinker step 0.9
+tinker step step
+tinker step 0.4
```
lonicerae: ntp seems to use bison for token scanning. Given the following example:

```
tinker step 0.9
tinker step step
tinker step 0.4
tinker step step
```

ntp can still parse the parameter well:

```
yylex: lexeme 'tinker' -> T_Tinker
yylex: lexeme 'step' -> T_Step
yylex: lexeme '0.9' -> T_Double
        <end of command>
yylex: lexeme 'tinker' -> T_Tinker
yylex: lexeme 'step' -> T_Step
yylex: lexeme 'step' -> T_Step
        <end of command>
yylex: lexeme 'tinker' -> T_Tinker
yylex: lexeme 'step' -> T_Step
yylex: lexeme '0.4' -> T_Double
        <end of command>
yylex: lexeme 'tinker' -> T_Tinker
yylex: lexeme 'step' -> T_Step
yylex: lexeme 'step' -> T_Step
        <end of command>
        <end of command>
Finished Parsing!!
```

while we get `ntp_obj.get_param('tinker', 'step') ==> 'step'` in current code.

PaulWay: @lonicerae I agree that `ntpd` can probably parse the parameter and work out that it needs to get '0.4' as a double precision floating point number there.  However, I don't think we should encoding the entire token scanning rules of `ntpd` into this parser, so the parser doesn't know that 'step' is not a valid value.  If the user has included invalid declarations, `ntpd` seems to silently ignore them, but the parser can't determine which of these is invalid.  The `get_param` method is simply a 'best effort' attempt at getting the right thing, which will work in 99% of the cases.

Maybe we should have a rule that checks for duplicate definitions of keywords we care about :-)

xiangce: Hey @PaulWay, as per @lonicerae 's example, I think the `get_last` name that @bfahr has mentioned makes more sense to me.
~~~
ntp_obj.get_last('tinker', 'step') ==> 'step'
~~~

For `get_param`, how about returning a list likes:
~~~
`ntp_obj.get_param('tinker', 'step') == ['0.9', 'step', '0.4', 'step']
~~~

At least we do not provide a value which is not incorrect or indicates any incorrection to the developer.

Just my 2 cents :)

xiangce: @lonicerae  @PaulWay - How do you think about implementing them in different 2 APIs `get_last and get_param` mentioned above?

lonicerae: Hey @xiangce , yep we can create two different APIs `get_last` and `get_param`. For the tinker command in ntpd config, it can accept multiple options, see here:

 - https://www.eecis.udel.edu/~mills/ntp/html/miscopt.html#tinker

If the `get_last` method can provide the last value for specific option, that would be nice. And maybe it's more useful to use `get_tinker` instead of `get_param` to get all arguments passed to tinker command. For example:

```
tinker dispersion 25
tinker panic step 0.4
```

we can get `['dispersion', '25'], ['panic', 'step', '0.4']` from `get_tinker`.

PaulWay: I'm working on `get_last` and `get_param` as @xiangce suggested - just got to get it all tested.  I'd prefer this over a `get_tinker` solution.  We can already get all the definitions of any keyword from the `data` dictionary anyway, so `get_param` and `get_last` are mainly there for getting specific parameters of one keyword.

lonicerae: >  We can already get all the definitions of any keyword from the data dictionary anyway

Yep, we have already had the ability to return all arguments for specific command, that's good.

insights/parsers/system_time.py
```diff
@@ -98,44 +101,74 @@ def parse_content(self, content):
...
         Returns:
-            str or None: The value of the given parameter, or None if not
-            found.
+            list: All the values of the given parameter, or an empty list if
+            not found.
         """
         if not keyword or keyword not in self.data:
-            return default
-        # keyword in data - may be None or list
+            print "no keyword or keyword not found - returning:", [default]
```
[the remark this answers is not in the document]
PaulWay: Dangit!  I forgot to use `noprints`!  Fixed now, thanks @xiangce

xiangce:
Approved pending resubmission to the master branch.

PR #744, opened 2017-11-23
Ps parsers update for 1.x branch

PaulWay:
There are four different parsers for ``ps`` output, depending on the flags used to invoke it.  All of them collect slightly different stuff.  Some are no longer in use.  Some collect insufficient data (e.g. a truncated command line). Some contain useless repeated flags ('ww' gets extra wide output, 'www' does not do anything more).

So of course the solution to this is to create a fifth parser that does everything that the other four do, and more.  The aim is to replace all use of the other four parsers in all rule code, and then deprecate the use of the previous parsers.

The new parser is ``PsAuxww``.  It collects all the statistics that ``PsAux``, ``PsAuxcww`` and ``PsAuxwww`` do, and gets full command lines (unlike ``PsAux`` and ``PsAuxcww``).  It does not get the full start time of the process (which ``PsAxcwwo`` does), but there are no recorded uses of ``PsAxcwwo`` so removing it just makes the code simpler.

In addition, ``PsAuxww`` implements a ``search`` method that allows keyword-based searching on any of the columns in the output in combination.  This, combined with the ``__contains`` keyword search functionality, makes it easy to perform sophisticated searches on the process list.

The other four parsers are marked as pending deprecation but should still work.

A separate branch has been tested of the internal repository converting all the existing rules to use the ``PsAuxww`` parser.


insights/parsers/ps.py
```diff
@@ -104,22 +105,76 @@ def users(self, proc):
...
-        """boolean: Returns ``True`` if ``proc`` is in the COMMAND column."""
-        return proc in "".join(self.running)
+        """boolean: Returns ``True`` if the word ``proc`` appears in the COMMAND column."""
+        return any(proc in row['COMMAND'] for row in self.data)
 
     def __contains__(self, proc):
-        return proc in self.running
+        # We should be able to take this if out when we only have the PsAuxww
+        # parser:
+        return any(proc == row['COMMAND'] for row in self.data if 'COMMAND' in row)
```
bfahr: Is this change necessary?

insights/parsers/ps.py
```diff
@@ -177,6 +234,9 @@ class PsAux(ProcessList):
             headers and each item in the list represents a process. The command
             and its args (if any) are kept together in the COMMAND key.
     """
+    def __init__(cls, *args, **kwargs):
```
bfahr: To be consistent this should be `self` instead of `cls`.

insights/parsers/ps.py
```diff
@@ -193,7 +253,9 @@ class PsAuxwww(PsAux):
         data (list):  List of dicts, where the keys in each dict are the column
             headers and each item in the list represents a process.
     """
-    pass
+    def __init__(cls, *args, **kwargs):
```
bfahr: Same comment above regarding `cls`.

insights/parsers/ps.py
```diff
@@ -209,6 +271,9 @@ class PsAxcwwo(ProcessList):
     Raises:
         ParseException: Raised if any error occurs parsing the content.
     """
+    def __init__(cls, *args, **kwargs):
```
bfahr: Same comment as above regarding `cls`.

insights/parsers/ps.py
```diff
@@ -104,22 +105,76 @@ def users(self, proc):
...
         for row in self.data:
             yield row
 
 
+@parser('ps_auxww')
+class PsAuxww(ProcessList):
+    """
+    Class to parse ``ps auxww`` command output.
+
+    Examples:
```
bfahr: Do you need to add the examples?

insights/parsers/ps.py
PaulWay: Yeah, probably not - sorry, got a bit carried away.

insights/parsers/ps.py
PaulWay: The examples are in the headings, so I've removed this.

insights/parsers/ps.py
PaulWay: OK, agreed - I've tidied this up.

PaulWay:
OK, @bfahr, let me know what you think.

PR #746, opened 2017-11-23
Add parser for audit_log

JoySnow:


insights/parsers/audit_log.py
```diff
@@ -0,0 +1,164 @@
...
+            s(str or list): one or more strings to search for.
+                If not supplied, all available lines are searched.
+
+        Yields:
+            (dict): the parsed data of lines with timestamps after this date in the
+            same format they were supplied.
+        """
+        for line in self.lines:
+            # If `s` is not None, keywords must be found in the line
+            if (s and ((type(s) == str and s not in line) or
```
PaulWay: 1) I believe it's best practice to compare types with 'isinstance' rather than directly.
2) I'd recommend for speed to set a comparison lambda function outside the loop and then use it inside the loop, e.g.:
```python
if not s:
    return []
if isinstance(s, str):
    comp=lambda line: s in line
elif isinstance(s, list):
    comp=lambda line: any(w in line for w in s)
else:
    raise TypeError('search items must be given as list or string')
for line in self.lines:
    if not comp(line):
        continue
    ...
```

PaulWay: In general it might be a good idea to think about how we can abstract this 'search from list or string' behaviour to all 'get_after' methods.

insights/parsers/audit_log.py
```diff
@@ -0,0 +1,164 @@
...
+                             else line)
+                unparsed2 = unparsed1.split(linesp[1])[-1]
+                info["unparsed"] = unparsed2.strip()
+                break
+            info[itemsp[0]] = itemsp[1]
+
+        info['raw_message'] = line
+        return info
+
+    def get(self, s):
```
xiangce: Then this `get` can be removed. Since once a line contains the keyword `s`, it's OK to return the `raw_message` even it cannot be parsed.

insights/parsers/audit_log.py
```diff
@@ -0,0 +1,164 @@
...
+                not (linesp[0] and linesp[0].startswith('type=')) or
+                not (linesp[1] and linesp[1].startswith('msg=audit('))):
+            return
+
+        info['type'] = linesp[0].lstrip('type=')
+
+        timestamp_id = linesp[1].lstrip('msg=audit(').rstrip('):')
+        timestamp_id_sp = timestamp_id.split(':')
+        if len(timestamp_id_sp) != 2:
+            return
```
xiangce: Better return the `raw_message`, no matter it can be parsed or not.

insights/parsers/audit_log.py
JoySnow: If you mean the 'search for list of strings or string' behaviour, this is already apllied to class `LogFileOutput`  in https://github.com/RedHatInsights/insights-core/pull/686 . 

And I'd like to take the 2 suggestions above. Thanks.

insights/parsers/audit_log.py
JoySnow: Add a flag named `is_valid` and return `{'raw_message': line, 'is_valid': False} ` if unparseable.
```

File: insights/parsers/audit_log.py [hunk 4]
```diff
@@ -0,0 +1,144 @@
...
+            * **type** - the log type
+            * **timestamp** - the timestamp of the log line (as a string)
+            * **msg_ID** - the serial number behind the timestamp
+            * **unparsed** - the unparsed part in log, may or maynot exist
+
+            If the line is not regular, for some reason, then as many fields as
+            possible are pulled from the line.
+        """
+        info = {}
+        info_basic = {'raw_message': line,
```
xiangce: Better use one `info` only (need to re-order below code logic)

insights/parsers/audit_log.py
```diff
@@ -0,0 +1,144 @@
...
+        Yields:
+            (dict): the parsed data of lines with timestamps after this date in the
+            same format they were supplied.
+        """
+        if isinstance(s, str):
+            invalid_search_result = lambda l: s not in l  # NOQA
+        elif isinstance(s, list):
+            invalid_search_result = lambda l: any(w not in l for w in s)  # NOQA
+        else:
+            invalid_search_result = lambda l: True  # NOQA
```
xiangce: As we discussed, 
- better raise a "TypeError" exception when its type is not `str` or `list`.
- better check the items in the `list`, if there are any non-string items, raise a "TypeError" exception.

PR #750, opened 2017-11-23
Remove --type option from docker inspect specs items

wushiqinlou:
Remove --type option from docker inspect specs items due to version before 1.8 has no such option


xiangce:
See #749

Issue #720, opened 2017-11-15
SkipComponent/SkipException Handling and Extension to Combiners


csams:
Insights Core 1.x has a `SkipException` class in `insights.parsers` that's supposed to allow a parser to signal that nothing is wrong but that it shouldn't be used in dependency resolution. However, the exception currently isn't handled any differently from other exceptions.  Also, it's scoped to parsers, which makes it awkward for [combiners](https://github.com/RedHatInsights/insights-core/pull/692) and other components to use.

Core 1.x needs a generic `SkipComponent` exception that mirrors `insights.core.dr.SkipComponent` in master. `insights.parsers.SkipException` should subclass it, and any instance of `SkipComponent` should be ignored regardless of the type of component that raises it.

csams:
Fixed in #751 and #755.

PR #751, opened 2017-11-24
Raise SkipComponent if unable to determine Satellite version

PaulWay:
Reading https://github.com/RedHatInsights/insights-core/issues/658 and https://github.com/RedHatInsights/insights-core/pull/692, according to @csams this should be returning a SkipException.

The lingering problem is that this SkipException still makes its way through to the rule, which it shouldn't according to the documentation.

Also fixing a minor spelling error.


xiangce:
Hey @PaulWay - Thanks for fixing the typos. I have raised these Exceptions in #693 because of #658. And just like @csams has clarified the `"Exception"` should be raised in this situation. #720 was opened to solve this problem (To skip these Exceptions)

csams:
Hey folks, I have a [PR](755) open for #720 that affects issues like this. Combiners should raise `insights.SkipComponent`.  `insights.parsers.SkipException` was needlessly specific to parsers, so it's been moved up to `insights.core`, made a subclass of `insights.SkipComponent`, and re-exported in the `parsers` package for backward compatibility.

insights/combiners/satellite_version.py
```diff
@@ -65,6 +65,7 @@
 from insights.core.plugins import combiner
 from insights.parsers.satellite_version import Satellite6Version as Sat6Ver
 from insights.parsers.installed_rpms import InstalledRpms
+from insights.parsers import SkipException
```
csams: Instead of using parser exceptions in combiners, let's move toward using `insights.SkipComponent`. I have a [PR](#755) open to support it.

PR #752, opened 2017-11-24
SSH parser has a new method for detecting the "+" syntax

jsvob:
- "+" for adding option values.


insights/parsers/ssh.py
```diff
@@ -176,6 +179,52 @@ def get_line(self, keyword, default=''):
...
+          options manually.
+
+        Parameters:
+            keyword(str): Keyword to find
+        """
+        lines = self.get(keyword)
+        if lines:
+            found_line = lines[-1].line
+            match = PLUS_PATTERN.match(found_line)
+            if match:
```
PaulWay: `if thing return True else return False` should always IMO be just `return thing`.

Personally, I'd put named groups in the regex and return the group dictionary (via `match.groupdict()`).  But if for some reason you want to make it harder for your users, then just `return bool(match)` or `return match is not None`.

insights/parsers/ssh.py
```diff
@@ -176,6 +179,52 @@ def get_line(self, keyword, default=''):
                 keyword=keyword, value=default
             )
 
+    def does_line_use_plus(self, keyword):
```
PaulWay: Since this is applies to keywords, and not just to lines, I would recommend this be called `plus_in_values_of` or `keyword_contains_plus`.

insights/parsers/ssh.py
```diff
@@ -53,6 +53,10 @@
 from collections import namedtuple
 from .. import Parser, parser, get_active_lines
 from insights.specs import sshd_config
+import re
+
+# optional whitespace, at least one non-whitespace (the keyword), at least one whitespace (space), a plus literal, anything
+PLUS_PATTERN = re.compile(r'^\s*\S+\s+\+.*$')
```
PaulWay: See below on why I think you should use named groups to provide the actual data from this value.

insights/parsers/ssh.py
skontar: @PaulWay I agree with what you have said. Just a small note, we usually do not have "your users" for our parsers. We usually write them for our rules only, so whatever difficulty Jakub will make in this parser method quite probably will only he suffer 😄.

insights/parsers/ssh.py
kylape: I prefer `keyword_contains_plus` of the two suggestions, and I think `does_keyword_use_plus` is also pretty good.

PaulWay:
@skontar I'd really recommend thinking about 'your users' being people other than just the security team within Red Hat.  This is an open source project that we want to encourage people from across the globe to use.  Therefore, IMO parsers should be written to think about the general case, not the specific, and should consider use cases outside their own author's plans for rules.

Insights started off with rules defining their own parsers, but that led to many parsers duplicating features and reading the same file many times and in general doing more work than they needed to.  So we moved to shared parsers, so that one parser could read the e.g. `/var/log/messages` file and then all rules based on that content could have access to it more easily.  So we've tried to move away from parsers that are specific to one use case and toward general parsers that collect information that can be useful to a wide variety of possible rules.

I see this change as a case in point.  Yes, you or Jakub might have one specific use case in mind - but many options in the sshd configuration can use the '+' notation to change the way they're interpreted.  We want to avoid other rule writers having to write more code to parse other options with '+' in them, or subsequent revisions to the SSHDConfig parser to add this feature to other options.  We always have to balance code complexity and performance against features and ease of use.  This seems to have been done pretty well here.

skontar:
@PaulWay I pretty much understand the concept - for us it is just very hard to apply all the time and usually very ineffective. I have a whole explanation about our (necessary) point of view. Because is coming up often, I think I will write it down and we can discuss it elsewhere. Thanks!

insights/parsers/ssh.py
jsvob: The "+" symbol with the meaning denoted by the comments can occur only in one place in an sshd.conf line. It really applies to the whole line at once and not to any individual value on its own. Of your suggestions, `keyword_contains_plus` is the most near being correct, but again, it's not the *keyword* that *contains* a "+", it's the *line* that contains it at a *very specific* place. So if you deem the current name `does_line_use_plus` unfit, I can change it either to `line_contains_plus` or `line_uses_plus` or `keyword_uses_plus` - please pick your choice. Please see the comments and the OpenSSH documentation for information how the "+" works.

insights/parsers/ssh.py
jsvob: @PaulWay As to your first sentence, you are right. I wanted to put the emphasis that I want to return exactly `True` or `False` no matter how regex is implemented in any version of Python that the framework happens to run on. The correct way here would then be to `return bool(match)` if I understand you correctly. I will change it that way.

insights/parsers/ssh.py
jsvob: I read the comments and I don't get it. I *think* that you mean that leaving the parsing of the individual values and stripping the leading "+" from the beginning of the first parsed value should be done by the parser rather than by the rule. In this interpretation then, modeling the line parsing through a regex with named groups would duplicate most of the parser's whole functionality. So, again continuing with that interpretation, I think it would make better sense to add a method that does the stripping of "+" or to extend the functionality of `get_values()`. However, I see `get_values()` as useless in my rules, since if I want it to split the values, it insists on joining all the lines with the given keyword - I rather use `last()` and split it myself, and that might be the source of our misunderstanding. Until `get_values()` behaves in a more fitting way, I'm not including the "+" parsing functionality. IIRC the testing I did back when I was beginning with my ssh rules, sshd, at least for some options, uses only the last line with the given keyword, which is why I do that in my rules. 

I understand your point about shared parsers preventing the same system file being read over and over again. It seems to me, however, that this parser tries to provide more functionality than what is necessary, and it complicates maintainability.

If I forewent with this MR, I could have done the "+" parsing solely in my rule and the parser would have been less correct and we wouldn't have this discussion. I agree with maintaining parsers well, but there's a business-practical balance to be met in my opinion.

jsvob:
@xiangce If I understand the code correctly, the options you specified instruct the `get_values()` method to really join the lines. I don't see it as very useful, but it behaves correctly according to the `get_values()` docstring IMO. Somewhat related - https://github.com/RedHatInsights/insights-core/pull/752/files#r154030401

PR #756, opened 2017-11-28
Modify fc-match command to make it work only on RHEL7

zhangyi733:


insights/parsers/fc_match.py
```diff
@@ -1,6 +1,6 @@
 """
-FCMatch - command ``/usr/bin/fc-match -sv 'sans:regular:roman' family fontformat``
-==================================================================================
+FCMatch - command ``/bin/uname -r | grep -F '3.10' > /dev/null && /bin/fc-match -sv 'sans:regular:roman' family fontformat``
```
xiangce: I don't think it's good to put the condition here. The title is too long...

PR #757, opened 2017-11-28
Modify fc-match command to make it work only on RHEL7

zhangyi733:


zhangyi733:
See https://github.com/RedHatInsights/insights-core/pull/756

PR #760, opened 2017-11-28
Add parser catalina_server_log to parser default catalina log file

zhangyi733:
Signed-off-by: Yi Zhang yizha@redhat.com


insights/config/specs.py
```diff
@@ -388,9 +388,11 @@
     "systemid"                  : First([SimpleFileSpec("etc/sysconfig/rhn/systemid"),
                                     SimpleFileSpec("conf/rhn/sysconfig/rhn/systemid")]),
     "teamdctl_state_dump"       : CommandSpec("/usr/bin/teamdctl {iface} state dump", iface=r"team\S+"),
-    "tmpfilesd"               : All([PatternSpec("etc/tmpfiles.d/.*\.conf"),
+    "tmpfilesd"                 : All([PatternSpec("etc/tmpfiles.d/.*\.conf"),
                                     PatternSpec(r"usr/lib/tmpfiles.d/.*\.conf"),
                                     PatternSpec(r"run/tmpfiles.d/.*\.conf")]),
+    "tomcat_server_log"         : All([PatternSpec("usr/share/tomcat/logs/catalina.*\.log"),
```
xiangce: Since the new `tomcat_server_log` will cover the old `catalina.out` (correct me if it's wrong plz), I suggest removing the old one in the master branch and mark it as deprecated (see #733) in the 1.x branch.

insights/parsers/tomcat_server_log.py
```diff
@@ -0,0 +1,41 @@
...
+
+from .. import LogFileOutput, parser
+from insights.specs import tomcat_server_log
+
+
+@parser(tomcat_server_log)
+class TomcatServerLog(LogFileOutput):
+    """
+    Read the default tomcat server log file.
+    """
```
xiangce: The default `time_format` of the base class `LogOutFile` is `'%Y-%m-%d %H:%M:%S'`, You should define it for TomcatServerLog since it's different with the default one.

zhangyi733:
In this commit, there are quite a few modifications.
1. Rename `tomcat_server_log` to `catalina_server_log` to consistent the name convention with existing parser `catalina.out`.
2. Add more PatternSpecs to  `catalina_server_log` to meet Satellite requirements.
3. Enhance existing spec `catalina.out` to make the corresponding command work well in current insights-client.

The reason why we keep both `catalina.out` and `catalina*.log` is that sometimes, the content in `catalina.out` is not totally included in catalina.log. To make sure we can really get an error message, it's better to check both these files.

PR #763, opened 2017-11-29
Revert "Modify fc-match command to make it work only on RHEL7"

xiangce:
- Since isn't evaluated by bash, the command will always fail.
Reverts RedHatInsights/insights-core#757


xiangce:
See #765

PR #764, opened 2017-11-29
No bashisms in command for 1.x branch

PaulWay:
Commands given to CommandSpec are run directly, rather than being given to `bash` or similar shells to execute.  Therefore, we should make sure that any inadvertent use of shell commands such as command chaining ('||' and '&&') and redirection ('|', '<' and '>') should raise an error.


PaulWay:
Once the problems with the current bashisms in commands are fixed, the tests should pass.

xiangce:
@PaulWay - There is a false positive on `lvs_noheadings`(and other `*_noheadings`)

```
lvs_noheadings : CommandSpec("/sbin/lvs --nameprefixes --noheadings --separator='|' ...)
```
The `|` in these commands is not redirection, but a special separator. This case should be skipped.

insights/config/__init__.py
```diff
@@ -588,6 +588,16 @@ def __str__(self):
     def __init__(self, command, multi_output=True, large_content=False, **kwargs):
         if command[0] != '/':
             raise SpecPathError(command, "CommandSpec command must start with '/'")
+        for shellism in ('&&', '||', '|', '>', '<'):
+            # The index will always work since we know from the previous test
+            # that index 0 is not one of these shellisms (because it's '/').
+            if shellism in command and command[command.index(shellism) - 1] == ' ':
+                raise SpecPathError(command,
```
bfahr: Since this is not really a path exception, should you create a new exception class for this such as `SpecSyntaxError`?  

Also this is failing the tests because the `SpecPathError` exception does not have a `__str__` method.  If you add something like this to the `SpecPathError` class then it should work correctly.
```
def __str__(self):
    return '{}: "{}"'.format(self.message, self.expression)
```

PaulWay: Hi @bfahr - good idea; done!

insights/config/__init__.py
```diff
@@ -28,6 +28,21 @@ def __init__(self, expression, message):
...
+class SpecSyntaxError(Exception):
+    """
+        Syntactic errors in the spec path, such as use of invalid characters.
+    """
+    def __init__(self, path, message):
+        self.path = path
+        self.message = message
+
+    def __str__(self):
+        return u"Error in path '{s.path}': {s.message}".format(s=self)
```
bfahr: @PaulWay not trying to be picky here, but is `path` the correct term here?  Should it be `cmd` instead since it represents invalid characters in the bash command?

PaulWay: I'm leaving the option open for syntax errors in file paths (e.g. $HOME or $MAILTO or a path starting with ~, which won't get interpolated).

bfahr: Ok by me then.

PR #765, opened 2017-11-29
Revert "Modify fc-match command to make it work only on RHEL7"

xiangce:
- Since isn't evaluated by bash, the command will always fail.
Reverts RedHatInsights/insights-core#756


xiangce:
Just like what Chirs.H mentioned in the Bugzilla, "it really isn't a bug in Insights. It is an issue with fc-match.". We'd better do NOT modify our specs. We'll provide other solutions to the CU.

PR #766, opened 2017-11-29
Modify fc_match to make command `fc-match` only work on RHEL 7

zhangyi733:


zhangyi733:
After discussion with @xiangce , we decide to use `/bin/fc-match` instead of `/usr/bin/fc-match` to make `fc-match` only work on RHEL7, not on RHEL6 because there is a fc-match on RHEL6 that can cause system segfault under certain conditions.

But this method is not that perfect in theory, if customer creates a soft link between `/bin/fc-match` and `/usr/bin/fc-match` on RHEL6, he can still face the segfault issue.

The reason why we revert previous modification `/bin/uname -r | grep -F '3.10' > /dev/null && /bin/fc-match -sv 'sans:regular:roman' family fontformat` is that this command is not evaluated by bash in insights-client and can not collect what we want. Thank you @PaulWay for finding this issue.

Glad to know if you guys have any better solution to make sure  `fc-match` only run on RHEL7 before the fc-match bug is fixed(maybe not be fixed since RHEL 6 is in Phase 3).

lonicerae:
Actually i would think it should be framework's business to control what commands should be sent to client according to the RHEL version it runs. Just as we discussed in the Beijing meeting. Currently it's a band-aid solution.
I could not think out a better solution at present either.

koalakangaroo:
Framework is not able to run different set of commands in different environment now. It is possible in the 3.x version as Chris Sams explained in Beijing. There would be a "flag" in spec file to specify an environment where each command will be ran. Currently, I think Yi's solution is the best.

insights/parsers/fc_match.py
```diff
@@ -40,7 +43,7 @@
 @parser(fc_match)
 class FCMatch(Parser):
     """
-    Class to parse command `/usr/bin/fc-match -sv 'sans:regular:roman' family fontformat`.
+    Class to parse command `/bin/fc-match -sv 'sans:regular:roman' family fontformat`.
     This object provides the ' __getitem__' and '__iter__' methods to allow it to
     be used as a list to iterate over the ``data`` data, e.g.::
```
xiangce: Before I merge it, @zhangyi733 - do you still want to remove the redundant `data` here?

zhangyi733:
@xiangce - I have removed the duplicated data. Please help to merge it and the PR for 1.x branch if they both look OK to you.

PR #767, opened 2017-11-29
Added new parsers for two options under /sys/kernel/mm/transparent_hugepage/.

jsvob:


insights/parsers/transparent_hugepage.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def get_line(self):
+        """
+        Returns the string value of the whole output.
+        E.g. `always [madvise] never`
+        """
+
+        return self.data
+
+    def get_opt(self):
```
bfahr: @jsvob have you considered using attributes or properties for all of these methods in both classes?

jsvob: No. Can you please point me to an existing parser that uses it in the way you'd like me to implement it?

insights/parsers/transparent_hugepage.py
```diff
@@ -0,0 +1,85 @@
...
+class ThpUseZeroPage(Parser):
+    """
+    Gets the contents of /sys/kernel/mm/transparent_hugepage/use_zero_page, which is either 0 or 1.
+    """
+    def parse_content(self, content):
+        self.data = " ".join(content).strip()
+
+    def get(self):
+        """Returns the string value."""
+        return self.data
```
bfahr: A couple of examples of how you might eliminate the `get` method are:
```
@property
def use_zero_page(self):
    return self.data
```
Alternatively you could set `self.use_zero_page` as an attribute in `parse_content`, and no method would be required.

insights/parsers/transparent_hugepage.py
bfahr: See my examples on above class.

insights/parsers/transparent_hugepage.py
jsvob: Right. I'll fix it. Thanks.

insights/parsers/transparent_hugepage.py
jsvob: I ended up using attributes because when I tried properties, the names of attributes and properties collided anyway.

xiangce:
Hi @jsvob, I merge it first, please do not forget to resubmit it to the master branch :) (#769)

PR #768, opened 2017-11-30
Modify fc-match to make it only run on RHEL7

zhangyi733:


zhangyi733:
See https://github.com/RedHatInsights/insights-core/pull/766

PR #772, opened 2017-11-30
Add parser for swift object-expirer.conf

chenlizhong:


insights/config/specs.py
```diff
@@ -371,6 +371,7 @@
     "sshd_config_perms"         : CommandSpec("/bin/ls -l /etc/ssh/sshd_config"),
     "sssd_config"               : SimpleFileSpec("etc/sssd/sssd.conf"),
     "sssd_logs"                 : PatternSpec(r"var/log/sssd/.*\.log$"),
+    "swift_object_expirer_conf": SimpleFileSpec("etc/swift/object-expirer.conf"),
```
xiangce: You might miss a whitespace before the ":" to keep them aligned.

insights/parsers/swift_conf.py
```diff
@@ -1,8 +1,12 @@
 """
-SwiftProxyServerConf - file ``/etc/swift/proxy-server.conf``
-============================================================
+Swift Conf Files - file ``/etc/swift/``
+=======================================
 
 This module provides parsers for swift config files under /etc/swift directory.
+
+SwiftProxyServerConf - file ``/etc/swift/proxy-server.conf``
```
xiangce: By adding a line of underlines to line-7 and line-8, the documation will look better.

insights/config/specs.py
xiangce: BTW, would you mind renaming the specs of `proxy_server.conf` to `swift_proxy_server_conf` in this PR? since the spec name doesn't affect plugins.

PR #773, opened 2017-11-30
Change mongod related specs to allow multi files parsing

JoySnow:


csams:
I don't think this will work as intended right now. The [original intent](https://github.com/RedHatInsights/insights-core/blob/1.x/insights/config/__init__.py#L310) of the All group was for files or commands that might be in different locations on different versions of RHEL. The client is told to collect them all, but in reality, the engine [expects only one element](https://github.com/RedHatInsights/insights-core/blob/1.x/insights/core/specs.py#L106) of an All group to be in a given archive. I'm bringing this up with Kyle to see if the behavior needs to change for 1.x specs.

xiangce:
Thank you @csams for correcting me. So, before this behavior is changed (if it will be changed), @JoySnow - better to split them into individual SimpleFileSpecs in 1.x for now.

JoySnow:
Since the 'All' is not functional, I changed the collecting method from 'All' to 'First', and split target files
into individual ones.
@xiangce , I still need to use PatternSpec instead of SimpleFileSpecs, to make the parsing result
to include all the items. Or, I will only get one parsed object instead of a list of them, no matter how many I feed it.

xiangce:
The current spec works well and @JoySnow has a plugin depend on it. I will go ahead and merge it.

PR #776, opened 2017-12-04
Update openshift get to include the command oc get bc

shzhou12:


bfahr:
As an enhancement you might consider adding a combiner to collect all of these `get...`'s into one set of information.  This may or may not make sense, depending on how the facts are used in rules.

shzhou12:
@bfahr Thanks for your suggestion. I also discussed with @xiangce, It makes sense to enhance the Openshift_get commands. I created a ToDo card https://trello.com/c/D0kswL41 and will enhance it later.

PR #794, opened 2017-12-05
Also parse the grub v1 config file for UEFI machine.

lonicerae:


insights/parsers/grub_conf.py
```diff
@@ -273,6 +276,42 @@ def get_current_title(self):
...
+@parser('grub-efi.conf')
+class Grub1EFIConfig(GrubConfig):
+    """
+    Parses grub v1 configuration for EFI-based systems
+    Content of grub-efi.conf is the same as grub.conf
+    """
+    def __init__(self, *args, **kwargs):
+        super(Grub1EFIConfig, self).__init__(*args, **kwargs)
+        self._version = 1
+        self._efi = True
```
bfahr: @lonicerae if you based this new class on `Grub1Config` instead of `GrubConfig` then you would only have to set `_efi` to `True` in `__init__` after calling `super`, and you would not need to implement `get_current_title`.

lonicerae: Yep, indeed.

lonicerae:
Hey @xiangce , please also review the enhancement to 3.x:

 - https://github.com/RedHatInsights/insights-core/pull/812

Thanks!

PR #799, opened 2017-12-06
WIP: Update the spec of `httpd_V` to get the correct command from `ps`

xiangce:
- In RHEL6, there are different httpd command binaries for different
  MPM_mode, we need to get the correct httpd command from `ps auxww`


xiangce:
The detail issue is:
- The old spec of HttpdV is `CommandSpec("/usr/sbin/httpd -V")`
- But, in RHEL6, for different MPM modes, there are different httpd command binaries provided, e.g. `/usr/sbin/httpd.worker` and `/usr/sbin/httpd.event`. 
- The `CommandSpec("/usr/sbin/httpd -V")` will only collect the output of `/usr/sbin/httpd -V`.
- So, if the customer has configured their httpd's to `worker` or `event` MPM mode, the parsing result of HttpdV will be wrong.

For insights-core_1.x, since the pre-command doesn't work, I added one new combiner HttpdV and two more individual parsers HttpdWorkerV and HttpdEventV to solve this issue.
- https://github.com/RedHatInsights/insights-core/pull/800

For insights-core_3.x, since the new "datasource" style spec is strong enough, I try to solve it inner the spec and without any combiners. So, I just updated its spec as follows:
- `httpd_cmd = sf.simple_command("/bin/ps auxww | grep /usr/sbin/httpd | grep -v grep | head -1 | awk '{print $11}'", name="httpd_cmd")`
- ` httpd_V = sf.foreach(httpd_cmd, "%s -V", name="http_V", alias="httpd-V")`

@csams - Would you please to take a look at this PR? I'm not sure if this solution will work as expected.

And since I didn't modify its old style specs in the `insights/config/specs`. This leads to the `insights/specs.py` and `insights/config/specs.py` don't match with each other. Is it OK in 3.x?

PaulWay:
Hi @xiangce - your simple_command won't work because the command doesn't get executed by `bash`, it gets executed directly.  See https://github.com/RedHatInsights/insights-core/pull/764

Can you give some examples of the different binaries in use?  And are they packaged by Red Hat?  Are they all in the `httpd` package?  Or are they in other packages?

xiangce:
@csams - Would you mind to confirm that if the simple_commands, like `httpd_cmd`, can be executed as expected?
If they won't work, I think I have to make corresponding `datasource` for them.

Thanks @PaulWay.

xiangce:
After #810 is OK and merged, the @datasource in this PR can be changed to use the `ps_lines`.

csams:
@xiangce commands aren't executed under a shell in 3.x, so "no bashisms" still applies.

xiangce:
@csams - I see, Thanks. I have updated the command contains "bashisms" to a datasource. While this PR depends on the update of #810. I will update this one after the 810 is merged.

PaulWay:
If you really need this kind of pipelined command processing, I'd suggest putting it in a script that we can deploy to the client.  A script is more testable, is self contained, and can do whatever it likes in there.

Otherwise, this data should be able to be retrieved from the `PsAuxwww` or upcoming `PsAuxww` parsers.

xiangce:
@PaulWay - As Chris Sams said "bahisms" are not supported for now, I have changed it to a @datasource and will update the spec in this PR as per #810 after it's merged. Thanks.

csams:
As far as I know, the client's update process doesn't support deployment of standalone bash scripts that specs can execute. @kylape or @jhjaggars can you confirm?

jhjaggars:
The new client deploys insights-core as an egg.  We don't put executables on the path, so a bash script for a custom pipeline is out.  Chaining datasources together is the desired mechansim for that type of operation.  If there is a case that chained datasources doesn't cover let's work on that.

xiangce:
Mark as `WIP` temporary, will rebase it to the master branch after the #810 is merged.

xiangce:
This PR is not merged. I closed it by mistake. And opened a new PR: #833

PR #804, opened 2017-12-07
Test example code in documentation for 1.x branch

PaulWay:
Test the documentation to make sure the example code works from the input data.

Insights parsers and combiners usually contain samples of the input data and code examples that are written as if a user had imported the parser or combiner and was trying commands out in an interpretive shell.  Not only is it useful to make sure that this code actually works and demonstrates the correct usage of the class under test, but the code also forms useful test code that should be checked by the test system.

We introduce the `doc_test_examples` function, which tests all the classes in a given module.  If the parser module given only contains one parser, the documentation for it is assumed to be in the parser module.  If there is more than one parser, the module documentation is ignored and the documentation of the parser itself is assumed to contain the input data and examples.

This work is done by the `doc_test_examples_in` function, which looks in the documentation given for an indented block (following a ``::``) and an `Examples:` section.  If these are not found, no testing is done.  Otherwise, the indented block is stripped of leading space and read as test data, and put in a 'shared' dictionary against the given parser using the `'context_wrap'` function.  This is stored in a dictionary of local declarations, to be used in later evaluation.

Each example line preceded by '`>>> `' is assumed to be an expression and `eval`uated, and the result is asserted to be equal to any non-'`>>> `' indented lines (which are concatenated together and then also `eval`'ed).  The `'variable' = shared['parser']` declaration is used to set the given name in the local declarations dictionary.  For example, the following example code:

```
>>> ns = shared[Netstat]
>>> hasattr(ns, 'data')
True
```
Will be interpreted in the following manner:

 * The 'ns' local variable will be set to the Netstat parser instance (which is populated from the input data given above).
 * The statement '`hasattr(ns, 'data')`' is evaluated and is checked against the value `True`.

Some attempt is made to set up the local evaluation environment but it is not assumed to be a complete Python environment.  In particular, other statements that are not 'variable = value' or 'expression' are not parsed.  Only 'variable = value' statements update the local environment.

To use this code, import the `doc_test_examples` function from the `insights.tests` module, and your parser module as itself, into your test module and then add the following code:

```python
def test_MODULE_docs():
    doc_test_examples(MODULE)
```

As an example, we have updated the `netstat` module parsers to use the correct format of documentation, and have verified that the documentation matches the parser output usage.

### DIFF
--- a/docs/docs_guidelines.rst
+++ b/docs/docs_guidelines.rst
@@ -34,7 +34,7 @@ only moving the documentation down into each parser when it is specific to the p
 For instance if there is only one parser in a module then almost all of the documentation
 will be in the module section.  If there are multiple parsers that are very similar then
 most of the documentation will be in the module section and only the unique details will
-be in each parser's class.  
+be in each parser's class.
 
 Look at the example code in this article and also review the source files for
 these parsers and see how the documentation has been organized.
@@ -94,7 +94,7 @@ in code contributed to the insights-core project:
        >>> "0d:00.0" in pci_info
        True
    """
-   from .. import LogFileOutput, parser 
+   from .. import LogFileOutput, parser
 
 
    @parser('lspci')
@@ -128,7 +128,7 @@ The second line is the name of the module and a descriptive phrase.  In this cas
 the file is **lspci.py**, the module is **lspci** and it is a command.  An example
 of a file parser would be file **fstab.py**, module name **fstab** and descriptive
 phrase'File /etc/fstab'.  The module
-name line is followed by a line of ``=`` characters that is the same length as the 
+name line is followed by a line of ``=`` characters that is the same length as the
 entire module line.  A blank line follows the module information.
 
 Description
@@ -155,7 +155,7 @@ Description
    whether a particular string is in the ``lspci`` output.  Other methods/operators
    are also supported, see the :py:class:`insights.core.LogFileOutput` class for more information.
 
-Next comes the description of the module. 
+Next comes the description of the module.
 Since this description is the first thing a developer will see when viewing
 the documentation it is important that the description is clear, concise and useful.
 Include elements of the module that would not be obvious from looking at the code.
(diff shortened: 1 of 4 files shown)


koalakangaroo:
@PaulWay - I like this idea. Good job! 👍

PaulWay:
I've just had a discussion with @bfahr where he pointed me to what I should have found before:

http://insights-core.readthedocs.io/en/1.x/customtut_parsers.html#parser-documentation

We can use doctest (http://www.sphinx-doc.org/en/stable/ext/doctest.html) to check the code in the documentation, writing it in such a way that it sets up the parser input and then uses `context_wrap` to create the new object.

The main downside I see to using doctest is that it allows imports - which it has to in order to get, say, context_wrap.  This could allow a malicious user to submit a PR which executes:
```python
from os import execl
execl("/bin/rm", "-rf", "/")
```
Or other more subtle attacks.  If this is done in the build system automatically, it may provide a route for compromising some of our build infrastructure.  I imagine this has already been thought of and dealt with, but I'd appreciate your thoughts on this.

My code also automatically detects and uses the input data, and uses context_wrap to create the parser object in a 'shared' dictionary.  This makes it easier to convert the existing documentation to being tested - these steps have to be done explicitly in doctest, as can be seen in our documentation example.

I think doctest is useful, and I think we should be using it more for our internal code such as examples of different spec uses or parser helper functions.  It is a better all-purpose code execution and test platform.  It's probably also got much nicer error messages :-)  So I'd like this proposal to be considered alongside doctest - it's not meant to replace it, but I do think it still has some advantages.  I probably wouldn't have written it if I'd thought of checking for doctest, but here it is now to be judged on its merits.

bfahr:
The doctest that is part of core python is useful for testing examples in documentation and is documented here https://docs.python.org/2/library/doctest.html.  Sphinx also provide support for this as referenced by @PaulWay.  The idea presented in this PR is a valid and I believe good, as discussed in the doctest documentation here https://docs.python.org/2/library/doctest.html#soapbox, although the implementation may be different that presented in this PR.

bfahr:
@PaulWay while I think testing the docs is a good idea I don't think this PR makes sense given the risk of changing production 1.x code vs the value, and given that doctest is already a feature provided by Python core libraries.

PaulWay:
Hi @bfahr - my usual habit is to submit to 1.x first; then I have a small script that duplicates that work to master via cherry-pick.  I didn't think of the fact that this is an enhancement, and so probably shouldn't go in 1.x which is (more) in maintenance mode.

How do you feel about this going into master instead?  Or would you really prefer that we use doctest?

I'm a little wary of arbitrary code execution, especially against our dev environment, using doctest.  Can we also make doctest failures an error in our test suite?

PaulWay:
OK, well, the consensus seems to be to use doctest, so I'm going to close this PR.

PR #817, opened 2017-12-12
Doctest example code - part A - for 1.x branch

PaulWay:
This implements doctest testing of the examples in:
 * netstat
 * ethtool
 * pam


insights/parsers/netstat.py
```diff
@@ -572,42 +614,51 @@ class SsTULPN(Parser):
     This class parse the input as a table with header:
         "Netid  State  Recv-Q  Send-Q  Local-Address-Port Peer-Address-Port  Process"
```
bfahr: You've duplicated this line after the examples so I think you need to remove this one.

bfahr:
@PaulWay in the future it would be helpful in expediting the review process if you would limit your PRs to one module, thanks!

PaulWay:
Hi @bfahr - sorry, I should have thought of that.  I was actually hoping to save you some trouble by reducing the number of individual PRs to review :-)  I'll do them separately in the future.

Do you want me to just do doctest work on master in the future as well?  Or do we want that in 1.x as well?

bfahr:
@PaulWay based on my conversations with @jhjaggars and @kylape I think you can add these to 1.x if it would help the user understand the parser, fix an issue, or provide some additional value to Insights.  Review of documentation PRs is lower priority than review of other types of PRs to please be patient if you do submit them.

PaulWay:
Sure thing, @bfahr  - understood.

insights/parsers/netstat.py
PaulWay: Yep, thanks, lost track of edits there.  Thanks @bfahr

PaulWay:
Updated with the much nicer hidden initialisation method from @bfahr - should be ready to go!

bfahr:
@PaulWay the changes look good to me.  FYI when you make changes to 3.x remember that `shared` no longer exists.

PaulWay:
Done, thanks @bfahr

PR #818, opened 2017-12-13
Nfnetlink queue documentation and code enhancements for master branch

PaulWay:
This is the master branch version of #814


insights/parsers/nfnetlink_queue.py
```diff
@@ -1,67 +1,67 @@
...
-        0  -4423     0 2 65535     0     0       22  1
-        1  -4424     0 2 65535     0     0       27  1
+    >>> # Set up of the environment - ignore this bit:
+    >>> nfnetlink_queue_data = '''
+    ... 0  -4423     0 2 65535     0     0       22  1
+    ... 1  -4424     0 2 65535     0     0       27  1
+    ... '''
+    >>> from insights.tests import context_wrap
+    >>> from insights.parsers.nfnetlink_queue import NfnetLinkQueue
+    >>> shared = {NfnetLinkQueue: NfnetLinkQueue(context_wrap(nfnetlink_queue_data))}
```
bfahr: @PaulWay since `shared` is no longer used by consumers of parsers in versions after 1.x, this could be simplified to be the suggested name that would be used in a rule, for example instead of `nf` you might have:
```
nfnetlink_queue = NfnetLinkQueue(context_wrap(nfnetlink_queue_data))
```

PaulWay: Good point!  I'll have to watch out for that in the future.  Thanks for picking that up @bfahr

xiangce:
I think the last commit has fixed the issue that @bfahr has commented, will go ahead and merge it today.

PR #819, opened 2017-12-13
DeprecationWarnings are now non-fatal and are the only type of deprecation - for 1.x branch

PaulWay:
This brings our use of `DeprecationWarning` into line with other parts of the Python ecosystem - they issue a DeprecationWarning if something is to be removed in a subsequent release, and the feature is removed in that later release.  The PendingDeprecation is no longer used.  DeprecationWarning is
no longer a fatal error, but will be seen in tests.

This is a cherry-pick of the functionality being pushed to the master branch in #726


xiangce:
Renamed the title and label to 1.x

PR #820, opened 2017-12-13
Add parser for nginx.conf 1x

wushiqinlou:
Function `create_parser()`  in this parser refer to the part code of package nginxparser_eb. This parser uses `create_parser()` to parse content of `nginx.conf` to list, then use `convert_nginx_list_to_dict` to convert the list to dict for convenience.  As it is hard to determine the path of `nginx.conf` if a customized version is used, we will parse only two versions, one is from Red Hat, and other is from `nginx.org`


xiangce:
The logic seems complex, I will add my comment after I digested it if I can :rofl:

bfahr:
@wushiqinlou we currently filter the Apache httpd.conf files because they may contain sensitive information.  Have you done any research to determine if sensitive information could be contained in the nginx.conf files?  If the file needs to be filtered then you would probably need to parse it differently, more like the httpd.conf files are parsed.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+    if isinstance(l, list):
+        return 1 + max(depth(item) for item in l)
+    else:
+        return 0
+
+
+def convert_nginx_list_to_dict(li):
+    """
+    After parsed by create_parser(), the result is like following, it is better to convert to dict for convenience.
+    Example: [['user', 'root'], ['worker_processes', '5'], ['error_log', 'logs/error.log'], ['pid',
```
PaulWay: You can probably afford to trim this down a bit to make your example more obvious.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+    [['location', '/'], [['proxy_pass', 'http://127.0.0.1:8080']]]]], [['map', '$http_upgrade',
+    '$connection_upgrade'], [['default', 'upgrade'], ["''", 'close']]], [['upstream', 'websocket'], [['server',
+    '10.66.208.205:8010']]], [['upstream', 'big_server_com'], [['server', '127.0.0.3:8000 weight=5'], ['server',
+    '127.0.0.3:8001 weight=5'], ['server', '192.168.0.1:8000'], ['server', '192.168.0.1:8001']]], [['server'],
+    [['listen', '80'], ['server_name', 'big.server.com'], ['access_log', 'logs/big.server.access.log main'],
+    [['location', '/'], [['proxy_pass', 'http://big_server_com']]]]]]]]
+
+    After converting, it should be like following:
+
+    For the section, the section name would be a dict key, and content embraced by brace would be value. In some
```
PaulWay: I think an actual example of the conversion of (some of) the above into dict format would be good here.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+    sections, the first item is like ['location', '/'], in this case, the convert rule is that 'location' would be
+    the key, and add {"name":'/'} key-value to the dict value.
+    """
+    dict_all = {}
+    for item in li:
+        if depth(item) == 1:
```
PaulWay: Since depth is recursive, I'd cache it rather than continually calling it.  Otherwise we may get unexpectedly quadratic behaviour and slow parsing.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+    """
+    Class for ``nginx.conf`` and ``conf.d`` configuration files.
+
+    Gerenally nginx.conf is writed as key-value format. It has a mail section and several sections,
+    http, mail, events, etc. They are unique, and subsection server and location in http section could
+    be duplicate, so the value of these subsections may be list.
+    """
+
+    def parse_content(self, content):
+        list1 = UnspacedList(create_parser().parseString("\n".join(get_active_lines(content))).asList())
```
PaulWay: Can pyparsing take a list of lines?  It seems a bit expensive to join the content together again only for it to be parsed...  I'd also recommend splitting this up so it's a bit easier to read.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+        listen          80;
+        server_name     big.server.com;
+        access_log      logs/big.server.access.log main;
+
+        location / {
+          proxy_pass      http://big_server_com;
+        }
+      }
+    }
+
```
PaulWay: I'd recommend only including a small part of the actual file, because you can make the examples work with just a small part.

I'd also recommend looking at the doctest usage in http://insights-core.readthedocs.io/en/1.x/customtut_parsers.html#parser-documentation to test the example code you write here.  The nfnetlink_queue parser also uses this, and there are other examples in https://github.com/RedHatInsights/insights-core/pull/817 .  I would only say that instead of having `if __name__ == '__main__'` in the parser code, test it explicitly using `doctest.testmod({module})` in your test module - see the above parser changes for how to do this and generate errors.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+Examples:
+
+    >>> nginxconf = shared[NginxConf]
+    >>> nginxconf['user'] # Quotes are not removed
```
PaulWay: You probably need an example to show a value in quotes :-)

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+    'root'
+    >>> nginxconf['events']['worker_connections']
```
PaulWay: I would recommend mentioning in a comment that values are all kept as strings.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+"""
+import string
+import copy
+from .. import parser, LegacyItemAccess, Parser, get_active_lines
+from insights.contrib.pyparsing import (
+    Literal, White, Word, alphanums, CharsNotIn, Combine, Forward, Group,
+    Optional, OneOrMore, ZeroOrMore, Regex, stringEnd, restOfLine)
+
+
+def create_parser():
```
PaulWay: I think your use of pyparsing is good here - you've documented it well and it's a horrible format to try and parse.  Good work!

PaulWay:
BTW, @bfahr , do we have any way of _excluding_ lines using our filtering?  Or do we only include?  The parsing of `httpd` config files is significantly broken because of this whitelist-only filtering.  Once we add filtering, it makes it very difficult to remove because the filtering is done on the client side and stored archives are no longer in the right format.  So I'd really like to see the complete file unless there's a very good reason to filter at this stage.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+                if depth(sub_item) == 3:
+                    sub2dict_all = {}
+                    for sub2_item in sub_item[1]:
+                        sub2dict_all[sub2_item[0]] = sub2_item[1]
+                    if len(sub_item[0]) > 1:
+                        sub2dict_all["name"] = ' '.join(sub_item[0][1:])
+                    subdict_all[sub_item[0][0]] = _handle_key_value(subdict_all, sub_item[0][0], sub2dict_all)
+            dict_all[item[0][0]] = subdict_all
+
+        if depth(item) == 7:
```
xiangce: There are some duplicated code slices in this function, how about scratching the duplicated code slices and defining several private-inner-functions to reuse them?

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,338 @@
...
+import copy
+from .. import parser, LegacyItemAccess, Parser, get_active_lines
+from insights.contrib.pyparsing import (
+    Literal, White, Word, alphanums, CharsNotIn, Combine, Forward, Group,
+    Optional, OneOrMore, ZeroOrMore, Regex, stringEnd, restOfLine)
+
+
+def create_parser():
+    """
+    A class that parses nginx configuration with pyparsing
```
xiangce: Hey @bfahr, after talked with @wushiqinlou, we have confirmed that this `create_parser()` is from [nginxparser_eb](https://pypi.python.org/pypi/nginxparser_eb/0.0.9). the `nginxparser_eb` is protected by MIT License, do you think whether it's necessary to add the MIT License content to the docstring title of this file?

bfahr: This is probably not a "copy or substantial portion", but I would go ahead and include it the docstring for `create_parser()`, not the whole module.

bfahr:
@PaulWay I don't think there is currently a capability to excluding lines via filtering.  This is something that happens in the client to clean files, but not in the engine.  I'll copy @csams to repond to the implementation for 3.x which may be different.

xiangce:
Since the logic is OK and the unit test is sufficient, and there is a new plugin depends on it, I go ahead and merge the PR to 1.x branch at first. Any enhancement on the code and docstring can be carried out on the PR (#822 ) to master branch and be backported after #822 is merged.

PR #822, opened 2017-12-13
Add parser for nginxconf

wushiqinlou:


insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,341 @@
...
+
+        location / {
+          proxy_pass      http://big_server_com;
+        }
+      }
+    }
+
+Examples:
+
+    >>> nginxconf = shared[NginxConf]
```
bfahr: Since `shared` does not exist anymore in 3.x would it make sense to change this to something more consistent with 3.x?

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,335 @@
...
+
+def create_parser():
+    """
+    A class that parses nginx configuration with pyparsing
+
+    The MIT License (MIT)
+
+    Copyright (c) 2014 Fatih Erikli
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy of
```
PaulWay: Just a note here - you can't apply the MIT license to just a part of a file AFAICS.  It applies to the whole file, and possibly the whole project.

If you're using code that you can't import from a standard location - i.e. that isn't installed from RPM package or via pip - then the usual form AFAIK is to say something like:
```
This parser is derived from the code by Fatih Erikli at:

http://url.to/code/

Used under the MIT license.
```
Check with someone else though - this may just be what I'm used to.

This also allows us to check on the upstream code for updates and possibly supply bug fixes.  We should always try to work directly with upstream where possible.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,335 @@
...
+                if "#" not in self[:i]:
+                    list.__delitem__(self, i)
+
+    def _spacey(self, x):
+        return (isinstance(x, str) and x.isspace()) or x == ''
+
+
+@parser('nginx_rh_conf')
+@parser('nginx_org_conf')
+class NginxConf(Parser, LegacyItemAccess):
```
PaulWay: Isn't there a problem with applying two specs to the one parser?

If you want to have the `nginx.conf` file found in multiple locations, I think this is where you use the `First` spec, and have only one spec that finds the first file it can in multiple locations.

However, if you want to have multiple files, and a parser for each, then you really need to have one parser class per spec, and one spec per file.  This is where you'd have a main class and then have a simple subclass for each file.

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,335 @@
...
+    def parse_content(self, content):
+        list_result = UnspacedList(create_parser().parseString("\n".join(get_active_lines(content))).asList())
+        self.data = self._convert_nginx_list_to_dict(list_result)
+
+    def _convert_nginx_list_to_dict(self, li):
+        """
+        After parsed by create_parser(), the result is a list, it is better to convert to dict for convenience.
+        """
+
+        def _listdepth_three(self, li):
```
PaulWay: This really looks like you need a `_list_to_dict()` method which is recursively applied to each substructure.  That approach would save you having very similar but slightly different functions for different layers of list - it would simply apply `_list_to_dict()` on each sublist value in turn.  I suspect this would also save you the `_depth()` method, which looks to me like it's going to cause recursion slow-down by being called many times on the same object.

I'd have to see what kind of structure comes out of the parsing and unspaced-list processing in order to draft up a suitable piece of test code.

insights/parsers/nginx_conf.py
wushiqinlou: @bfahr - Do we have a standard workflow to handle open source code reference? Add license  to the function docstring, or do as Paul said?

insights/parsers/nginx_conf.py
```diff
@@ -0,0 +1,333 @@
...
+                self.spaced[i] = sublist.spaced
+            elif self._spacey(entry):
+                list.__delitem__(self, i)
+
+    def _spacey(self, x):
+        return (isinstance(x, str) and x.isspace()) or x == ''
+
+
+@parser('nginx_rh_conf')
+@parser('nginx_org_conf')
```
xiangce: you missed this part.

insights/parsers/nginx_conf.py
bfahr: @klape @jhjaggars how do you recommend we handle adding code from other open source projects?

bfahr: @wushiqinlou @jhjaggars recommends that this function be placed in the `insights/contrib` directory.  Please backport this change to the 1.x branch as well.  Take a look at the modules that are already there for examples.

insights/parsers/nginx_conf.py
bfahr: @wushiqinlou I think @PaulWay has a good point here, and I'm not sure that a dictionary representation is accurate, especially if order matters which I believe it does in some cases.  I had to dump out the two representations to understand what was going on, here's the list before conversion to a dict:
```
[   ['user', 'root'],
    ['worker_processes', '5'],
    ['error_log', 'logs/error.log'],
    ['pid', 'logs/nginx.pid'],
    ['worker_rlimit_nofile', '8192'],
    [['events'], [['worker_connections', '4096']]],
    [   ['mail'],
        [   ['server_name', 'mail.example.com'],
            ['auth_http', 'localhost:9000/cgi-bin/auth'],
            [['server'], [['listen', '143'], ['protocol', 'imap']]]]],
    [   ['http'],
        [   ['include', 'conf/mime.types'],
            ['include', '/etc/nginx/proxy.conf'],
            ['include', '/etc/nginx/fastcgi.conf'],
            ['index', 'index.html index.htm index.php'],
            [   'log_format',
                'main  \'$remote_addr - $remote_user [$time_local] "$request" \'\n\'$status $body_bytes_sent "$http_referer" \'\n\'"$http_user_agent" "$http_x_forwarded_for"\''],
            ['default_type', 'application/octet-stream'],
            ['access_log', 'logs/access.log  main'],
            ['sendfile', 'on'],
            ['tcp_nopush', 'on'],
            ['server_names_hash_bucket_size', '128'],
            [   ['server'],
                [   ['listen', '80'],
                    ['server_name', 'domain1.com www.domain1.com'],
                    ['access_log', 'logs/domain1.access.log  main'],
                    ['root', 'html'],
                    [   ['location', '~', '\\.php$'],
                        [['fastcgi_pass', '127.0.0.1:1025']]]]],
            [   ['server'],
                [   ['listen', '80'],
                    ['server_name', 'domain2.com www.domain2.com'],
                    ['access_log', 'logs/domain2.access.log  main'],
                    [   [   'location',
                            '~',
                            '^/(images|javascript|js|css|flash|media|static)/'],
                        [   [   'root',
                                '/var/www/virtual/big.server.com/htdocs'],
                            ['expires', '30d']]],
                    [   ['location', '/'],
                        [['proxy_pass', 'http://127.0.0.1:8080']]]]],
            [   ['map', '$http_upgrade', '$connection_upgrade'],
                [['default', 'upgrade'], ["''", 'close']]],
            [['upstream', 'websocket'], [['server', '10.66.208.205:8010']]],
            [   ['upstream', 'big_server_com'],
                [   ['server', '127.0.0.3:8000 weight=5'],
                    ['server', '127.0.0.3:8001 weight=5'],
                    ['server', '192.168.0.1:8000'],
                    ['server', '192.168.0.1:8001']]],
            [   ['server'],
                [   ['listen', '80'],
                    ['server_name', 'big.server.com'],
                    ['access_log', 'logs/big.server.access.log main'],
                    [   ['location', '/'],
                        [['proxy_pass', 'http://big_server_com']]]]]]]]
```

And here's the that data converted to a dict by the parser:

```
{   'error_log': 'logs/error.log',
    'events': {   'worker_connections': '4096'},
    'http': {   'access_log': 'logs/access.log  main',
                'default_type': 'application/octet-stream',
                'include': [   'conf/mime.types',
                               '/etc/nginx/proxy.conf',
                               '/etc/nginx/fastcgi.conf'],
                'index': 'index.html index.htm index.php',
                'log_format': 'main  \'$remote_addr - $remote_user [$time_local] "$request" \'\n\'$status $body_bytes_sent "$http_referer" \'\n\'"$http_user_agent" "$http_x_forwarded_for"\'',
                'map': {   "''": 'close',
                           'default': 'upgrade',
                           'name': '$http_upgrade $connection_upgrade'},
                'sendfile': 'on',
                'server': [   {   'access_log': 'logs/domain1.access.log  main',
                                  'listen': '80',
                                  'location': [   {   'fastcgi_pass': '127.0.0.1:1025',
                                                      'name': '~ \\.php$'}],
                                  'root': 'html',
                                  'server_name': [   'domain1.com www.domain1.com']},
                              {   'access_log': 'logs/domain2.access.log  main',
                                  'listen': '80',
                                  'location': [   {   'expires': '30d',
                                                      'name': '~ ^/(images|javascript|js|css|flash|media|static)/',
                                                      'root': '/var/www/virtual/big.server.com/htdocs'},
                                                  {   'name': '/',
                                                      'proxy_pass': 'http://127.0.0.1:8080'}],
                                  'server_name': [   'domain2.com www.domain2.com']},
                              {   'access_log': 'logs/big.server.access.log main',
                                  'listen': '80',
                                  'location': [   {   'name': '/',
                                                      'proxy_pass': 'http://big_server_com'}],
                                  'server_name': ['big.server.com']}],
                'server_names_hash_bucket_size': ['128'],
                'tcp_nopush': 'on',
                'upstream': [   {   'name': 'websocket',
                                    'server': ['10.66.208.205:8010']},
                                {   'name': 'big_server_com',
                                    'server': [   '127.0.0.3:8000 weight=5',
                                                  '127.0.0.3:8001 weight=5',
                                                  '192.168.0.1:8000',
                                                  '192.168.0.1:8001']}]},
    'mail': {   'auth_http': 'localhost:9000/cgi-bin/auth',
                'server': [{   'listen': '143', 'protocol': 'imap'}],
                'server_name': ['mail.example.com']},
    'pid': 'logs/nginx.pid',
    'user': 'root',
    'worker_processes': '5',
    'worker_rlimit_nofile': '8192'}
```

It seems to me that a the list might be a better internal representation and have accessor methods that simplify access to the data by rules.

insights/contrib/nginxparser.py
```diff
@@ -0,0 +1,119 @@
+# Copyright (c) 2014 Fatih Erikli
```
bfahr: This looks good, thanks.

insights/parsers/nginx_conf.py
wushiqinlou: @bfahr - Thanks for your point. However, I can not understand why order matters. In nginx.conf, I set the values of key which could be duplicate as list in the result(function `_handle_key_value`), and they are all valuable, no overwrite. For keys which could not be duplicate, nginx would report error while starting service. For example:

~~~
# cat /etc/nginx/nginx.conf
...
worker_shutdown_timeout 10ms;
worker_shutdown_timeout 11ms;
...

# systemctl restart nginx
Job for nginx.service failed because the control process exited with error code. See "systemctl status nginx.service" and "journalctl -xe" for details.

# cat /var/log/messages
Jan  5 17:57:08 dns nginx: nginx: [emerg] "worker_shutdown_timeout" directive is duplicate in /etc/nginx/nginx.conf:2
~~~ 

If there is any other order matters, please kindly info me.

Also, For list format, if we want to access the the value of key `sendfile` in `http` section, we need to loop the whole list in plugin, right?  If any useful accessor methods, please kindly info me.

xiangce: After a discussion with @wushiqinlou, we confirmed that for different options (e.g `user` and `pid` and etc.) in nginx.conf, the nginx app does not care their order. And just like @wushiqinlou mentioned above, for duplicated options, their values are stored in a list which will keep the original order.

@bfahr does it make sense to you? Please kindly correct us if there is something wrong. Thanks.

bfahr: @wushiqinlou @xiangce thanks for checking this.  I'm satisified with your confirmation that the current implementation is correct.

PR #831, opened 2017-12-18
Let `foreach` to support glob files

xiangce:
- rename the old `foreach` to `foreach_command` and add a new `foreach_file`
  add a new `foreach` to call them correspondingly


xiangce:
Hey @csams, there are 2 PRs (#832 and #840) depend on this PR, and Jaylin wants us to make several plugins about Java/Jboss asap. Would you please to review it today?

xiangce:
My bad, sometimes paths may contain spaces,  "foreach_execute" and "foreach_collect" make sense to me, will modify it as per your suggestion. Thanks.

PR #832, opened 2017-12-18
 Merge catalina log and update its specs

xiangce:
- merge the catalina log parers to `catalina_log.py`
- use the new `foreach` updated in #831
  To avoid false positive, we should get the `tomcat_home` from the command line gotten from `ps`


insights/parsers/catalina_log.py
```diff
@@ -0,0 +1,109 @@
...
+    Please refer to its super-class :class:`insights.core.LogFileOutput`.
+
+    When using this parser, consider using a filter or a scan method, e.g.::
+
+        CatalinaServerlog.filters.append('initializing ProtocolHandler')
+        CatalinaServerlog.keep_scan('init_pro', 'initializing ProtocolHandler')
+
+    Examples:
+        >>> CatalinaServerlog.filters.append('Failed to initialize')
+        >>> log = shared[CatalinaServerlog]
```
bfahr: @xiangce since `shared` is no longer used for 3.x is there a better way to demonstrate this in an example?

insights/parsers/catalina_log.py
```diff
@@ -0,0 +1,109 @@
...
+
+    However, this parser only recognises single lines.
+
+    When using this parser, consider using a filter or a scan method, e.g.::
+
+        CatalinaOut.filters.append('BAD packet signature')
+        CatalinaOut.keep_scan('bad_signatures', 'BAD packet signature')
+
+    Example:
+        >>> catalina = shared[CatalinaOut]
```
bfahr: Same question as above regarding `shared`.

insights/specs.py
```diff
@@ -34,8 +34,35 @@
...
 candlepin_log = sf.simple_file("/var/log/candlepin/candlepin.log", name="candlepin_log", alias="candlepin.log")
 candlepin_error_log = sf.first_of([sf.simple_command("/var/log/candlepin/error.log"), sf.simple_file(r"sos_commands/foreman/foreman-debug/var/log/candlepin/error.log", context=HostArchiveContext)], name="candlepin_error_log")
-catalina_out = sf.glob_file(["/var/log/tomcat/catalina.out", "/var/log/tomcat6/catalina.out", "/tomcat-logs/tomcat/catalina.out", "/tomcat-logs/tomcat6/catalina.out"], name="catalina_out", alias="catalina.out")
-catalina_server_log = sf.glob_file(["/var/log/tomcat/logs/catalina*.log", "/var/log/tomcat6/logs/catalina*.log", "/tomcat-logs/tomcat/catalina*.log", "/tomcat-logs/tomcat6/catalina*.log"], name="catalina_server_log")
+ps_auxww = sf.first_of([
+    sf.simple_command("/bin/ps auxww", name="ps_auxww"),
+    sf.simple_file('sos_commands/process/ps_aux', alias='ps_aux', context=HostArchiveContext),
+    sf.simple_file('sos_commands/process/ps_auxwww', alias='ps_auxwww', context=HostArchiveContext),
+    sf.simple_file('sos_commands/process/ps_auxcww', alias='ps_auxcww', context=HostArchiveContext),
+])
```
bfahr: I wonder if there is a better way to organize this file so we can keep specs in close to sorted order?  Maybe functions like `tomcat_base` could be moved after the spec factories like `ps_auxww` are defined?

insights/parsers/catalina_log.py
xiangce: My fault, will correct them. Thank you @bfahr.

insights/specs.py
xiangce: Yep, the specs sort is a bit messy currently. How about classifying them into different groups, such as `@datasource` group, `provider-factories` (e.g. ps_auxww) group and reorder them in each group?

insights/specs.py
```diff
@@ -34,8 +34,35 @@
...
+    findall = re.compile(r"\-Dcatalina\.base=(\S+)").findall
+    for p in ps:
+        found = findall(p)
+        if found:
+            # Only get the path which is absolute
+            results.extend(f for f in found if f[0] == '/')
+    return list(set(results))
+
+
+catalina_out = sf.first_of([sf.foreach(tomcat_base, "%s/catalina.out"),
```
zhangyi733: `catalina.out` should locate in $tomcat_base/logs/

xiangce: Thanks @zhangyi733, I updated it in the new commit.

PR #833, opened 2017-12-18
Update the spec of `httpd_V` to get the correct command from `ps`

xiangce:


insights/specs.py
```diff
@@ -34,6 +34,12 @@
 brctl_show = sf.simple_command("/usr/sbin/brctl show", name="brctl_show")
 candlepin_log = sf.simple_file("/var/log/candlepin/candlepin.log", name="candlepin_log", alias="candlepin.log")
 candlepin_error_log = sf.first_of([sf.simple_command("/var/log/candlepin/error.log"), sf.simple_file(r"sos_commands/foreman/foreman-debug/var/log/candlepin/error.log", context=HostArchiveContext)], name="candlepin_error_log")
+ps_auxww = sf.first_of([
+    sf.simple_command("/bin/ps auxww", name="ps_auxww"),
+    sf.simple_file('sos_commands/process/ps_aux', alias='ps_aux', context=HostArchiveContext),
+    sf.simple_file('sos_commands/process/ps_auxwww', alias='ps_auxwww', context=HostArchiveContext),
+    sf.simple_file('sos_commands/process/ps_auxcww', alias='ps_auxcww', context=HostArchiveContext),
+])
```
bfahr: See my comment in #832 regarding how to organize the things in this file.

xiangce: Got it. Thank you, Bob. I will adjust the order after we got a common conclusion in #832.

xiangce: Since the #832 is closed temporarily, I reopen the reordering conversation to here:

---
From @bfahr:
I wonder if there is a better way to organize this file so we can keep specs in close to sorted order? Maybe functions like tomcat_base could be moved after the spec factories like ps_auxww are defined? @csams and @kylape do you have any opinions?
---

How about classifying them into different groups, such as @datasource group, provider-factories (e.g. ps_auxww) group and reorder them in each group?

@csams, @kylape  any suggestions?

insights/specs.py
```diff
@@ -181,7 +187,23 @@ def docker_container_ids(broker):
...
+        if len(p_splits) >= 11:
+            cmd = p_splits[10].split()[0]
+            # Should compatible with RHEL6
+            # e.g. /usr/sbin/httpd, /usr/sbin/httpd.worker and /usr/sbin/httpd.event
+            if 'httpd' in os.path.basename(cmd):
+                return [cmd]
+    return []
+
+
+httpd_V = sf.foreach(httpd_cmd, "%s -V", name="http_V", alias="httpd-V")
```
bfahr: @xiangce I believe this will need to change now that #831 has been merged.

xiangce: Sure, I changed to foreach_execute, thanks.

PR #835, opened 2017-12-18
Added non-recursive spec for /var/www.

jsvob:


xiangce:
I cannot see any shared parsers or local-mappers in `insights-prodsec-rules` repo used this specs, then the insights client will never collect it.  Would you mind to add corresponding shared parsers as well?

skontar:
@xiangce I think Jakub forgot to mention BZ https://bugzilla.redhat.com/show_bug.cgi?id=1508567 (and maybe https://bugzilla.redhat.com/show_bug.cgi?id=1461065 for more information). Parsers will be developed in https://github.com/RedHatInsights/insights-prodsec-rules first and moved to core later if they would seem to benefit others.

skontar:
Parser is currently in private security repository, under review.

https://git.prodsec.redhat.com/Insights/insights-prodsec-rules/merge_requests/36

xiangce:
@skontar - I see, thanks for the clarification. Makes sense to me.

PR #841, opened 2017-12-19
Freeradius configuration files parser (/etc/raddb/*)

rmetrich:
FreeRadius is a parser for the freeradius configuration files usually in `/etc/raddb`.
The list of files is limited to active files:
- `/etc/raddb/*.conf`
- `/etc/raddb/mod-enabled/*`
- `/etc/raddb/policy.d/*`
- `/etc/raddb/sites-enabled/*`

Files ending with `.rpmnew` are excluded because they are backups made when
ugprading but customer modified the configuration files, so are not used by the
service.

Because this parser reads multiple files, the files are stored as a list within
the parser and need to be iterated through in order to find specific
configuration files.


eduardocerqueira:

bfahr:
ok to test

insights/config/specs.py
```diff
@@ -118,6 +118,7 @@
                                     SimpleFileSpec("sos_commands/foreman/foreman-debug/etc/foreman-proxy/settings.yml")]),
     "foreman_proxy.log"         : First([SimpleFileSpec("var/log/foreman-proxy/proxy.log", large_content=True),
                                     SimpleFileSpec("sos_commands/foreman/foreman-debug/var/log/foreman-proxy/proxy.log", large_content=True)]),
+    "freeradius"                : NoneGroup([PatternSpec(r"etc/raddb/(.+\.conf|(mods-enabled|policy\.d|sites-enabled)/.+)")]),  # /etc/raddb/*.conf + /etc/raddb/mod-enabled/* + /etc/raddb/policy.d/* + /etc/raddb/sites-enabled/*
```
csams: `insights/config/specs.py` doesn't exist in Insights Core 3.x. All specs (now known as datasources) should have a RegistryPoint entry in the Specs class found in `insights/specs/__init__.py`. Datasources for files that should be collected directly from a host, insights archive, or sos report (and share a relative path in all those contexts) are defined in insights/specs/default.py. If the path is different in insights archives or sos reports, put definitions in `insights/specs/insights_archive.py` and `insights/specs/sos_archive.py` respectively. If a datasource will collect multiple files, set `multi_output=True` on its `RegistryPoint`. If the data should be raw and not split line by line in advance, set `raw=True` on its `RegistryPoint` and pass `kind=RawFileProvider` in the definition. Have a look at the [Rule Migration Guide](https://github.com/RedHatInsights/insights-core/blob/master/notebooks/Rule%20Migration%20Guide.ipynb) and the [Datasource Registry Notebook](https://github.com/RedHatInsights/insights-core/blob/master/notebooks/Datasource%20Registry.ipynb) for details on how datasources work in 3.x.

Also, make sure all the files returned by a single datasource make sense for a single parser (they should share the same syntax and semantics, the first of which the parser interprets and the later of which the parser exposes with a higher level API than the raw lines).

insights/parsers/freeradius.py
```diff
@@ -0,0 +1,53 @@
...
+    Path: etc/raddb/proxy.conf
+    Path: etc/raddb/policy.d/filter
+
+"""
+
+from .. import Parser, parser
+from . import SkipException
+
+
+@parser("freeradius")
```
csams: Parsers in 3.x depend directly on their entry from the `Specs` class in `insights/specs/__init__.py`.

insights/parsers/freeradius.py
```diff
@@ -0,0 +1,53 @@
...
+class FreeRadius(Parser):
```
csams: The parser doesn't provide an API above the raw data of the datasource behind it.

The objects returned by datasources already contain file path information and contents.

Ignoring files can be done at collection time by passing a regular expression to `glob_file` with its `ignore` keyword (see `glob_file`'s entry in the [Insights Core Tutorial](https://github.com/RedHatInsights/insights-core/blob/master/notebooks/Insights%20Core%20Tutorial.ipynb)), and basic line cleansing can be handled with [get_active_lines](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/__init__.py#L30).

Parsers and combiners should provide a higher level API to rules and other components so they don't have to deal with parsing individual lines themselves.

PR #845, opened 2017-12-20
Updating documentation with current doctest usage code and discussion

PaulWay:
This updates the main documentation pages that discuss writing parsers with the `doctest` examples-as-tests work that @bfahr and I have done and that is in https://github.com/RedHatInsights/insights-core/pull/817  I've tried to follow Bob's style as much as possible, but I anticipate having to improve this to fit in more with the overall style.

PR #847, opened 2017-12-21
Add parser jboss_standalone_main_conf to parse JBoss standlone mode main

zhangyi733:
conf file


insights/specs.py
```diff
@@ -203,6 +203,36 @@ def docker_container_ids(broker):
 ipv4_neigh = sf.simple_command("/sbin/ip -4 neighbor show nud all", name="ipv4_neigh")
 ipv6_neigh = sf.simple_command("/sbin/ip -6 neighbor show nud all", name="ipv6_neigh")
 iscsiadm_m_session = sf.simple_command("/usr/sbin/iscsiadm -m session", name="iscsiadm_m_session")
+ps_auxww = sf.first_of([
```
PaulWay: I'd _really strongly_ recommend not reordering the ps parser specs just to be above the jboss_standalone_main_config_files spec.  I can see why you're doing it, but we're trying to keep the parser specs in alphabetical order for ease of reference.

What I'd recommend doing here is putting the more complex data sources at the bottom of the specs file, in a specific place so that we know to look there.  Then leave a comment where the 'jboss_standalone_main_config_files' spec would normally be defined to say that it's down below.

However, I think @csams was talking about a better way to include data sources in parsers, so that we didn't have this kind of complex dep

insights/parsers/tests/test_jboss_standalone_main_conf.py
```diff
@@ -0,0 +1,71 @@
...
+                    </include>
+                </role>
+            </role-mapping>
+        </access-control>
+    </management>
+</server>
+"""
+
+
+def test_jboss_standalone_conf():
```
PaulWay: If this is the same content that's in your examples, then setting the environment up for documentation testing is easy!

insights/parsers/jboss_standalone_main_conf.py
```diff
@@ -0,0 +1,109 @@
...
+    self.xmlns is the using namespace
+    """
+
+    def parse_content(self, content):
+        dom = ET.fromstring("\n".join(content))
+        self.dom = dom
+        xmlns = dom.tag.split("server")[0]
+        self.xmlns = xmlns
+
+    def get_properties(self, property_path):
```
PaulWay: Doesn't the DOM support XPath searching?  Maybe we should get users to fetch properties using XPath queries?

insights/parsers/jboss_standalone_main_conf.py
```diff
@@ -0,0 +1,109 @@
...
+                </security-realm>
+            </security-realms>
+        </management>
+    </server>
+
+Examples:
+
+    >>> from insights.tests import context_wrap
+    >>> from insights.parsers.jboss_standalone_main_conf import JbossStandaloneConf
+    >>> JBOSS_STANDALONE_CONFIG = '''
```
PaulWay: This is fine, but I'd recommend following the current practices for example testing explained in https://github.com/RedHatInsights/insights-core/pull/845 - you don't need to re-include this content or do the environment set-up within the documentation.  It also allows you to do some of the standard testing of properties, etc. in your examples rather than duplicate that in the test suite.

zhangyi733: @PaulWay Good suggestion!

insights/parsers/jboss_standalone_main_conf.py
zhangyi733: @PaulWay A good suggestion. According to https://docs.python.org/2/library/xml.etree.elementtree.html,  limited support XPath syntax is supported.  Xiangce and I agreed to create a base class for all XML related parsers in . So I will enhance this func when https://github.com/RedHatInsights/insights-core/pull/855 is ready to merge.

insights/parsers/tests/test_jboss_standalone_main_conf.py
zhangyi733: Yes and I use doctest module instead now.

insights/specs.py
zhangyi733: @PaulWay Agree and I add my datasource to the end of the spec for now before Chris give us a better way.

zhangyi733:
@xiangce @bfahr @PaulWay , I have updated class JbossStandaloneConf to inherit from xml parser base class XMLParser. Would you mind to review it again when you are available?

bfahr:
@zhangyi733 please resolve the conflicts and this is ready to merge.

PR #849, opened 2017-12-22
Documentation and improvements to OVSvsctlshow parser for 1.x branch 

PaulWay:
* Documentation improvements including sample input and examples
* Testing of the example code
* Some improvements to the code, including not reinventing wheels

### DIFF
--- a/insights/parsers/ovs_vsctl_show.py
+++ b/insights/parsers/ovs_vsctl_show.py
@@ -1,90 +1,69 @@
+"""
+OVSvsctlshow - command ``/usr/bin/ovs-vsctl show``
+==================================================
+
+This class parses the output of the ``/usr/bin/ovs-vsctl show`` command.  The
+output looks like::
+
+    aa3f16d3-9534-4b74-8e9e-8b0ce94038c5
+        Bridge br-int
+            fail_mode: secure
+            Port br-int
+                Interface br-int
+                    type: internal
+            Port int-br-ctlplane
+                Interface int-br-ctlplane
+                type: patch
+                options: {peer=phy-br-ctlplane}
+            Port "tap293d29ec-d1"
+                tag: 1
+                Interface "tap293d29ec-d1"
+                    type: internal
+        Bridge br-tun
+            fail_mode: secure
+            Port "vxlan-aca80118"
+                Interface "vxlan-aca80118"
+                    type: vxlan
+                    options: {df_default="true", in_key=flow, local_ip="172.168.1.26", out_key=flow, remote_ip="172.168.1.24"}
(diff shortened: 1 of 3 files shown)


insights/parsers/ovs_vsctl_show.py
```diff
@@ -1,90 +1,69 @@
...
-                        "interface": "tap293d29ec-d1",
-                        "port": "tap293d29ec-d1",
-                        "tag": "1",
-                        "type": "internal"
-                    }
-                ]
-            }
-        },
-        "ovs_version": "2.3.2",
-    }
```
bfahr: @PaulWay I realize this is verbose, but I feel like removing it makes it harder to understand the format of the output.  It takes some imagination to see this in the examples.  If you remove this could you show more detail in the examples sections to demonstrate the structure of the output?

PaulWay: A fair point.  It's always the problem with this kind of structured information - it's hard to see what you're getting :-)  I'm not convinced that the output structure is that unintuitively different from the original indented input that it's hard to understand what's available, but I will add more examples :-)

PaulWay: Let me know what you think of that now, @bfahr

insights/parsers/ovs_vsctl_show.py
```diff
@@ -95,8 +88,7 @@ def parse_content(self, content):
         bridge_dict = {}
         port_dict = {}
         for line in content[1:-1]:
-            line = line.strip()
-            prefix, _, value = line.partition(" ")
+            prefix, value = line.strip().split(None, 1)
```
bfahr: @PaulWay if the uuid is on the second line, which happens when there's a blank line at the top of the input then this line will fail when paring the uuid.  Since the parser assume that the first line is always the uuid and the last line is always the version would it make sense to ensure there are no leading or trailing blank lines first?

PaulWay: Yeah, I know.  This is a problem with the original parser.  It assumes that the UUID will be on the first line, the version will be on the last line, and doesn't handle '/bin/ovs: command or file not found'.  I didn't want to change the parser in documenting it, to be honest.  What would you prefer to do, @bfahr - should we leave it, fix it up in another PR or do it in this one?

bfahr: But you already changed the code, that is why this error came up when testing with rules.  I think it might be better if you take out your code changes here and limit to only documentation changes, and make the code changes in a separate PR.

xiangce:
As talked with @PaulWay , I will close this PR, since the 1.x branch will be deprecated in this week.

PR #850, opened 2017-12-22
Add parser to parse domain server log

zhangyi733:


insights/parsers/jboss_domain_server_log.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+
+from .. import LogFileOutput, parser
+from insights.specs import jboss_domain_server_log
+import re
+from datetime import time
+
+
+@parser(jboss_domain_server_log)
+class JbossDomainServerlog(LogFileOutput):
```
xiangce: Are there any other `Log` specs should be parsed for "Jboss"? Since these log Parsers are all inherited from "LogFileOutput" and are basically the same, we'd better put all the similar Parsers into one python module. (e.g. foreman_log and nova_log, etc). I mean, before adding other log Parsers, you can just make a base module for Jboss log files, e.g "jboss_log.py".

And BTW, would you mind use upcase "L" for the name "JbossDomainServerLog"?

zhangyi733:
Hi @xiangce, thank you for your good suggestion. I have updated the code according to your comments. Please help to review it again.

xiangce:
Since the spec of this parser depends on the pre-command (which is called so in 1.x)  `ps auxww`, but in fact, the pre-command doesn't work well in 1.x, we'd like to add these new parsers (include this one) only to the master branch.

@bfahr, how do you think?

I'll go ahead and merge this PR to the master branch at first. If you think it still necessary to add it to the 1.x branch too, please feel free to tell us. Thanks :)

PR #851, opened 2017-12-25
Update TomcatWebXml for 3.x

xiangce:
- update its specs as per the new foreach in #831
- move it to `tomcat_xml.py`, since there are several `xml` can be
  handle in this file
- update documentation


insights/parsers/tomcat_xml.py
```diff
@@ -0,0 +1,86 @@
...
+    The tomcat XML files are gotten from the directory specified in the java
+    commands
+
+"""
+import xml.etree.ElementTree as ET
+from .. import Parser, parser, LegacyItemAccess
+from insights.specs import tomcat_web_xml
+
+
+@parser(tomcat_web_xml)
```
zhangyi733: Hey @xiangce , as we discussed yesterday, we may need to create a base class for XMLParser, like https://github.com/RedHatInsights/insights-core/pull/855. I think we'd better re-update this PR when the base class is merged.

xiangce: Sure, Looking forward to the new XMLParser you created.

PR #852, opened 2017-12-26
Enhance parser nginx_conf_1x

wushiqinlou:
- Set value of "server" and "location" section as list.


insights/parsers/nginx_conf.py
```diff
@@ -291,11 +294,19 @@ def _listdepth_seven(self, li):
...
+                        dict_result[tmp_key] = tmp_dict[tmp_key]
                 if self._depth(sub_item) == 5:
                     tmp_dict = _listdepth_five(self, sub_item)
                     tmp_key = tmp_dict.keys()[0]
-                    dict_result[tmp_key] = self._handle_key_value(dict_result, tmp_key, tmp_dict[tmp_key])
+                    # As it is possible that multi key "server" and "location" exist, set the value of dict as list.
+                    if "server" in tmp_key or "location" in tmp_key:
+                        dict_result[tmp_key] = self._handle_key_value(dict_result, tmp_key, tmp_dict[tmp_key])
+                    else:
+                        dict_result[tmp_key] = tmp_dict[tmp_key]
```
xiangce: Since the codes to handle the special key are the same for above 3 modification, how above shift the checking of 'server' or 'location' into the `_handle_key_value`?

PR #855, opened 2017-12-29
Add xml file base class

zhangyi733:


zhangyi733:
Just a draft one, no need to review now.

zhangyi733:
@xiangce Please help to review the code logic and structure. I have not added detailed document in case there are any big changes needed to do.

insights/core/__init__.py
```diff
@@ -284,6 +285,36 @@ def get(self, item, default=None):
...
+def get_xmlns(self, root):
+        return root.tag.strip("{").split("}")[0] if all(c in root.tag for c in ["{", "}"]) else ""
+
+
+class XMLParser(Parser):
+    """
+    A parser class that reads XML files.  Base your own parser on this.
+    """
+
+    def parse_content(self, content):
```
xiangce: As a public `parse_content` method of the base class, better implement it a bit comment to avoid the sub-classes to re-write it.

Here is an example:
```python
    def _parse_dom(self, dom):
        return {}

        # ignore empty xml file
        self.dom = self.xmlns = None
        self.data = {}
        if len(content) > 3:
            root = ET.fromstring('\n'.join(content))
            self.dom = root
            self.data = self._parse_dom(root)
            self.xmlns = self.get_xmlns(root)
```

BTW, better add the empty content check to the base class to avoid duplicate it in sub-classes.

insights/core/__init__.py
```diff
@@ -284,6 +285,36 @@ def get(self, item, default=None):
...
+        root = ET.fromstring('\n'.join(content))
+        self.dom = root
+        self.xmlns = self.get_xmlns(root)
+
+    def get_xmlns(self, root):
+        return root.tag.strip("{").split("}")[0] if all(c in root.tag for c in ["{", "}"]) else ""
+
+    def get_elements(self, element):
```
xiangce: How about adding the default `namespaces` as the default argument:
`def get_elements(self, element, xmlns=self.xmlns)`?

insights/parsers/rhn_entitlement_cert_xml.py
```diff
@@ -1,9 +1,30 @@
 import xml.etree.ElementTree as ET
-from .. import Parser, parser, LegacyItemAccess
+from .. import XMLParser, parser, LegacyItemAccess
+
+
+def _parse_content(content):
```
xiangce: As the example of the base class I gave above, this method should be moved into the class (as a class method "self._parse_dom") and can be renamed to "_parser_dom". And just return the `rhn_cert`

insights/core/__init__.py
```diff
@@ -284,6 +285,36 @@ def get(self, item, default=None):
         return self.data.get(item, default)
 
 
+def get_xmlns(self, root):
+        return root.tag.strip("{").split("}")[0] if all(c in root.tag for c in ["{", "}"]) else ""
+
+
+class XMLParser(Parser):
```
xiangce: We can let the LegacyItemAccess mixin here on the base class:
` class XMLParser(Parser, LegacyItemAccess):`

insights/parsers/rhn_entitlement_cert_xml.py
```diff
@@ -64,22 +85,5 @@ class RHNCertConf(LegacyItemAccess, Parser):
...
-                        k: v for k, v in field.items() if k not in ('name', 'family')}
-                elif field.text:
-                    rhn_cert[field.get('name')] = field.text
-            # for all channel families
-            rhn_cert['channel-families'] = channel_familes
-            singature = rhntree.findall(".//rhn-cert-signature")
-            rhn_cert['signature'] = singature[0].text
-
-        self.data = rhn_cert
+        self.dom, self.data = _parse_content(content)
```
xiangce: And this `parser_content` can be omitted from all the base classes.

insights/parsers/tomcat_web_xml.py
```diff
@@ -17,19 +34,5 @@ def parse_content(self, content):
...
-        key_field = xmltree.findall(keyword_0)
-        key_field = key_field if key_field else xmltree.findall(keyword_1)
-        field_text = key_field[0].text if key_field else None
-
-        tmo_dict = {}
-        if field_text and field_text.isdigit():
-            tmo_dict = {TIMEOUT_KEYWORD: int(field_text)}
-        self.data = tmo_dict
+        self.dom, self.data = _parse_content(content)
+        self.xmlns = self.get_xmlns(self.dom)
```
xiangce: Take this sub-class as an example:

```python
import xml.etree.ElementTree as ET
from .. import XMLParser, parser

TIMEOUT_KEYWORD = 'session-timeout'


@parser('tomcat_web.xml')
class TomcatWebXml(XMLParser):
    def _parse_dom(self, dom):
        # default namespace
        xmlns_0 = 'http://java.sun.com/xml/ns/javaee'
        xmlns_1 = 'http://java.sun.com/xml/ns/j2ee'
        keyword_0 = '*{%s}%s' % (xmlns_0, TIMEOUT_KEYWORD)
        keyword_1 = '*{%s}%s' % (xmlns_1, TIMEOUT_KEYWORD)

        key_field = dom.findall(keyword_0)
        key_field = key_field if key_field else dom.findall(keyword_1)
        field_text = key_field[0].text if key_field else None

        tmo_dict = {}
        if field_text and field_text.isdigit():
            tmo_dict = {TIMEOUT_KEYWORD: int(field_text)}
        return tmo_dict

```

insights/core/__init__.py
```diff
@@ -284,6 +285,36 @@ def get(self, item, default=None):
...
+    """
+    A parser class that reads XML files.  Base your own parser on this.
+    """
+
+    def parse_content(self, content):
+        root = ET.fromstring('\n'.join(content))
+        self.dom = root
+        self.xmlns = self.get_xmlns(root)
+
+    def get_xmlns(self, root):
```
xiangce: This `root` should be removed

insights/core/__init__.py
xiangce: Sorry, _parse_dom(self, dom) should be _parser_dom(self). `dom` should not be passed in

zhangyi733:
@xiangce According to your comment and our discussion, I re-upload the code, please help to review it again.

bfahr:
@zhangyi733 @xiangce I'm not sure we want to be storing `ElementTree` and `Element` objects in the parser data.  @csams @jhjaggars and @kylape would you mind to take a look at this PR?

xiangce:
@bfahr thanks for heads up. The reason we want to keep the original ElementTree in the class is that we are not sure if the `get_elements` interface could satisfy the unknown requests from future plugins. But, of course, we are very much looking forward to your suggestions.

bfahr:
@zhangyi733 @xiangce I discussed with @jhjaggars and he convinced me that implementation is ok but recommended that the parser should also keep the raw XML as an attribute of the parser.

zhangyi733:
@bfahr and @xiangce I have added attribute raw to keep XML content. Please help to review it again.

insights/core/__init__.py
```diff
@@ -284,6 +285,127 @@ def get(self, item, default=None):
...
+        3
+        >>> len(xml_parser.get_elements(".//year/..[@name='Singapore']")[0]) # Nodes with name='Singapore' that have a 'year' child
+        1
+        >>> xml_parser.get_elements(".//*[@name='Singapore']/year")[0].text # 'year' nodes that are children of nodes with name='Singapore'
+        '2011'
+        >>> xml_parser.get_elements(".//neighbor[2]", "http://people.example.com")[0].get('name') # All 'neighbor' nodes that are the second child of their parent
+        'Switzerland'
+
+    Attributes:
+        dom (Element): Root element of parsed XML file
```
bfahr: Don't forget to add `raw` to the attributes.

zhangyi733:
Thank you @bfahr for the good catch. I have added the raw attribute into docstring. 
@xiangce If it looks good to you, please help to merge this PR.

PR #856, opened 2017-12-29
Update TomcatWebXml for 1.x

xiangce:
- move it to `tomcat_xml.py`, since there are several `xml` can be
  handle in this file
- update documentation


xiangce:
This is a temporary PR, I will update it after the base class made by @zhangyi733 is merged.

xiangce:
Since #855 is merged and I updated this PR as per it, @zhangyi733 @bfahr please take a look at this one when you are free, thanks.

insights/parsers/tomcat_xml.py
```diff
@@ -0,0 +1,76 @@
...
+    Examples:
+        >>> type(web_xml)
+        <class 'insights.parsers.tomcat_xml.TomcatWebXml'>
+        >>> web_xml.get('session-timeout') == 30
+        True
+    """
+
+    TIMEOUT_KEYWORD = ''
+
+    def _parse_dom(self):
```
bfahr: Before we start using this method, I think we should consider whether it should be a public method in the base class.  Currently its private so its not included in any of the documentation and a user would not know that its there and could be implemented in a child class.  Now that I understand how its used I think the base class should be changed to make this a public method, and the documentation should be clear about how to use it.

xiangce: Thanks @bfahr , that makes sense to us. We'll make it public.

PR #861, opened 2018-01-04
Delete the decorator on GrubConf that already fixed in plugins

xiangce:
- and update the corresponding combiner: selinux


insights/combiners/selinux.py
```diff
@@ -43,7 +40,7 @@
 
 
 @combiner(requires=[SEStatus, SelinuxConfig],
-          optional=[GrubConfig, Grub1Config, Grub2Config, Grub2EFIConfig])
+          optional=[Grub1Config, Grub2EFIConfig, Grub2Config, Grub2EFIConfig])
```
lonicerae: Should be `Grub1Config, Grub1EFIConfig, Grub2Config, Grub2EFIConfig` ?

xiangce: Oh... my :x: :bangbang:, will fix it

PR #867, opened 2018-01-08
Merge catalina log and update its specs

xiangce:
- Continue of #832, see #858


xiangce:
This is the re-submission of #832, @bfahr has reverted it since the dependent #831 was not merged that time. But it's already approved by @bfahr, and this one is just the re-submission of the original #832. I'll go ahead and merge it.

PR #868, opened 2018-01-08
Parser net snmp

vishwanathjadhav:
The ``IpSats`` class implements the parsing of ``/proc/net/snmp``
file, which contains TCP/IP stats of individual layer.


insights/parsers/snmp.py
```diff
@@ -0,0 +1,140 @@
...
+            'OutDestUnreachs', 'OutTimeExcds', 'OutParmProbs', 'OutSrcQuenchs', 'OutRedirects',
+            'OutEchos', 'OutEchoReps', 'OutTimestamps', 'OutTimestampReps', 'OutAddrMasks', 'OutAddrMaskReps']
+
+
+def parse_snmp_contents(content):
+    snmp_stats = {}
+    for line in content:
+        line_split = line.split()
+        if line_split[1].isdigit():
+            if "Ip:" in line_split:
```
bfahr: Have you considered putting these search strings into a list so that you could just iterate over them instead of having all the separate `if` tests?  You could use a tuple to associate the keys with the search string, like `('Ip:', IP_KEYS)`.

insights/parsers/snmp.py
```diff
@@ -0,0 +1,140 @@
+"""
+IpSats - file ``/proc/net/snmp``
```
bfahr: Should the class be `IpStats`?

insights/parsers/tests/test_snmp.py
```diff
@@ -0,0 +1,52 @@
...
+    assert snmp_stats["OutType3"] == 2
+
+    snmp_stats = stats.get_nstats("Udp")
+    assert snmp_stats["OutDatagrams"] == 95
+    assert snmp_stats["RcvbufErrors"] == 1
+    assert snmp_stats["NoPorts"] == 0
+
+    stats = IpSats(context_wrap(PROC_SNMP_NO))
+    snmp_stats = stats.get_nstats("Ip")
+    assert not snmp_stats
```
bfahr: If you are testing for `None` this should be `assert snmp_stats is None`.

insights/parsers/snmp.py
```diff
@@ -0,0 +1,112 @@
...
+    18905
+"""
+
+from .. import Parser, parser, LegacyItemAccess
+
+
+def get_snmp_stats(nkeys, line_split):
+    nstats = {}
+    nstats = dict(zip(nkeys, map(int, line_split)))
+    return nstats
```
xiangce: seems this function can be simplified enough to be removed:
line-97:
```python
snmp_stats[line_split[0].replace(":", "")] = dict(zip(data_id, map(int, line_split[1:])))
```

insights/parsers/snmp.py
```diff
@@ -0,0 +1,112 @@
...
+        data_id = []
+        for line in content:
+            line_split = line.split()
+            if line_split[1].isdigit():
+                snmp_stats[line_split[0].replace(":", "")] = get_snmp_stats(data_id, line_split[1:])
+            else:
+                data_id = line_split[1:]
+        self.data = snmp_stats
+
+    def get_nstats(self, layer):
```
xiangce: Since the default `get()` method of the `LegacyItemsAccess` already implemented the logic of this `get_nstats()`. Do you think it's worth being added here?

~~~
"stats.get('Ip')" is totally equal to "stats.get_nstats('Ip')"
~~~

bfahr:
@vishwanathjadhav your changes look very good, thanks!

PR #869, opened 2018-01-09
Add xml base parse class XMLParser

zhangyi733:
* Add xml base parse class
* Add testcase for base class
* Update existing XML related parsers to inherit from XMLParser


zhangyi733:
see https://github.com/RedHatInsights/insights-core/pull/855

insights/core/__init__.py
```diff
@@ -275,9 +284,9 @@ class XMLParser(LegacyItemAccess, Parser):
     Attributes:
         raw (str): raw XML content
         dom (Element): Root element of parsed XML file
-        data (dict): All required specific properties can be included in data.
         xmlns (str): The default XML namespace, an empty string when no
             namespace is declared.
+        data (dict): All required specific properties can be included in data.
```
xiangce: After rebasing to the master, this modification will be ignored.

insights/core/__init__.py
```diff
@@ -290,17 +299,18 @@ def _parse_dom(self):
     def parse_content(self, content):
         """
         All child classes inherit this function to parse XML file automatically.
-        It will call the private function :attr:`_parse_dom` by default to parser all
-        necessary data to :attr:`data`.
+        It will call the private function :attr:`_parse_dom` by default to
+        parser all necessary data to :attr:`data` and the :attr:`xmlns` (the
+        default namespace) is ready for this function.
```
xiangce: The same to above.

insights/core/__init__.py
```diff
@@ -6,12 +6,21 @@
...
+import sys
+# Since XPath expression is not supported by the ElementTree in Python 2.6,
+# import insights.contrib.ElementTree when running python is prior to 2.6 for compatibility.
+# Script insights.contrib.ElementTree is the same with xml.etree.ElementTree in Python 2.7.14
+# Otherwise, import xml.etree.ElementTree instead.
+if sys.version_info[0] == 2 and sys.version_info[1] <= 6:
+    import insights.contrib.ElementTree as ET
+else:
+    import xml.etree.ElementTree as ET
+
```
xiangce: Hey @bfahr, @kylape or @csams, Would you please take a look at this modification?

The two files ElementTree.py and ElementPath.py are copied from the latest "/usr/lib64/python2.7/xml/etree/" module.

xiangce: @bfahr @kylape @csams - If this kind of python version checking is OK, could we also carry it out on the `import` of `ConfigParser` and `RawConfigParser`?

bfahr: @xiangce would you mind to clarify the reason for including these from 2.7?  Is it because implementation changes for `ElementTree` and `ElementPath` between 2.6 and 2.7?  https://docs.python.org/2/whatsnew/2.7.html#updated-module-elementtree-1-3

insights/core/__init__.py
```diff
@@ -229,6 +239,124 @@ def get(self, item, default=None):
...
+        'Switzerland'
+    Attributes:
+        raw (str): raw XML content
+        dom (Element): Root element of parsed XML file
+        xmlns (str): The default XML namespace, an empty string when no
+            namespace is declared.
+        data (dict): All required specific properties can be included in data.
+    """
+
+    def _parse_dom(self):
```
bfahr: @zhangyi733 see my comment in #856, since this can be implemented in a child class to change functionality it should either be a public method so that it will show up in the documentation or at least described in the class docstring so that users will be aware of its functionality.  Either way, the documentation should be clear as to how it is used.

insights/core/__init__.py
xiangce: @bfahr - Sorry for missing the explanation. It's because that the `get_elements` used the XPath feature which is only available in ElementTree-1.3.  When testing on python-2.6, it reported errors like: `expected path separator ([)`, so we try to backport the ElementTree-1.3 for python-2.6 by copying them to the `contrib` folder. (we got that from https://stackoverflow.com/questions/222375/elementtree-xpath-select-element-based-on-attribute)

For `ElementPath`, it's because that the `ElementTree` depends on it.

HTH, Thanks.

bfahr: Thanks for the explanation @xiangce.

PR #871, opened 2018-01-09
[doc] Fixed missing 'self' as parameter in parse_content()

psachin:
In section "Creating the Initial Parser Files", we don't pass `self`
as a first parameter to `parse_content()` which is OK until test
method `test_sshd_config()` just have `pass` but running `py.test -k
test_ssh` will soon fail if user includes few assertions with error,

```
E       TypeError: parse_content() takes exactly 1 argument (2 given)
```

An experience programmer will figure this out but may get difficult
for a newbie. It is better to include `self` from start.


xiangce:
@psachin @bfahr - I cherry-picked #875 to the 3.x branch. will merge them then.

PR #883, opened 2018-01-11
Parser net snmp ipv6

vishwanathjadhav:
Class TcpIpStatsIPV6 Parser for /proc/net/snmp6 file to get ipv6 stats of TCP/IP layer protocols


insights/parsers/tests/test_snmp.py
```diff
@@ -50,3 +152,15 @@ def test_snmp():
     stats = TcpIpStats(context_wrap(PROC_SNMP_NO))
     snmp_stats = stats.get("Ip")
     assert snmp_stats is None
+
+    stats = TcpIpStatsIPV6(context_wrap(PROC_SNMP6))
```
xiangce: Better separate the test for TcpIpStatsIPV6 to another test function e.g. `test_snmp6`?

PR #826, opened 2017-12-15
Hammer task list parser for 1.x branch

PaulWay:
The output of the `hammer task list` command is useful for diagnosing problems with Satellite running tasks.

Full documentation and test coverage included.

PR #896, opened 2018-01-15
Hammer task list parser for master branch

PaulWay:
This is the master branch version of #826


insights/config/specs.py
```diff
@@ -137,6 +137,7 @@
     "grub_config_perms"         : CommandSpec("/bin/ls -l /boot/grub2/grub.cfg"),  # only RHEL7 and updwards
     "grub1_config_perms"        : CommandSpec("/bin/ls -l /boot/grub/grub.conf"),  # RHEL6
     "hammer_ping"               : CommandSpec("/usr/bin/hammer ping"),
+    "hammer_task_list"          : CommandSpec("/usr/bin/hammer --csv task list"),
```
xiangce: @csams - Should it be added to the "InsightsArchiveSpecs" to keep synchronize?

xiangce: @PaulWay - As Chris.Sams's suggestion, for simple_command we'd better add it to the "InsightsArchiveSpecs" in "insights/specs_insights_archive.py"

PaulWay: I think I've done this correctly by adding the '/usr/bin/hammer --csv task list' command to `insights/spec_default.py`.  How does that look @xiangce

xiangce: For 3.x, the `insights/config/specs.py` will not be supported. To keep compatible with the 1.x and parse the 1.x's archives, we should add a corresponding `simple_file` to the `insights/specs_insights_archive.py` for each `simple_command` in `insights/specs_default.py`.

So, we should add it's simple_file to the `insights/specs_insights_archive.py`. After rerun the `pip install -e .[develop]` for 3.x, you'll have a "bin/mangle" script that can converts simple_command strings to simple_file strings.

PR #897, opened 2018-01-15
Use datasource instead of `|` for some command specs

xiangce:


xiangce:
@csams - Since I modified some command spec in this PR, is it necessary to update the corresponding collected-specs-names in `InsightsArchiveSpecs` defined in "insights/specs_insights_archive.py"?

I had thought these collected-specs-names were generated (and mapped between the `core` and `client`) automatically. Why you add such a class then? Should we manually edit/add each corresponding collected-specs-name when updating/adding new specs? There will be some risks once someone of us make some typos in these names and would cause the framework cannot find the special collected spec file.

zhangyi733:
Hi @csams, after going through the guide https://github.com/RedHatInsights/insights-core/blob/master/notebooks/Datasource%20Registry.ipynb, I know some basic knowledge about how to define/modify datasource. But to be honest, after discussion with @xiangce, we don't quite understand why we do this modification and what we can benefit from it. Could you please tell us a little more about the background? Is it prepared to make source be able to utilize parser/combiner result or for other purpose?

csams:
Hey @xiangce, thanks for the questions! I'll answer the last one first.

Parsers in 3.x should depend directly on the datasource they parse. That datasource, until now, was the function that actually had the code for collecting data on a target machine. That's okay if we're using 3.x everywhere, including on customer machines, because the files in archives generated by 3.x are self-describing. Each document contains the contents produced by its datasource along with several attributes that let us process it directly without searching through the archive looking for files.

However, we still have to be able to process archives generated by the old insights clients, sos reports, JDR reports, etc., which are just bags of raw files. To allow parsers to continue depending on just one thing but also be able to adjust the datasource implementation depending on context, we developed the registry. We could achieve something similar with the "first_of" helper datasource, but that wouldn't allow others to provide their own extensions to the system without modifying the source code. 

Now to the first question. I'm sorry for missing that part in my initial email. The idea of having a separate implementation for the old insights archives is new, and yes, we'd like to keep those up to date as well. However, all that has to be added there are the locations of simple commands. The file based datasources will just work on both the client machine and for old insights archives, so they don't have to be specified twice.

xiangce:
@bfahr - I have added the "simple_file" for the updated `httpd_pid`, please take a look at this PR when you are free. Thanks.

xiangce:
OK, Thanks @bfahr .

PR #901, opened 2018-01-17
take into account rotated logs of directory server

germanparente:
The logs in directory server are of this shape:

/var/log/dirsrv/.*/access
/var/log/dirsrv/.*/access.YYYYMMDD-HHMMSS
/var/log/dirsrv/.*/access.rotationinfo
/var/log/dirsrv/.*/errors
/var/log/dirsrv/.*/errors.YYYYMMDD-HHMMSS
/var/log/dirsrv/.*/errors.rotationinfo

when we parse errors or access logs, we want to use also the rotated ones. Skip the *.rorationinfo files that are irrelevant.


insights/config/specs.py
```diff
@@ -66,8 +66,8 @@
     "df_-li"                    : CommandSpec("/bin/df -li"),
     "dig"                       : CommandSpec("/usr/bin/dig +dnssec . DNSKEY"),
     "dirsrv"                    : SimpleFileSpec("etc/sysconfig/dirsrv"),
-    "dirsrv_access"             : PatternSpec("var/log/dirsrv/.*/access"),
-    "dirsrv_errors"             : PatternSpec("var/log/dirsrv/.*/errors"),
+    "dirsrv_access"             : PatternSpec("var/log/dirsrv/.*/(access|access\.2.*)"),
+    "dirsrv_errors"             : PatternSpec("var/log/dirsrv/.*/(errors|errors\.2.*)"),
```
xiangce: This modification looks good, but I'm afraid it doesn't work with the current client 1.x. For details, see https://mojo.redhat.com/groups/red-hat-insights/blog/2016/09/13/recent-framework-updates-to-falafel-and-ceeph

It's a very old issue, but unfortunately not fixed yet. However, it's OK to add it to the 1.x branch. @bfahr do you think so?

germanparente: Hi,

thanks a lot for your comment. It has worked fine for me when I have tested it. I am not using the right client ?

thanks a lot and regards,

German

bfahr: @xiangce as long as the modified specs will collect at least the same files as the previous specs then its OK to add them.  Just don't want make the collection worse.

xiangce: Thanks @bfahr

PR #902, opened 2018-01-17
take into account rotated logs of directory server

germanparente:
The logs in directory server are of this shape:

/var/log/dirsrv/./access
/var/log/dirsrv/./access.YYYYMMDD-HHMMSS
/var/log/dirsrv/./access.rotationinfo
/var/log/dirsrv/./errors
/var/log/dirsrv/./errors.YYYYMMDD-HHMMSS
/var/log/dirsrv/./errors.rotationinfo

when we parse errors or access logs, we want to use also the rotated ones. Skip the *.rorationinfo files that are irrelevant.


insights/config/specs.py
```diff
@@ -66,8 +66,8 @@
     "df_-li"                    : CommandSpec("/bin/df -li"),
     "dig"                       : CommandSpec("/usr/bin/dig +dnssec . DNSKEY"),
     "dirsrv"                    : SimpleFileSpec("etc/sysconfig/dirsrv"),
-    "dirsrv_access"             : PatternSpec("var/log/dirsrv/.*/access"),
-    "dirsrv_errors"             : PatternSpec("var/log/dirsrv/.*/errors"),
+    "dirsrv_access"             : PatternSpec("var/log/dirsrv/.*/(access|access\.2.*)"),
+    "dirsrv_errors"             : PatternSpec("var/log/dirsrv/.*/(errors|errors\.2.*)"),
```
xiangce: Would you please to update the corresponding items in `insights/specs_default.py` for the master (3.x) branch? Thanks.

germanparente: HI,

thanks a lot. When is this file used so as I could test it. Or is it used first before parsing config/specs.py ?

thanks.

German

xiangce: The '3.x' contains many new features but it is not released yet. The "insights/specs_default.py" and "insights/specs.py" are  the new styles spec files. And to keep compatibility with the current '1.x', we need to edit both the new style in "insights/specs_default.py" and "insights/specs.py" and the old one in "insights/config/specs.py".

You can test the updated spec written in new style by referring to corresponding step in https://github.com/RedHatInsights/insights-core/blob/master/notebooks/Datasource%20Registry.ipynb

PaulWay: The new specs use `glob()`, so they should be written as:

```python
     "dirsrv_access"             : PatternSpec("var/log/dirsrv/*/access", "var/log/dirsrv/*/access.2*)"),
     "dirsrv_errors"             : PatternSpec("var/log/dirsrv/.*/errors", "var/log/dirsrv/*/errors.2*)"),
```

bfahr:
@germanparente insights/config/specs.py is gone, and the new specs have been moved to the package insights/specs/ so you need to make your changes to insights/specs/__init__.py and insights/specs/default.py.  It might be easier to close this PR, rebase and submit a new PR.

germanparente:
Hi,

thanks a lot for clarifying.

Best regards,

German.

PR #905, opened 2018-01-23
Netstat combiner master

vishwanathjadhav:
Added combiner in master (3.x) to keep it in sync with (1.x)
Previous reference #903


insights/combiners/netstat.py
```diff
@@ -0,0 +1,96 @@
...
+Link Layer stats
+================
+
+Combiner for link layer stats. It uses the results of
+the ``netstat -i`` parser and the ``ip -s link`` parser to determine the
+network stats of link layer. ``ip -s link`` is the preferred source
+of data and return object which can be used to access ``group_by_iface``.
+
+Examples:
+    >>> nstat = shared[NetworkStats]
```
bfahr: @vishwanathjadhav since `shared` is no longer used in 3.x please take a look at #890 to see an example of how to demonstrate the object in 3.x without using `shared`.

PR #906, opened 2018-01-23
Add systemd timer scheduling

brantleyr:


kylape:
Since we're already using a class here, why not take advantage of class features and write two implementations for `InsightsSchedule` -- one from cron and another for systemd?

Issue #904, opened 2018-01-22
Modify behavior of --support


gravitypriest:
Use `--support` to collect the multiple logs (client log + systemd logs) into one tarfile for upload

YakovLittle:
>insights-client --vers
Client: 3.0.3-9
Core: 3.0.35-1

>insights-client --support
Fatal error
Traceback (most recent call last):
  File "/var/lib/insights/last_stable.egg/insights/client/phase/v1.py", line 34, in _f
    func(client, config)
  File "/var/lib/insights/last_stable.egg/insights/client/phase/v1.py", line 103, in pre_update
    support = InsightsSupport()
TypeError: __init__() takes exactly 2 arguments (1 given)

Glutexo:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1607803

Glutexo:
@gravitypriest Is this issue still relevant? If so, would you please tell me, what logs exactly are going to be uploaded? Like, the full output of the `journalctl` command or the commands mentioned in the related [Bugzilla ticket](https://bugzilla.redhat.com/show_bug.cgi?id=1607803)?

```
systemctl cat insights-client.timer
systemctl cat insights-client.service
systemctl status insights-client.timer
systemctl status insights-client.service
```

Could you please take a look at the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1607803) and tell me whether it describes the same problem as this GitHub issue? To me it looks like that this issue describes a RFE that the _systemd_ logs should be added, but the BZ on the other hand looks like it is already there, but not working. Thanks!

PR #907, opened 2018-01-23
change --support behavior, save all logs to a tar

gravitypriest:


kylape:
This resolves https://github.com/RedHatInsights/insights-core/issues/904

gravitypriest:
Found a bug with the registration check change, closing until I fix it.

kylape:
A few questions:

- Is there a reason the `systemctl` commands can't be combined with the other set of command being executed in `_support_diag_dump`?
- Should we try not to execute the `systemtl` commands on RHEL 6?  It would obviously take more code to detect, but it would clean up the support output a bit.
- Is there really much value of creating a tar of the logs?  It seems trivial for a user to tar up the logs in `/var/log/insights-client`, but on the other hand I guess it doesn't hurt to do it for them.

gravitypriest:
- Good catch, I went through a few iterations when I was refactoring and moving them to the rest of the cmds seems to make sense.
- If the concern is log output I'll just remove the exception log dump when it fails to run.
- My thinking was, this one file could end up being the the only thing they would need to upload when doing a support case.

kylape:
I think it's no big deal to just run systemctl and print that it can't be found on rhel 6.

Eh, sosreport does the same thing as what you're doing here, so let's just leave it.

We can merge once the build passes.

kylape:
```
./insights/client/support.py:13:1: F401 'utilities.run_command_get_output' imported but unused
./insights/client/support.py:139:31: F841 local variable 'o' is assigned to but never used
./insights/client/support.py:142:72: F821 undefined name 'e'
```

Issue #827, opened 2017-12-15
`FileListing` ignores `ls` listings that don't contain the name of the directory.


jsvob:
If an `ls` listing doesn't contain the name of the directory before the individual files, it skips the files.

https://github.com/RedHatInsights/insights-core/blob/f89748ffabc7a4b24159c9ccfdee2067bfab1c47/insights/core/__init__.py#L1046

The test shows that this is indeed intentional:
https://github.com/RedHatInsights/insights-core/blob/master/insights/tests/test_file_listing.py#L60

A listing of a single directory doesn't print the header with the directory name:

```
$ ls -la /var/www
total 0
drwxr-xr-x.  4 root root  33 Dec 15 08:12 .
drwxr-xr-x. 20 root root 278 Dec 15 08:12 ..
drwxr-xr-x.  2 root root   6 Oct  3 09:37 cgi-bin
drwxr-xr-x.  2 root root   6 Oct  3 09:37 html
```

When multiple directories are specified, or when the `-R` option is used, the directory names are printed:
```
ls -laR /var/www
/var/www:
total 0
drwxr-xr-x.  4 root root  33 Dec 15 08:12 .
drwxr-xr-x. 20 root root 278 Dec 15 08:12 ..
drwxr-xr-x.  2 root root   6 Oct  3 09:37 cgi-bin
drwxr-xr-x.  2 root root   6 Oct  3 09:37 html

/var/www/cgi-bin:
total 0
drwxr-xr-x. 2 root root  6 Oct  3 09:37 .
drwxr-xr-x. 4 root root 33 Dec 15 08:12 ..

/var/www/html:
total 0
drwxr-xr-x. 2 root root  6 Oct  3 09:37 .
drwxr-xr-x. 4 root root 33 Dec 15 08:12 ..
```

In order to work around the limitation, there are currently three options:

* using a spec like `ls -la /var/www /var/www` - specify the directory twice, it is printed twice and with the directory name
* listing e.g. /dev/null - `ls -la /var/www /dev/null`
* using the `-R` switch - `ls -laR /var/www`

The first two are inelegant. The third one cannot be used in some scenarios, like in the example - if a rule needs only first-level information about the contents of /var/www, I don't want to go recursively, because there *might* be millions of files (performance reasons) and because there might be data that the rule doesn't need and will just needlessly gather potentially sensitive information only to discard it later.

So the issue is that `FileListing` should support simple one-directory listings that don't contain the directory name.

[...]

PR #910, opened 2018-01-30
FileListing class to handle single directory listings for 1.x branch

PaulWay:
Jakub Svoboda identified in issue #827 that the FileListing parser wasn't able to read single directory listings, because a single directory listing doesn't include the directory name in the output and only if the directory name is read will anything else be captured.

This patch fixes that by checking for a path on the command line or in the file path in sosreport format - e.g. 'ls_-la_.etc.pki.tls', with '.' substituted for '/'.  That dotted suffix, converted back to path format, is used as a 'first guess' of the directory.  If a directory heading is found in the output that is used in preference to the first guess.

Tests have been added to check this functionality.


jsvob:
@PaulWay Hi, I like it. What if we have a spec like `ls -la /etc/yum.repos.d`? The path would get converted to /etc/yum/repos/d IIUC. I would be OK with a solution to that through just documenting that it is a tradeoff and that a consumer of such a `ls` has to be aware of it or has to use `ls -la /dev/null /etc/yum.repos.d`, or having some sort of method that does search and replace in the parsed data to correct it back.

PaulWay:
Hi @jsvob - good point!  Obviously `/etc/yum.repos.d` is captured already in the `LsEtc` parser, but I take your point that a directory listing of a single path that contains dots will be incorrectly back-translated.  In general if you want a directory listed exactly, I think we shouldn't rely on the back-translation - at this stage the best option is to do a recursive listing of the parent.  If you know the directory that you're listing, then you can find that listing in the parser output fairly easily - I don't think we need to concern ourselves with whether its path is exactly as originally given.

In the future I think we should aim to not use `ls` at all and instead have a datasource that captures directory listings in an efficient, safe and unambiguous manner.

PaulWay:
I updated the code to reverse the ' ' to '_' translation (just in case someone's listing a directory with spaces in its name) and with a note in the documentation about the path translations.

I don't know why I'm getting a build error on `sphinx-build -W -b html -qa -E docs docs/_build/html` saying that:
```
Warning, treated as error:
/home/travis/build/RedHatInsights/insights-core/insights/parsers/xfs_info.py:docstring of insights.parsers.xfs_info.XFSInfo.parse_content:3:Unexpected indentation.
```
Because when I run the same command from the command line, I get no error.  Can someone else check this for me?

xiangce:
@PaulWay - I think it's caused by the class XFSInfo missed the docstring entry, then the sphnix works weird. I have fixed it in https://github.com/RedHatInsights/insights-core/pull/941, please take a look at #941. And rebase this PR after #941 is OK and merged.

PaulWay:
Hi @xiangce - thanks for that fix - I've re-submitted this and hopefully it should be good to go now.

PaulWay:
@xiangce Should be all good to go now.

jsvob:
Hi @PaulWay, thanks for documenting that. I agree with your arguments.

Just a note that one might not always want a recursive listing - in cases where the directory might be unpredictably excessively large and only one level of listing is needed anyway.

Issue #912, opened 2018-01-30
PSAuxcww need to skip warning line


chenlizhong:
Exception:

ps.py, line 222, in parse_content

output of ps auxcww
~~~
911 your 131072x1 screen size is bogus. expect trouble
USER        PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root          1  0.0  0.1 193704  5088 ?        Ss   Jan10   6:52 systemd
root          2  0.0  0.0      0     0 ?        S    Jan10   0:02 kthreadd
root          3  0.0  0.0      0     0 ?        S    Jan10   0:29 ksoftirqd/0
root          5  0.0  0.0      0     0 ?        S<   Jan10   0:00 kworker/0:0H
~~~

output of ps auxww
~~~
your 131072x1 screen size is bogus. expect trouble
USER        PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root          1  0.0  0.1 193704  5088 ?        Ss   Jan10   6:52 /usr/lib/systemd/systemd --switched-root --system --deserialize 21
root          2  0.0  0.0      0     0 ?        S    Jan10   0:02 [kthreadd]
root          3  0.0  0.0      0     0 ?        S    Jan10   0:29 [ksoftirqd/0]
root          5  0.0  0.0      0     0 ?        S<   Jan10   0:00 [kworker/0:0H]
root          7  0.0  0.0      0     0 ?        S    Jan10   0:00 [migration/0]
~~~

PaulWay:
The PsAuxcww parser is deprecated in v3, so I don't think it's worth patching that.  The PsAuxww parser  should parse this correctly because it looks for the line starting with 'USER' and treats that as the first line of the table.  I'll add tests for handling this output correctly.

PaulWay:
These tests have been merged - this issue can be closed.

PR #914, opened 2018-01-30
Update PsAuxcww parser to skip heading warn message

chenlizhong:
#913


bfahr:
@chenlizhong would you mind to provide more information about the purpose of this PR.  Is it to fix an existing reported issue, if so could you provide a reference to the issue report? Thanks.

chenlizhong:
@bfahr 
I open the issue  https://github.com/RedHatInsights/insights-core/issues/912
If you need more info, let me know.

bfahr:
@chenlizhong thank you for the explanation and thank you for adding the reference to the issue, that is sufficient.

chenlizhong:
@bfahr 
You are experienced!!!
Just test the rules. There is one test data with "UER" no "USER".  I just submit PR to fix it.

xiangce:
3.x is already OK, do not need this patch.

PR #917, opened 2018-01-30
FileListing class to handle single directory listings for master branch

PaulWay:
Jakub Svoboda identified in issue #827 that the FileListing parser wasn't able to read single directory listings, because a single directory listing doesn't include the directory name in the output and only if the directory name is read will anything else be captured.

This patch fixes that by checking for a path on the command line or in the file path in sosreport format - e.g. 'ls_-la_.etc.pki.tls', with '.' substituted for '/'. That dotted suffix, converted back to path format, is used as a 'first guess' of the directory. If a directory heading is found in the output that is used in preference to the first guess.

Tests have been added to check this functionality.


jsvob:
Discussion in https://github.com/RedHatInsights/insights-core/pull/910

insights/core/__init__.py
```diff
@@ -1031,6 +1031,15 @@ class FileListing(Parser):
       directory, in the order found in the listing
     * total blocks allocated to all the entities in this directory
 
+    Note: For listings that only contain one directory, ``ls`` does not
```
xiangce: Would you mind change this `Note` format to `.. note::` (See http://www.sphinx-doc.org/en/stable/markup/para.html), sorry for the late comment.

PaulWay: Ah, of course.  Hold on, I'll do that in both requests.

PR #918, opened 2018-01-31
VDSM version 3 & 4 log file parser

psachin:
The present class `VDSMLog` don't really parse logs as per format
specified in `/etc/vdsm/logger.conf`(vdsm-4.19) which is as below,

```
[formatter_long]
format: %(asctime)s %(levelname)-5s (%(threadName)s) [%(name)s] %(message)s (%(module)s:%(lineno)d)
```

I was not able to find logs with below format even when DEBUG is enabled,
```
Thread-60::DEBUG::2015-05-08 18:01:03,071::blockSD::600::Storage.Misc.excCmd::(getReadDelay) '/bin/dd if=/dev/5a30691d-4fae-4023-ae96-50704f6b253c/metadata iflag=direct of=/dev/null bs=4096 count=1' (cwd None)
Thread-60::DEBUG::2015-05-08 18:01:03,090::blockSD::600::Storage.Misc.excCmd::(getReadDelay) SUCCESS: <err> = '1+0 records in\\n1+0 records out\\n4096 bytes (4.1 kB) copied, 0.00038933 s, 10.5 MB/s\\n'; <rc> = 0
Thread-65::DEBUG::2015-05-08 18:01:04,835::blockSD::600::Storage.Misc.excCmd::(getReadDelay) '/bin/dd if=/dev/e70cce65-0d02-4da4-8781-6aeeef5c86ff/metadata iflag=direct of=/dev/null bs=4096 count=1' (cwd None)
Thread-65::DEBUG::2015-05-08 18:01:04,857::blockSD::600::Storage.Misc.excCmd::(getReadDelay) SUCCESS: <err> = '1+0 records in\\n1+0 records out\\n4096 bytes (4.1 kB) copied, 0.000157193 s, 26.1 MB/s\\n'; <rc> = 0
Thread-4662::DEBUG::2015-05-08 18:01:05,560::task::595::TaskManager.Task::(_updateState) Task=`9a7948f6-b6d9-42c2-b91f-7e0346dfc1d6`::moving from state init -> state preparing
```

Current logs are similar as below ,
```
2018-01-31 09:20:02,658+0530 DEBUG (check/loop) [storage.check] FINISH check u'/rhev/data-center/mnt/dell-r530-17.gsslab.pnq2.redhat.com:_home_exports_iso/be94b389-000f-4487-baf4-35ded403e579/dom_md/metadata' (rc=0, elapsed=0.05) (check:328)
2018-01-31 09:20:02,955+0530 DEBUG (mailbox-spm) [storage.Misc.excCmd] /usr/bin/taskset --cpu-list 0-15 dd if=/rhev/data-center/5a5e5d2a-0076-00b1-00b8-00000000006f/mastersd/dom_md/inbox iflag=direct,fullblock count=1 bs=1024000 (cwd None) (commands:69)
2018-01-31 09:20:03,006+0530 DEBUG (mailbox-spm) [storage.Misc.excCmd] SUCCESS: <err> = '1+0 records in\n1+0 records out\n1024000 bytes (1.0 MB) copied, 0.0350045 s, 29.3 MB/s\n'; <rc> = 0 (commands:93)
2018-01-31 09:20:05,016+0530 DEBUG (mailbox-spm) [storage.Misc.excCmd] /usr/bin/taskset --cpu-list 0-15 dd if=/rhev/data-center/5a5e5d2a-0076-00b1-00b8-00000000006f/mastersd/dom_md/inbox iflag=direct,fullblock count=1 bs=1024000 (cwd None) (commands:69)
2018-01-31 09:20:05,040+0530 DEBUG (mailbox-spm) [storage.Misc.excCmd] SUCCESS: <err> = '1+0 records in\n1+0 records out\n1024000 bytes (1.0 MB) copied, 0.0171592 s, 59.7 MB/s\n'; <rc> = 0 (commands:93)
```

The new class `VDSMInfoLog` should parse above logs format and return
a dict with keys matching formatter.


bfahr:
@psachin please don't include links that are not publically accessible such as the trello link in your comment.

bfahr:
@psachin if you've confirmed that the existing parser does not correctly parse the vdsm logs then it would be better to replace the existing parser instead of adding a new parser.

psachin:
@bfahr Thanks, I removed trello link from commit message.

I was thinking to check with @PaulWay before implementing a new parser logic. As @PaulWay wrote initial logic, we need to discuss if that can also be merge so that the parser is able to parse both old(the logs used for testing is from 2015) and new log formats

PaulWay:
Sorry, but I didn't write the initial logic - you'll have to see @kylape for that.  From the old falafel repository:

```commit 1db30bc1f9178e1b7193ca387ff7c4dbcc7e6d42
Author: Kyle Lape <klape@redhat.com>
Date:   Fri Jul 15 13:50:00 2016 -0500

    Initial commit
```

All I did was document it.

PaulWay:
We've talked with Kyle and he didn't write it either - just did the merge into the old repository.

What's more important than archaeology is that I've talked to Germano Veit Michel, an SME in APAC SD, who tells me that the current format is version 3 of VDSM and the new format is version 4.  I think the one parser should read either.

It shouldn't be too difficult - I'd do it by regexes but you could just check if the first character in the line is '2' and the ninth is ' ' and that'll probably recognise most of v4 format.

What I'd also recommend here is to try and make the fields captured in v4 match those found in v3 as closely as possible.  That way a rule can apply well to either format.  If you're populating a dictionary, you may like to fill in blank fields for things that are only in one format, so that rule writers don't have to deal with checking if the line dictionary has the fields its expecting.

I'd also recommend capturing the backtraces, at least in a separate `errors` or `backtraces` property.  They're probably things people want to check for too.

insights/parsers/vdsm_log.py
```diff
@@ -29,66 +28,108 @@ class VDSMLog(LogFileOutput):
...
+                    fields['message'] = msg
+                # Did we get a timestamp in there?
+                if 'timestamp' in fields:
+                    # Try to convert the datetime if possible
+                    try:
+                        fields['asctime'] = datetime.strptime(
+                            fields['timestamp'], time_format)
+                        del fields['timestamp']
+                    except:
+                        pass
```
bfahr: @psachin would it make sense to store the raw value of `timestamp` in `fields[asctime]` or in `fields[timestamp]` if the conversion fails?  Would that information still be useful?

psachin: Hi @bfahr 

I decided to use `fields[asctime]`(datetime) compared to `field[timestamp]`(raw) because `get_after()` takes `datetime`. This also makes both version 3 and 4 output same which makes job easy for rule writer. This is the reason I'm deliberately deleting that key-value after conversion. The conversion will fail only if `field[timestamp]` does not comply `time_format` which is rare.

`field[timestamp]` might be useful to show user when exactly the error occurred but it will be more appropriate to convert `datetime` to human readable format. Apart from this I think having an additional `field[timestamp]` is of no use. It is quite easy to include raw time in future if we feel so. 

Let me know your opinion.

bfahr: @psachin that sounds good to me.

insights/parsers/vdsm_log.py
```diff
@@ -29,66 +28,108 @@ class VDSMLog(LogFileOutput):
...
-                # Try to convert the datetime if possible
-                try:
-                    fields['datetime'] = datetime.strptime(
-                        fields['timestamp'][:-4],  # skip milliseconds
-                        self.time_format
-                    )
-                except:
-                    pass
-            yield fields
+            if re_datetime_obj.match(line):
```
lonicerae: Per our discussion on using the regex, please double confirm the input data when doing the parsing job. And please add more description on the regex rule.

xiangce:
@psachin - There is an error while building the document and caused the test of Shippable and Travis failed, you can run command `(cd docs && make html)` to reproduce it in your local.

xiangce:
The doc building error is caused by a document issue in XFSInfo, I have fixed it in #941. Please rebase this PR and after #941 is merged.

xiangce:
@psachin - I go ahead and merge this PR now, please resubmit the same updates to the master branch as well thanks.

PR #919, opened 2018-02-01
Parser for logfile var log up2date 1.x

TripleJQK:
@xiangce need review on new parser fro /var/log/up2date log file


insights/config/specs.py
```diff
@@ -437,6 +437,7 @@
     "uname"                     : First([CommandSpec("/bin/uname -a"),
                                     SimpleFileSpec("uname")]),
     "up2date"                   : SimpleFileSpec("etc/sysconfig/rhn/up2date"),
+    "up2date.log"               : SimpleFileSpec("var/log/up2date", large_content=True),
```
xiangce: Better use the same spec name in 1.x and 3.x for the newly added spec.

insights/parsers/up2date_log.py
```diff
@@ -0,0 +1,7 @@
+from .. import LogFileOutput, parser
+
+
+@parser('up2date.log')
+class Up2dateLog(LogFileOutput):
```
xiangce: `Up2DateLog`?

insights/parsers/up2date_log.py
```diff
@@ -0,0 +1,7 @@
...
+    """Class for parsing the log file: ``/var/log/up2date`` """
+    pass
```
xiangce: The new sub-classes inherits from LogFileOutput, you should check if the default `time_format` can cover the actual time format of the new log-file. The default time format is 

`time_format = '%Y-%m-%d %H:%M:%S'`

It does not match with the time stamp string in up2date. Please modify it correspondingly and use `get_after` in test to verify it.

insights/parsers/up2date_log.py
TripleJQK: the log file name is /var/log/up2date,  as the up2date is one word, so Up2dateLog

insights/config/specs.py
TripleJQK: I check the parser for rhsm.log in both 1.x and 3.x:
In 1.x, the spec is "rhsm.log"
In 3.x, the spec is rhsm_log with alias rhsm.log(using Specs.rhsm_log in parser file. If rhsm.log, Specs.rhsm.log seems a little werid).

So use up2date_log in both 1.x and 3.x will be better?

insights/parsers/up2date_log.py
TripleJQK: I will modify this

insights/config/specs.py
TripleJQK: use up2date_log instead of up2date.log as same as 3.x

insights/parsers/up2date_log.py
TripleJQK: not change this as my last comment

insights/parsers/up2date_log.py
TripleJQK: done

xiangce:
See comments: https://github.com/RedHatInsights/insights-core/pull/920#discussion_r165290734

TripleJQK:
@xiangce modified as your request, please review when free

PR #920, opened 2018-02-01
Parser for logfile var log up2date 3.x

TripleJQK:


insights/specs.py
```diff
@@ -355,6 +355,7 @@ class Specs(SpecSet):
     udev_persistent_net_rules = RegistryPoint(alias="udev-persistent-net.rules")
     uname = RegistryPoint()
     up2date = RegistryPoint()
+    up2date_log = RegistryPoint(alias="up2date_log")
```
xiangce: `alias` is used to keep compatible with the name in 1.x which contains invalid characters (`.` or `-`), For the newly added specs, it's good to use the same name in both 1.x and 3.x

TripleJQK:
changed as your requested:
delele alias, use up2date_log as same as 1.x
edit the timestamp

insights/parsers/up2date_log.py
```diff
@@ -0,0 +1,29 @@
+from .. import LogFileOutput, parser
```
xiangce: Please add the corresponding docstring file header. Refer to the `insights/parsers/httpd_V.py` (rebase to the latest master please)

bfahr: Also please add a `.rst` file to include this module in the documentation as described here: http://insights-core.readthedocs.io/en/latest/docs_guidelines.html#documentation-guidelines

TripleJQK:
@xiangce modified as your request, please review when free

insights/parsers/up2date_log.py
```diff
@@ -0,0 +1,38 @@
+"""
+up2date Logs -  Files ``/var/log/up2date``
+==========================================
+
+Modules for parsing the content of log file ``/var/log/up2date`` in sosreport archives of RHEL.
+
+.. note::
```
xiangce: This `Note` is duplicated with the one inner the class,  better remove this one.

PR #923, opened 2018-02-02
Enhance parser blkid.py

zhangyi733:
Fix bug when device name contains ":"


xiangce:
1.x: #922

PR #924, opened 2018-02-02
Correct issue with lssap parser

SteveHNH:
Bug was found regarding the way this command is parsed. It's possible
that the command dumps some content prior to the `- lssap version - `
line, which throws off the index used to start needed content.

This fix checks for the starting line and uses it as the intial index.

Will also need to submit a fix for 1.x


insights/parsers/lssap.py
```diff
@@ -59,8 +59,14 @@ def __init__(self, *args, **kwargs):
...
+        # Ensure we only parse content after the lssap version line. Default the
+        # starting index to 2 for expected lssap output
+        start = 2
+        for row in content:
+            if 'lssap' and 'version' in row:
+                start = content.index(row) + 2
+        clean_content = content[start:-1]
+        print(clean_content)
         if len(clean_content) > 0 and "SID" in clean_content[0]:
             self.data = parse_delimited_table(clean_content, delim='|', header_delim=None)
```
PaulWay: Firstly, you'll need to remove your debugging `print` statement :-)

Secondly, just wondering if you can use the `heading_ignore` option to `parse_delimited_table` to do this kind of search automatically.  `heading_ignore` actually indicates the line of the header (rather than ignoring it), so you could do something like:
```python
self.data = parse_delimited_table(content, delim='|', header_delim=None, heading_ignore=['  SID'])
```
That might be an easier way of getting the header line right 'automatically'.

bfahr:
@SteveHNH if there is a issue/bug report for this problem would you mind to link it, assuming its publically available?

csams:
@SteveHNH Is this PR still relevant?

PR #929, opened 2018-02-05
add commands to the oc_get parser

shzhou12:


insights/config/specs.py
```diff
@@ -495,16 +495,20 @@
...
     "oc_get_service"            : CommandSpec("/usr/bin/oc get service -o yaml --all-namespaces"),
+    "oc_get_event"              : CommandSpec("/usr/bin/oc get event -o yaml --all-namespaces"),
     "oc_get_rolebinding"        : CommandSpec("/usr/bin/oc get rolebinding -o yaml --all-namespaces"),
     "oc_get_project"            : CommandSpec("/usr/bin/oc get project -o yaml --all-namespaces"),
     "oc_get_role"               : CommandSpec("/usr/bin/oc get role -o yaml --all-namespaces"),
     "oc_get_pv"                 : CommandSpec("/usr/bin/oc get pv -o yaml --all-namespaces"),
     "oc_get_pvc"                : CommandSpec("/usr/bin/oc get pvc -o yaml --all-namespaces"),
-    "oc_get_endpoints"          : CommandSpec("/usr/bin/oc get endpoints -o yaml --all-namespaces")
+    "oc_get_endpoints"          : CommandSpec("/usr/bin/oc get endpoints -o yaml --all-namespaces"),
+    "oc_get_egressnetworkpolicy": CommandSpec("/usr/bin/oc get egressnetworkpolicy -o yaml --all-namespaces")
```
bfahr: @shzhou12 since you are making multiple changes to this, would you mind to put the specs in order by name?

insights/parsers/openshift_get.py
```diff
@@ -61,6 +73,15 @@ def metadata_name_items(data):
     return {item['metadata']['name']: item for item in data['items']}
 
 
+@parser('oc_get_node')
+class OcGetNode(YAMLParser):
+    """Class to parse ``oc get node -o yaml --all-namespaces``"""
+
+    def get_node(self):
+        """ dict: Returns a dictionary of openshift nodes information."""
+        return metadata_name_items(self.data)
```
bfahr: @shzhou12 if you used the `@property` decorator and renamed the method you could make this a bit cleaner and more pythonic, maybe something like:
```
@property
def nodes(self):
    return metadata_name_items(self.data)
```
You could also consider making this an attribute by overloading the `parse_content` method and adding it there.  This applies to all of the classes in this module.

insights/parsers/openshift_get.py
```diff
@@ -40,6 +49,9 @@
 OcGetService - command ``oc get service -o yaml --all-namespaces``
 ------------------------------------------------------------------
 
+OcGetEgressNetworkPolicy - command ``oc get egressnetworkpolicy -o yaml --all-namespaces``
+------------------------------------------------------------------------------------------
+
```
bfahr: These subheadings should be in sorted order since they show up in the documentation in the same order they appear in this file.

shzhou12:
@bfahr Thanks for your comments. I changed the code according to your comments. Please help to review the changes.

insights/parsers/openshift_get.py
xiangce: Hi @bfahr, is it OK that only update these @property methods (and remove the old non-@property ones) in the 3.x branches? and in this 1.x we can keep these old (and not @property) methods?

bfahr: @xiangce yes that would be OK with me.

xiangce: @bfahr - Thanks for your advice, @shzhou12 has updated the code and document according to your comments. I'm going to merge it.

PR #935, opened 2018-02-07
Remove the filter on PsAuxww

xiangce:
- PsAuxww should not be filtered.


xiangce:
@bfahr - I have made a mistake that pushed commits to the 3.0 branch directly. Hope that will not break the rebase to the 3.0 branch. Sorry for the inconvenience.

csams:
@xiangce @bfahr Please revert this commit. psauxww should be filtered as it contains the complete command line including arguments. Any rules that depend on this parser should need the command line arguments and should add an appropriate filter. If a rule doesn't need the command line arguments, it should use one of the other parsers we've added back that don't include them.

xiangce:
@csams done in #940

PR #961, opened 2018-02-21
Subscription-manager list parser for 1.x branch

PaulWay:
This module provides a base class for parsing the output of ``subscription-manager list``, with a variety of tools to make it useful.

Documentation and complete testing is provided.

### DIFF
--- a/insights/config/specs.py
+++ b/insights/config/specs.py
@@ -398,6 +398,8 @@ static_specs = {
     "sshd_config_perms"         : CommandSpec("/bin/ls -l /etc/ssh/sshd_config"),
     "sssd_config"               : SimpleFileSpec("etc/sssd/sssd.conf"),
     "sssd_logs"                 : PatternSpec(r"var/log/sssd/.*\.log$"),
+    "subscription_manager_list_consumed": CommandSpec("/usr/bin/subscription-manager list --consumed"),
+    "subscription_manager_list_installed": CommandSpec("/usr/bin/subscription-manager list --installed"),
     "swift_object_expirer_conf" : SimpleFileSpec("etc/swift/object-expirer.conf"),
     "swift_proxy_server_conf"   : SimpleFileSpec("etc/swift/proxy-server.conf"),
     "sysconfig_chronyd"         : SimpleFileSpec("etc/sysconfig/chronyd"),
--- a/insights/parsers/subscription_manager_list.py
+++ b/insights/parsers/subscription_manager_list.py
@@ -10,7 +10,7 @@ SubscriptionManagerListConsumed - command ``subscription-manager list --consumed
 ----------------------------------------------------------------------------------
 
 SubscriptionManagerListInstalled - command ``subscription-manager list --installed``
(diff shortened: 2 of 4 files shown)


insights/config/specs.py
```diff
@@ -398,6 +398,8 @@
     "sshd_config_perms"         : CommandSpec("/bin/ls -l /etc/ssh/sshd_config"),
     "sssd_config"               : SimpleFileSpec("etc/sssd/sssd.conf"),
     "sssd_logs"                 : PatternSpec(r"var/log/sssd/.*\.log$"),
+    "subscription_manager_list_consumed": CommandSpec("subscription-manager list --consumed"),
+    "subscription_manager_list_installed": CommandSpec("subscription-manager list --installed"),
```
bfahr: @PaulWay you'll need to add the path to this command.

PaulWay: That's stupid of me, especially since I wrote the code that checked for that!  Oops!  I've fixed it in this update.  Thanks @bfahr

insights/parsers/subscription_manager_list.py
```diff
@@ -0,0 +1,231 @@
...
+        '11/14/14'
+        >>> type(sub1['Starts timestamp'])  # Converted to date
+        <type 'datetime.date'>
+        >>> sub1['Starts timestamp'].year
+        2014
+        >>> consumed.all_current  # Are all subscriptions listed as current?
+        True
+    """
+    @property
+    def all_current(self):
```
xiangce: Just a suggestion too, Instead of calculating the "Boolean" result each time when calling this interface, how about just calculating it once and assign the result to an attribute to save some calling workload?

insights/parsers/subscription_manager_list.py
```diff
@@ -0,0 +1,231 @@
...
+                    # If we have an existing record, save it
+                    if current_record:
+                        self.records.append(current_record)
+                    current_record = {}
+                current_record[key] = value
+                # Do some type conversions and add-ons
+                if key == 'Active' and value in ('True', 'False'):
+                    current_record[key] = (value == 'True')
+                elif key in ('Starts', 'Ends'):
+                    date_match = mdy_re.search(value)
```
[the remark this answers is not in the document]
PaulWay: That's a much better idea - thanks @xiangce, I don't know why I didn't think of using that myself!

insights/parsers/subscription_manager_list.py
PaulWay: I thought about that, but those properties are only relevant for the child, not the parent.  So to make it an attribute would mean overriding the `parse_content` method in the subclass and doing the post-processing in that.  To me that's a little less elegant.

I agree that it will have to rescan each time it's called, but a) hopefully that won't be too often, and b) there aren't usually a lot of records to scan.

The other benefit of doing it this way is that if someone never uses this, it's not called and doesn't waste any time pre-calculating.

xiangce: sounds good to me :)

PR #978, opened 2018-03-01
Remove deprecated installed-rpms specs

zhangyi733:


zhangyi733:
In master branch, these deprecated specs have been removed.

insights/config/specs.py
```diff
@@ -8,8 +8,8 @@
 specs listed here are collected.
 """
 
-from insights.config import SimpleFileSpec, PatternSpec, CommandSpec, format_rpm, json_format, First, All, NoneGroup, format_rpm_old
```
shzhou12: I am not sure if it is correct to remove the json_format. If it is correct,  I think you should also remove the definition of josn_format.

xiangce: @zhangyi733 - I prefer to remove the definition of `josn_format`, how do you think?

zhangyi733: @shzhou12 and @xiangce - Since no functions involve `json_format`, I agree to remove it.

xiangce:
Hey @bfahr,

@zhangyi733 raised this PR is to ensure the 1.x and 3.0 used the same "mangled file name" in the insights archive. And try to avoid different hit count caused by this issue that we met when migrating rules to 3.0. Please take look and merge it if it's OK with you. 

Thanks.

csams:
@zhangyi733 @xiangce @bfahr, in 3.0 we have the following spec for installed rpms when reading from an archive:
```
  4     installed_rpms = first_file([                                               
  3         "insights_commands/rpm_-qa_--qf_NAME_-_VERSION_-_RELEASE_._ARCH_INSTALLTIME_date_BUILDTIME_VENDOR_BUILDHOST_DUMMY_SIGPGP_pgpsig",
  2         "insights_commands/rpm_-qa_--qf_NAME_-_VERSION_-_RELEASE_._ARCH_INSTALLTIME_date_BUILDTIME_RSAHEADER_pgpsig_DSAHEADER_pgpsig"
  1     ])
```
Does that fix the differences you were seeing? If not, should resolve the issue by adding to the 3.x definition because it means some archives have an old format that 3.x doesn't recognize.

zhangyi733:
@csams - Your fix in 3.x can solve the issue.

Issue #988, opened 2018-03-07
Add check for brctl "No such file or directory"


vishwanathjadhav:
There are some archives where it is noticed that the ``brctl show`` command not found, but the error message is different ``timeout: failed to run command `/usr/sbin/brctl': No such file or directory``. Need to add a check for such error message.

csams:
Fixed in master in 0e0bb08f73867bf76e0fb539843c535e0aabeea6. Changes for 1.x are no longer being accepted.

PR #989, opened 2018-03-07
Enhance brctl_show parser 

vishwanathjadhav:
Parser ``brctl`` is through an exception ``index out of range`` when try to access empty list, hence added a condition to check if the list has a proper number of elements.


insights/parsers/brctl_show.py
```diff
@@ -77,28 +80,29 @@ def group_by_iface(self):
     def parse_content(self, content):
         self._group_by_iface = {}
         self.data = []
-        if "/usr/sbin/brctl: file not found" in content[0]:
-            return
-        elif "\t" in content[0]:
+        for err_srt in ERRORS:
```
xiangce: A more pythonic way:

```python
 if any(err_srt in content[0] for err_srt in ERRORS):
   retrun
```

xiangce:
Hey @vishwanathjadhav - Chris has raised a [patch](https://github.com/RedHatInsights/insights-core/commit/0e0bb08f73867bf76e0fb539843c535e0aabeea6) to fix this problem in framework level, then, I think this PR can be closed.

See: https://github.com/RedHatInsights/insights-core/issues/988

xiangce:
Hey @vishwanathjadhav , would you please to check if we can remove the corresponding invalid lines check since @csams has already handled them in datasource.

And better update this parser in 3.0 one more time to raise an exception when can't find any meaningful data.

PR #991, opened 2018-03-07
New parser for `engine-config --all` command

psachin:
Very early version of parser for command `engine-config`. I intended to design parser where one could pass the key on-the-fly and the command will execute it as

```
engine-config --get=KEY
```
to output value. But this seem not possible at the moment as the specs file run the command and collect the output as an insight archive.

The implementation still works on the output from `engine-config --all`. The parser tries its best to get 'value' & 'version' for 'key' specified(Please see usage example in doc-string). It ignores keys
where is fails(simple try...expect). If key with same name exist, it creates a list of values and versions accordingly


psachin:
@JoySnow 

Here's the initial implementation of parser for command `ovrit-engine --all` that we discuss on https://trello.com/c/6iXpYkg9. Please review

insights/config/specs.py
```diff
@@ -94,6 +94,7 @@
     "docker_sysconfig"          : DockerHostSimpleFileSpec("etc/sysconfig/docker"),
     "dumpe2fs-h"                : CommandSpec("/sbin/dumpe2fs -h {dumpdev}", dumpdev=r"\S+"),
     "engine.log"                : SimpleFileSpec("var/log/ovirt-engine/engine.log", large_content=True),
+    "engine-config_all"         : CommandSpec("/usr/bin/engine-config --all"),
```
xiangce: In the title and description, you wrote command `ovirt-engine --all`, but here, it's `engine-config --all`. I don't know if they are the same, but would you please to explain a bit?

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,83 @@
+from collections import namedtuple
```
xiangce: Docstring file header is necessary, please add it (see example in other parses  e.g. "tomcat_xml.py")

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,83 @@
...
+    []
+    >>> output.get_version('ClusterRequiredRngSourcesDefault')
+    ['3.6', '4.0', '4.1']
+    >>> output.last('HotUnplugCpuSupported')
+    '{"x86":"true","ppc":"true"}'
+    """
+    KeyValue = namedtuple('KeyValue',
+                          ['keyword', 'value', 'version', 'kw_lower'])
+
+    def parse_content(self, content):
```
xiangce: Documents for the attributes like `fields`, `keywords` should be added to the docstring. Please refer to examples in other parses or http://www.sphinx-doc.org/en/stable/ext/example_google.html

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,83 @@
...
+        self.fields = []
+        for line in get_active_lines(content):
+            try:
+                key, val, ver = itemgetter(0, 1, -1)(line.split(" "))
+                self.fields.append(self.KeyValue(key.strip(':'), val, ver, key.strip(':').lower()))  # noqa
+            except:
+                pass
+        self.keywords = set([kw.kw_lower for kw in self.fields])
```
xiangce: Better explain it clearly you have converted all keywords to lower cases in the document.

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,83 @@
...
+        return iter(self.fields)
+
+    def __getitem__(self, keyword):
+        kw = keyword.lower()
+        if kw in self.keywords:
+            return filter(
+                None,
+                [kv.value for kv in self.fields if kv.kw_lower == kw])
+
+    def head(self, keyword):
```
xiangce: Better add documentation for each public interface.

insights/config/specs.py
psachin: Done

insights/parsers/engine_config.py
psachin: Done

insights/parsers/engine_config.py
psachin: Sure

bfahr:
@psachin I didn't see any issues beyond those pointed out by @xiangce.

psachin:
@xiangce @bfahr The documentation is in place now.

insights/parsers/engine_config.py
bfahr: @psachin I don't see the commit with the changes, have you pushed it?

bfahr:
@psachin you still need to add docstrings to `head`, `last` and `get_version`.

psachin:
@bfahr I added docstring for those methods. Please check.

insights/parsers/tests/test_engine_config.py
```diff
@@ -0,0 +1,296 @@
...
+:INPUT ACCEPT [0:0]
+:FORWARD ACCEPT [0:0]
+:OUTPUT ACCEPT [0:0]
+-A INPUT -m state --state ESTABLISHED,RELATED -j ACCEPT
+-A INPUT -p icmp -j ACCEPT
+-A INPUT -i lo -j ACCEPT
+# vdsm
+-A INPUT -p tcp --dport @VDSM_PORT@ -j ACCEPT
+# ovirt-imageio-daemon
+-A INPUT -p tcp --dport 54322 -j ACCEPT
```
JoySnow: This current parsing logic is good enough for Line-122 type parameters.
Well, for this 'IPTablesConfig' block, it will go wrong in parsing.
If we don't have a better way to solve this, maybe it will be better to add some explanation in docstrings to notice the users pay attention to this.
@xiangce  Any ideas about this?

psachin: Hi @JoySnow Yeah! that the a part which is not been parsed at-the-moment. I even added an assertion to test `IPTablesConfig`. Let me think over it a bit else I'll add appropriate explanation.

psachin: @JoySnow 

I tried few regex patterns(they can be any multi-line command output in future) and they don't seems reliable. As for now I'll skip this kind of blocks and update the docs so that the user is informed.

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __iter__(self):
+        return iter(self.fields)
+
+    def __getitem__(self, keyword):
+        kw = keyword.lower()
+        if kw in self.keywords:
+            return filter(
+                None,
+                [kv.value for kv in self.fields if kv.kw_lower == kw])
```
bfahr: @psachin I noticed that in your example code above at line 44 an empty list is returned but here if `keyword` is not in `self.keywords` then `None` is returned.  Should you return an empty list instead?  Also could you could simplify this return as just `return [kv.value for kv in self.fields if kv.kw_lower == kw]` or is there some reason why `filter` is necessary?

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,187 @@
...
+
+        Returns:
+
+            list: List of versions associated with a key.
+        """
+        kw = keyword.lower()
+        if kw in self.keywords:
+            return filter(
+                None,
+                [kv.version for kv in self.fields if kv.kw_lower == kw])
```
bfahr: Same questions as above regarding the default return value and use of the filter.

insights/parsers/engine_config.py
psachin: Hi @bfahr I use `filter(None, ...)` on purpose. As you see in the example, the keyword 'ClusterRequiredRngSourcesDefault' has empty values(returns `['', '', '']`). I want `__getitem__` to return empty list in case all values in list are empty(now documented). 

The current implementation returns `False` if keyword don't exists but values and versions will be empty list for non-existing keyword(documented). 

Another approach is to keep the list as it is(`['', '', '']`) rather then returning `[]` and map them to versions(say `['3.0', '3.6', '4.0']`) so that user can fetch values specific to version(or vice versa) for a keyword. But I think that is not needed at this time and I don't have a use case. 

Let me know you view

bfahr:
@psachin we found out last night that since 3.x is going live this week, we are not merging any more changes to the 1.x branch unless they fix critical errors.  You may resubmit this PR against the 3.x branch.

psachin:
Closed in favour of https://github.com/RedHatInsights/insights-core/pull/1019

PR #994, opened 2018-03-07
Initial spec + parser for /etc/sysconfig/memcached

kgrant-rh:

PR #995, opened 2018-03-07
Initial spec + parser for /etc/sysconfig/memcached for insights-core 1x

jsvob:


jsvob:
The same for 3.x - https://github.com/RedHatInsights/insights-core/pull/994

PR #998, opened 2018-03-08
* Add parsers LvsAll, PvsAll, VgsAll and combiner LvmAll for getting LVM information for all accepted and rejected devices.

lonicerae:
- When executing the lvs, pvs, vgs commands on system, the collected data is
   generated following the 'filter' list in /etc/lvm/lvm.conf by default. But
   this make it impossible to find out the bad filters that blacklist normal
   disks. So that we need to add extra directives to collect LVM information
   for all devices.
 - Note that these parsers and combiner should not be used in ordinary plugins.
   They're used for checking the LVM filter issues.


insights/combiners/tests/test_lvm.py
```diff
@@ -456,3 +457,160 @@ def test_combiner_lvs(lvm_data):
...
+            assert len(list(lvm_info.volume_groups)) == 8
+            assert lvm_info.volume_groups['data7']['VG_UUID'] == 'r58iLs-sYla-lC4h-UURW-bZOG-P2Yd-lambpM'
+            assert lvm_info.volume_group_names == {
+                'data1', 'data2', 'data3', 'data4', 'data5', 'data6', 'data7',
+                'rhel_ceehadoop1'
+            }
+            assert lvm_info.volume_groups['data1']['VPerms'] == 'writeable'
+            filtered_lvm_info = lvm_info.filter_volume_groups('data')
+            assert set(filtered_lvm_info.keys()) == {
+                'data1', 'data2', 'data3', 'data4', 'data5', 'data6', 'data7'
```
xiangce: python2.6 test failed here.

insights/specs/default.py
```diff
@@ -315,6 +315,7 @@ def semid(broker):
     lvm_conf = simple_file("/etc/lvm/lvm.conf")
     lvs = None  # simple_command('/sbin/lvs -a -o +lv_tags,devices --config="global{locking_type=0}"')
     lvs_noheadings = simple_command("/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_name,lv_size,lv_attr,mirror_log,vg_name,devices,region_size,data_percent,metadata_percent --config=\"global{locking_type=0}\"")
+    lvs_noheadings_all = simple_command("/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_name,lv_size,lv_attr,mirror_log,vg_name,devices,region_size,data_percent,metadata_percent --config='global{locking_type=0} devices{filter=[\"a|.*|\"]}'")
```
xiangce: You should add its corresponding `simple_file` to specs/insights_archive.py too.

insights/specs/default.py
```diff
@@ -438,6 +439,7 @@ def package_and_java(broker):
     puppet_ssl_cert_ca_pem = None
     pvs = simple_command('/sbin/pvs -a -v -o +pv_mda_free,pv_mda_size,pv_mda_count,pv_mda_used_count,pe_count --config="global{locking_type=0}"')
     pvs_noheadings = simple_command("/sbin/pvs --nameprefixes --noheadings --separator='|' -a -o pv_all,vg_name --config=\"global{locking_type=0}\"")
+    pvs_noheadings_all = simple_command("/sbin/pvs --nameprefixes --noheadings --separator='|' -a -o pv_all,vg_name --config='global{locking_type=0} devices{filter=[\"a|.*|\"]}'")
```
xiangce: The same.

insights/combiners/tests/test_lvm.py
lonicerae: Got it.

insights/specs/default.py
lonicerae: Fix in next commit.

insights/specs/default.py
lonicerae: Fix in next commit.

PR #1000, opened 2018-03-08
Doc changes for notebooks in 3.x

bfahr:
* Make changes to the documentation build and structure to support inclusion of Jupyter notebooks.
* Add nbsphinx extension to support inclusion of notebooks


bfahr:
I set this PR back to WIP because I need to change the build to install pandoc via pypandoc since the `readthedocs.io` does not allow installs on its build images.

bfahr:
I confirmed that the docker build image that is used by readthedocs.io already contains pandoc so this should work fine.

Issue #1007, opened 2018-03-13
WARNING:insights.tests:Replacing ps_auxcww in --appdebug output


PaulWay:
Running the test suite on the Insights Plugins Demo ruleset with the options '-s -v --appdebug' gives multiple warnings in the output of the form:

```
WARNING:insights.tests:Replacing ps_auxcww
[...]

PR #1008, opened 2018-03-13
Only warn if spec is already in InputData object

PaulWay:
Fixes issue #1007


bfahr:
This change looks reasonable to me.  It would be good if @csams or @kylape could also review.

PaulWay:
Looks like @csams has created a patch independently of this that fixes this bug in the same way, so we can close this :-)

PR #1010, opened 2018-03-13
Strip all lines instead of lines contain ':'

xiangce:


xiangce:
Merge it quickly to make a test

xiangce:
Hit result of the is the same to 1.x now, this PR is OK.

PR #1013, opened 2018-03-13
Changes do file, command and filter operations outside of python code

lhuett:
Changes in spec_factory.py to improve performance:
- Changed TextFileProvider class to use grep for reading and filtering files 
- Changed simple_commad function to use grep for filtering results of command.
-  Changed foreach_execute function to use grep for filtering results of each command.
Change in context.py:
- Added override of ExecutionContext.check_output function in HostContext class to allow subprocess piping


insights/core/spec_factory.py
```diff
@@ -142,18 +143,20 @@ def validate_lines(self, results):
...
-                # This should shell out to a grep pipeline
-                results = [l.rstrip("\n") for l in apply_filters(self.ds, f)]
+
+        filters = False
+        if self.ds:
+            filters = "|".join(get_filters(self.ds))
+        if filters:
+            rc, out = subproc.call("grep -E '{0}' {1} ".format(filters, self.path), shell=False, keep_rc=True)
+            if rc == 0 and out != '':
+                results = [l.rstrip() for l in out.splitlines()]
```
jhjaggars: I think we just changed the behavior of datasources to *not* `rstrip`.

insights/core/spec_factory.py
```diff
@@ -452,14 +455,18 @@ def simple_command(cmd, context=HostContext, split=True, keep_rc=False, timeout=
     def inner(broker):
         ctx = broker[context]
         rc = None
-        raw = ctx.shell_out(cmd, split=split, keep_rc=keep_rc, timeout=timeout)
+        if split:
+            filters = "|".join(get_filters(inner))
+        if filters:
+            raw = ctx.shell_out("{0} | grep -E '{1}'".format(cmd, filters), split=split, keep_rc=keep_rc,
```
jhjaggars: Have you considered chaining `Popen` objects together rather than relying on a shell to do the pipelining?

lhuett: Thanks Jesse,
I actually do that further down in the calls. shell_out calls check_output in the ExecutionContext. Check_output then calls the call function in supbroc and the subproc class does not support chaining Popens. To get around that I override check_output in the HostContext and it parses the arguments and does the chaining of the Popens (does not uses subproc).
I was trying not to change current behavior in the upper functions that other classes might be using.
Hope all this babble makes sense.

insights/core/spec_factory.py
lhuett: Hey @jhjaggars ,
I noticed that in master there is still one instance of rstrip in the load() function in TextFileProvider class in the master branch.
Do I still need to make the changes to remove the rstrip()?

https://github.com/RedHatInsights/insights-core/blob/f99b9fccb9fece4fa8083505a29e77bf432a31f6/insights/core/spec_factory.py#L149-L151

jhjaggars: Calling @csams.

insights/core/spec_factory.py
```diff
@@ -142,18 +143,20 @@ def validate_lines(self, results):
...
-            results = []
-            if self.ds:
-                # This should shell out to a grep pipeline
-                results = [l.rstrip("\n") for l in apply_filters(self.ds, f)]
+
+        filters = False
+        if self.ds:
+            filters = "|".join(get_filters(self.ds))
+        if filters:
+            rc, out = subproc.call("grep -E '{0}' {1} ".format(filters, self.path), shell=False, keep_rc=True)
```
kylape: Are we making a conscious decision to change from fixed string to regex grepping?

insights/core/spec_factory.py
kylape: I think we should remove the strip and test the change.  It seems weird to keep this one while all the others are taken out.  Testing the change should show us if it has adverse affects on rule hits.

insights/core/context.py
```diff
@@ -1,7 +1,15 @@
...
 from insights.util.subproc import call
-from subprocess import STDOUT
+
+from subprocess import CalledProcessError, PIPE, Popen
+import subprocess
+import shlex
+import signal
+
+STDOUT = subprocess.PIPE
+STDERR = subprocess.STDOUT
```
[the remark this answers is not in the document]
kylape: I now realize you were copying this from the subproc module, but I think my suggestion just applies there as well.

insights/core/spec_factory.py
kylape: > I was trying not to change current behavior in the upper functions that other classes might be using.

I think making your changes in subproc makes the most sense to me.  Having code duplication like this is probably a maintenance burden long-term.  We can always perform regression testing to ensure you don't break backwards compat.  As long as you aren't changing/breaking APIs used by plugin repos.

```
$ ag 'subproc[^e]' -l insights
insights/core/context.py
insights/core/spec_factory.py
insights/core/archives.py
insights/tests/test_subproc.py
insights/util/fs.py
```

Not too many consumers of `subproc` it seems.

insights/core/spec_factory.py
lhuett: I will remove the strip from the new code which should take care of all of the rstrips() in the spec_fatctory.

insights/core/spec_factory.py
lhuett: That sounds good to me, just trying not to stir up too much dust. Removing the override if check_output) in the context class will eliminate the the issue you mentioned about...
STDOUT = subprocess.PIPE
STDERR = subprocess.STDOUT
While I'm at it I'll see if I can change that in subproc().

insights/core/context.py
```diff
@@ -155,7 +163,64 @@ def __repr__(self):
...
+            cmd = cmd.encode('utf-8', 'replace')
+            if timeout is not None and sys.platform != "darwin":
+                cmd = "timeout -s {0} {1} {2}".format(signal.SIGKILL, timeout, cmd)
+
+            log.debug(cmd)
+
+            if not shell:
+                rcmd = cmd = shlex.split(cmd)
+                try:
+                    pidx = rcmd.index("|")
```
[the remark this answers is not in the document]
lhuett: I'll look at this.

insights/core/spec_factory.py
lhuett: The idea was to remove the file open/read and separate apply_filters() to improve performance. Reformatting the list of filters and using grep to traverse and filter the lines in the file based on the filters seemed like the simplest and fastest way top do it.

kylape: Sorry, I should have made it clear that I was referring specifically to your `-E` option.  Historically we've used `-F`, and I was wondering if this is something that we are intentionally changing at this point.

bfahr:
@kylape @lhuett whats the status on this PR?  Reading the comments it seems like there are still a few changes to be made, is that correct?

PR #1015, opened 2018-03-13
Add dynamic doc generation for components

bfahr:
* Add new utility component_graph to build component cross reference
* Include docs in index and add specs to api index
* Add call to utility to sphinx setup


bfahr:
Example screenshot of the cross reference:

components_xref_screenshot

PR #1017, opened 2018-03-13
validate obfuscate_hostname, catch config errs

gravitypriest:


insights/client/phase/v1.py
```diff
@@ -20,7 +20,11 @@
 def phase(func):
     @functools.wraps(func)
     def _f():
-        compile_config()
+        try:
+            compile_config()
+        except ValueError as e:
+            print('ERROR:' + e)
```
kylape: looks like other messages are printed to stderr via logging.

kylape: oh shoot.  I see that logging may not be set up at this point.  carry on.

kylape: Actually, thinking about it even more, I believe the right thing to do is to print this message to stderr.  Since logging isn't guaranteed to be set up, you probably need to do `sys.stderr.write('ERROR: ' + e + '\n')`

PR #1018, opened 2018-03-14
Refine the HttpdV parser

xiangce:
In the new framework 3.0, HttpdV parser is based on the new datasource
which can get the unique and correct `httpd` command by analyzing the
`ps auxww` output. In RHEL6, the command could be `httpd`, `httpd.worker`
or `httpd.event`, but only one of them will be collected. In RHEL7, only
`httpd` is provided.
- Remove the specs: `httpd_event_V` and `httpd_worker_V`
- Update the `httpd_V` spec to use wildcard in `insights_archive.py`
- Raise `SkipComponent` in the `HttpdWorkerV` and `HttpdEventV` when necessary
- Mark the `HttpdV` combiner as deprecated
- Update the examples in `HttpdV` parser as per doctest


insights/specs/default.py
```diff
@@ -223,21 +223,22 @@ def docker_container_ids(broker):
...
-                # e.g. /usr/sbin/httpd, /usr/sbin/httpd.worker and /usr/sbin/httpd.event
-                if 'httpd' in os.path.basename(cmd):
-                    return [cmd]
-        return []
+                # e.g. httpd, httpd.worker and httpd.event
+                #      and SCL's httpd24-httpd
+                if os.path.basename(cmd).startswith('httpd'):
+                    ps_httpds.add(cmd)
+        # collect nothing if multiple httpd instance are running
+        return ps_httpds if len(ps_httpds) == 1 else []
```
csams: return list(ps_httpds) here or ps_httpds will be a set in one case and a list in the other. If the set is empty, list() will convert it to an empty list automatically.

PR #1019, opened 2018-03-14
New parser for `engine-config --all` command

psachin:
Very early version of parser for command `engine-config`. I intended
to design parser where one could pass the key on-the-fly and the
command will execute it as

```
engine-config --get=KEY
```

to output value. But this seem not possible at the moment as the specs
file run the command and collect the output as an insight archive.

The implementation still works on the output from `engine-config
--all`. The parser tries its best to get 'value' & 'version' for 'key'
specified(Please see examples in doc-string). It ignores keys where is
fails(try...expect). If key with same name exist, it creates a list of
values and versions accordingly.

Includes:
- Tests
- Documentation & Examples


insights/parsers/engine_config.py
```diff
@@ -0,0 +1,231 @@
...
+            ['30']
+
+        Args:
+
+            keyword (str): A key. Ex: 'HotUnplugCpuSupported'.
+
+        Returns:
+
+            list: Values associated with a keyword. Returns an empty list if,
+                  1. All the values are empty.
```
xiangce: Better move this two lines forward and keep them the same indent with `list:`. See:  http://www.sphinx-doc.org/en/stable/ext/example_google.html?highlight=Returns

insights/specs/default.py
```diff
@@ -169,6 +169,7 @@ def docker_container_ids(broker):
     docker_sysconfig = simple_file("/etc/sysconfig/docker")
     dumpdev = simple_command("/bin/awk '/ext[234]/ { print $1; }' /proc/mounts")
     dumpe2fs_h = foreach_execute(dumpdev, "/sbin/dumpe2fs -h %s")
+    engine_config_all = simple_command("/usr/bin/engine-config --all")
```
xiangce: In 3.x, for simple_command, should add its `simple_file` to the `specs/insights_archives.py`. You can use the 'bin/mangle' command to generate it.

~~~
mangle "/usr/bin/engine-config --all"
~~~

psachin: @xiangce I don't find a guide to create rule which has example as this in http://insights-core.readthedocs.io/en/latest/. Where is this documented?

psachin: Besides, what is the package which provides `mangle` command? 

I see "no match found" on RHVH and below output on Fedora,
```
$ sudo dnf whatprovides */mangle
Last metadata expiration check: 0:07:36 ago on Wed 14 Mar 2018 03:18:16 PM IST.
shorewall-5.1.10.2-1.fc27.noarch : An iptables front end for firewall configuration
Repo        : updates
Matched from:
Filename    : /etc/shorewall/mangle
Matched from:
Filename    : /usr/share/shorewall/configfiles/mangle

shorewall6-5.1.10.2-1.fc27.noarch : Files for the IPV6 Shorewall Firewall
Repo        : updates
Matched from:
Filename    : /etc/shorewall6/mangle
Matched from:
Filename    : /usr/share/shorewall6/configfiles/mangle

shorewall-5.0.14.1-5.fc27.noarch : An iptables front end for firewall configuration
Repo        : fedora
Matched from:
Filename    : /etc/shorewall/mangle
Matched from:
Filename    : /usr/share/shorewall/configfiles/mangle

shorewall6-5.0.14.1-5.fc27.noarch : Files for the IPV6 Shorewall Firewall
Repo        : fedora
Matched from:
Filename    : /etc/shorewall6/mangle
Matched from:
Filename    : /usr/share/shorewall6/configfiles/mangle
```

xiangce:
Sorry for the poor explanation, it's a utility tool inner the
insights-core, after re-run the `pip install -e .[develop]` you can
find it here
`insights-core/bin/mangle`.



Regards,
Xiangce Liu

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,231 @@
...
+    ClusterRequiredRngSourcesDefault:  version: 4.1
+    HotUnplugCpuSupported: {"x86":"false","ppc":"false"} version: 3.6
+    HotUnplugCpuSupported: {"x86":"false","ppc":"false"} version: 4.0
+    HotUnplugCpuSupported: {"x86":"true","ppc":"true"} version: 4.1
+
+Examples::
+
+    >>> from insights.parsers.engine_config import EngineConfigAll
+    >>> from insights.tests import context_wrap
+    >>> shared = {EngineConfigAll: EngineConfigAll(context_wrap(OUTPUT))}
```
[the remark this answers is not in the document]
psachin: Done

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,231 @@
...
+    >>> output = shared[EngineConfigAll]
+
+The ``in`` operator can be used to test whether a keyword is present
```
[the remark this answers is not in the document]
psachin: Sure

insights/parsers/engine_config.py
psachin: Done

insights/parsers/engine_config.py
```diff
@@ -0,0 +1,204 @@
...
+        MaxStorageVdsTimeoutCheckSec: 30 version: general
+        ClusterRequiredRngSourcesDefault:  version: 3.6
+        ClusterRequiredRngSourcesDefault:  version: 4.0
+        ClusterRequiredRngSourcesDefault:  version: 4.1
+        HotUnplugCpuSupported: {"x86":"false","ppc":"false"} version: 3.6
+        HotUnplugCpuSupported: {"x86":"false","ppc":"false"} version: 4.0
+        HotUnplugCpuSupported: {"x86":"true","ppc":"true"} version: 4.1
+
+
+    Examples::
```
bfahr: @psachin because `Examples` is a recognized header it should only have 1 colon, if you look at the generated documentation you can see the problem.  Line 24 above it correct because you are designating the following indented block as a literal block.  Treat `Examples` like `Parameters`, `Attributes`, `Raises` and etc.

psachin: @bfahr You are right. I see the problem in highlighting. I wondered why it display `Examples` header in different way. Thanks for pointing out.

psachin:
@xiangce Sure. Not a problem. 

As per my previous experience, `--amend` is hard for reviewers to track changes but good to merge as a single commit to master. It saves work to squash commits before merge.

xiangce:
@psachin Yes, exactly.

As you may know, GitHub provided a "Squash and Merge" button for us to do that. So, don't worry about that.

PR #1020, opened 2018-03-14
Removed check all the check for unwanted command output

vishwanathjadhav:
Removed check all the check for unwanted command output

Previous Reference:
https://github.com/RedHatInsights/insights-core/pull/989


insights/parsers/brctl_show.py
```diff
@@ -102,6 +98,8 @@ def parse_content(self, content):
                 else:
                     iface_lst.append(line.strip())
                     br_mapping[iface] = iface_lst
+        else:
+            raise ParseException("Invalid Data Found")
 
         for entry in self.data:
```
xiangce: Better shift this `for` into the `if iface` clause.
and remove the last `return`.

PR #1028, opened 2018-03-21
Fix ZeroDivisionError for empty mongod.conf in mongod_conf.py

chenlizhong:


insights/parsers/mongod_conf.py
```diff
@@ -109,9 +109,12 @@ def _file_type_is_yaml(self, content):
             any unexpected lines with wrong settings.
         """
         content = get_active_lines(content)
-        cnt = sum([1 for line in content if "=" in line])
-        percent = float(cnt) / len(content)
-        return True if percent < 0.9 else False
+        if content:
```
xiangce: Seems the checking of `content` is a much high-level condition, Better move it out this inner method to the beginning of `parse_content` instead.

And when `content` is empty which means there is no meaningful data, better raise a ParseException or SkipComponent for this case

PR #1029, opened 2018-03-22
Update HttpdV parser and spec to support multiple httpd instances

xiangce:
- The httpd_v datasource could handle multiple httpd instances
- Retain the HttpdV parser only, since it could parser both the
  HttpdEventV and HttpdWorkerV for RHEL6 too.
- Mark the HttpdEventV and HttpdWorkerV as deprecated
  The combiner HttpdV is useless too and was marked as deprecated already
- Raise Exceptions when there is no meaningful data.
- Add several useful attributes and test


xiangce:
Hey @csams - I have marked the parsers (HttpdEventV and HttpdWorkerV) and combiner HttpdV as deprecated since they not so useful to the 3.x. Could you please to take a look at to see if it's suitable? Thanks.

PR #1030, opened 2018-03-22
Fix the ValueError for "AllowUsers" without value

chenlizhong:


insights/parsers/ssh.py
```diff
@@ -83,7 +83,10 @@ class SshDConfig(Parser):
     def parse_content(self, content):
         self.lines = []
         for line in get_active_lines(content):
-            kw, val = [s.strip() for s in line.split(None, 1)]
+            try:
```
xiangce: I don't think your logic is incorrect, however, I don't think `try ... catch ...` is a good way to solve this kind of problems that we can predict. How about e.g.:
```python
line_splits =[s.strip() for s in line.split(None, 1)]
kw, val = line_splits if len(line_splits) == 2 else line_splits[0], ''
```

chenlizhong: ok

bfahr:
@xiangce I've asked @chenlizhong to report this issue in insights-core and link it to this PR since its a parser error.

chenlizhong:
Fix the issue: https://github.com/RedHatInsights/insights-core/issues/1034

bfahr:
@chenlizhong thanks for the extra effort.

PR #1032, opened 2018-03-22
Enhance NtpqLeap parser for more output content

chenlizhong:


insights/parsers/ntp_sources.py
```diff
@@ -76,10 +76,22 @@ class NtpqLeap(Parser):
 
     def parse_content(self, content):
         self.data = {}
-        line = content[0].strip()
-        if "=" in line:
-            k, v = line.split("=")
-            self.data = {k: v}
+        other_status = []
```
xiangce: Here is the manual of command `ntpq`: [man ntpq](https://www.freebsd.org/cgi/man.cgi?query=ntpq&apropos=0&sektion=0&manpath=FreeBSD%206.1-RELEASE&format=html), the command `ntpq -c rv 0 leap` only aims at outputting the value of the system variable `leap`. Any other outputs are not related. 

Just talked this with @lonicerae , and confirmed this is a defect of the old versions of `ntpq` (ver<4.2.4), these old `ntpq` haven't this feature to filter what you requested in the `readvar` list.

So, we'd better filter the `leap=##` manually in this parser and ignore any other outputs that have nothing to do with the `leap`.

chenlizhong: ok. we can just get the leap value

bfahr:
@chenlizhong there's no need to submit an issue report.  I approve this PR.

PR #1035, opened 2018-03-23
To skip the command output with Unrecognised field error

chenlizhong:


insights/parsers/lvm.py
```diff
@@ -34,6 +34,7 @@
 from .. import add_filter
 from .. import Parser, parser, get_active_lines, LegacyItemAccess
 from . import parse_fixed_table
+from insights.core.dr import SkipComponent
```
xiangce: Seems SkipException is more suitable than SkipComponent. See http://insights-core.readthedocs.io/en/latest/exception_model.html#skipcomponent-and-skipexception

chenlizhong: Below is the code in dr.py,  Only SkipComponent is pass exception. All other exceptions will be logged into log file. I do not want to log anything for this as this is normal for some lvm version does not support one or two fields.

~~~
       try:
            if component not in broker and component in DELEGATES and ENABLED[component]:
                log.info("Trying %s" % get_name(component))
                result = DELEGATES[component](broker)
                broker[component] = result
        except MissingRequirements as mr:
            if log.isEnabledFor(logging.DEBUG):
                name = get_name(component)
                reqs = stringify_requirements(mr.requirements)
                log.debug("%s missing requirements %s" % (name, reqs))
            broker.add_exception(component, mr)
        except SkipComponent:
            pass
        except Exception as ex:
            log.exception(ex)
            broker.add_exception(component, ex, traceback.format_exc())
~~~

chenlizhong: Yes

insights/parsers/lvm.py
```diff
@@ -92,6 +93,8 @@ class Lvm(Parser):
     """Base class for parsing LVM data in key=value format."""
 
     def parse_content(self, content):
+        if "Unrecognised field:" in content[-1]:
```
bfahr: @chenlizhong I'm wondering if there should be a more generalized solution to this problem?  This message string is very specific and to some extent localized.

chenlizhong: @bfahr 
Look at the source code of lvm.  "Unrecognised field:" is hardcode.  I also tested vgs, pvs, lvs they all give "Unrecognised field:" when the -o parameter following unsupport field.

~~~
if (!_field_match(rh, ws, (size_t) (we - ws), report_type_only)) {
    _display_fields(rh, 1, 0);
    log_warn(" ");
    log_error("Unrecognised field: %.*s", (int) (we - ws), ws);
~~~

https://github.com/lvmteam/lvm2/search?utf8=%E2%9C%93&q=Unrecognised+field&type=

bfahr: Thank you @chenlizhong, that seems more reasonable then.

xiangce: @chenlizhong - how about changing to ParseException?

PR #1037, opened 2018-03-23
Raise ParseException instead of ValueError

xiangce:
- for empty content and other invalid JSON content


xiangce:
@bfahr - I'm a little confused, do you think this is necessary?

bfahr:
@xiangce not I do not believe it is necessary if all you are doing is adding a try/except around the entire method.  Any exceptions are caught by the infrastructure.  We may need to have a broader discussion about this however.

bfahr:
@csams @kylape related to exception discussion in #1040

xiangce:
Close this PR as per the discussion with NA members.

PR #1039, opened 2018-03-23
New method in HttpdConfAll combiner to list sections.

jsvob:
Needed by upcoming httpd hardening rule.


jsvob:
Also, guys, who prefers to be included and who prefers to not be included in the list of reviewers for these kinds of pull requests in the future?

insights/combiners/httpd_conf.py
```diff
@@ -235,9 +237,64 @@ def get_active_setting(self, directive, section=None):
...
+                    # From each section, preserve only the last ParsedData
+                    # {(section, ""): [ParsedData, ParsedData]} ---> ParsedData
                     values_list[i] = val.values()[0][-1]
                 return values_list
             return []
         else:
             if values_list:
                 return values_list[-1]
+
+    def get_section_list(self, section):
```
jsvob: Just a note that it is a partially inspired by `get_setting_list` because it is quite complex matter and this approach helped me to not get lost.

insights/combiners/httpd_conf.py
```diff
@@ -235,9 +237,64 @@ def get_active_setting(self, directive, section=None):
...
+        Returns the specified sections.
+
+        Parameters:
+            section (str): The section to look for, e.g. "Directory"
+
+        Returns:
+            (list of dict): List of tuples, each tuple has three elements - the
+                first being a tuple of the section and section name, the second
+                being the file name of the file where that section resides,
+                the third being the full file path of the file.
```
bfahr: @jsvob the indentation on this should match the first line (list of dict), you can see the problem in the rendered output.  Also, is it a list of dict or list of tuples?

jsvob: @bfahr Good point, it's a list of tuples. Fixed. I tried to compile the documentation according to https://github.com/RedHatInsights/insights-core/wiki and it errored out (used to work for me before but I can't rule out I didn't miss something during the recent insights transition to 3.0)), so I hope it'll be fine.

PR #1040, opened 2018-03-26
User SkipException instead of ParserException for empty file

chenlizhong:


xiangce:
Close this PR as per the discussion with NA

PR #1041, opened 2018-03-26
Handle "failed to get device path" in output of lsblk

chenlizhong:


insights/parsers/lsblk.py
```diff
@@ -294,17 +294,22 @@ class LSBlockPairs(BlockDevices):
     """
     def parse_content(self, content):
         self.rows = []
+        self.failed_device_paths = set()
```
xiangce: Better add corresponding description for the newly added `failed_device_paths` attribute in above "Attributes:" docstring.

insights/parsers/lsblk.py
```diff
@@ -286,6 +286,7 @@ class LSBlockPairs(BlockDevices):
                 Column Name     Key Name
                 RM              removable
                 RO              read_only
+        failed_device_paths: Set of device names that failed to get device path
```
xiangce: Better ad the type of the attributes `failed_device_paths (set)`,

PR #1049, opened 2018-03-27
replace mountpoint only once in the string

gravitypriest:


kylape:
@gravitypriest Is this fixing the issue where the file paths end up all jammed together like one big long file?

gravitypriest:
Yep. With the default mountpoint of `/` it would previously erase all instances of `/` in the archive path, ending up with the weird concatenated version.

kylape:
Can you please add that explanation to the commit message?

PR #1051, opened 2018-03-28
[vdsm_log] Removed examples using `shared` in docstring

psachin:
As we no longer use `shared` in 3.x, the lines using `shared` are removed in docstring.


psachin:
I see `redhat_release` parser also using `shared` in docstring. I'll send changes once this is OK.

PR #1057, opened 2018-03-29
New specs to support retpo kernels for upcoming security parsers&comb…

jsvob:
…iners&rules.


jsvob:
Hi @bfahr it is ok and ready to be merged now. Thank you.

Issue #1079, opened 2018-04-04
[ls_etc] Test can be merged into docstring


psachin:
The tests included in `insights/parsers/tests/test_ls_etc.py` can be merged to docstring and can be tested using `doctest`

PR #1080, opened 2018-04-04
[ls_etc] Merged tests to docstring

psachin:
File `insights/parsers/tests/test_ls_etc.py` contains tests which can
be test using `doctest`. This patch merges those tests into docstring.

I also suggest we enable `sphinx.ext.doctest` extension and use `make doctest` in shippable. This will help us to fix docstring using `shared` in examples. 

Fixes: #1079 

### DIFF
--- a/insights/parsers/tests/test_ls_etc.py
+++ b/insights/parsers/tests/test_ls_etc.py
@@ -1,0 +1,49 @@
+from insights.parsers.ls_etc import LsEtc
+from insights.tests import context_wrap
+
+
+LS_ETC = """
+/etc/sysconfig:
+total 96
+drwxr-xr-x.  7 0 0 4096 Jul  6 23:41 .
+drwxr-xr-x. 77 0 0 8192 Jul 13 03:55 ..
+drwxr-xr-x.  2 0 0   41 Jul  6 23:32 cbq
+drwxr-xr-x.  2 0 0    6 Sep 16  2015 console
+-rw-------.  1 0 0 1390 Mar  4  2014 ebtables-config
+-rw-r--r--.  1 0 0   72 Sep 15  2015 firewalld
+lrwxrwxrwx.  1 0 0   17 Jul  6 23:32 grub -> /etc/default/grub
+
+/etc/rc.d/rc3.d:
+total 4
+drwxr-xr-x.  2 0 0   58 Jul  6 23:32 .
+drwxr-xr-x. 10 0 0 4096 Sep 16  2015 ..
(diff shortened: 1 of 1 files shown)


insights/parsers/ls_etc.py
```diff
@@ -29,12 +31,12 @@
     ... lrwxrwxrwx.  1 0 0   15 Jul  6 23:32 S97rhnsd -> ../init.d/rhnsd
     ... '''
     >>> ls_etc = LsEtc(context_wrap(LS_ETC))
```
PaulWay: I thought the whole point of the new documentation examples was that we weren't importing the parser or `context_wrap` and creating the parser arbitrarily, we were showing that (in this case) the `ls_etc` parser object exists by a type check, e.g.:
```python
>>> type(ls_etc)
<class 'insights.parsers.ls_etc.LsEtc'>
```
When using `doctest.testmod` in the tests, this would then be simulated by doing the `context_wrap` and parser object creation in a global environment for the test module.  Look for `testmod` in other parser tests for examples of how this can be done.

insights/parsers/ls_etc.py
```diff
@@ -43,20 +45,22 @@
     []
     >>> ls_etc.total_of("/etc/sysconfig")
     96
+    >>> ls_etc.dir_entry('/etc/sysconfig', 'grub')
```
PaulWay: One of the downsides of using `doctest` is that it compares literal string outputs, rather than semantic object representation.  Thus, the problem with testing the output of dictionaries is that the ordering of elements is arbitrary and can change.  It is better to test the type of the `dir_entry` return and check the existence of and look up items in the dictionary - and these are also more likely uses of the parser object in rules anyway.

insights/parsers/ls_etc.py
```diff
@@ -71,3 +75,8 @@
 class LsEtc(FileListing):
     """Parses output of ``ls -lanR /etc`` command."""
     pass
+
+
+if __name__ == '__main__':
```
PaulWay: I would strongly discourage this usage as very few people run the doctests manually.  Instead, one can test the documentation using an explicit test, e.g.:
```python
def test_ls_etc_documentation():
    failed, total = doctest.testmod(ls_etc, globs={
        'ls_etc': LsEtc(context_wrap(LS_ETC_DOCUMENTATION))
    })
    assert failed == 0
```
When the tests are run, if any of the example code fails then the user will be alerted.

psachin:
Closing this PR as it was made when base repository was private. Please see new PR https://github.com/RedHatInsights/insights-core/pull/1092

PR #1083, opened 2018-04-08
Fix bugs in IPTablesConfiguration

JoySnow:
The old code failed to pass the newly added test-data.
    Code is modified to fix the bugs.


xiangce:
This is from an old PR which is approved before BFG but closed after that. However, I cannot find the close history today, will go ahead and merge it.

PR #1084, opened 2018-04-08
Fix ValueError of smbstatus for special content

xiangce:
- raise ParseException when output is empty or does not contain header line
- refine the docstring as per 3.0

### DIFF
--- a/insights/parsers/smbstatus.py
+++ b/insights/parsers/smbstatus.py
@@ -49,6 +49,10 @@ class SmbstatusS(Statuslist):
         >>> smbstatuss_info.data[0]['pid']
         '13668'
 
+    Raises:
+        ParseException: When there is no usefull data or the input content is
+        empty,  or does contain the header line.
+
     Attributes:
         data (list): List of dicts, where the keys in each dict are the column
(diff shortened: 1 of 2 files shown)


insights/parsers/smbstatus.py
```diff
@@ -50,36 +22,81 @@ def __iter__(self):
...
+        >>> smbstatuss_info.data[0]
+        {'Signing': '-', 'Service': 'share_test', 'Encryption': '-', 'pid': '13668', 'Machine': '10.66.208.149', 'Connected_at': 'Wed Sep 27 10:33:55 AM 2017 CST'}
+        >>> smbstatuss_info.data[0]['pid']
+        '13668'
 
-        The format of ``smbstatus -S`` is like table, and function `parse_fixed_table`
-        could parse it.
+    Raises:
+        ParseException: When there is no usefull data or the input content is
+        empty,  or does contain the header line.
```
bfahr: @xiangce you need to indent the second line here and on line 91 below.

xiangce: Done, Thanks @bfahr

PR #1089, opened 2018-04-10
Parser for `/etc/vmware-tools/tools.conf`

psachin:
A configuration file parser based on IniConfigFile.

I've also included doctest based on conversation in https://github.com/RedHatInsights/insights-core/pull/1080


insights/parsers/vmware_tools_conf.py
```diff
@@ -0,0 +1,44 @@
...
+=======================================================
+
+The VMware tools configuration file ``/etc/vmware-tools/tools.conf``
+is in the standard 'ini' format and is read by the IniConfigFile
+parser. ``vmtoolsd.service`` provided by ``open-vm-tools`` package is
+configured using ``/etc/vmware-tools/tools.conf``.
+
+Examples:
+
+    >>> from insights.parsers.vmware_tools_conf import VMwareToolsConf
```
PaulWay: I'd suggest removing this kind of test set-up code from the examples, as it's not really how a rule writer would use the parser.  Instead, put the sample content as a separate section (headed 'Sample tools.conf file::' or similar), copy its content into the test file, and use the `globs` option to `doctest.testmod` to set an environment which contains a `result` object (preferably finding a better name for it :-) ) that can then be tested.

PaulWay: Bonus points will be awarded for having the same sample content in the parser documentation as the configuration text in the test module :-)

psachin: @PaulWay I implemented your suggestion. Please have a look. I wish if can avoid duplicating sample content in docstring as well as in test. I looked into `doctest.DocTestFinder()` but as the doc says it looks for lines starting with `>>>` or `...` I think it is not a good idea to include sample content as

```
>>> CONF = '''
... CONTENT
'''
``` 

Hence at this moment, I'm happy with current implementation that also test docstring within tests.

PR #1092, opened 2018-04-11
[ls_etc] Test docstring using pytest

psachin:
The idea was to test 'Examples' section in docstring with minimal changes possible. Most of the discussion & suggestions can be found in https://github.com/RedHatInsights/insights-core/pull/1080 and https://github.com/RedHatInsights/insights-core/pull/1089

Fixes: #1079

PR #1093, opened 2018-04-12
Add gnocchi module

chenlizhong:
- GnocchiConf
- GnocchiMetricdLog

### DIFF
--- a/insights/parsers/gnocchi.py
+++ b/insights/parsers/gnocchi.py
@@ -62,7 +62,6 @@ class GnocchiConf(IniConfigFile):
 
     Examples:
 
-        >>> conf = shared[GnocchiConf]
         >>> conf.sections()
         ['api', 'archive_policy', 'indexer', 'metricd', 'oslo_middleware', 'oslo_policy', 'statsd', 'storage', 'keystone_authtoken']
         >>> 'storage' in conf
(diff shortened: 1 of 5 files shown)


insights/parsers/tests/test_gnocchi.py
```diff
@@ -0,0 +1,66 @@
...
+    gnocchi_conf = GnocchiConf(context_wrap(GNOCCHI_CONF))
+    assert "storage" in gnocchi_conf.sections()
+    assert gnocchi_conf.get("indexer", "url") == "mysql+pymysql://gnocchi:exampleabckeystring@192.168.0.1/gnocchi?charset=utf8"
+    assert gnocchi_conf.getint("statsd", "flush_delay") == 10
+
+
+def test_metrics_log():
+    log = GnocchiMetricdLog(context_wrap(METRICD_LOG))
+    assert len(log.get('INFO')) == 3
+    assert len(list(log.get_after(datetime(2017, 04, 12, 19, 36, 38)))) == 1
```
bfahr: @chenlizhong would you mind to go ahead and add the tests for your examples?

insights/parsers/gnocchi.py
```diff
@@ -0,0 +1,109 @@
...
+        username=gnocchi
+        password=yourpassword23432
+        user_domain_name=Default
+        project_name=services
+        project_domain_name=Default
+
+
+    Examples:
+
+        >>> conf = shared[GnocchiConf]
```
xiangce: `shared` is no longer used in 3.x. please modify the examples as per 3.0

insights/parsers/tests/test_gnocchi.py
xiangce: @chenlizhong - I think @bfahr might mean you'd better add the test for the `Examples` in docstring to the unit test with e.g. `doctest`

insights/parsers/gnocchi.py
psachin: Hi @chenlizhong Sorry for jumping in but I see in some of parsers we have started including doctest. Please refer https://github.com/RedHatInsights/insights-core/pull/1092/files#diff-340ae940da9a54f85293da193efe7a17R54. This is a good time to include those in this parser as well.

PR #1094, opened 2018-04-16
parser for ls_osroot

TripleJQK:


TripleJQK:
https://github.com/RedHatInsights/insights-frontend-assets/issues/28

insights/parsers/ls_tmp.py
```diff
@@ -0,0 +1,52 @@
+"""
+LsTmp - command ``ls -ld /tmp``
+===============================
+
+The ``ls -l /tmp`` command provides information for only the ``/tmp`` directory.
```
bfahr: Should it be `-ld` instead of `-l`?

docs/components.rst
```diff
@@ -0,0 +1,6849 @@
+Components Cross-Reference
```
bfahr: @TripleJQK this file is generated automatically by the doc build process and should not be checked into the repo.  Please remove it from your PR.

insights/parsers/ls_tmp.py
```diff
@@ -0,0 +1,52 @@
...
+            /dummy-sysroot:
+            drwxrwxrwt. 8 root root 160 Apr 16 12:37 tmp
+
+        """
+        if len(content) > 0 and not (content[0].startswith('/') and content[0].endswith(':')):
+            # for the output "drwxrwxrwt. 8 root root 160 Apr 16 12:37 /tmp", the '/tmp' should be replaced by 'tmp', or it will not pass the FileListing check
+            # for ls -ld /tmp, can not analysis the raw_list without line "/dir:", so append "dummy-sysroot" at head
+            content[0] = content[0].replace('/tmp', 'tmp')
+            content.insert(0, '/dummy-sysroot:')
+        super(LsTmp, self).parse_content(content)
```
bfahr: I'm wondering if it would be clearer to just parser this directly?  Alternatively, you could overload the methods that will provide useful data since not all of the methods in `FileListing` will return useful info.  Since you are explicitly at the `tmp` directory only and not the contents of the directory is `FileListing` the appropriate base class?

TripleJQK: @bfahr I use the `Filelisting` as all series of `ls` parser are using, and I want to make all of them identical. If you think it is not necessary, I will think about to use other base class. May I have your further advice?

docs/components.rst
TripleJQK: I also think it is a automatic created file, as I have never modify this file by hands. When did the git status, I was also a little surprised that why the file is in the list.... I will remove this

insights/parsers/ls_tmp.py
TripleJQK: sorry ,my misstake

insights/parsers/ls_tmp.py
xiangce: @TripleJQK @bfahr, Just came into my mind, do you guys think `LsTmp` is a bit limited? Once in the future, there is rule would like to check the permission settings of other directories, such as `/var` or `/opt`, we need to add individual parsers `LsVar` `LsOpt`. How about up it a level to `ls -lan /` (LsROOT)? How do you think?

TripleJQK: @xiangce @bfahr I have thinked about this ,then this will be `ls -lan /`. and:
~~~
insights_archive.py:    ls_var_log = simple_file("insights_commands/ls_-la_.")
~~~
about `ls_la_.`, will there be any protential risk on core code or just looks weird?

xiangce: @TripleJQK - by using the `mangle` command, I got `insights_commands/ls_-lan` but not `ls_-lan_.`

TripleJQK: OK

TripleJQK: @xiangce @bfahr I think the mangle give a bad return:
~~~
└> mangle "ls -lan /"
insights_commands/ls_-lan
~~~
As:
In the insights/core/__init__.py, the code for `FileListing`:
~~~
:
1119         self.first_path = None
1120         path_re = re.compile(r'ls_-\w+_(?P<path>.*)$')
1121         match = path_re.search(context.path)
1122         if match:
1123             self.first_path = match.group('path').replace('.', '/').replace('_', ' ')
1124         super(FileListing, self).__init__(context)
:
1174     def parse_content(self, content):
1175         """
1176         Called automatically to process the directory listing(s) contained in
1177         the content.
1178         """
1179         listings = {}
1180         this_dir = {}
1181         for line in content:
1182             l = line.strip()
1183             if not l:
1184                 continue
1185             if (l.startswith('/') and l.endswith(':')) or (self.first_path):
1186                 # Directory name from output first and context path second
1187                 if (l.startswith('/') and l.endswith(':')):
1188                     name = l[:-1]
1189                 else:
1190                     name = self.first_path
1191                 # Unset the first path so we don't come here again.
1192                 self.first_path = None
1193                 # New structures for a new directory
1194                 this_dir = {'entries': {}, 'files': [], 'dirs': [],
1195                             'specials': [], 'total': 0, 'raw_list': [],
1196                             'name': name}
1197                 listings[name] = this_dir
1198             elif l.startswith('total') and l[6:].isdigit():
1199                 this_dir['total'] = int(l[6:])
1200             elif not this_dir:
1201                 # This state can happen if processing an archive that filtered
1202                 # a file listing due to an old spec definition.
1203                 # Let's just skip these lines.
1204                 continue
1205             else:
1206                 self.parse_file_match(this_dir, l)

~~~
I think the `context.path` is based on `insights_commands/ls_xxxx`, if use `insights_commands/ls_-lan`, the line 1122 and 1123 will not work and the `self.path_first` could still be None, and make the `parse_content` not work as expected.
So I think the correct one should be `insights_commands/ls_-la_.`

xiangce: @TripleJQK - Good finding! I will look into it to check if it's just an issue of the `mangle` script only. If the client really mangled the simple command `ls -lan /` as `ls_-lan`, we need to update the `reg-exp` logic in `FileListing`.

xiangce: @bfahr @csams - for this issue, the current `FileListing` works well when the simple command of `ls -lan /` is defined as `insights_commands/ls_-la_.` (it's what @TripleJQK just implemented in this patch).
But the `mangle` script strips all the trailing ` . _ -`, and it mangled the simple command as `insights_commands/ls_-la` which will cause the current  `FileListing` failed to get the `self.first_path`.

Can we just forcibly define it as `insights_commands/ls_-la_.` regardless of the mangled result? I‘m not sure if this definition which is different with the mangled result will bring any implicit risks to the future @datasource.

Or if you have any better solutions?

insights/specs/insights_archive.py
```diff
@@ -98,6 +98,7 @@ class InsightsArchiveSpecs(Specs):
     ls_var_log = simple_file("insights_commands/ls_-la_.var.log_.var.log.audit")
     ls_var_www = simple_file("insights_commands/ls_-la_.dev.null_.var.www")
     ls_var_tmp = simple_file("insights_commands/ls_-ln_.var.tmp")
+    ls_osroot = simple_file("insights_commands/ls_-lan")
```
xiangce: Please modify it to `insights_commands/ls_-lan.`

PR #1096, opened 2018-04-16
Foreman prod log

lhuett:
Added spec (foreman_production_log) to sos_archive.py


lhuett:
Reverted the formatting done in sos_archive.py

PR #1115, opened 2018-04-19
Update the file_path attribute in Parser class to actual path in host

chenlizhong:
- previous file_path is the file path -- archive extract directory path
- Update failed tests


chenlizhong:
Parser is logic layer. It do not need to care about extract file path. So update the path to relative_path.

insights/core/__init__.py
```diff
@@ -60,7 +60,7 @@ class Parser(object):
     """
 
     def __init__(self, context):
-        self.file_path = context.path
+        self.file_path = os.path.join("/", context.relative_path) if context.relative_path is not None else None
```
xiangce: Should we just add a `self.relative_path` to `Parser` and leave the `self.file_path` as it was?

csams:
Should we have the / prefix on the relative paths?

insights/core/__init__.py
chenlizhong: Originally, we think keep these two. But when I do the update. I think file_path and relative_path that will confuse others that consume the parser. In fact, the file_path in context is just exact temp dir, to evaluate one same archive two times, the value of file_path will be different also. So do not keep the value.  just keep relative_path as path here.

The parser

xiangce: Yep, currently, I also cannot imagine what the path that contains the temp dir can be used for yet. That makes sense to me.

PR #1116, opened 2018-04-19
Enhance parser fstab with method match_device_of_path()

JoySnow:

### DIFF
--- a/insights/parsers/tests/test_fstab.py
+++ b/insights/parsers/tests/test_fstab.py
@@ -99,3 +99,30 @@ def test_fstab():
     assert sitedata_mount.fs_mntops['defaults'] is True
     assert sitedata_mount.fs_vfstype == "nfs"
     assert sitedata_mount.fs_spec == "192.168.48.65:/cellSiteData"
+
+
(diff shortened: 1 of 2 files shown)


insights/parsers/tests/test_fstab.py
```diff
@@ -99,3 +99,30 @@ def test_fstab():
...
+
+
+def test_match_device_of_path():
+    fstab_info = fstab.FSTab(context_wrap(FSTAB_DEVICE_PATH_TEST_INFO))
+    path_device_map = {'/var/crash': '/dev/sdb3',
+                       '/var/some/path': '/dev/sdb2',
+                       '/kdump/crash': 'UUID=qX0bSg-p8CN-cWER-i8qY-cETN-jiZL-LDt93V',
+                       '/some/path': '/dev/sda2',
+                       '/lv_test': '/dev/mapper/vgext-lv--test',
+                       '/lv': None,
```
xiangce: Is the `/lv` different with `/some/path`? The result of `/lv` should be `/dev/sda2`, right?

JoySnow: Yes, you are right. I will change the function's logic.

PR #1118, opened 2018-04-19
Fix (spec_factory.py, subproc.py) to handle special characters in fil…

lhuett:


insights/util/subproc.py
```diff
@@ -79,26 +79,35 @@ def call(cmd, timeout=None, signum=signal.SIGKILL, shell=False, stdout=subproces
...
-            if len(spltcmd) > 1:
-                cmd = shlex.split(spltcmd[0])
-                cout = Popen(cmd, stdout=stdout)
-                del spltcmd[0]
-
-                for next in spltcmd:
-                    nxt = shlex.split(next)
-                    cout = Popen(nxt, stdout=stdout, stderr=stderr, stdin=cout.stdout)
+            if not clist:
+                spltcmd = cmd.split("|")
```
csams: I'd take the split on pipe out since it'll break if the regular string has a pipe in a quoted string or something. Calls can just pass a list of lists for piping so we don't have to worry about parsing.

lhuett: I left the split in there for backward compat. The original version of subproc had the split in it 'if not shell' so I figured there may be some other things using subproc passing a string with a pipe that count on the split. Just didn't want to break something.

lhuett: Removed split on pipe from subproc.

PR #1119, opened 2018-04-20
Enhance parser KDumpConf

JoySnow:
Fix the wrong logic of 'using_local_disk' property.
  Add two more properties to KDumpConf.
  Adjust the test-cases.

### DIFF
--- a/insights/parsers/kdump.py
+++ b/insights/parsers/kdump.py
@@ -18,6 +18,7 @@ SysconfigKdump - file ``/etc/sysconfig/kdump``
 import re
 from urlparse import urlparse
 from .. import Parser, parser
+from insights.parsers import ParseException
 from insights.specs import Specs
 
 
@@ -63,7 +64,7 @@ class KDumpConf(Parser):
         True
     """
     NET_COMMANDS = set(['nfs', 'net', 'ssh'])
-    FS_TYPES = ['ext2', 'ext3', 'ext4', 'btrfs', 'xfs']
+    SUPPORTED_FS_TYPES = ['ext2', 'ext3', 'ext4', 'btrfs', 'xfs']
 
     def parse_content(self, content):
         lines = list(content)
@@ -114,6 +115,7 @@ class KDumpConf(Parser):
         self.data = items
         self.comments = comments
         self.inline_comments = inline_comments
+        self._check_target_conflict()
 
     def options(self, module):
         """
@@ -206,17 +208,16 @@ class KDumpConf(Parser):
         """
         Is kdump configured to only use local disk?
 
-        More than one dump targets will lead to kudmp service start failure.
-        So let's suppose that only one target is set here.
+        Several target types:
 
-        The logic used here is:
-
-        * If 'raw' is given as an option, then the dump is local.
+        * If 'raw' is given, then the dump is local.
         * If 'ssh', 'net', 'nfs', or 'nfs4' is given, then the dump is NOT local.
         * If '<fs type> <partition>' is given, then the dump is local.
         * Otherwise, the dump is local.
+
+        Since only one target could be set, the logic used here is checking
+        if remote target is used, return True for not.
         """
-        # Simplify the logic to check by:
         return not ('ssh' in self.data or 'net' in self.data or
                     'nfs' in self.data or 'nfs4' in self.data)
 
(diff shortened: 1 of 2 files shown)


insights/parsers/kdump.py
```diff
@@ -205,19 +206,37 @@ def using_local_disk(self):
...
         * Otherwise, the dump is local.
         """
-        # The previous version used iteration across self.data.keys(), which
-        # is of course non-repeatable because hash key ordering may change.
-        # So we reverted to logic.
-        return ('raw' in self.data) or (not (
-            'ssh' in self.data or 'net' in self.data or
-            'nfs' in self.data or 'nfs4' in self.data)
-        )
+        # Simplify the logic to check by:
```
xiangce: Though the code looks being simplified, the efficiency is different, better revert the check for `raw`.

insights/parsers/kdump.py
```diff
@@ -205,19 +206,37 @@ def using_local_disk(self):
...
 
-        The logic used here is the first of these conditions:
+        More than one dump targets will lead to kudmp service start failure.
+        So let's suppose that only one target is set here.
+
+        The logic used here is:
 
         * If 'raw' is given as an option, then the dump is local.
         * If 'ssh', 'net', 'nfs', or 'nfs4' is given, then the dump is NOT local.
+        * If '<fs type> <partition>' is given, then the dump is local.
```
xiangce: Don't you want to implement this logic here? but in fact its in `is_using_local_fs`

insights/parsers/kdump.py
```diff
@@ -205,19 +206,37 @@ def using_local_disk(self):
...
+        """
+        Return (fs_type, partation) as a tuple if set, else None.
+        https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/kernel_administration_guide/kernel_crash_dump_guide#sect-supported-kdump-targets
+        """
+        for t in self.FS_TYPES:
+            if t in self.data:
+                return (t, self.data[t])
+
+    @property
+    def is_using_local_fs(self):
```
xiangce: What's the different between `is_using_local_fs` and `using_local_disk`?

insights/parsers/kdump.py
```diff
@@ -63,6 +63,7 @@ class KDumpConf(Parser):
         True
     """
     NET_COMMANDS = set(['nfs', 'net', 'ssh'])
+    FS_TYPES = ['ext2', 'ext3', 'ext4', 'btrfs', 'xfs']
```
xiangce: Better name it as `SUPPORTED_FS_TYPES`

insights/parsers/kdump.py
```diff
@@ -226,17 +227,19 @@ def fs_and_partation(self):
...
         """
-        Is kdump configured to use local file system as dump target?
-        Tell by if '<fs type> <partition>' is set.
+        More than one dump targets will lead to kudmp service start failure.
+        So raise an exception here if more than one target is set here.
         """
-        return bool(self.fs_and_partation)
+        used_targets = (self.is_ssh(), self.is_nfs(),
+                        'raw' in self, bool(self.fs_and_partation))
+        if len([v for v in used_targets if v]) > 1:
```
xiangce: A more simple and pythonic way:
```
if len(filter(None, used_targets)) > 1:
```

insights/parsers/kdump.py
```diff
@@ -114,6 +115,7 @@ def parse_content(self, content):
         self.data = items
         self.comments = comments
         self.inline_comments = inline_comments
+        self._check_target_conflict()
```
xiangce: Looks good, but for this purpose, it could be implemented as an inner function but not a inner member function.
~~~
    def check_target_conflict():
          xxx
    xxx
    check_target_conflict()
~~~

insights/parsers/kdump.py
```diff
@@ -237,10 +237,20 @@ def fs_and_partation(self):
         """
         Return (fs_type, partation) as a tuple if set, else None.
         https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/kernel_administration_guide/kernel_crash_dump_guide#sect-supported-kdump-targets
+        If more than one target is configured, kudmp service fails to start.
         """
-        for t in self.SUPPORTED_FS_TYPES:
-            if t in self.data:
-                return (t, self.data[t])
+        targets = []
```
xiangce: I think the original `fs_and_partation` is good enough. in the contrary better do the conflict check in `parse_content`

insights/parsers/kdump.py
```diff
@@ -205,19 +218,39 @@ def using_local_disk(self):
...
-        # is of course non-repeatable because hash key ordering may change.
-        # So we reverted to logic.
-        return ('raw' in self.data) or (not (
-            'ssh' in self.data or 'net' in self.data or
-            'nfs' in self.data or 'nfs4' in self.data)
-        )
+        return not ('ssh' in self.data or 'net' in self.data or
+                    'nfs' in self.data or 'nfs4' in self.data)
+
+    def _check_target_conflict(self):
```
xiangce: As an effort, the `self.target` is got from this method. Better return it as a return value directly.

Btw, how do you think to rename it to `_parse_target`?, e.g.:
```
self.target = self._parse_target
```

insights/parsers/kdump.py
```diff
@@ -32,6 +33,7 @@ class KDumpConf(Parser):
         data (dict): a dictionary of options set in the data.
         comments(list): fully commented lines
         inline_comments(list): lines containing inline comments
+        target(tuple): target line parsed as a (x, y) tuple if set, else None
```
xiangce: The double colons "::" in line-30 must be a typo, would you please correct it in your patch?

PR #1124, opened 2018-04-23
Update AttributeDict

xiangce:
- Try not deprecate the AttributeDict to minimize the effect to plugins
- Only the `keys` included in `fixed_attrs` will be set as attributes
  Because of this, the tests of some rules in insights-plugins are broken, will fix them soon

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -1289,7 +1289,10 @@ class FileListing(Parser):
 
 class AttributeDict(dict):
     """
-    Class to convert the access to the keys in ``attrs`` dictionary as attributes.
+    Class to convert the access to each item in a dict as attribute.
+
+    .. warning::
+        Deprecated class, please set attributes explicitly.
 
     Examples:
         >>> data = {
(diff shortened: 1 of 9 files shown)


insights/tests/test_attribute_dict.py
```diff
@@ -1,19 +1,33 @@
...
     assert d_obj.get('relatime') is True
-    assert d_obj.nosuid is True
     assert 'uid' in d_obj
     assert d_obj.get('mode', None) == '0400'
-    assert d_obj.get('xmode', None) is None
-    for d, v in d_obj:
-        if d.startswith('nod'):
-            assert v is True
+    assert 'mode' in d_obj
+    assert 'xmode' not in d_obj
```
xiangce: This looks a little weird that `xmode` not in the obj but we could get the value of `obj.xmode`  :cry:

insights/core/__init__.py
```diff
@@ -1289,34 +1293,67 @@ def raw_directory(self, directory):
...
 
+    Raises:
+        TypeError: When more than one directories are passed
+        ValueError: When no or an empty 'fixed_attrs' is passed
+
+    """
+    type_info = namedtuple('type_info', field_names=['type', 'default'])
+    """namedtuple: Type for the values of the ``fixed_attrs`` dictionary"""
     def __init__(self, *args, **kwargs):
+        if len(args) > 1:
```
bfahr: @xiangce if args can only be 1, couldn't you just use a variable name for the parameter?

insights/parsers/fstab.py
```diff
@@ -22,13 +22,28 @@
...
 This data, as above, is available in the ``data`` property:
 
-* As wrapped as an AttributeDict, each column can also be accessed as a property
-  with the same name.
-* The mount options are also an AttributeDict object with properties
-  corresponding to the common mount options.
-
-The data for each mount point is also available via the ``mounted_on``
-property; the data is the same as that stored in the ``data`` list.
+* As wrapped as an :class:`FSTabEntry`, each column can also be accessed as a
```
bfahr: `Wrapped as an` or `As an :class`

insights/parsers/fstab.py
```diff
@@ -22,13 +22,28 @@
...
+  property with the same name.
+* The `` fs_mntops`` is wrpped as a :class:`insights.parsers.mount.MountOpts`
```
bfahr: `is wrapped as a`

insights/parsers/mount.py
```diff
@@ -10,30 +10,40 @@
...
+are listed in the same order as in the command output:
+
+ * ``filesystem`` - Name of filesystem
+ * ``mount_point`` - Name of mount point for filesystem
+ * ``mount_type`` - Name of filesystem type
+ * ``mount_options`` - Mount options as a dictionary
+ * ``mount_label`` - Only present if optional label is present
+ * ``mount_clause`` - Full string from command output
+
+The `` mount_options`` is stored as a :class:`MountOpts` object contains below
```
bfahr: Space before `mount_options` disables the rendering emphasis in docs.

insights/core/__init__.py
xiangce: That makes sense to me, will modify it. Thanks

insights/parsers/mount.py
```diff
@@ -10,30 +10,40 @@
...
+* ``relatime`` - Inode access times relative to modify or change time
+* ``seclabel`` - `seclabel` is enabled or not
+* ``attr2`` - `opportunistic` improvement is enabled or not
+* ``inode64`` - `inode64` is enabled or not
+* ``noquota`` - Disk quotas are enforced or not
+
+For instance, the option ``rw`` in ``(rw,dmode=0500)`` may be accessed as
+``mnt_row_info.rw`` with the value ``True``, but the ``dmode`` can only be
+accessed as ``mnt_row_info.get('dmode')`` or ``mnt_row_info['dmode']`` with the
+value ``0500``.
```
bfahr: @xiangce why wouldn't you just add `dmode` (and other non-fixed attributes) to the attributes instead of changing to dictionary access, to maintain compatibility with the previous version?

PaulWay:
Hi @xiangce - I appreciate the work you're doing here but to me you seem to be doing a lot of work around a very simple problem.  The key incompatibility with AttributeDict is that it uses `iteritems()` (which provides key,value pairs) to provide an `__iter__` interface) which normally iterates through keys only).

One solution to this would be:
https://github.com/RedHatInsights/insights-core/pull/1140
which simply removes the override of the `__iter__` method in AttributeDict so it behaves exactly as a dictionary normally does.  The downside to this is that it breaks any rule which has been written to expect the `iteritems()`-like interface.  The upside is that once that override is removed the interface is more consistent.

Another solution to this is simply to make `unordered_compare` iterate specifically over the keys of the dictionary, which AttributeDict does not override.  This solution would look like:
https://github.com/RedHatInsights/insights-core/pull/1122

Let me know what you think!

insights/parsers/mount.py
xiangce: @bfahr - just wanted to only set those explicitly documented `fixed attributes` as attributes (try to fix the issue that Mike raised in email), so I didn't add these options no in the `fixed attributes`, such as `dmode`, to attributes.

Now I fixed it as all keys in `data` and `fixed attributes` will be added as attributes.

xiangce:
@PaulWay @bfahr - I deprecated the AttributeDict finally (please forgive my stubborn persistence just want to minimize the effect to the existed plugins) please check it again, thanks.

xiangce:
Working on fixing the broken of plugins. It's a bit complex.

PR #1127, opened 2018-04-24
WIP: Support old specs

kylape:
The code is still in "debug mode", meaning I modified the actual execution of `InsightsSpec` objects to just print the command that would have been executed.  You can see the test running I wrote in `test_collector.py`.

The current state is that it properly reads `files` and `commands`, but it does not do anything with the mountpoint yet.

Testing should be done against https://github.com/RedHatInsights/insights-frontend-assets/blob/symbolic_names/uploader.json


kylape:
"Debug mode" has been removed with this last commit.

PR #1128, opened 2018-04-25
Add parser for odbc.ini and odbcinst.ini

JoySnow:


insights/parsers/odbc.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+
+from .. import parser, IniConfigFile, add_filter
+from insights.specs import Specs
+
+filter_list = [
+        '[',
+        'driver',
+        'server',
+        'no_sspa',
```
xiangce: As your test data for ODBCIni, there are "Driver", "SERVER", "NO_SSPS" but in your filter all of them is lowercase, this will cause your excepted lines absent in the archive. Please double confirm it.

PR #1131, opened 2018-04-25
Nuking 2016

gravitypriest:


insights/client/client.py
```diff
@@ -243,8 +241,7 @@ def get_branch_info():
     # or we are analyzing a running container/image
     # or tar file, mountpoint, simply return the default branch info
     if (config['offline'] or
-            config['analyze_container'] or
```
brantleyr: I still see some references to `analyze_container.` Does that affect execution?

brantleyr: NVM, it looks like you merged this config/flag with some other container config/flag stuff correct?

gravitypriest: Yup. Most of the things from the deprecated `container_mode` switch were changed to `analyze_container` or a related flag. I switched or deleted the remaining ones.

kylape:
```
./insights/client/client.py:268:20: E121 continuation line under-indented for hanging indent
./insights/client/client.py:276:20: E121 continuation line under-indented for hanging indent
./insights/client/client.py:290:9: F841 local variable 'targets' is assigned to but never used
./insights/client/client.py:371:13: F841 local variable 'machine_id' is assigned to but never used
./insights/client/containers.py:313:5: E303 too many blank lines (2)
./insights/client/utilities.py:17:1: F401 'config.CONFIG as config' imported but unused
```

insights/client/data_collector.py
```diff
@@ -234,38 +190,36 @@ def run_collection(self, conf, rm_conf, branch_info):
...
-        self._write_branch_info(conf, branch_info)
-        self._write_analysis_target_id(conf)
+        self._write_branch_info(branch_info)
         logger.debug('Metadata collection finished.')
 
     def done(self, conf, rm_conf):
         """
         Do finalization stuff
         """
-        self._write_uploader_log(conf)
```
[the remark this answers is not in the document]
gravitypriest: Rather, I think we ought to add `/var/log/insights-client/insights-client.log` to uploader.json, but I am fine with either.

insights/client/client.py
```diff
@@ -325,135 +306,89 @@ def collect(rc=0):
         return False
 
     collection_rules, rm_conf = pc.get_conf(False, stdin_config)
-    individual_archives = []
-
-    for t in targets:
```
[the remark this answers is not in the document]
gravitypriest: This was for the uber-archive yeah. We wanted to include data from the hosts + the images and containers on it in a single upload.

insights/client/client.py
```diff
@@ -325,135 +306,89 @@ def collect(rc=0):
...
+            target_type = target['type']
+        logger.debug("Inferring target_type '%s' for SPEC collection", target_type)
+        logger.debug("Inferred from '%s'", target['type'])
+        dc = DataCollector(archive, mountpoint=mp)
 
-            machine_id = generate_analysis_target_id(t['type'], t['name'])
-            archive_meta['system_id'] = machine_id
-            archive_meta['machine_id'] = machine_id
+        logger.info('Starting to collect Insights data for %s', logging_name)
+        dc.run_collection(collection_rules, rm_conf, branch_info)
```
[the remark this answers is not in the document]
gravitypriest: Agreed

insights/client/client.py
```diff
@@ -325,135 +306,89 @@ def collect(rc=0):
...
+                logger.error('Could not open %s for analysis', logging_name)
+                return False
 
-                from containers import docker_display_name
-                archive_meta['display_name'] = docker_display_name(
-                    t['name'], t['type'].replace('docker_', ''))
-                logger.debug('Docker display_name: %s', archive_meta['display_name'])
-                logger.debug('Docker docker_id: %s', archive_meta['docker_id'])
+        # analyze docker containers
+        elif target['type'] == 'docker_container':
```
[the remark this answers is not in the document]
gravitypriest: Nope. I misunderstood the `--analyze-container` option before clarifying with @brantleyr . I'll take this out.

insights/client/containers.py
```diff
@@ -347,11 +309,6 @@ def _docker_all_container_ids():
         the_verbiage = "Docker"
         the_exception = HaveDockerException
 
-    def get_targets():
-        logger.error('Could not connect to ' + the_verbiage + ' to collect from images and containers')
-        logger.error(the_verbiage + ' is either not installed or not accessable: %s' %
-                     (the_exception if the_exception else ''))
-        return []
```
gravitypriest: note to myself: re-add this default func back in

insights/client/utilities.py
```diff
@@ -89,11 +88,6 @@ def generate_machine_id(new=False,
     machine_id_file = None
     logging_name = 'machine-id'
 
-    if docker_group:
-        # generate a docker group ("docking station") id
-        #   for the API to parse this group of systems
-        destination_file = constants.docker_group_id_file
```
gravitypriest: note to myself: remove unused constants like this one

gravitypriest:
Things that were removed and why:

**"Spec" collection**

**"Analysis target ids"**
Originally we thought images & containers should have their own unique IDs, based on the namespace of the host's machine-id. This is no longer the case, and we can go back to just generating machine-ids for the host like the old days. This is also the case for "docker group" machine-ids.

**The `--container-mode` option**
This was originally purposed for collecting data about a host & the images/containers that live on it. 

Since then, that requirement has become null and void. Much of the code related to it has been deleted or short circuited, e.g., `get_targets()` returns an array of only 1 item. Most of the control flow checks for "container mode" have been replaced with the newer "analyze image/container/mountpoint/file" options, so it can be safely removed.

Specifically, each `analyze_*` option implies `analyze_container`, so this functionality can be left completely to depending on `analyze_container`.

**Multiple analysis targets**
Related to `--container-mode` and part of the "uber archive" schema. Removed the array + loop and simplified the collection function to a single target.

**Archive metadata**
Related to `--container-mode` and part of the "uber archive" schema. Removed everything except `branch_info`, which is an Insights archive requirement.

**Cluster upload**
Removal of calls to the unsupported "cluster" API endpoint.

**Docker display name & image links**
Related to `--container-mode` and part of the "uber archive" schema. Part of the archive metadata which was also removed.

**Appending archive path and re-mangling commands with "pre-commands" in `InsightsCommand`**
This is a rather small change; `archive_path` is no longer a thing we use, and each each incoming command spec to `InsightsCommand` is an independent copy of the original command with added args. No need to treat it specially.

PR #1132, opened 2018-04-26
Add parser software_collections_list

matysek:
Updated spec to collect installed software collections. Output of command 'scl --list'.


skontar:
Looks good to me. But I think you also need a file in `shared_parsers_catalog` for docs to work.

insights/specs/insights_archive.py
```diff
@@ -169,6 +169,7 @@ class InsightsArchiveSpecs(Specs):
     sestatus = simple_file("insights_commands/sestatus_-b")
     smbstatus_S = simple_file("insights_commands/smbstatus_-S")
     smbstatus_p = simple_file("insights_commands/smbstatus_-p")
+    software_collections_list = simple_file('insights_commands/scl_--list')
```
jsvob: How do the entries in insights_archive.py work? What are the guidelines for adding/not adding entries here? What program generates the files in "insights_commands/" that are referenced here? Because I never added entries there and I don't know whether that was wrong. Thanks!

bfahr: @jsvob if you add a `simple_command` to defaults.py it will be collected as a file in the insights archive under the `insights_commands` directory.  In order to be able to access that information from an archive you need to add the appropriate `simple_file` entry to the insights_archive.py file.  You can use the command mangle.py to determine the appropriate string to use for the filename.

```
$ python insights/util/mangle.py '/usr/bin/scl --list'
insights_commands/scl_--list
```

bfahr:
@matysek we are working on instructions for additional steps you'll need to take in order to update the uploader JSON for your spec changes.  This PR will remain pending until those changes are made.  Sorry for any delay.  We'll send out the instructions on the mailing list as soon as they are complete.

insights/parsers/software_collections_list.py
```diff
@@ -0,0 +1,67 @@
+"""
+Software Collections list output - command ``scl --list``
+=========================================================================
+
+This module provides parser for list output of ``scl``.
+
+Parser provided by this module is:
+
+SoftwareCollectionsListInstalled - command ``scl --list``
+------------------------------------------------------------------------------------
```
bfahr: Since you already have this information in the title, this sub heading is not really necessary but its not necessary to change it unless you are making other changes.

bfahr:
Process for spec changes is now documented in the [wiki here](https://github.com/RedHatInsights/insights-core/wiki#spec-changes).

insights/specs/insights_archive.py
jsvob: @bfahr I have always had only the frontend-visible insights rules in mind, and these have always worked with only entries in defaults.py, and nothing in insights_archive.py. So what usecases this would help that I don't see? If it is command-line one-off analysis of an sos report, what do I do to make the tool that generates the sos report add the appropriate files? Is there documentation about this? Is it mandatory that I always add a command both to defaults.py and insights_archive.py?

Sorry for the multiple questions; I am a bit lost and I don't want to hinder wider processes I might be unaware of, and for that, I need to understand it a bit.

bfahr: @jsvob as to documentation, there are bits and pieces but I don't believe they answer your question directly.  @csams would you mind to respond to @jsvob's questions regarding requirements for spec declaration in `insights_archive.py`?

csams: @jsvob Datasources that collect files only need to be defined in `default.py` so long as the relative location of the file(s) is the same on the host, in old insights archives (produced by the client interpreting `uploader.json`), and in sos reports. For example, `var/log/messages` is in the same relative location in all three contexts, so it only needs one definition.

Command data sources are different because they actually execute something to produce data. They're not just reading it from disk. However, they aren't being run in the wild yet - the old client is still out there making archives from `uploader.json`. Reading the data captured from commands executed by the old client is the job of datasources defined in `insights_archive.py`. At the top of that file you can see that the `simple_file` and `glob_file` have been curried to run under a `HostArchiveContext`.

If you add a file collecting datasource to `default.py`, and it can be added to the `uploader.json` to produce the same file in an insights archive, then you're done. If it can't be added to uploader.json at all, then you're still done, but it won't be collected until we have a client that directly runs datasources.

If you add a *command* datasource to `default.py`, but the same command can't be added to `uploader.json`, then you're done. If the command can be added to `uploader.json`, then you need to add a datasource to `insights_archive.py` that will capture the file(s) that the old client will produce when it runs the command.

It works like this: for a given datasource that a parser uses, you have a RegistryPoint at the very top in the Specs class. The classes in default.py, insights_archive.py, and sos_archive.py provide entries that are named the same as those in the Specs class. As those classes load, the datasources they contain register themselves as implementations for their RegistryPoints.

Notice that the implementations in `default.py` typically don't specify a context. That's because commands default to `HostContext`, and file collection defaults to a list of contexts each of which provide a directory that acts as a root for the file locations (hosts, archives, exploded and mounted docker images, etc.). However, nearly all of the datasources in `insights_archive.py` and `sos_archive.py` specify a context. In fact, the helper functions are curried to a context at the tops of the modules. That's so those implementations will only run under those contexts.

If you register two implementations for the same `RegistryPoint`, and they both depend on the same execution context, then the last one to load wins. This provides a way to override a datasource without having to change core code. You just write a module and add it to the end of the plugin package list. This is really useful when developing or deploying quick workarounds for bugs in the datasources that come with core.

For more info on this stuff, check out the [Datasource Registry notebook](https://github.com/RedHatInsights/insights-core/blob/master/docs/notebooks/Datasource%20Registry.ipynb) and the New Style Specs section of the [Rule Migration Guide](https://github.com/RedHatInsights/insights-core/blob/master/docs/notebooks/Rule%20Migration%20Guide.ipynb).

jsvob: @csams Thank you for the valuable response. I *think* I *almost* get it. I don't mainly understand this thing (related to http://localhost:8888/notebooks/Rule%20Migration%20Guide.ipynb#Adding-a-New-Datasource ):

* If I add a data source to `insights/specs/insights_archive.py`, something in the Insights infrastructure that I don't see will automagically update the `uploader.json` for insights 1.x clients?
* `insights/specs/insights_archive.py` exists only because of insights 1.x clients and will go away once they are deprecated?
* If I add a data source to `insights/specs/sos_archive.py`, something somewhere will automagically change how sos reports are gathered on insights systems to contain that source?

Thank you!

csams: @jsvob 
* Updating the uploader.json isn't automatic. It has to be updated manually to reflect updates to the new style datasources. Some datasources can't be expressed in uploader.json, while others might have to be expressed differently.
* It's correct that insights_archive.py only exists to support clients that still generate archives from uploader.json. Theoretically, insights_archive.py would go away once the old client is deprecated, but practically it'll probably never happen.
* Adding a datasource to sos_archive.py just tells core how to collect files from a sosreport. It doesn't do any magic to update sos collections or anything like that.

Core inspects an archive's contents to determine whether to process it under an InsightsArchiveContext or SosArchiveContext. It manually creates the right context as the first dependency and then evaluates everything whose dependencies are met. If processing a sosreport, all datasources that depend on SosArchiveContext will try to fire.

Note that file based datasources in default.py depend on "at least one" of several contexts like HostContext, InsightsArchiveContext, SosArchiveContext, etc. and so will fire for any of them. That dependency happens automatically if a context isn't specified. Datasources in `sos_archive.py` explicitly depend on SosArchiveContext (see the curried functions at the top of the module). Since they load after default.py, if they have the same name as those in default.py and depend on the same context (SosArchiveContext in this case), they'll override those definitions when running under that context.

Since parsers depend on RegistryPoints in Specs, and RegistryPoints are just glorified first_of datasources that return the first implementation for them that succeeds, the parser will be passed the correct datasource that actually ran regardless of the context it ran under.

jsvob: @csams Thank you for the explanation. The last bit I'm missing is updating uploader.json. So I'm curious whether the following is correct for the case of adding a new command that is compatible with uploader.json (just as an example, out of the multiple cases you described in your previous post):

* add it to default.py for insights 3x clients
* add it to insights_archive.py for insights 1x clients
* add it to uploader.json for insights 1x clients by following http://post-office.corp.redhat.com/archives/insights-dev/2018-April/msg00545.html

bfahr:
@matysek that is what I was waiting on, thanks.

PR #1135, opened 2018-04-26
hide root in archive

gravitypriest:


kylape:
Thankfully we were able to come up with a solution to this on the back end.

See https://github.com/RedHatInsights/insights-core/commit/c5e3d19222db9d53615057792456c2b2bacb161d

PR #1136, opened 2018-04-27
Update Jenkinsfile

SteveHNH:
- Break into additional steps
- Add doc testing


Jenkinsfile
```diff
@@ -10,13 +10,16 @@ pipeline {
             }
           }
           steps {
+            echo "Building environment..."
             sh """
                virtualenv .
                source bin/activate
                pip install 'flake8==3.3.0' 'coverage==4.3.4' 'pytest==3.0.6' 'pytest-cov==2.4.0' 'futures==3.0.5' 'requests==2.13.0' 'six' 'wheel' 'pyyaml>=3.10,<=3.12' 'pyOpenSSL' 'importlib' 'Jinja2==2.9.6'
```
kylape: Is there a reason you don't remove this pip install but you remove the pip installs in the other steps?

Jenkinsfile
```diff
@@ -10,13 +10,16 @@ pipeline {
...
-               pytest
-               flake8 
-            """
+               """
+            echo "Testing with Pytest..."
+            sh 'source bin/activate && pytest'
```
kylape: I wonder if you could just do `bin/pytest` rather than `source bin/activate && pytest`

Jenkinsfile
SteveHNH: The rhel6 and the rhel7 (python 3.6) pip lines had issues with just doing `pip install '.[develop]'. For instance, RHEL6 didn't support Sphinx, and Python 3.6 didn't support importlib. Importing needed packages individually was the workaround I came up with.

Jenkinsfile
SteveHNH: that should work. I'll update the PR

Jenkinsfile
kylape: Ah, I misread the diff.  I thought you weren't running `pip` at all on the other steps.  Aren't most of the requirements already installed in the builder image?

SteveHNH: @kylape I'm fixing the images now. That's much better plan.

PR #1140, opened 2018-04-30
AttributeDict should iterate as a dictionary

PaulWay:
AttributeDict should iterate as a dictionary, and should not override `__iter__` to provide an `iteritems()`-like interface.


insights/parsers/tests/test_fstab.py
```diff
@@ -79,7 +79,7 @@ def test_fstab():
     assert nfs_host.fs_passno == 0
     assert dev_vg0.fs_mntops.data == 'writeback'
     assert dev_vg0.raw == '/dev/mapper/vg0-lv2 /test1             ext4 defaults,data=writeback     1 1'
-    for opt, v in dev_vg0.fs_mntops:
+    for opt, v in dev_vg0.fs_mntops.iteritems():
```
jhjaggars: In order to support python3 you should use `items()` rather than `iteritems()`.

PaulWay: Hi @jhjaggars - thanks for that!  You're much more up to date on Python 3 than I am :-)

Do you have a preferred solution to #1124 - that, this or #1122 ?

PR #1149, opened 2018-05-03
Added check to glob_file() function in limit number of files

lhuett:
Added check to glob_file() function in spec_factory.py to throw 
exception if more than 1000 files collected
Added a test for ContentException thrown when files > 1000


lhuett:
Great idea, max_files keyword arg
Thanks.

PR #1152, opened 2018-05-08
Initial specs for NetworkManager dhclient scripts

kgrant-rh:
These files are needed for parsers/rules currently in development.

### DIFF
--- a/insights/specs/__init__.py
+++ b/insights/specs/__init__.py
@@ -237,6 +237,7 @@ class Specs(SpecSet):
     netstat_i = RegistryPoint()
     netstat = RegistryPoint()
     netstat_s = RegistryPoint()
+    networkmanager_dispatch_d = RegistryPoint(multi_output=True)
     neutron_conf = RegistryPoint()
     neutron_l3_agent_log = RegistryPoint()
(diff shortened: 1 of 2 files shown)


kylape:
I believe you'll need a spec in `insights/specs/insights_archive.py` to pull the file(s) out of the archive.

kylape:
If you only ever expect one file, you can grab the first one using the `head` function, similar to how `installed_rpms` works.

kgrant-rh:
While I don't expect it to return more than one file, if there are multiple, we need to examine them all.

Regarding `insights_archive.py`, that's only necessary for command output, right? There don't appear to be entries for similar file collections such as `nginx_conf`.

kylape:
@csams can validate my assertions, but that sounds like a problem for `nginx_conf` then.  AFAIK if you want your spec to be read out of the archive, then it needs to be in `insights_archive.py`.  `default.py` is for client collection only, not for server-side archive mapping.

csams:
File datasources (simple_file, glob_file, etc) in default.py are also used server side, although insights_archive.py can override them. However, insights_archive.py is used almost exclusively for getting command output out of file in archives. I think the spec looks okay.

Issue #1126, opened 2018-04-24
fstab parser: parsing broken with ValueError


JoySnow:
When testing with following line contained in `/etc/fstab` content, 
```
/dev/sdb5                    /l\040ok/at                ext4    defaults        1 1
```
the parsing will broken at `parse_content()` function like:
```
        for line in fstab_output:
>           line['fs_freq'] = int(line['fs_freq']) if 'fs_freq' in line else 0
E           ValueError: invalid literal for int() with base 10: 'defaults'

insights/parsers/fstab.py:120: ValueError
```

* Root cause:
In manpage of fstab, it allow the spaces' usage, currently, our parser cann't support it correctly.
```
     The second field (fs_file).
              This  field  describes  the mount point for the filesystem.  For swap partitions, this field should be specified as `none'. If the name of the mount point contains spaces these can be escaped as `\040'.
```

JoySnow:
Rsolved at https://github.com/RedHatInsights/insights-core/pull/1158

Issue #1157, opened 2018-05-09
Remove reminder comments after fixed mount-point-space issue from fstab's test


xiangce:
Please uncomment the test cases prepared for mount-point which contains spaces and remove any other reminder comments as well after the #1126 is fixed.

xiangce:
Resolved in #1158

PR #1158, opened 2018-05-09
Enhance parser for fstab to handle space in fs_file column

JoySnow:
Resolve issue#1126.
  Tested with archive provided files in insights-cli.
  Without the escape decode of fs_file, tests for fsspec_of_path with
  space in path will fail.


JoySnow:
Resolve https://github.com/RedHatInsights/insights-core/issues/1126 and https://github.com/RedHatInsights/insights-core/issues/1157

JoySnow:
Tested: Plugins are working well with this changes.

PR #1160, opened 2018-05-10
Add parser for command 'mysqladmin variables'

JoySnow:


insights/parsers/mysqladmin.py
```diff
@@ -0,0 +1,132 @@
...
+        Set each variable as an class attribute.
+        """
+        if len(content) < 5:
+            raise ParseException("Empty or wrong content in table.")
+
+        self._data = {}
+        for l in content[3:-1]:
+            lsp = l.split('|')
+            if len(lsp) != 4:
+                raise ParseException("Unparseable line in table.")
```
xiangce: Better display the current line in the Exception together with the prompt info.

insights/parsers/mysqladmin.py
```diff
@@ -0,0 +1,132 @@
...
+            If `keyword` does not exist or it's not a number, returns specific
+            value in `default`.
+        """
+        if default and not isinstance(default, int):
+            raise TypeError("Default value should be int type.")
+        v = self._data.get(keyword.lower())
+        return int(v) if v and v.isdigit() else default
+
+    def __contains__(self, keyword):
+        return keyword.lower() in self._data
```
xiangce: It's different with the common `__contains__` since it's case insensitive, better add corresponding description in the class's docstring.

JoySnow:
SEPC change: https://github.com/RedHatInsights/insights-frontend-assets/pull/46

insights/parsers/mysqladmin.py
```diff
@@ -0,0 +1,132 @@
...
+            raise ParseException("Empty or wrong content in table.")
+
+        self._data = {}
+        for l in content[3:-1]:
+            lsp = l.split('|')
+            if len(lsp) != 4:
+                raise ParseException("Unparseable line in table.")
+            self._data[lsp[1].strip().lower()] = lsp[2].strip()
+
+        for k, v in self._data.items():
```
xiangce: This syntactic sugar could be useful, but to use it, you have to check for the key's existence to avoid an AttributeError being raised which reduced the value of the functionality. Better just fix several common attributes (e.g version, datadir etc.) and explicitly document them in the docstring.

insights/parsers/mysqladmin.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+    Example:
+        >>> output.version
+        '5.5.56-MariaDB'
+        >>> 'datadir' in output
+        True
+        >>> output.get('what', '233')
+        '233'
+        >>> config.getint('aria_block_size', '4096')
```
bfahr: @JoySnow did you mean to use `output` here instead of `config`?

Issue #1076, opened 2018-04-03
Better handling of files containing failed command output


csams:
If the current collector tries to execute a command that doesn't exist, it captures the "command not found" or "No such file or directory" message as if that's what the command returned.

So every parser doesn't have to guard against this, we handle it for all of them in the spec code. If a file is one line long and contains any of those messages, its spec is skipped. It's not perfect, but I didn't expect that corner case to come up often enough to matter.

Turns out, it comes up often enough to matter when dealing with filtered output.

```
subsystem_a: this is a line
subsystem_b: oh no, something happened! No such file or directory!
subsystem_a: some other line
subsystem_a: another line
```
If the output is filtered to subsystem_b, it'll reduce to one line that contains the bad phrases and be skipped. Checking the number of lines before the filter isn't an option.

The bad command check should be scoped to only commands, and we should handle it in a common way while still allowing individual parsers to opt in. @PaulWay had the right idea a while back with a mixin class.

The corners of the corner case are commands that return log like output: journalctl, etc. They need to opt out of the mixin and handle true missing command messages individually.

csams:
Fixed in #1162

PR #1162, opened 2018-05-11
Removing bad_lines check from TextFileProvider in spec_factory to

lhuett:
WIP 
CommandParser (Parser subclass that checks for bad_lines)
in core __init__
Replaced Parser with CommandParser in subclasses in parser files
Changed parser tests that were asserting for bad_lines situations
Added a CommandParser test to vaidate bad_lines functionality


bfahr:
This is a fix for issue #1076

insights/core/__init__.py
```diff
@@ -241,6 +241,37 @@ def get(self, item, default=None):
         return self.data.get(item, default)
 
 
+class CommandParser(Parser):
+    """
+    This class checks output from the command defined in th spec to be sure line
+    do not contain certain string values and that there is only one line.
```
bfahr: @lhuett you may need to check spelling, and is it checking that there is only one line or that if there is one line, it doesn't contain the bad strings?  Be sure to check the output documentation to see if it looks like you expected.

insights/core/__init__.py
```diff
@@ -241,6 +241,37 @@ def get(self, item, default=None):
...
+    """
+
+    bad_lines = ["No such file or directory", "Command not found", "command not found"]
+
+    def validate_lines(self, results):
+        if results and len(results) == 1:
```
bfahr: You should add a docstring to this function so that it will have a description in the documentation.  Also `bad_lines` should probably be documented.  You can do this by adding a docstring comment in triple quotes under the declaration of the attribute.

insights/core/spec_factory.py
```diff
@@ -135,15 +135,15 @@ class TextFileProvider(FileProvider):
...
-        return True
+    # bad_lines = ["No such file or directory", "Command not found"]
+    #
+    # @classmethod
+    # def validate_lines(self, results):
+    #     if results and len(results) == 1:
+    #         first = results[0]
+    #         if any(l in first for l in self.bad_lines):
+    #             return False
+    #     return True
```
bfahr: Make sure to remove the commented lines in your final commit.  Same with lines below.

insights/specs/default.py
```diff
@@ -538,7 +538,8 @@ def block(broker):
     sysconfig_virt_who = simple_file("/etc/sysconfig/virt-who")
     sysctl = simple_command("/sbin/sysctl -a")
     sysctl_conf = simple_file("/etc/sysctl.conf")
-    sysctl_conf_initramfs = simple_command("/bin/lsinitrd /boot/initramfs-*kdump.img -f /etc/sysctl.conf /etc/sysctl.d/*.conf")
+    sysctl_conf_initramfs = simple_command(
+        "/bin/lsinitrd /boot/initramfs-*kdump.img -f /etc/sysctl.conf /etc/sysctl.d/*.conf")
```
bfahr: Did you mean to change this line?

insights/tests/test_commandparser.py
```diff
@@ -0,0 +1,24 @@
...
+def test_command_not_found():
+    with pytest.raises(ParseException) as e:
+        MockParser(context_wrap(CMF))
+    assert "Command not found" in str(e.value)
+
+
+def test_no_such_file_or_directory():
+    with pytest.raises(ParseException) as e:
+        MockParser(context_wrap(NO_SUCH_FILE))
+    assert "No such file or directory" in str(e.value)
```
bfahr: It would probably be a good idea to test with multi-line files containing the bad words as well to make sure it allows them to parse OK.

insights/parsers/tests/test_chkconfig.py
```diff
@@ -61,14 +61,14 @@ def test_chkconfig():
...
-    assert not chkconfig.is_on('kdump')
-    assert not chkconfig.is_on('restorecond')
+# def test_command_not_found():
+#     context = context_wrap(ERROR)
+#     chkconfig = ChkConfig(context)
+#     assert len(chkconfig.services) == 0
+#     assert len(chkconfig.parsed_lines) == 0
+#     assert not chkconfig.is_on('crond')
+#     assert not chkconfig.is_on('kdump')
+#     assert not chkconfig.is_on('restorecond')
```
bfahr: Parsers should not break backward backward compatibility so you may need to look at each of these parsers to see what they returned when the bad words were present in the input data, if they were not throwing an exception. (if you haven't already done that.)

insights/specs/default.py
kylape: Looks like a formatting-only change to me.

insights/parsers/tests/test_tomcat_virtual_dir_context.py
```diff
@@ -70,20 +70,20 @@ def test_tomcat_virtual_dir_context_not_found():
 
 def test_tomcat_virtual_dir_context_exceptions():
     for parser in [TomcatVirtualDirContextFallback, TomcatVirtualDirContextTargeted]:
-        with pytest.raises(SkipException) as excinfo:
-            parser(context_wrap(ERRORS_1))
-            assert 'VirtualDirContext not used.' in str(excinfo.value)
+        # with pytest.raises(SkipException) as excinfo:
+        #     parser(context_wrap(ERRORS_1))
+        #     assert 'VirtualDirContext not used.' in str(excinfo.value)
```
kylape: I suggest just taking out commented code.  We can always restore from a previous version.

insights/parsers/tests/test_chkconfig.py
lhuett: The parser (parse_context()) should never get called, if a `bad_line` is found, since the CommandParser is catching the issue before parser_context is called and is raising a ContentException (which is based on SkipComponent).

insights/specs/default.py
lhuett: I did not, will revert.

insights/parsers/tests/test_tomcat_virtual_dir_context.py
lhuett: Yea, was leaving it in for the formal walk through, removing.

PR #1163, opened 2018-05-11
branch_info parser now inherits YAMLParser for parsing

lhuett:
Changed branch_info parser to use YAMLParser to do the parsing.
Fixed an issue with YAMLParser that surfaced while applying to branch_info that
caused a failure if content was not a list, handles lists and strings now.
There was no test for branch_info so one was added.


insights/core/__init__.py
```diff
@@ -365,7 +365,10 @@ class YAMLParser(Parser, LegacyItemAccess):
     A parser class that reads YAML files.  Base your own parser on this.
     """
     def parse_content(self, content):
-        self.data = yaml.safe_load('\n'.join(content))
+        if type(content) is list:
+            self.data = yaml.safe_load('\n'.join(content))
+        else:
+            self.data = yaml.safe_load(content)
```
bfahr: @lhuett do you need to add a test for this?

lhuett: @bfahr I added test_branch_info.py that tests the non list situation I added to support branch_info.py. I believe that the only parser class defined in __init__ that actually had a test for it individually is XMLParser. I could add one for this a guess.

bfahr: Would you mind to add a test for this parser?

lhuett: Created a new test for core Parser class YAMLParser(Parser, LegacyItemAccess)
test_yamlparser.py

PR #1165, opened 2018-05-15
Add github notify step to Jenkinsfile

SteveHNH:


SteveHNH:
@bfahr No worries! I was just optimizing one last thing and it triggered another test.

PR #1166, opened 2018-05-15
WIP - Support external component metadata and a new component for service requests

csams:
This PR enables external component configuration. You can enable or disable a component and provide arbitrary metadata for it to inspect at runtime.

For example, the following rule gets the RPM it should inspect from its metadata and falls back to a default if none is provided:
~~~python
#!/usr/bin/env python
# check_ovirt.py

from insights import dr, make_response, rule, run                               

@rule(InstalledRpms)                                                            
def report(rpms):                                                               
    rpm = dr.get_metadata(report).get("rpm", "ovirt-hosted-engine-ha")          
    if rpm in rpms:                                                             
        return make_response("RUNNING", version=rpms.get_max(rpm).nvr)          
    return make_response("NOT_RUNNING")                                         
                                                                                
if __name__ == "__main__":                                                      
    run(report, print_summary=True)
~~~

It can be configured with a yaml file like this:
~~~yaml
check_ovirt.report:
    enabled: true
    metadata:
        rpm: bash
~~~
Run it with `python -m insights --config config.yml -p check_ovirt`

If multiple config files are passed, they are applied in sorted order.

Included in this PR is a new component type that takes advantage of the above functionality to encapsulate http(s) requests.

It includes a component type (`@webservice`)  and a base class `WebService` that automatically inspects subclass metdata to pass the following along to `requests.get`:
-  verify (bool), optional, default = False
-  timeout (int) seconds, optional, default = 5
-  username (string), optional
-  password (string), optional

`WebService` subclasses should provide a suitable API that delegates to the `WebService.get` method. Subclasses are responsible for defining appropriate URLs that back the service. They should be included in the service metadata.

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -4,12 +4,14 @@ import json
 import logging
 import os
 import re
+import requests
 import shlex
 import yaml
 import six
 from insights.contrib.ConfigParser import RawConfigParser
 
 from insights.parsers import ParseException
+from insights.core import dr
 from insights.core.serde import deserializer, serializer
 from insights.util import deprecated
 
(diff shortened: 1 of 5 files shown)


insights/core/__init__.py
```diff
@@ -4,12 +4,14 @@
 import logging
 import os
 import re
+import requests
```
kylape: This makes `requests` a runtime dependency.

kylape: My opinion is that requests should be strictly optional.

bfahr:
@csams is this PR still applicable with all of the other changes you have made?

ivarin:
test this please

jhjaggars:
This PR is so old now, I want to suggest we abandon it.

bfahr:
@csams what do you think about @jhjaggars comment?

csams:
Closing as obsolete.

PR #1167, opened 2018-05-16
new parser for tuned_adm active

TripleJQK:


insights/parsers/tuned_adm.py
```diff
@@ -0,0 +1,38 @@
...
+    Current active profile: default
+    Service tuned: disabled, stopped
+    Service ktune: disabled, stopped
+
+Examples::
+
+    >>> 'Service tuned' in tuned_adm_active.data
+    True
+    >>> tuned_adm_active.data['Service ktune']
+    'disabled, stopped'
```
bfahr: Since you have based the parser on `LegacyItemAccess` you could also show examples without the `.data`.

bfahr:
@TripleJQK I don't believe this parser is necessary since this existing `Tuned` parser already returns the active profile information.  Is this parser somehow different?

PR #1170, opened 2018-05-16
New parser and combiner for logrotate configuration file

xiangce:


xiangce:
WIP: The combiner is not finished.

insights/parsers/logrotate_conf.py
```diff
@@ -0,0 +1,163 @@
...
+        Return a list of global options configured in this configuration file.
+        """
+        return self._options
+
+    @property
+    def log_files(self):
+        """
+        Return a list of log files configured in this configuration file.
+        """
+        return self._log_files
```
bfahr: @xiangce would it make sense to just have `self.options` and `self.log_files` as attributes?

insights/parsers/logrotate_conf.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+
+    def parse_content(self, content):
+
+        def _parse_opts(line):
+            if '=' in line:
+                l_sp = line.split('=')
+            else:
+                l_sp = line.split(None, 1)
+            return l_sp[0], l_sp[1] if len(l_sp) == 2 else True
```
bfahr: @xiangce I believe you'll get an exception if you return `True` since the calling code expects a tuple.

bfahr:
@xiangce the title indicates you are also adding a combiner, are you adding that as part of this PR?

insights/parsers/logrotate_conf.py
xiangce: @bfahr - Yep, I think I did return a tuple with two elements:
- The first element is `l_sp[0]`
- The second one is `l_sp[1]`, but when there is only one element in `l_sp`, `True` will be the second element.
   `True` is for the options without a value, such as `compress`

xiangce:
@bfahr - Yes, there will be a  combiner be wrote in this PR, and that's why I marked it as [WIP](https://github.com/RedHatInsights/insights-core/pull/1170#issuecomment-389502197) :smile: , Thanks.

insights/parsers/logrotate_conf.py
xiangce: The only reason I used the `@property` is trying to avoid writing data back to the object. But seems it's not a big deal at all. I will use attributes directly. Thanks @bfahr .

insights/parsers/logrotate_conf.py
bfahr: Thanks for the clarification @xiangce, that looks correct.

Issue #1175, opened 2018-05-18
httpd_conf parser and combiner cannot handle directives in nest and parallel sections correct


xiangce:
1. parallel
~~~
<IfModule !php5_module>
  Testphp php5_1
  <IfModule !php4_module>
    Testphp php4_1
  </IfModule>
  Testphp php5_2
</IfModule>
<IfModule !php5_module>
  Testphp php5_3
</IfModule>
~~~

Expect result:

```python
{
 ('IfModule', '!php5_module'): {
  'Testphp': [
   ('php5_1', 'Testphp php5_1', 'IfModule', '!php5_m odule', 'httpd.conf', '/etc/httpd/conf/httpd.conf'),
   ('php5_2', 'Testphp php5_2', 'IfModule', '!php5_module', 'httpd.conf', '/etc/httpd/conf/httpd.conf'),
   ('php5_3', 'Testphp php5_3', 'IfModule', '!php5_module', 'httpd.conf', '/etc/httpd/conf/httpd.conf')],
  ('IfModule', '!php4_module'): {
    'Testphp': [
      ('php4_1', 'Testphp php4_1', 'IfModule', '!php4_module', 'httpd.conf', '/etc/httpd/conf/httpd.conf')]
 }
}
```

Actual result: (one more line of `Testphp php5_3` is stored)
```python
{
 ('IfModule', '!php5_module'): {
  'Testphp': [
   ('php5_1', 'Testphp php5_1', 'IfModule', '!php5_m odule', 'httpd.conf', '/etc/httpd/conf/httpd.conf'),
   ('php5_2', 'Testphp php5_2', 'IfModule', '!php5_module', 'httpd.conf', '/etc/httpd/conf/httpd.conf'),
   ('php5_3', 'Testphp php5_3', 'IfModule', '!php5_module', 'httpd.conf', '/etc/httpd/conf/httpd.conf')
   ('php5_3', 'Testphp php5_3', 'IfModule', '!php5_module', 'httpd.conf', '/etc/httpd/conf/httpd.conf')],
  ('IfModule', '!php4_module'): {
    'Testphp': [
      ('php4_1', 'Testphp php4_1', 'IfModule', '!php4_module', 'httpd.conf', '/etc/httpd/conf/httpd.conf')]
 }
}
```

2. nest

~~~
# /etc/httpd/conf/httpd.conf
<VirtualHost 128.39.140.28>
    <IfModule !php5_module>
        Testphp php5_v3_1
[...]

PR #1176, opened 2018-05-18
Fix httpd_conf cannot handle directives in nest and parallel sections correct

xiangce:
- fix the issue #1175


xiangce:
Hi @bfahr - For the detailed debugging progress, please see the comments added in #1175.

PR #1177, opened 2018-05-21
Added make_fingerprint as response

pratikgadiya12:


pratikgadiya12:
@xiangce :Done

insights/core/plugins.py
```diff
@@ -240,47 +240,12 @@ def make_response(error_key, **kwargs):
 
 
 def make_fingerprint(**kwargs):
```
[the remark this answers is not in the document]
xiangce: I surmised there is no `error_key` from your initial internal rules. Since the `error_key` is necessary, it could be the `manufacture` + `cert_id`.

insights/core/plugins.py
```diff
@@ -191,6 +191,23 @@ def make_skip(rule_fqdn, reason, details=None):
             "type": "skip"}
 
 
+def is_make_reponse_too_long(key, kwargs):
```
xiangce: Nitpick: would you mind to add a leading underscore to the function name? which indicates it's an internal but not global interface.

insights/core/plugins.py
```diff
@@ -223,16 +240,51 @@ def make_response(error_key, **kwargs):
...
+        {
+            "type": "fingerprint",
+            "fingerprint_key": "FINGERPRINT",
+            "manufacturer": "Red Hat",
+            "product": "Insights"
+        }
+    """
+
+    if "fingerprint_key" in kwargs or "type" in kwargs:
+        raise Exception("Can't use an invalid argument for make_fingerprint")
```
bfahr: @pratikgadiya12 to me consistent with the similar functions shouldn't this be a `ValueError` exception?

PR #1184, opened 2018-05-29
Fix and remove unnecessary secondary specs from DefaultSpecs

xiangce:
- '/conf/*', '/rhn-logs/*', '/database/' and '/config-defaults' are
  for `spacewalk-debug` archive
- '/cpuinfo' and '/meminfo' are for `foreman-debug` archive
- and add '/' for all simple_file


xiangce:
This PR is for https://github.com/RedHatInsights/insights-core/issues/1186

xiangce:
@bfahr - Thanks for adding the `SPEC Change` label. However, I would like to explain that although this PR updated some specs, we do need to modify them in the `uploader.json`, because that the removed secondary specs had not been added to the `uploader.json` and the newly added `/` had been added to the `uploader.json` (I think it's at the first time when they are added).

xiangce:
Just pushed once by force to resolve newly generated conflicts

insights/specs/default.py
```diff
@@ -129,8 +126,8 @@ def tomcat_base(broker):
     dig_edns = simple_command("/usr/bin/dig +edns=0 . SOA")
     dig_noedns = simple_command("/usr/bin/dig +noedns . SOA")
     dirsrv = simple_file("/etc/sysconfig/dirsrv")
-    dirsrv_access = glob_file("var/log/dirsrv/*/access")
-    dirsrv_errors = glob_file("var/log/dirsrv/*/errors")
+    dirsrv_access = glob_file("/var/log/dirsrv/*/access")
```
bfahr: This spec is not currently in the uploader.json.

insights/specs/default.py
```diff
@@ -129,8 +126,8 @@ def tomcat_base(broker):
...
+    dirsrv_errors = glob_file("/var/log/dirsrv/*/errors")
```
bfahr: This doesn't appear to match what is in the uploader.json, is the json correct?

insights/specs/default.py
```diff
@@ -187,15 +184,15 @@ def docker_container_ids(broker):
     ethtool_g = foreach_execute(ethernet_interfaces, "/sbin/ethtool -g %s")
     ethtool_i = foreach_execute(ethernet_interfaces, "/sbin/ethtool -i %s")
     ethtool_k = foreach_execute(ethernet_interfaces, "/sbin/ethtool -k %s")
-    exim_conf = simple_file("etc/exim.conf")
+    exim_conf = simple_file("/etc/exim.conf")
```
bfahr: This spec is not in the json.

insights/specs/default.py
xiangce: @bfahr - I'm not sure about why the uploader.json is updated as below:
```
"file": "/var/log/dirsrv/.*/()*(errors|errors\\.2.*)"
```
in https://github.com/RedHatInsights/insights-frontend-assets/commit/550bb7a

@wcmitchell, would you mind to take a look at the above commit? Why the spec is modified:
from `"/var/log/dirsrv/.*/()*errors"` to `"/var/log/dirsrv/.*/()*(errors|errors\\.2.*)"`?
I'm sorry that I can't find any corresponding information about this modification.

bfahr: @xiangce I discussed this with @kylape and this spec is broken in the uploader.json because the regex will not work for the directory segment.  I'm going to log an issue on that project and will go ahead and merge this PR.

PR #1188, opened 2018-05-30
[yum] Added new property to retrive RHEL repo names

psachin:
Often we need a list of repo-ids and not entire repo dictionary to
check if a repo is available in customer's environment. This patch
provides that property.


bfahr:
@psachin you'll need to fix the test failure for RHEL 3.6.

PR #1192, opened 2018-06-05
Update specs to rules script to work with version 3 of core

lhuett:
Took the Insights Core 1.x version of specs_to_rules (insights-core/insights/tools/specs_to_rules.py) and refactored to work with Insights Core 3.x

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -633,8 +633,8 @@ def first_of(deps):
                 return broker[c]
 
     docs = [escape(d.__doc__) for d in deps if d.__doc__]
-    docs = ["Returns the first of the following:"] + docs
-    inner.__doc__ = "\n".join(docs)
+    docs = [b"Returns the first of the following:"] + docs
+    inner.__doc__ = b"\n".join(docs)
     return inner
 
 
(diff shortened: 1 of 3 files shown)


insights/specs/default.py
```diff
@@ -267,6 +271,7 @@ def httpd_cmd(broker):
 
     @datasource(ipcs_s)
     def semid(broker):
+        """Command - semids"""
```
bfahr: Should this be `Command:` to be consistent with the others?

insights/core/spec_factory.py
```diff
@@ -334,6 +339,7 @@ def simple_file(path, context=None, kind=TextFileProvider):
     def inner(broker):
         ctx = _get_context(context, FSRoots, broker)
         return kind(ctx.locate_path(path), root=ctx.root, ds=inner)
+    inner.__doc__ = 'Path: ' + re.sub(r"([=\(\)|\-_!@*~\"&/\\\^\$\=])", r"\\\1", path)
```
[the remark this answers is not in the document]
lhuett: Yea, good idea

insights/specs/default.py
lhuett: Oh yea, I changed format at one point and missed that one. Thanks Bob, good catch.

PR #1193, opened 2018-06-06
fix the wrong command of systemctl_show_qdrouterd

TripleJQK:


xiangce:
@TripleJQK - Please open an issue for this PR and add the corresponding description for it. Thanks.

PR #1194, opened 2018-06-06
parser for foreman in /etc/passwd

TripleJQK:


insights/parsers/passwd_foreman.py
```diff
@@ -0,0 +1,31 @@
...
+@parser(Specs.passwd_foreman)
+class PasswdForeman(Parser):
+    """
+    Parse data from the ``/bin/grep  '^foreman:' /etc/passwd`` command.
+    """
+
+    def parse_content(self, content):
+        if len(content) == 1 and content[0].startswith('foreman:'):
+            self.data = content[0].split(':')
+            print self.data
```
xiangce: Please remove this line

TripleJQK:
@xiangce @kylape 
first I saw the following parser existed:

~~~
numeric_user_group_name = simple_command("/bin/grep -c '^[[:digit:]]' /etc/passwd /etc/group")
~~~

I have considered the parser which `just get the /etc/passwd and add a filter`, even though it also get just one line from cu side as same as `grep command`, but in fact it incluedes two parts - first the whole file, second the filter, and I think cu will always put their eyes on `redhat get the whole file`. 
But if use grep command, the cu will easily understood just one line collected by redhat. and will feel more comfortable

xiangce:
A bit difference between the new one for `foreman` and the old one `numeric_user_group_name` is:
- the `numeric_user_group_name` just counted the number of the users
- but the new one will collect one whole line.
This's my major concern.

TripleJQK:
https://projects.engineering.redhat.com/browse/INSPEC-38

xiangce:
Closed since the spec is not suitable to be collected.

PR #1195, opened 2018-06-06
Add item for docker_image_inspect

wushiqinlou:


bfahr:
@wushiqinlou you won't be able to add the specification necessary for this parser to the the uploader.json since it requires pre_commands which are no longer being added to the uploader.json.  @kylape can correct me if I got this wrong.

xiangce:
close this pr for now, since the spec need a precommand that the current client doesn't suppport.

PR #1196, opened 2018-06-06
Add item for docker_container_inspect

wushiqinlou:


insights/specs/insights_archive.py
```diff
@@ -39,6 +39,7 @@ class InsightsArchiveSpecs(Specs):
     docker_info = simple_file("insights_commands/docker_info")
     docker_list_containers = simple_file("insights_commands/docker_ps_--all_--no-trunc")
     docker_list_images = simple_file("insights_commands/docker_images_--all_--no-trunc_--digests")
+    docker_container_inspect = glob_file("insights_commands/docker_inspect_*")
```
xiangce: The file name is the same as the one in https://github.com/RedHatInsights/insights-core/pull/1195, it will puzzle the parser to parse both specs I think.

xiangce: Two workarounds I could think for this issue:
1. Add different options for each of them and make sure the option does not affect the actual output.
2. Update the parser and distinguish them inner the parser by analyzing their content.

xiangce:
@wushiqinlou - Please open an issue for this update and add corresponding description for it. Thanks.

bfahr:
@wushiqinlou you won't be able to add the specification necessary for this parser to the the uploader.json since it requires pre_commands which are no longer being added to the uploader.json. @kylape can correct me if I got this wrong.

xiangce:
close this pr for now, since the spec need a precommand that the current client doesn't suppport.

PR #1199, opened 2018-06-06
re-exporting common rpm bits for convenience

jhjaggars:


README.rst
```diff
@@ -99,21 +99,21 @@ installed within the range.
...
-        if rpm and rpm >= lower and rpm < upper:
-            return make_response("BASH_AFFECTED", version=rpm.nvr)
-        elif rpm:
-            return make_response("BASH_UNAFFECTED", version=rpm.nvr)
-        else:
+        bash = rpms.get_max("bash")
+        if not bash:
             return make_response("NO_BASH_FOUND")
 
+        msg = "BASH_AFFECTED" if lower <= bash < upper else "BASH_UNAFFECTED"
```
kylape: It seems unusual to me to return a response at all when bash is unaffected.

miclark: I went back and forth on this.  From a support standalone script, it seemed natural to give the "green light."   It certainly is unusual from a usual plugin.  And, because of that, we may want to leave it off here to help set the idiom.

kylape: My opinion is to leave it off for two reasons:

1. As you said, it's unnatural for a "standard" plugin
2. It make the example simpler to digest.

We can always explain how to make a diagnostic response elsewhere.

miclark: Ack, @jhjaggars feel free to simplify the pluging stuff to drop the  `BASH_UNAFFECTED` or even `BASH_NOT_FOUND`.

PR #1201, opened 2018-06-07
put dotfiles where they're needed

gravitypriest:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1570172

### DIFF
--- a/insights/client/constants.py
+++ b/insights/client/constants.py
@@ -4,6 +4,7 @@ from insights import package_info
 
 
 class InsightsConstants(object):
+    sfr_version = os.environ.get('INSIGHTS_SFR', 'insights-client')
     app_name = 'insights-client'
     auth_method = 'BASIC'
     package_path = os.path.dirname(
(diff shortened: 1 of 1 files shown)


insights/client/constants.py
```diff
@@ -4,6 +4,7 @@
 
 
 class InsightsConstants(object):
+    sfr_version = os.environ.get('INSIGHTS_SFR', 'insights-client')
```
kylape: What's sfr?

gravitypriest:
That's correct. I have another PR open in the client repo that sets INSIGHTS_SFR based on the executable ran.

SFR = Simple Find Replace

insights/client/constants.py
gravitypriest: simple find replace

kylape:
ooOOOooo can you link that PR pretty please?

gravitypriest:
https://github.com/RedHatInsights/insights-client/pull/57

PR #1202, opened 2018-06-11
Added support for fingerprint response type in insights-core evaluators

pratikgadiya12:


csams:
Hey @pratikgadiya12, after discussing this some more, we'd like fingerprint responses to go into their own section of the response at the same level as reports instead of adding a type to each report and mixing responses in the same section. Rule responses will still go in the reports section of the dictionary, and fingerprint responses should go in a sibling fingerprints section. Make sense?

pratikgadiya12:
@csams : Make sense, I have updated PR. Let me know if this is what you were trying to suggest

csams:
@pratikgadiya12, looks good to me!

pratikgadiya12:
@csams : Thanks for the review. It seems like I don't have access to merge pull request.
Could you merge this pull request? Thanks in advance

PR #1207, opened 2018-06-18
XML parser for /etc/libvirt/qemu/*.xml files

psachin:
Included:

- Basic tests
- Docs
- Doc test


xiangce:
@psachin - please open a Jira request to add this spec to uploader.json at first, thanks.

psachin:
https://projects.engineering.redhat.com/browse/INSPEC-46

insights/parsers/etc_libvirt_qemu_xml.py
```diff
@@ -0,0 +1,157 @@
...
+EtcLibvirtQemuXML file ``/etc/libvirt/qemu/*.xml``
+--------------------------------------------------
+"""
+from .. import XMLParser, parser
+from insights.specs import Specs
+
+
+@parser(Specs.etc_libvirt_qemu_xml)
+class EtcLibvirtQemuXML(XMLParser):
+    """This class parses xml files under ``/etc/libvirt/qemu/`` using
```
bfahr: @psachin what do you think about shortening this class name to `QemuXML` instead?

insights/parsers/etc_libvirt_qemu_xml.py
```diff
@@ -0,0 +1,157 @@
+"""
+EtcLibvirtQemuXML file ``/etc/libvirt/qemu/*.xml``
+--------------------------------------------------
+"""
```
bfahr: Please add a `.rst` file to the docs directory for this parser to include it in the documentation, and confirm that the rendered documentation looks correct.

insights/parsers/etc_libvirt_qemu_xml.py
```diff
@@ -0,0 +1,157 @@
...
+              <address type='pci' domain='0x0000' bus='0x00' slot='0x07' function='0x0'/>
+            </watchdog>
+            <memballoon model='virtio'>
+              <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x0'/>
+            </memballoon>
+          </devices>
+        </domain>
+
+    Examples:
+        >>> xml_numa.file_name = 'vm.xml'
```
bfahr: Should this be an assert?

insights/parsers/etc_libvirt_qemu_xml.py
```diff
@@ -0,0 +1,157 @@
...
+                if not child.getchildren():
+                    domain[child.tag] = child.text
+                else:
+                    domain[child.tag] = [c.items() for c in child.getchildren()]
+
+            return domain
+
+    @property
+    def vm_name(self):
+        if self.parse_dom():
```
bfahr: I believe you can just use `self.data` here since the base class `parse_content()` calls `parse_dom()` and saves the results in `self.data`.

insights/parsers/etc_libvirt_qemu_xml.py
psachin: Done

insights/parsers/etc_libvirt_qemu_xml.py
psachin: I had the same opinion in mind, the long file and class name does not look so good but they clearly indicates the file path the parser will parse.

xiangce: Is it necessary to express the file path things in the class name and module name? This kind of name style is quite different from all the existing parsers. Unless the parser's logic depends on the specified `file_path`, I think it would be better to user QemuXML instead.

psachin: class renamed to QemuXML

insights/parsers/tests/test_qemu_xml.py
```diff
@@ -0,0 +1,294 @@
...
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x08' function='0x0'/>
+    </memballoon>
+  </devices>
+</domain>
+"""
+
+
+def test_vm_xml():
+    xml = qemu_xml.QemuXML(context_wrap(XML_NUMA, path='/etc/libvirt/qemu/vm.xml'))
+    xml.file_name = 'vm.xml'
```
bfahr: @psachin `.file_name` and `.file_path` attributes are set by the parser based on the `path=` attribute to `context_wrap`.  There's no need to set it here.  Did you mean to use `assert` instead?  Same comment for the other tests.

psachin: Oops! My intention was to check assertion. (I need coffee)

PR #1208, opened 2018-06-18
Dotfiles fix mk2

gravitypriest:
electric boogaloo


insights/client/utilities.py
```diff
@@ -52,27 +52,51 @@ def get_time():
...
+def write_registered_file():
+    delete_unregistered_file()
+    for f in constants.registered_files:
+        if os.path.lexists(f):
+            if os.path.islink(f):
+                # kill symlinks and regenerate
+                os.remove(f)
+                write_to_disk(f)
+        else:
+            write_to_disk(f)
```
kylape: nit: seems like the logic could be a bit simpler:

```
if os.path.islink(f):
    os.remove(f)
write_to_disk(f)
```

insights/client/utilities.py
```diff
@@ -52,27 +52,51 @@ def get_time():
...
-    else:
-        rc = 1
+    for f in constants.unregistered_files:
+        if os.path.lexists(f):
+            if os.path.islink(f):
+                # kill symlinks and regenerate
+                os.remove(f)
+                write_to_disk(f, content=str(date))
+        else:
+            write_to_disk(f, content=str(date))
```
kylape: same nit as above

insights/client/constants.py
```diff
@@ -13,15 +13,18 @@ class InsightsConstants(object):
     default_conf_file = '/etc/insights-client/insights-client.conf'
     user_agent = os.path.join(app_name, package_info["VERSION"])
     log_dir = os.path.join(os.sep, 'var', 'log', app_name)
+    simple_find_replace_dir = '/etc/redhat-access-insights'
```
kylape: Can you explain why you chose `simple_find_replace` for the name?

insights/client/connection.py
```diff
@@ -626,9 +626,6 @@ def register(self):
         """
         Register this machine
         """
-
-        write_to_disk(constants.unregistered_file, delete=True)
-
```
kylape: I take it this can be removed because you're doing it in `write_registered_file`?

insights/client/utilities.py
kylape: Nevermind, I now see that your code doesn't write to disk when the file exists and is not a symlink.  There's probably no harm in writing the file again if it already exists, but either way is fine.

insights/client/constants.py
gravitypriest: Once upon a time it was said that renaming the RPM would be a simple find and replace 😏

insights/client/utilities.py
gravitypriest: The main reason is because we write the current time as time of registration to the file and regenerating it every time would make that data meaningless

insights/client/connection.py
gravitypriest: Yep

PR #1211, opened 2018-06-19
Add parser for grubby_--default-index

JoySnow:


insights/parsers/grubby.py
```diff
@@ -0,0 +1,25 @@
...
+    >>> default_index
+    0
+"""
+
+from .. import parser
+from insights.specs import Specs
+
+
+@parser(Specs.grubby_default_index)
+def grub2_default_index(context):
```
xiangce: I do suggest you wrap this parser as a Class Parser and raise a SkipException when below condition@line-24 is not satisfied.

PR #1216, opened 2018-06-20
new parser for qpidd_conf

TripleJQK:


insights/specs/__init__.py
```diff
@@ -341,6 +341,7 @@ class Specs(SpecSet):
     rhn_taskomatic_daemon_log = RegistryPoint()
     rhsm_conf = RegistryPoint()
     rhsm_log = RegistryPoint()
+    qpidd_conf = RegistryPoint()
```
xiangce: please add the new spec as per its alphabet order.

insights/specs/default.py
```diff
@@ -507,6 +507,7 @@ def package_and_java(broker):
                                                "rhn-logs/rhn/rhn_taskomatic_daemon.log"])
     rhsm_conf = simple_file("/etc/rhsm/rhsm.conf")
     rhsm_log = simple_file("/var/log/rhsm/rhsm.log")
+    qpidd_conf = simple_file("/etc/qpid/qpidd.conf")
```
xiangce: This one should be re-sorted too.

PR #1229, opened 2018-06-26
Corrected issue for simple_command data source in spec_factory where

lhuett:
'ps' command used with filters returned grep command used for filters.

### DIFF
--- a/insights/parsers/ps.py
+++ b/insights/parsers/ps.py
@@ -7,7 +7,7 @@ This module provides processing for the various outputs of the ``ps`` command.
 from .. import parser, CommandParser
 from . import ParseException, parse_delimited_table, keyword_search
 from insights.specs import Specs
-from insights.core.filters import add_filter
+from insights.core.filters import add_filter, get_filters
 
 
 class Ps(CommandParser):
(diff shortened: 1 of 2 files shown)


kylape:
@lhuett do you think it would make more sense to put a ps-specific fix in the ps parser?  Or maybe create a custom datasource function to put this custom logic in?

lhuett:
@kylape  I don't know of any other commands that return process information the way 'ps' does so doing it in the 'ps' parser might be ok as long as only filtering out lines that contain " grep -F" would be safe. The problem is that I do not believe we have access to all of the original filter values in the 'ps' parser to build the exact string to compare with.
As for the separate function, are you suggesting creating a separate function within the simple_command data source to remove the line? I think that would be fine to isolate the logic for this. 
Thanks

insights/parsers/ps.py
```diff
@@ -53,8 +53,22 @@ def __init__(self, *args, **kwargs):
...
+
+        filters = [get_filters(s) for s in (Specs.ps_ef, Specs.ps_aux, Specs.ps_auxww)]
+        fgrep_lines = ["grep -F " + " ".join(fters) for fters in filters]
+
+        for c in content:
+            for g in fgrep_lines:
+                if g in c:
+                    content.remove(c)
+        return content
+
```
bfahr: @lhuett would you mind to write a test for this function?

klape:
Please let Bob Fahr know he has the wrong Kevin

lhuett:
Changes are not needed in core. Will be made at the condition in plugins

PR #1230, opened 2018-06-27
[wip] Refactor connection_rules.InsightsUploadConf.get_conf

Glutexo:
When fixing #1226, I found the `get_conf` method very bulky with too many conditionals.

Reduced its cognitive complexity by splitting the method into smaller ones. Used separate methods for different configuration sources: file, url (with fallback to file) and stdin for regular configuration, and one for the remove_file configuration.

`InsightsUploadConf` class interface changed, existing behavior should remain the same. It’s just syntax shuffling that should not have any impact on an actual execution. I don’t have much confidence in the test coverage though: even when I intentionally break the code, no additional test fails. Thus I’d like to ask for a thorough review of the execution paths of former `get_conf` method calls.

My goal was not to make the code polished and beautiful, only to split that one big method into more readable pieces. I didn’t touch anything extra.

WIP: Python 2/3 compatibility issues. Reported by @gravitypriest.

### DIFF
--- a/insights/client/client.py
+++ b/insights/client/client.py
@@ -201,8 +201,12 @@ def get_machine_id():
 
 
 def update_rules(config, pconn):
+    if not pconn:
+        raise ValueError('ERROR: Cannot update rules in --offline mode. '
+                         'Disable auto_update in config file.')
+
     pc = InsightsUploadConf(config, conn=pconn)
-    return pc.get_conf(True, {})
+    return pc.get_conf_update()
 
 
 def get_branch_info(config, pconn):
@@ -293,7 +297,11 @@ def collect(config, pconn):
                      ('--from-file' if config.from_file else '--from-stdin'))
         return False
 
-    collection_rules, rm_conf = pc.get_conf(False, stdin_config)
+    if stdin_config:
+        collection_rules = pc.get_conf_stdin(stdin_config)
+    else:
+        collection_rules = pc.get_conf_file()
+    rm_conf = pc.get_rm_conf()
     if rm_conf:
         logger.warn("WARNING: Excluding data from files")
 
--- a/insights/client/collection_rules.py
+++ b/insights/client/collection_rules.py
@@ -167,57 +167,78 @@ class InsightsUploadConf(object):
         with os.fdopen(fd, 'w') as dyn_conf_file:
             dyn_conf_file.write(data)
 
-    def get_conf(self, update, stdin_config=None):
+    def get_conf_file(self):
         """
-        Get the config
+        Get config from local config file, first try cache, then fallback.
         """
-        rm_conf = None
-        # Convert config object into dict
-        if os.path.isfile(self.remove_file):
-            parsedconfig = ConfigParser.RawConfigParser()
-            parsedconfig.read(self.remove_file)
-            rm_conf = {}
-            for item, value in parsedconfig.items('remove'):
-                rm_conf[item] = value.strip().split(',')
-        if stdin_config:
-            rules_fp = NamedTemporaryFile(delete=False)
-            rules_fp.write(stdin_config["uploader.json"].encode('utf-8'))
-            rules_fp.flush()
-            sig_fp = NamedTemporaryFile(delete=False)
-            sig_fp.write(stdin_config["sig"].encode('utf-8'))
-            sig_fp.flush()
-            if not self.gpg or self.validate_gpg_sig(rules_fp.name, sig_fp.name):
-                return json.loads(stdin_config["uploader.json"]), rm_conf
-            else:
-                logger.error("Unable to validate GPG signature in from_stdin mode.")
-                raise Exception("from_stdin mode failed to validate GPG sig")
-        elif update:
-            if not self.conn:
-                raise ValueError('ERROR: Cannot update rules in --offline mode. '
-                                 'Disable auto_update in config file.')
-            dyn_conf = self.get_collection_rules()
-            if dyn_conf:
-                version = dyn_conf.get('version', None)
-                if version is None:
-                    raise ValueError("ERROR: Could not find version in json")
-                dyn_conf['file'] = self.collection_rules_file
-                logger.debug("Success reading config")
-                config_hash = hashlib.sha1(json.dumps(dyn_conf).encode('utf-8')).hexdigest()
-                logger.debug('sha1 of config: %s', config_hash)
-                return dyn_conf, rm_conf
         for conf_file in [self.collection_rules_file, self.fallback_file]:
             logger.debug("trying to read conf from: " + conf_file)
             conf = self.try_disk(conf_file, self.gpg)
-            if conf:
-                version = conf.get('version', None)
-                if version is None:
-                    raise ValueError("ERROR: Could not find version in json")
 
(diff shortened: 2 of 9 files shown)


Glutexo:
There are some failing tests in master (https://github.com/RedHatInsights/insights-core/issues/1231). When I run `pytest` locally, I don’t see more failing tests than on master. Are there other tests suite to check? I’ll at least look into Jenkins.

gravitypriest:
Yeah they're the same as the local tests, but if those tests fail GitHub
won't allow the merge. :)

Glutexo:
Still I am confused that the tests that fail in Jenkins are different ones than those that fail for me locally:

[Here in Jenkins](http://10.8.252.209:8080/job/CI-PR-insights-core/33/) you can see 38 failing tests. All of them are failing asserts after running some external command `run(cmd)`. My local test suite on the other hand reports 2–4 failures (2 of them are happening only on some runs), both in `test_text_file_provider.py` failing with `AttributeError: type object 'TextFileProvider' has no attribute 'validate_lines'`. This test however does not fail in Jenkins.

How is that possible? This distillates to two questions:

* Why there are so many failing tests in Jenkins and why those don’t fail locally?
* Why the tests that fail locally, don’t fail in Jenkins?

Is there something I can do to fix that? Without having clear test results that are coherent between my local machine and Jenkins it is difficult to trust them. Thanks!

kylape:
@Glutexo we have two jenkins jobs that are triggered to run against insights-core PRs.  The job labeled "continuous-integration" runs flake8 and the unit/integration tests found in insights-core that you can run on your local box.  The job labeled "default" is managed by our QE team and runs tests that QE has written in a separate repository.  Hope that clears up some confusion, at least.

Glutexo:
Rebased on curent master and resolved conflicts.

eduardocerqueira:
test this please

Glutexo:
@eduardocerqueira K. I’ll try to figure out something. Originally I didn’t add any tests just because there already weren’t any. I was satisfied with the fact that the changes didn’t introduce any failure.

eduardocerqueira:
test this please

Glutexo:
@eduardocerqueira I‘ve finished regression unit tests and verified by them that my changes don’t change the behavior. I just have to DRY them before submitting to this PR.

@gravitypriest Thanks for the review! I’ll ask you for one more review of the tests once I submit them.

Glutexo:
I returned to this to finally finish the tests, so this can get properly reviewed and merged.

[First](
https://github.com/RedHatInsights/insights-core/pull/1230/commits/245ff9e9183dcd39a8e86f639ac5c36473ce1182), I created some tests that work both with the old and new code, testing the internal workings of the `get_conf` method. [Then](
https://github.com/RedHatInsights/insights-core/pull/1230/commits/e31b558e8084e7b4ff5cd622c6ef9de63cecbdef), I wrote unit tests for the new [`get_conf_*`](https://github.com/RedHatInsights/insights-core/pull/1230/files#diff-7b50997b6b5a9fcee662820ddb2de96eR170) and [`get_rm_conf`](https://github.com/RedHatInsights/insights-core/pull/1230/files#diff-7b50997b6b5a9fcee662820ddb2de96eR227) methods and adapted the original tests to trust those currently tested and thus trusted methods. Still, I felt like there are the regression tests now missing and that’s why I wasn’t brave enough for a quite long time to get to that.

So [finally](
https://github.com/RedHatInsights/insights-core/pull/1230/commits/cc95c9873292488838d86515794565c90e10ed5e), I added some new regression tests, that still work both before and after the changes. Those however don’t test the gory details of the already trusted methods and focus only on the returned values and some important points that should be verified that actually happen: throwing errors, signature validation and such, but no every single line of code.

Unfortunately the underlying codebase has changed a bit, so It‘s not a conflictless merge. I‘m going to rebase this on the current master. I don’t expect this to be hard, because only [`update_rules`](https://github.com/RedHatInsights/insights-core/pull/1230/files#diff-ee9a2b119a69ec61a776b377f494e5e2R224) method interface changed.

Glutexo:
It’s there, rebased and linted! Thanks to the changes in the [update_rules](https://github.com/RedHatInsights/insights-core/blob/fef4c234c8e97ae1a8702679b2c348a14b685b50/insights/client/client.py#L203) and [collect](https://github.com/RedHatInsights/insights-core/blob/fef4c234c8e97ae1a8702679b2c348a14b685b50/insights/client/client.py#L232) methods there is less patching necessary, saving lines of code and cognitive complexity. That‘s great!

The regression tests are marked and can be run with `pytest -m regression`. I don’t expect them to live there too long, when the code they are checking changes more. They serve their purpose now, verifying that the refactoring didn't break the functionality.

I’d like to ask @gravitypriest or @kylape for a new review. @gravitypriest already checked the changes and the rebase didn’t change much. But it would be worth reviewing the new test suites. Thanks!

gravitypriest:
@Glutexo check out the failing test in Python 3 in the pr-merge tests. Tempfiles in are opened in 'wb' mode by default and it's a lot pickier in Py3. Just tack an `.encode()` on there and you'll be good.

Glutexo:
@gravitypriest I’ll take a look. Thanks for pointing out!

Glutexo:
So the main problem is that Python 2 reads stdin as bytes, but Python 3 as Unicode. In both cases this is the _str_ type. However _json.load_ refuses to deal with bytes on Python 3.

Glutexo:
I investigated the situation a bit:

Sticking `.encode` to the string in the test is not viable: The temporary file would be still read as bytes on Python 3, but _stdin_ is opened as Unicode. _StringIO_ on the other hand reads always Unicode, even on Python 2.

The only working and also simplest solution (Occam’s razor for the win) is to open the temporary file mocking the _stdin_ in text mode `w+t`. This returns bytes (str) on Python 2 and Unicode (str) on Python 3, just as the real _stdin_ does. 🎉 

As you can see, the tests are still failing. The bad news is that my test unraveled some bugs that were lingering in the code unnoticed. Reading configuration from _stdin_ simply fails on Python 3, because exactly the same mistake I did in that test of mine. Configuration is read as Unicode, but a _NamedTemporaryFile_ used to write it to for signature validation is opened in binary mode and thus failing. https://github.com/RedHatInsights/insights-core/blob/9e2c8e548128e53e4a15ce599c11d3586594f211/insights/client/collection_rules.py#L183

There are more failures on Python 3 and it is possible that some of them are my buggy tests and not bugs in the actual code. But this one is surely not my fault and happens with current master. If you want to see it yourself, run [this script](https://gist.github.com/Glutexo/d6a5944bc0bf28f08ae46de1bcd78f0e) in the _insights-core_ Python 3 environment with a following command. There is _get_branch_info_ mocked out to avoid network communication, everything else is untouched.

```
$ echo '{"uploader.json": "{\"some key\": \"some value\"}", "sig": "some signature"}' | python bug.py
…
TypeError: 'str' does not support the buffer interface
```

Glutexo:
Rebased, tests pass on Python 3. @gravitypriest fixed the encoding bugs.

Party hard

Review? Merge? Comments? @eduardocerqueira, @kylape?

gravitypriest:
Ship it

PR #1232, opened 2018-06-27
few tweaks for py3

gravitypriest:
- `shlex.split()` in python 3 barfs if you give it unicode
- `hashlib.sha1()` in python 3 expects bytes
- Need to explicitly write bytes to file instead of str


kylape:
@gravitypriest did you make these changes because of specific errors you encountered when running the client with py3?

gravitypriest:
Yep, and tested with py2 as well to confirm the changes worked in both.

PR #1234, opened 2018-06-28
new parser for rhv_log_collector_analyzer

dougsland:
Parser for rhv-log-collector-analyzer tool


insights/archive/repository/base_archives/rhel7/insights_commands/rhv-log-collector-analyzer_--json
```diff
@@ -0,0 +1,188 @@
+{
```
bfahr: @dougsland it is not necessary to add this to the archive template.  It is used to test the archive code.

insights/parsers/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,22 @@
+"""
+RHV Log Collector Analyzer
+=================================
+
```
bfahr: In order to include this parser in the documentation, please add a new file `docs/shared_parsers_catalog/rhv_log_collector_analyzer.rst` which looks like this:
```
.. automodule:: insights.parsers.rhv_log_collector_analyzer
   :members:
   :show-inheritance:
```
You can find information on writing and testing documentation here: https://insights-core.readthedocs.io/en/latest/docs_guidelines.html

I believe when you attempt to build the docs you will get an error from Sphinx caused by the `===...` on line 3 being longer than the title string above it.  They must be the same length.

insights/archive/repository/base_archives/rhel7/insights_commands/rhv-log-collector-analyzer_--json
dougsland: Thanks, removing.

insights/parsers/rhv_log_collector_analyzer.py
dougsland: good catch, done

PR #1236, opened 2018-06-29
Added some more fs tests

psav:
* Added test to change touch times
* Added parametrized test of modes

### DIFF
--- a/insights/tests/test_fs.py
+++ b/insights/tests/test_fs.py
@@ -1,13 +1,14 @@
-import os
 import shutil
 import tempfile
-
 from contextlib import closing
+
+import os
+import pytest
+
 from insights.util import fs
 
 
 def test_remove_exists():
-    full_path = None
     with closing(tempfile.NamedTemporaryFile(delete=False)) as f:
         full_path = f.name
         f.write(b'test')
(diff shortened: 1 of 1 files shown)


psav:
So I expected at least one failure - On my local system, running as a standard user, it won't set mode 777, and running as root, it won't set 777, or 775. This is an issue with Python more than the library, but it worries me that we expect it to work and it potentially doesnt. We even state that the default is 0o777, but I have not seen this work. What should we do here?

insights/tests/test_fs.py
```diff
@@ -42,3 +43,28 @@ def test_exists_ensure_path_exists():
...
+    fs.ensure_path(mode_dir, int(mode, 8))
+    assert oct(os.stat(mode_dir).st_mode)[-3:] == mode
+
+
+def test_touch_times():
+    path = os.path.join('/tmp/insights-test-touch-times')
+    fs.touch(path, (1259798405, 1259798400))
+    assert os.stat(path).st_atime == 1259798405
+    assert os.stat(path).st_mtime == 1259798400
+
```
bfahr: @psav if you remove this blank line at the end of the file you can fix the flake8 error.  You can run `flake8` check from the root directory of the insights-core project to check for these errors.

psav:
Sure thing

PR #1237, opened 2018-06-29
Added multi line syslog entries into tests

psav:
Added a test here for multi line as I experienced in logs before if we are expecting the second/third lines not to related to the first, then this test can be removed I guess.

### DIFF
--- a/insights/tests/test_syslog.py
+++ b/insights/tests/test_syslog.py
@@ -11,6 +11,12 @@ Apr 22 10:35:01 boy-bona CROND[27921]: (root) CMD (/usr/lib64/sa/sa1 -S DISK 1 1
 Apr 22 10:37:32 boy-bona crontab[28951]: (root) LIST (root)
 Apr 22 10:40:01 boy-bona CROND[30677]: (root) CMD (/usr/lib64/sa/sa1 -S DISK 1 1)
 Apr 22 10:41:13 boy-bona crontab[32515]: (root) LIST (root)
+Jul 10 15:08:35 shelly audit[1]: USER_AVC pid=1 uid=0 auid=4112341244 ses=49287573824 subj=system_u:system_r:init_t:s0 msg='avc:  received policyload notice (seqno=2)
+                                 exe="/usr/lib/systemd/systemd" sauid=0 hostname=? addr=? terminal=?'
+Jul 10 15:08:35 shelly audit[1]: USER_AVC pid=1 uid=0 auid=4112341244 ses=49287573824 subj=system_u:system_r:init_t:s0
+                                 msg='avc:  received policyload notice (seqno=2)
+                                 exe="/usr/lib/systemd/systemd" sauid=0 hostname=? addr=? terminal=?'
+Jul 10 15:08:35 shelly systemd[1]: Reloading.
 """.strip()
 
 
@@ -27,3 +33,5 @@ def test_syslog():
     assert msg_info.get('Wrapper')[0].get('message') == "--> Wrapper Started as Daemon"
     assert msg_info.get('Launching')[0].get('raw_message') == "May 18 15:13:36 lxc-rhel68-sat56 wrapper[11375]: Launching a JVM..."
     assert 2 == len(msg_info.get('yum'))
+    import pdb; pdb.set_trace()
+    assert "exe" in msg_info.get('audit')[0].get('raw_message')


insights/tests/test_syslog.py
```diff
@@ -27,3 +33,5 @@ def test_syslog():
     assert msg_info.get('Wrapper')[0].get('message') == "--> Wrapper Started as Daemon"
     assert msg_info.get('Launching')[0].get('raw_message') == "May 18 15:13:36 lxc-rhel68-sat56 wrapper[11375]: Launching a JVM..."
     assert 2 == len(msg_info.get('yum'))
+    import pdb; pdb.set_trace()
```
bfahr: @psav you need to remove this line from the test so that it won't interfere with running tests.

insights/tests/test_syslog.py
```diff
@@ -11,6 +11,12 @@
 Apr 22 10:37:32 boy-bona crontab[28951]: (root) LIST (root)
 Apr 22 10:40:01 boy-bona CROND[30677]: (root) CMD (/usr/lib64/sa/sa1 -S DISK 1 1)
 Apr 22 10:41:13 boy-bona crontab[32515]: (root) LIST (root)
+Jul 10 15:08:35 shelly audit[1]: USER_AVC pid=1 uid=0 auid=4112341244 ses=49287573824 subj=system_u:system_r:init_t:s0 msg='avc:  received policyload notice (seqno=2)
+                                 exe="/usr/lib/systemd/systemd" sauid=0 hostname=? addr=? terminal=?'
+Jul 10 15:08:35 shelly audit[1]: USER_AVC pid=1 uid=0 auid=4112341244 ses=49287573824 subj=system_u:system_r:init_t:s0
+                                 msg='avc:  received policyload notice (seqno=2)
+                                 exe="/usr/lib/systemd/systemd" sauid=0 hostname=? addr=? terminal=?'
+Jul 10 15:08:35 shelly systemd[1]: Reloading.
```
bfahr: @psav are messages actually stored in the file this way or is this an artifact of how you view the messages?  Insights core captures an unfiltered file like this:
```
with open(path, "rU") as f:
    results = [l.rstrip("\n") for l in f]
return results
```
Would you get the same results with continued lines if the file is read in this manner?

bfahr:
@psav this is an interesting test case since most log files in syslog format are filtered, meaning that only the lines containing filter strings are returned.  Filtering could produce interesting results if lines in the log are continued in this manner.  For unfiltered files, the current syslog parser does not support lines that are continued as shown in these tests.  So if the continuation is not an artifact of how the log file is viewed, then I'd say this is an error in the Syslog parser for unfiltered log files.

insights/tests/test_syslog.py
psav: oh my goodness - did I really leave that in there.....crikey - I'm so sorry - clearly my brain was in amateur hour

psav:
I did just find this in a journalctl, it's in the syslog format, but not sure where it resides, so I guess it's possible. I'm happy to just close the PR if you don't think it's a valid test.
```
Jul 10 15:07:13 machine python3[4693]: SELinux is preventing systemd from create access on the unix_stream_socket Unknown.
                                     
                                     *****  Plugin catchall (100. confidence) suggests   **************************
                                     
                                     If you believe that systemd should be allowed create access on the Unknown unix_stream_socket by default.
                                     Then you should report this as a bug.
                                     You can generate a local policy module to allow this access.
                                     Do
                                     allow this access for now by executing:
                                     # ausearch -c 'systemd' --raw | audit2allow -M my-systemd
                                     # semodule -X 300 -i my-systemd.pp
```

xiangce:
Seems the base class `LogFileOutput` still doesn't support analyzing continuation multiple lines, they will be treated as separated normal lines.

Glutexo:
From the [discussion](https://github.com/RedHatInsights/insights-core/pull/1237#issuecomment-402651869) I understand that _LogFileOutput_, which _Syslog_ is a child of, does not support messages spanning multiple lines. The fact that the [test](https://github.com/RedHatInsights/insights-core/pull/1237/files#diff-e0cc51b328f6a20859f526395a802b8763157b1f66cd8f8cb71eac0cb0ed7d3cR14) starts failing with the patch applied only supports this.

```
FAIL insights/tests/test_syslog.py::test_syslog

============================================================================================ FAILURES =============================================================================================
___________________________________________________________________________________________ test_syslog ___________________________________________________________________________________________

    def test_syslog():
        msg_info = Syslog(context_wrap(MSGINFO))
        bona_list = msg_info.get('(root) LIST (root)')
        assert 2 == len(bona_list)
        assert bona_list[0].get('timestamp') == "Apr 22 10:37:32"
        assert bona_list[1].get('timestamp') == "Apr 22 10:41:13"
        crond = msg_info.get('CROND')
        assert 2 == len(crond)
        assert crond[0].get('procname') == "CROND[27921]"
        assert msg_info.get('jabberd/sm[11057]')[0].get('hostname') == "lxc-rhel68-sat56"
        assert msg_info.get('Wrapper')[0].get('message') == "--> Wrapper Started as Daemon"
        assert msg_info.get('Launching')[0].get('raw_message') == "May 18 15:13:36 lxc-rhel68-sat56 wrapper[11375]: Launching a JVM..."
        assert 2 == len(msg_info.get('yum'))
        # import pdb; pdb.set_trace()
>       assert "exe" in msg_info.get('audit')[0].get('raw_message')
E       AssertionError

insights/tests/test_syslog.py:37: AssertionError
```

As @psav himself [mentions](https://github.com/RedHatInsights/insights-core/pull/1237#issuecomment-401774440), this is an invalid test case then. A test like this would be a valid part of a _LogFileOutput_ patch introducing the ability to capture multi-line messages.

If anybody (@xiangce, @bfahr, @psav) of you guys think we need the feature, feel free to create an Issue, or a JIRA ticket and link this PR. Since there is no bug report or a use case linked, I am closing it as invalid and currently not needed.

Issue #1238, opened 2018-06-29
insights.utils.fs.ensure_path is not setting requested mode on dirs


bfahr:
#1236 exposes an potential problem in `fs.ensure_path`.  The umask could prevent the function from setting the requested mode.  This patch will enable the function to successfully pass the test assuming we want to change the umask value:
```diff
diff --git a/insights/util/fs.py b/insights/util/fs.py
index fc5cc818..e4557e1c 100644
--- a/insights/util/fs.py
+++ b/insights/util/fs.py
@@ -94,7 +94,9 @@ def ensure_path(path, mode=0o777):
 
     if path:
         try:
+            umask = os.umask(000)
             os.makedirs(path, mode)
+            os.umask(umask)
         except OSError as e:
             if e.errno != errno.EEXIST:
                 raise
```

bfahr:
Fixed with merge referenced above.

PR #1239, opened 2018-06-29
Fixed problem with setting tmp file permissions

bfahr:
* Fixes issue #1238 by clearing umask before setting mode and then
  resetting umask to the original value after setting mode.


bfahr:
This will fix the test failure in #1236.

PR #1241, opened 2018-07-02
Enhance parser mysqladmin with bad_lines property

JoySnow:
Record bad lines and keep on parsing instead of raise exception.


miclark:
Closing the related issue #1242

PR #1243, opened 2018-07-05
Sap combiner

xiangce:
Since most of the SAP rules would use this logic to check is the host an SAP server, make it a shared combiner in the core.


eduardocerqueira:
test this please

PR #1244, opened 2018-07-05
Fix the dead loop in Mount.get_dir() in some cases

JoySnow:

### DIFF
--- a/insights/parsers/tests/test_mount.py
+++ b/insights/parsers/tests/test_mount.py
@@ -2,6 +2,7 @@
 test mount
 ==========
 """
+from insights.parsers import ParseException
 from insights.parsers.mount import Mount
 from insights.tests import context_wrap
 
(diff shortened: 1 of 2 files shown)


JoySnow:
https://github.com/RedHatInsights/insights-core/issues/1245

eduardocerqueira:
test this please

insights/parsers/tests/test_mount.py
```diff
@@ -97,3 +97,20 @@ def test_mount():
     assert results.search(mount_options__contains='seclabel') == [
         results.rows[n] for n in (0, 1, 3, 4, 5, 7, 8, 11)
     ]
+
+
+MOUNT_WITHOUT_ROOT = """
```
xiangce: This kind of mount content is not correct, for `mount` without the root '/', it's totally impossible and incorrect, for this cases, we should raise an Exception when finding this issue at the end of the parser.

bfahr:
@JoySnow this change will break backward compatibility with existing plugins.  In this instance the change is to fix a bug which is an exception to the rule about breaking backward compatibility.  I just want to confirm that this is the correct approach.

JoySnow:
@bfahr 
Those tests cases in existing plugins are not right without `/` mount point. (As we had supposed that  '/' must exists. ) 
I had fixed those test cases. We can merge this PR without any  incompatibility .

PR #1246, opened 2018-07-05
Parser for command: `virsh --readonly list --all`

psachin:
- [X] Parser
- [X] Tests
- [X] Docs
- [X] Specs


insights/parsers/virsh_list_all.py
```diff
@@ -0,0 +1,138 @@
...
+        Args:
+
+            vmname (str): A key. For ex. ``rhel7.0``.
+
+        Returns:
+
+            str: State of VM. Returns None if, ``vmname`` does not exist.
+        '''
+        vmname = vmname.lower()
+        if vmname in self.keywords:
```
xiangce: Though the code looks simple, don't you think `search` is too heavy? how about implement it by iterating the `self.cols` like:

```
st = [i['state'] for i in self.cols if i['name'] == vmname]
return st[0] if st else None
```

insights/specs/default.py
```diff
@@ -618,6 +618,7 @@ def tomcat_home_base(broker):
     vgs = None  # simple_command('/sbin/vgs -v -o +vg_mda_count,vg_mda_free,vg_mda_size,vg_mda_used_count,vg_tags --config="global{locking_type=0}"')
     vgs_noheadings = simple_command("/sbin/vgs --nameprefixes --noheadings --separator='|' -a -o vg_all --config=\"global{locking_type=0}\"")
     vgs_noheadings_all = simple_command("/sbin/vgs --nameprefixes --noheadings --separator='|' -a -o vg_all --config='global{locking_type=0} devices{filter=[\"a|.*|\"]}'")
+    virsh_list_all = simple_command("/usr/bin/virsh --readonly list --all")
```
xiangce: For new `simple_command`, please add its corresponding `simple_file` to the `insights_archives.py` as well

bfahr:
@psachin I discussed this with @kylape and @lhuett and we were wondering if you had considered obtaining this information from the `PsAux` parser instead by filtering on the command `/usr/bin/qemu-system-x86_64`? It appears to provide a significant amount of information for each running VM.

bfahr:
@psachin what's the status of this PR?  Looks like @xiangce recommended a few changes.

bfahr:
@xiangce this PR looks good to me if all of your questions have been resolved.

psachin:
@bfahr Yes. changes requested by @xiangce are included.

PR #1251, opened 2018-07-10
Do not retry if server returns 412 on upload

Glutexo:
Status code 412 means that the machine is unregistered. Thus there is almost no point in retrying as this is not a temporary condition.

This is rather a “hotfix”, since the `--retry` parameter is ignored regardless of whether the `.unregistered` file is actually written. This is because the `InsightsConnection.handle_fail_rcs` actually does not treat a 412 without a valid API response body as a reason to unregister. I’d address this separately.

If the assumption that 412 should result in not retrying at all is correct, then this solves https://bugzilla.redhat.com/show_bug.cgi?id=1554794 and https://projects.engineering.redhat.com/browse/MAASPLAT-205. @kylape and @YakovLittle can confirm this.

Added some tests and along with that [mock](https://pypi.org/project/mock/) as a development requirement.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,6 +10,8 @@ pipeline {
             }
           }
           steps {
+            echo "Installing Insights..."
+            sh 'pip install --user -e .[develop]'
             echo "Testing with Pytest..."
             sh 'pytest'
           }
@@ -22,7 +24,7 @@ pipeline {
           }
           steps {
             echo "Installing Insights..."
-            sh 'pip install --user -e .'
+            sh 'pip install --user -e .[develop]'
             echo "Testing with Pytest..."
             sh 'pytest'
             echo "Testing with flake8..."   
@@ -37,7 +39,7 @@ pipeline {
           }
           steps {
             echo "Installing Insights..."
-            sh '/bin/python36 -m pip install --user -e .'
+            sh '/bin/python36 -m pip install --user -e .[develop]'
             echo "Testing with Pytest..."
             sh '/bin/python36 -m pytest'
             echo "Testing with flake8..."   
--- a/insights/client/client.py
+++ b/insights/client/client.py
@@ -448,6 +448,7 @@ def upload(config, tar_file, collection_duration=None):
 
         elif upload.status_code == 412:
             pconn.handle_fail_rcs(upload)
+            break
         else:
             logger.error("Upload attempt %d of %d failed! Status Code: %s",
                          tries + 1, config.retries, upload.status_code)
--- a/insights/tests/client/test_client.py
+++ b/insights/tests/client/test_client.py
@@ -3,7 +3,11 @@ import sys
 from insights.client import InsightsClient
 from insights.client.config import InsightsConfig
 from insights import package_info
+from mock.mock import patch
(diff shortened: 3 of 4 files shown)


Glutexo:
@eduardocerqueira How can I introduce [mock](https://pypi.org/project/mock/) to the build? I [added](missing) `'mock==2.0.0'` to the `develop` dependencies in `setup.py`, but apparently it is not enough. Jenkins  still [complains](https://jenkins-insights-ci.1b13.insights.openshiftapps.com/blue/organizations/jenkins/insights-core/detail/PR-1251/3/pipeline/) about missing `mock.mock`.

YakovLittle:
test this please

Glutexo:
@YakovLittle I tested it locally. Unfortunately I don’t know how to make PyTest in Jenkins use the mock library. Hopefully we will find out soon.

eduardocerqueira:
Hi @Glutexo I assume @YakovLittle already chat with you but just to make sure you got it, our suggestion would be to change in your Jenkinsfile to run: `pip install --user -e ./[develop]` for all stages and amend it here to this PR.

Glutexo:
@eduardocerqueira Got it! I didn’t realize that the Jenkinsfile is in the repo and not somewhere in the GitHub bot or in the Jenkins instance. The check is green now. 🍏 So, ready to merge? @gravitypriest 

Still I have some concerns that I‘m going to address in separate issues:

* The _develop_ extras bundle is too big to include for just running tests. I’d prefer to extract a _test_ bundle and make it a part of _develop_. The same applies to documentation etc. The reasoning behind this is that it makes easy to include a runtime dependency that wouldn’t make the check fail, because the same dependency would be listed in the _develop_ bundle.
* RHEL6 check is missing the _flake8_ lint. I’m not sure however whether it is necessary to run _flake8_ for all three platforms. As far as I know, it checks only the syntax regardless of the Python version installed, so it might be possible to run it only once, just as we run the Docs Check only once.

One more thing: When there is something wrong with the PR and it might seem untested, please write me some more detail in your comments. It might not be always obvious and saying just “test this” when there are new tests and they are passing results in a rather detective work. I’d appreciate something like “Your new tests are failing in Jenkins checks, complaining about missing mock.mock,  fix it.” more. Thanks, guys!

gravitypriest:
Please rebase against master first.

Glutexo:
Rebased. (And I have to correct myself: not ready to merge, but ready for a review. 😊)

eduardocerqueira:
test this please

Glutexo:
@eduardocerqueira There are tests for (not) retrying on different HTTP status codes. We were discussing that because of the additional dependencies. As I [wrote](https://github.com/RedHatInsights/insights-core/pull/1251#issuecomment-404093973) to @YakovLittle, I’d like to ask you as well: if there is something missing from the tests, please point me in the right direction like „you’re not testing x“ or such. Thanks! What can I do here to improve the tests, please?

Glutexo:
Cool. Merge? Who has the merge rights? Maybe @kylape?

gravitypriest:
Merged because this passed checks the other day, but the Jenkins pipelines are acting up today.

PR #1257, opened 2018-07-11
new parser for ls_usr_sbin with two filter

TripleJQK:


insights/parsers/ls_usr_sbin.py
```diff
@@ -0,0 +1,46 @@
...
+
+from insights.core.filters import add_filter
+from insights.specs import Specs
+
+from .. import CommandParser, parser
+from .. import FileListing
+
+
+add_filter(Specs.ls_usr_sbin, "total")
+add_filter(Specs.ls_usr_sbin, "postdrop")
```
xiangce: I think it's OK to add the 'total' here, but better add the 'postdrop' in the corresponding plugins, since different plugins require a different filter.

insights/parsers/ls_usr_sbin.py
```diff
@@ -0,0 +1,46 @@
...
+
+Examples:
+
+    >>> "accessdb" in ls_usr_sbin
+    False
+    >>> "/usr/sbin" in ls_usr_sbin
+    True
+    >>> ls_usr_sbin.dir_entry('/usr/sbin', 'accessdb')['type']
+    '-'
+
```
bfahr: @TripleJQK based on the filter you have added below, these examples are invalid.  I think the best approach would be for you to include documentation that clearly states this parser is being filtered and describe how someone using this parser should use `add_filter` to include additional data.  You could also add examples showing how adding a filter would change the parser output.

xiangce: @bfahr - Would you please to check if the new modification could satisify your requests?

insights/parsers/ls_usr_sbin.py
```diff
@@ -0,0 +1,46 @@
...
+
+from insights.core.filters import add_filter
+from insights.specs import Specs
+
+from .. import CommandParser, parser
+from .. import FileListing
+
+
+add_filter(Specs.ls_usr_sbin, "total")
```
bfahr: @TripleJQK if this is being filtered for just this information then shouldn't it just be a `CommandParser` and not a `FileListing` parser?  The filter would probably cause the `FileListing` parser to fail and no data would be provided.

TripleJQK: @bfahr hi, thank you for your advice.
At first, I put two filter here:
~~~
add_filter(Specs.ls_usr_sbin, "total")
add_filter(Specs.ls_usr_sbin, "postdrop")
~~~
for more, please see: https://github.com/RedHatInsights/insights-core/pull/1257/commits/b6fce0a63bf2f56c68660851c3bad9cce78e07ea

But by xiangce's advice, he thought:"I think it's OK to add the 'total' here, but better add the 'postdrop' in the corresponding plugins, since different plugins require a different filter." So I delete the postdrop filter from the parser and will add it to the corresponding plugin(which is not coded yet, as pr is not merged). @bfahr do think it is good to do it as xiangce's way? I think other other parser like ps are also in this way

TripleJQK: @bfahr I have make a plugin code with only filter first here:
https://github.com/RedHatInsights/insights-plugins/pull/959

insights/parsers/ls_usr_sbin.py
bfahr: I believe the example is still confusing because you would never use the `add_filter` command in that manner.

TripleJQK:
@bfahr @xiangce I think jenkins has some changed and make the insights pytest fail, my pr is ok in my local env

esakaiev:
test this please

insights/parsers/ls_usr_sbin.py
```diff
@@ -0,0 +1,76 @@
...
+
+Sample added filter:
+
+    >>> add_filter(Specs.ls_usr_sbin, "accessdb")
+
+Sample directory list collected::
+
+    total 41472
+    -rwxr-xr-x. 1 0  0   11720 Mar 18  2014 accessdb
+    -rwxr-sr-x. 1 0 90  218552 Jan 27  2014 postdrop
```
xiangce: Should this "postdrop" be here?

insights/parsers/ls_usr_sbin.py
```diff
@@ -0,0 +1,76 @@
...
+
+Examples:
+
+    >>> add_filter(Specs.ls_usr_sbin, "accessdb")
```
xiangce: Did you missed "postdrop" here?

PR #1268, opened 2018-07-16
Update spec factory helpers to be classes instead of functions.

csams:
Allows
```
In [1]: from insights.specs.default import DefaultSpecs

In [2]: DefaultSpecs.hosts.path
Out[2]: '/etc/hosts'

In [3]: DefaultSpecs.hostname.cmd
Out[3]: '/usr/bin/hostname -f'
```

Also updated `insights/tools/specs_to_rules.py`


csams:
Will open another PR that includes more simplifying changes.

PR #1271, opened 2018-07-17
Update plugin decorators and datasource helpers to be class based

csams:
*  Remove the need for metaclasses to name the standard plugin types correctly (datasource, parser, combiner, etc. don't have to be defined under a `TypeSet` class - they're just classes that inherit from `ComponentType`).
*  Provide a more intuitive way to customize decorator interfaces and how components defined with them get executed.
* Expand observer type checking so its possible to add an observer for any subtype of a given component type.
* Expand component type checking to include subclasses instead of exact match.

In addition to simplifying internals, the changes allow things like:
```python
In [1]: from insights.specs.default import DefaultSpecs
In [2]: DefaultSpecs.hosts.path
Out[2]: '/etc/hosts'

In [3]: DefaultSpecs.hostname.cmd
Out[3]: '/usr/bin/hostname -f'
```
or
```python
In [1]: from insights.parsers.installed_rpms import InstalledRpms
In [2]: from insights import dr
In [3]: d = dr.get_delegate(InstalledRpms)
In [4]: d.requires
Out[4]: [insights.specs.Specs.installed_rpms]

In [5]: d.optional
Out[5]: []

In [6]: d.at_least_one
Out[6]: []
```


insights/core/dr.py
```diff
@@ -376,8 +361,119 @@ def register_component(delegate):
...
 
-    name = get_name(component)
-    COMPONENT_NAME_CACHE[name] = component
+
+class ComponentType(object):
+    """
+    ComponentType is the base class for all component type decorators.
+    """
+
+    requires = []
```
jhjaggars: Setting these names up as mutable state types here seems like a source of possible confusion and errors in the future.

csams: They're meant to be overridden in base classes, and the initialization of the base classes hide them. They're the "auto_requires" of the previous implementation.

Issue #1272, opened 2018-07-18
insights hang when extract jdr archive


chenlizhong:
The process hang at below code in archives.py

   command = "unzip -q -d %s %s" % (self.tmp_dir, path)

Try unzip command at shell and wait answer like below:

```
$unzip -q -d /tmp/mytempdir ./jdr_18-07-13_11-11-11_mylinux.zip
replace /tmp/mytempdir/jdr_18-07-13_11-11-11_mylinux/JBOSS_HOME/standalone/configuration/standalone_xml_history? [y]es, [n]o, [A]ll, [N]one, [r]ename: n
```

Add -n parameter can resolve this issue.

 -n     never  overwrite existing files.

PR #1273, opened 2018-07-18
Add -n parameter into unzip command in archives.py

chenlizhong:
-n means never overwrite existing files

Resolve issue: https://github.com/RedHatInsights/insights-core/issues/1272


csams:
@chenlizhong When are you running into this issue? Extractions should always happen in a random temporary directory that doesn't conflict with other directories.

PR #1274, opened 2018-07-18
Add jdr_archive.py and three specs

chenlizhong:
- jdr_archive.py is specs file for archive from JBoss Diagnostic Reporter


insights/specs/jdr_archive.py
```diff
@@ -0,0 +1,48 @@
...
+                    results.append(line)
+            if results:
+                # default is standalone.xml
+                config_xml = 'standalone.xml'
+                java_command = results[-1]
+                if '--server-config' in java_command:
+                    config_xml = java_command.split('--server-config=')[1].split()[0]
+                elif '-c ' in java_command:
+                    config_xml = java_command.split('-c ')[1].split()[0]
+                return config_xml
```
csams: This should return a list like `[config_xml]` for consistency with the default return value `[]` and use in a `foreach_collect`.

insights/specs/jdr_archive.py
```diff
@@ -0,0 +1,48 @@
...
+        return []
+
+    # use foreach is to return list type result as jboss sepcs in defaults.py
+    jboss_standalone_main_config = foreach_collect(jboss_standlone_conf_file,
```
csams: Maybe we should add `foreach_collect = partial(foreach_collect, context=JDRContext)` to the top of the file. It's not necessary to specify a context with `foreach_collect` since it won't execute if its dependencies don't execute, and they already depend on `JDRContext`. It doesn't hurt for it to depend on `JDRContext` anyway, but let's make it consistent with the other datasources in the file if we do that.

insights/specs/jdr_archive.py
chenlizhong: Yes. Agree with you.

insights/specs/jdr_archive.py
chenlizhong: I prefer the foreach_collect = partial(foreach_collect, context=JDRContext) way.

insights/specs/jdr_archive.py
```diff
@@ -0,0 +1,45 @@
...
+from insights.core.spec_factory import simple_file, foreach_collect, first_file, glob_file, listdir
+
+first_file = partial(first_file, context=JDRContext)
+glob_file = partial(glob_file, context=JDRContext)
+simple_file = partial(simple_file, context=JDRContext)
+foreach_collect = partial(foreach_collect, context=JDRContext)
+
+
+class JDRSpecs(Specs):
+    # Here use glob_file is to return list type result to keep same with  jboss specs default.py
```
bfahr: @chenlizhong please add docstrings to this module and class so they will be included in the documentation.  You will need to also add a reference to this module in `docs/api_index.rst`.

insights/specs/jdr_archive.py
```diff
@@ -0,0 +1,45 @@
...
+    # jdr just collect files under one jboss home,
+    # For insights-client, we will collect files under different jboss home, return list is necessary.
+    jboss_standalone_server_log = glob_file("JBOSS_HOME/standalone/log/server.log")
+
+    @datasource(jboss_standalone_server_log, context=JDRContext, multi_output=True)
+    def jboss_standlone_conf_file(broker):
+        # read the server.log to detect if specific xml is using in the startup command
```
bfahr: Please add a docstring to this datasource to explain what is being collected by this datasource.

insights/util/component_graph.py
```diff
@@ -100,6 +100,7 @@ def main(filename):
     dr.load_components("insights.specs.default")
     dr.load_components("insights.specs.insights_archive")
     dr.load_components("insights.specs.sos_archive")
+    dr.load_components("insights.specs.jdr_archive")
```
bfahr: Also, please add this line to `insights/tools/specs_to_rules.py`.

chenlizhong: @bfahr 
I checked the specs_to_rules.py, we do not load jdr_archive, sos_archive and insights_archive. It works well.

insights/specs/jdr_archive.py
chenlizhong: Yes. will add that.

insights/util/component_graph.py
bfahr: You are correct, sorry about that.

PR #1279, opened 2018-07-19
Fix unregister on timeout

gravitypriest:
This began as simply fixing the dotfiles being erased on timeout and turned into so much more...

Ended up refactoring registration functions a bit. I modified some of the "vision" functions as some of them were either incompletely implemented or copied code from elsewhere that was being unused.

Also removed the connection singleton due to an issue where its config was stale; the config in the connection class was a different object than the one it was initialized with. Moved the connection persistence to the `InsightsClient` class instead. Used a decorator to mark any methods that need network functionality. Also added an `http_timeout` option.

Finally, tests! I wanted to make sure registration was still behaving as intended.

There's a lot of passing of connection along with config now, and I'd like to refactor some more at some point (maybe move stuff from client.py to __init__.py) but that's a philosophical question for another day.

### DIFF
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -290,6 +290,7 @@ class InsightsClient(object):
         logger.debug("The new Insights Core was installed successfully.")
         return {'success': True}
 
+    @_net
     def update_rules(self):
         """
             returns (dict): new client rules
@@ -337,6 +338,7 @@ class InsightsClient(object):
         """
         return client.get_registration_status(self.config, self.connection)
 
+    @_net
     def upload(self, path, rotate_eggs=True):
         """
             returns (int): upload status code
@@ -424,6 +426,7 @@ class InsightsClient(object):
         """
         return client.delete_archive(path)
 
+    @_net
     def get_registration_status(self):
         return client.get_registration_status(self.config, self.connection)
 
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -42,6 +42,15 @@ class InsightsClient(object):
         self.session = None
         self.connection = None
 
+    def _net(func):
+        def _init_connection(self, *args, **kwargs):
+            # setup a request session
+            if not self.config.offline and not self.session:
+                self.connection = client.get_connection(self.config)
+                self.session = self.connection.session
+            return func(self, *args, **kwargs)
+        return _init_connection
+
     def get_conf(self):
         # need this getter to maintain compatability with RPM wrapper
         return self.config
@@ -52,17 +61,19 @@ class InsightsClient(object):
     def version(self):
         return "%s-%s" % (package_info["VERSION"], package_info["RELEASE"])
 
+    @_net
     def test_connection(self):
         """
             returns (int): 0 if success 1 if failure
         """
-        return client.test_connection(self.config)
+        return self.connection.test_connection()
 
+    @_net
     def branch_info(self):
         """
             returns (dict): {'remote_leaf': -1, 'remote_branch': -1}
         """
-        return client.get_branch_info(self.config)
+        return client.get_branch_info(self.config, self.connection)
 
     def fetch(self, force=False):
         """
@@ -94,15 +105,11 @@ class InsightsClient(object):
 
             return fetch_results
 
+    @_net
     def _fetch(self, path, etag_file, target_path, force):
         """
             returns (str): path to new egg. None if no update.
         """
-        # setup a request session
-        if not self.session:
-            self.connection = client.get_connection(self.config)
-            self.session = self.connection.session
-
         url = self.connection.base_url + path
         # Searched for cached etag information
         current_etag = None
@@ -291,50 +298,51 @@ class InsightsClient(object):
             logger.debug("Bypassing rule update due to config "
                 "running in offline mode or auto updating turned off.")
         else:
-            return client.update_rules(self.config)
+            return client.update_rules(self.config, self.connection)
 
+    @_net
     def collect(self):
         # return collection results
-        tar_file = client.collect(self.config)
+        tar_file = client.collect(self.config, self.connection)
 
         # it is important to note that --to-stdout is utilized via the wrapper RPM
         # this file is received and then we invoke shutil.copyfileobj
         return tar_file
 
-    def register(self, force_register=False):
+    @_net
+    def register(self):
         """
-            returns (json): {'success': bool,
-                            'machine-id': uuid from API,
-                            'response': response from API,
-                            'code': http code}
+            returns (bool | None):
+                True - machine is registered
+                False - machine is unregistered
+                None - could not reach the API
         """
-        self.config.register = True
-        if force_register:
-            self.config.reregister = True
-        return client.handle_registration(self.config)
+        return client.handle_registration(self.config, self.connection)
 
+    @_net
     def unregister(self):
         """
             returns (bool): True success, False failure
         """
-        return client.handle_unregistration(self.config)
+        return client.handle_unregistration(self.config, self.connection)
 
+    @_net
     def get_registration_information(self):
         """
-            returns (json): {'machine-id': uuid from API,
-                            'response': response from API}
+            returns (json):
+                {'messages': [dotfile message, api message],
+                 'status': (bool) registered = true; unregistered = false
+                 'unreg_date': Date the machine was unregistered | None,
+                 'unreachable': API could not be reached}
         """
-        registration_status = client.get_registration_status(self.config)
-        return {'machine-id': client.get_machine_id(),
-                'registration_status': registration_status,
-                'is_registered': registration_status['status']}
+        return client.get_registration_status(self.config, self.connection)
 
     def upload(self, path, rotate_eggs=True):
         """
             returns (int): upload status code
         """
         # do the upload
-        upload_results = client.upload(self.config, path)
+        upload_results = client.upload(self.config, self.connection, path)
         if upload_results:
 
             # delete the archive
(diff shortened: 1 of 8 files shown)


kylape:
Looks like your tests are busted in CI.  They failed on my local box, too.

insights/tests/client/test_client.py
```diff
@@ -18,3 +26,155 @@ def test_version():
         assert result == "%s-%s" % (package_info["VERSION"], package_info["RELEASE"])
     finally:
         sys.argv = tmp
+
+
+def test_register():
+    config = InsightsConfig(register=True)
+    client = InsightsClient(config)
+    try_auto_configuration(config)
+    assert client.register() is True
```
kylape: Does this actually make network calls?

insights/tests/client/test_client.py
```diff
@@ -18,3 +26,155 @@ def test_version():
...
+    config.http_timeout = 1
+
+    # set net delay and try to check registration
+    subprocess.call(
+        shlex.split('tc qdisc add dev eth0 root netem delay 1001ms'))
+    with pytest.raises(requests.exceptions.Timeout):
+        assert client.get_registration_information()['unreachable'] is True
+        assert client.register() is None
+    subprocess.call(
+        shlex.split('tc qdisc del dev eth0 root netem delay 1001ms'))
```
kylape: I don't think this fits in a unit test.  Pretty sure this requires root to run correctly.

insights/client/new_refg.py
```diff
@@ -0,0 +1,90 @@
+def handle_registration(config):
```
kylape: Is `new_refg.py` a typo?

insights/client/client.py
```diff
@@ -88,79 +87,88 @@ def set_up_logging(config):
...
+
+def test_connection(pconn):
     """
     Test the connection
     """
-    pconn = get_connection(config)
     return pconn.test_connection()
 
 
-def _is_client_registered(config):
```
kylape: Did you change the order of the functions?  This diff is so hard to read.

kylape:
I wonder if we could create a mock connection object to run our unit tests.

insights/tests/client/test_client.py
gravitypriest: It ought to...

insights/tests/client/test_client.py
gravitypriest: Hmm alright. Wonder if there's a better way to force a timeout

insights/client/new_refg.py
gravitypriest: d'oh I meant to delete this file. It was a placeholder while I was changing stuff

insights/client/client.py
gravitypriest: yeah my bad. lemme put them back in the old order so it's easier to parse

insights/client/client.py
```diff
@@ -171,63 +105,107 @@ def register(config):
...
+        Returns:
+            True - machine is registered
+            False - machine is unregistered
+            None - could not reach the API
+    '''
     # force-reregister -- remove machine-id files and registration files
     # before trying to register again
-    new = False
     if config.reregister:
-        logger.debug('Re-register set, forcing registration.')
```
gravitypriest: should restore the missing log messages in this function

insights/client/client.py
kylape: Ah! So much better! ❤️

insights/client/__init__.py
```diff
@@ -47,17 +56,19 @@ def set_up_logging(self):
     def version(self):
         return "%s-%s" % (package_info["VERSION"], package_info["RELEASE"])
 
+    @_net
     def test_connection(self):
         """
             returns (int): 0 if success 1 if failure
         """
-        return client.test_connection(self.config)
+        return client.test_connection(self.connection)
```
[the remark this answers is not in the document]
brantleyr: This will also break `post_update` function in `v1.py` I believe.

https://github.com/RedHatInsights/insights-core/blob/c1165678a91d58a2fe5d5358fcff3b42453ba1b7/insights/client/phase/v1.py#L148

kylape:
test this please

insights/client/__init__.py
```diff
@@ -291,50 +298,51 @@ def update_rules(self):
...
+                 'unreg_date': Date the machine was unregistered | None,
+                 'unreachable': API could not be reached}
         """
-        registration_status = client.get_registration_status(self.config)
-        return {'machine-id': client.get_machine_id(),
-                'registration_status': registration_status,
-                'is_registered': registration_status['status']}
+        return client.get_registration_status(self.config, self.connection)
 
     def upload(self, path, rotate_eggs=True):
```
Glutexo: Shouldn’t this use the new __net_ decorator too? Even though it is not necessary, because some other method with it is always called before this one, it still uses and needs _self.connection_.

insights/tests/client/test_client.py
```diff
@@ -1,10 +1,34 @@
...
 from insights.client.config import InsightsConfig
 from insights import package_info
-from mock.mock import patch
-from mock.mock import Mock
+from insights.client.constants import InsightsConstants as constants
+# from mock.mock import patch
+# from mock.mock import Mock
+
+
+class FakeConnection(object):
```
Glutexo: I’d prefer to use _Mock_ instead. It’s not 100% necessary, but in the tests already present (introduced by my PR #1251) it‘s already being mocked/patched, so we can stick to that.

insights/tests/client/test_client.py
```diff
@@ -24,63 +48,212 @@ def test_version():
...
-@patch('insights.client.client.constants.sleep_time', 0)
-@patch('insights.client.client.InsightsConnection.upload_archive',
-       return_value=Mock(status_code=500))
-def test_upload_500_retry(upload_archive):
+def test_register():
+    config = InsightsConfig(register=True)
+    client = InsightsClient(config)
+    client.connection = FakeConnection()
+    client.session = True
+    assert client.register() is True
```
Glutexo: Without any mocking here, we are touching the real filesystem. The tests are actually reading and **writing** files in the real _/etc_. That’s really bad. Not only in would need either root access or some special permissions setup, but more importantly it breaks out of the sandbox, doing mess in a real system.

insights/tests/client/test_client.py
```diff
@@ -24,63 +48,212 @@ def test_version():
...
+    client.connection = FakeConnection(registered=False)
+    assert client.get_registration_information()['unreachable'] is True
+    assert client.register() is None
+    for r in constants.registered_files:
+        assert os.path.isfile(r) is False
+    for u in constants.unregistered_files:
+        assert os.path.isfile(u) is True
+
+
+# @patch('insights.client.client.constants.sleep_time', 0)
```
Glutexo: If you cherry-pick my changes to these tests, it won’t be necessary to comment them out. They resolve merge conflicts.

https://github.com/Glutexo/insights-core/commit/f10ffed5b4e399dc22e99d0e95f0f3adcae015a7

insights/client/__init__.py
gravitypriest: agreed

esakaiev:
test this please

Glutexo:
Hey, @gravitypriest. I created a “demo” [commit](https://github.com/Glutexo/insights-core/commit/7f552998e73661ed2e7b068665e810489acc2c7e) as a possible inspiration to improve the tests. My goal was to mock away touching the real filesystem and also to use a _Mock_ instant of a fake class for _InsightsConnection_ replacement.

See https://github.com/Glutexo/insights-core/commit/7f552998e73661ed2e7b068665e810489acc2c7e. I added some thought dump comments preceded by `***`, not to be put in the real code.

The `utilities` module is currently untested and thus can’t be trusted. Thus I went the way of patching and asserting the actual file reads and writes. We trust those _stdlib_ calls. Testing the tiny bits separately and then relying on their correct functionality would be better, but not very feasible for now.

Many of those changed lines could be refactored if used by more tests. I changed just the one for inspiration. It’s not very polished; I just spent a few dreamless hours (or… days) trying to grasp how to use _unittest.mock_, so I now have an unsuppressable urge to sprinkle this around.

gravitypriest:
Just mocked the file paths in the constants to /tmp for now. Would be open to improving the tests in the future, but this ought to suffice to get it merged in.

insights/tests/client/test_client.py
```diff
@@ -49,6 +49,7 @@ def test_version():
         sys.argv = tmp
 
 
+@pytest.mark.skip(reason="Mocked paths not working in QE jenkins")
```
Glutexo: Maybe consider using [NamedTemporaryFile](https://docs.python.org/ja/2.7/library/tempfile.html#tempfile.NamedTemporaryFile). I‘m not sure whether it’d work, since it actually writes files to _/tmp_ just as you do, but it might be worth trying.

Glutexo:
There is one thing I wanted to add:

The original approach of testing writing directly to _/etc_ is great for functional/integration testing. Unit tests are great, but you still want to test the application as a whole and see that the registration process really works from the beginning to the end. Such tests are however slow, heavy on requirements and often require sandboxes, fake services and such. You want to run them in CI or manually, but probably not among fast and atomic unit tests.

Unfortunately I have no idea how to ideally and pythonically split unit tests from the complex ones. Just wanted too say that it would be a pity to throw away your original tests, I’d just like to move them to another realm.

esakaiev:
test this please

eduardocerqueira:
test this please

eduardocerqueira:
test this please

eduardocerqueira:
test this please

eduardocerqueira:
test this please

insights/tests/client/test_client.py
kylape: I think this is resolved now

insights/client/client.py
kylape: do you need to follow up with this?

kylape:
@Glutexo I agree on your testing points. Our QE team has a suite of tests they run, and so I think any functional test that Jeremy wrote should to in their test suite.

insights/client/client.py
gravitypriest: Already did. Added in a commit either this morning or yesterday.

PR #1280, opened 2018-07-23
imply disable_schedule with no_schedule

gravitypriest:
`no_schedule` doesn't exist anymore and the new config is more strict about invalid options... let's just carry over and imply the newer option


gravitypriest:
Closed in favor of #1289

Issue #1281, opened 2018-07-24
Wrong comparison method is used in get_localport and get_peerport of SsTULPN parser  


vishwanathjadhav:
Wrong comparison method is used in get_localport and get_peerport of SsTULPN parser, when port='22' and value of l["Local-Address-Port"]='22450'  then '22 ' is present in l["Local-Address-Port"]. which leads to the wrong conclusion.

xiangce:
Resolved in #1281

PR #1282, opened 2018-07-24
Updated code to fix issue insights_core_#1281

vishwanathjadhav:
Issue Link: https://github.com/RedHatInsights/insights-core/issues/1281


insights/parsers/netstat.py
```diff
@@ -681,12 +681,22 @@ def get_service(self, service):
         return [l for l in self.data if l.get("Process", None) and service in l["Process"]]
 
     def get_localport(self, port):
-        return [l for l in self.data if l.get("Local-Address-Port") and
-                port in l["Local-Address-Port"]]
+        list_conn = []
+        for line in self.data:
```
xiangce: Since line.get('Local-Address-Port') being fetched times, how about simplifying as:
```python
            local_port = line.get('Local-Address-Port')
            if local_port and ':*' not in local_port and\
                    int(port) == int(local_port.split(':')[-1]):
                list_conn.append(line)
```

insights/parsers/netstat.py
```diff
@@ -681,12 +681,22 @@ def get_service(self, service):
...
+                    int(port) == int(line.get("Local-Address-Port").split(':')[-1]):
+                list_conn.append(line)
+        return list_conn
 
     def get_peerport(self, port):
-        return [l for l in self.data if l.get("Peer-Address-Port") and
-                port in l["Peer-Address-Port"]]
+        list_conn = []
+        for line in self.data:
+            if line.get('Peer-Address-Port') and\
```
xiangce: The same

```python
            peer_port = line.get('Peer-Address-Port')
            if peer_port and ':*' not in peer_port and\
                    int(port) == int(peer_port.split(':')[-1]):
                list_conn.append(line)
```

insights/parsers/netstat.py
vishwanathjadhav: Done

insights/parsers/netstat.py
vishwanathjadhav: Done

PR #1283, opened 2018-07-24
Modify the spec "lspci" to simple command "lspci -k"

xiangce:
For details please refer to the Spec's Jira Request


kylape:
Hey @xiangce I think it would make more sense to just keep one `Lspci` class and add the driver parsing in the current class.  Does that make sense to you?

xiangce:
Hi @kylape - That sounds good to me. However, we need to modify a new rule (not activated yet) which used the new `LsPciDriver` back to use `LsPci`.

xiangce:
@kylape - I made the update as per your comment, would you please double confirm it?

bfahr:
This can be merged once https://github.com/RedHatInsights/insights-frontend-assets/pull/110 has been merged.

xiangce:
@bfahr - The patch of this PR will make the new `LsPci` compatible with both `lspci` and `lspci -k` commands, but the old `LsPci` Parser can't work with the new `lspci -k` command. So, I think it should be merged before the spec.

bfahr:
@xiangce we can't merge this PR yet because it would break production.  I've added a [spec change to the json[(https://github.com/RedHatInsights/insights-frontend-assets/pull/141) to implement the `lspci_kernel` spec so that the existing parser will work.  Once that is in production we can coordinate this change.

eduardocerqueira:
The CI failed above is not related to change on this PR. It was a change in the insights-client that changed the status return and we are updating our tests now.

bfahr:
@xiangce I think we can go ahead with this change as long as the corresponding change is made in https://github.com/RedHatInsights/insights-frontend-assets/pull/110.

YakovLittle:
test this please

bfahr:
@xiangce I ran tests with this new parser and it appears to break backward compatibility with the existing `LsPci` class.  In order to merge the two classes we need to maintain backward compatibility.

vishwanathjadhav:
@bfahr I have tested its backward compatibility and it works well now. Please check and let me know

insights/specs/default.py
```diff
@@ -330,8 +330,7 @@ def semid(broker):
                                  simple_command("/usr/bin/lsinitrd -f /etc/lvm/lvm.conf")
                                  ])
     lsmod = simple_command("/sbin/lsmod")
-    lspci = simple_command("/sbin/lspci")
-    lspci_kernel = simple_command("/sbin/lspci -k")
+    lspci = first_of([simple_command("/sbin/lspci -k"), simple_command("/sbin/lspci")]),
```
bfahr: @vishwanathjadhav this should be a `simple_command` instead of `first_of` to be consistent with the changes you are making in https://github.com/RedHatInsights/insights-frontend-assets/pull/110.

insights/specs/insights_archive.py
```diff
@@ -108,8 +108,7 @@ class InsightsArchiveSpecs(Specs):
     lscpu = simple_file("insights_commands/lscpu")
     lsmod = simple_file("insights_commands/lsmod")
     lsof = simple_file("insights_commands/lsof")
-    lspci = simple_file("insights_commands/lspci")
-    lspci_kernel = simple_file("insights_commands/lspci_-k")
+    lspci = first_file(["insights_commands/lspci_-k", "insights_commands/lspci"])
```
bfahr: Consistent with the commend above this should be a `simple_file`.

insights/specs/default.py
vishwanathjadhav: Done

insights/specs/insights_archive.py
vishwanathjadhav: Done

esakaiev:
test this please

bfahr:
This PR is approved and it will be merged during the production release this week along with https://github.com/RedHatInsights/insights-frontend-assets/pull/110.

PR #1286, opened 2018-07-24
Added parser for specs "ss -tupna"

vishwanathjadhav:
Added parser for specs "ss -tupna" it will list all (listening + non-listing)connections


xiangce:
I have no idea why the CI being blocked and seems there is no entry for me to take a look.

PR #1287, opened 2018-07-24
Implement console output formats as loadable modules.

csams:
Implement output formats as loadable modules. Cleans up the run function and allows us to extend with more formats if needed. The next formatter probably would be one that prints rule results or errors as they execute.

Examples:
*  `python -m insights -f json`
*  `python -m insights -f yaml`
*  `python -m insights -f text` (the default)


eduardocerqueira:
test this please

PR #1294, opened 2018-07-26
BUGREPORT: configtree breaks on quotes

jsvob:
The diff says it all - I figured it is more efficient to write exact failing tests than to explain otherwise.

The majority of hits for one rule are machines with such failing config files (thousands of machines), so a particular rule has wildly different results on the old httpd_conf combiner vs. the new configtree one.

@csams Can you fix it, please?

### DIFF
--- a/insights/combiners/tests/test_httpd_conf_tree.py
+++ b/insights/combiners/tests/test_httpd_conf_tree.py
@@ -91,6 +91,21 @@ Bar 3C
 </IfModule>
 '''.strip()
 
+HTTPD_CONF_WITH_NONSTANDARD_INCLUDE_PATH = '''
+JustFotTest_NoSec "/var/www/cgi"
+Include blaaaa.blaaa.blabla.d/*.conf
+Include conf.modules.d/*.conf
+# prefork MPM
+<IfModule prefork.c>
+ServerLimit      256
+ThreadsPerChild  16
+JustForTest      "AB"
+MaxClients       256
+</IfModule>
+
+IncludeOptional conf.d/*.conf
+'''.strip()
+
 HTTPD_CONF_MAIN_1 = '''
 ServerRoot "/etc/httpd"
 Listen 80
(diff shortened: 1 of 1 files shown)


jsvob:
Oh, I just realized I should probably do these things against master. I always use what's in production, though, because that's in production. I've been working on this all day until I just isolated the problem, so I am sorry and I'll leave it against 3.0 if that's not a big problem.

insights/combiners/tests/test_httpd_conf_tree.py
```diff
@@ -91,6 +91,21 @@
 </IfModule>
 '''.strip()
 
+HTTPD_CONF_WITH_NONSTANDARD_INCLUDE_PATH = '''
```
jsvob: Oh, this is from another branch in my testing, but this is not a failing test, it actually does what I expected.

insights/combiners/tests/test_httpd_conf_tree.py
```diff
@@ -256,11 +271,396 @@
 """.strip()
 
 
+# TODO
+# These are real-world config files - you can delete them after reviewing the MR, I provide them in case the minimum crashing example seems inscrutable to you.
+# There are at least 983 customer machines with such a config file.
+
+BZ1591241C7_CRASHING_FULL_ETC_HTTPD_CONF_HTTPD_CONF = "/etc/httpd/conf/httpd.conf"
```
jsvob: Note: I did read about 3 times, slowly, whether there are any potential PII, before submitting the MR. I found none.
Note 2: I found which sets of highly filtered config files make it crash based on some of their common features and then I wrote a second custom combiner and rule that pulled just that one crashing config file sans comments into analysis, I never pulled all of them.

bfahr:
@jsvob you are correct about needing to submit this against master.

bfahr:
@jsvob would you mind to create an issue for this bug?

jsvob:
Agreed, I will close this and submit a new better PR and file an issue for it.

jsvob:
@bfahr Resubmitted against master and in a cleaner form. Thank you.

https://github.com/RedHatInsights/insights-core/issues/1300
https://github.com/RedHatInsights/insights-core/pull/1301

PR #1295, opened 2018-07-26
Generalize make_response to be class based and extendable.

csams:
We've seen a need for rules to return more than `make_response()`. `make_metadata` and `make_fingerprint` already exist, and we're being asked to add `make_pass` and `make_fail`.

To better manage these kinds of extensions, all `make_*` components are now simple specializations of a base class.


bfahr:
test this

PR #1298, opened 2018-07-27
Fix the simple_file in insights-archive for saphostctl

xiangce:
Correct the simple_file for `saphostctl` command as per the result of `mangle`


lhuett:
Please resolve conflicts.

Issue #1300, opened 2018-07-27
Quotes and doublequotes break configtree parser


jsvob:
So far tested only with HttpdConfTree (not with other configtree parsers&combiners).

Quotes and doublequotes make the parser crash.

```
CustomLog logs/ssl_request_log \
"%t %h %{SSL_PROTOCOL}x %{SSL_CIPHER}x \"%r\" %b"
```

More context is in https://github.com/RedHatInsights/insights-core/pull/1294/files

In a few minutes I will submit a new PR with a test that shows the bug.

EDIT: Corrected the code to include backslashes I previously mistakenly stripped out.

csams:
Fixed in https://github.com/RedHatInsights/insights-core/pull/1302. Reopen if you still see the issue.

PR #1301, opened 2018-07-27
A new test for configtree httpd_conf combiner. It should not fail, but it currently does.

jsvob:
issue - https://github.com/RedHatInsights/insights-core/issues/1300

more details in a previous wrongly submitted PR (should not be important in the long term) - https://github.com/RedHatInsights/insights-core/pull/1294/files


csams:
This is fixed in https://github.com/RedHatInsights/insights-core/pull/1302

PR #1310, opened 2018-07-31
change type check to isinstance

gravitypriest:
fix traceback because it was being treated as a string so it was looking for a file called "false"


gravitypriest:
Merged without passing checks because the test nodes are acting up right now.

PR #1316, opened 2018-08-01
parser for /etc/sysconfig/foreman-tasks

TripleJQK:


bfahr:
@TripleJQK these changes look OK assuming the spec change is approved.

Issue #1315, opened 2018-08-01
About the parser 'pcs_status', the spec of it does include the full correct file path


crackcodecamp:
For most recent sosreport, the file path would be 'sos_commands/pacemaker/pcs_status' (for example: sosreport-overcloud-controller-0-20180703073332/sos_commands/pacemaker/pcs_status), so this need to be fixed first with adding the path to the sos_archive in insights-core repo.

insights/specs/sos_archive.py:
    pcs_status = simple_file("sos_commands/cluster/pcs_status")
                        ---> this path is for old rhel version

insights/specs/sos_archive.py:
    pcs_status = simple_file("sos_commands/pacemaker/pcs_status")
                        ---> this path is for New rhel version

bfahr:
PR #1317 has been merged with master and 3.0 branches.

PR #1317, opened 2018-08-01
About the parser 'pcs_status' updated with correct path

crackcodecamp:
For most recent sosreport, the file path would be 'sos_commands/pacemaker/pcs_status'(for example sosreport-overcloud-controller-0-20180703073332/sos_commands/pacemaker/pcs_status), so this need to be fixed first with adding the path to the sos_archive in the insights-core repo.

The raised issue for this enhancement https://github.com/RedHatInsights/insights-core/issues/1315

### DIFF
--- a/insights/specs/sos_archive.py
+++ b/insights/specs/sos_archive.py
@@ -62,7 +62,7 @@ class SosSpecs(Specs):
     netstat_s = simple_file("sos_commands/networking/netstat_-s")
     nmcli_dev_show = simple_file("sos_commands/networking/nmcli_dev_show")
     ntptime = simple_file("sos_commands/ntp/ntptime")
-    pcs_status = simple_file("sos_commands/cluster/pcs_status")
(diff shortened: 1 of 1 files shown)


xiangce:
Hi @crackcodecamp,

As you mentioned in #1315 , 
```
pcs_status = simple_file("sos_commands/cluster/pcs_status")
---> this path is for old rhel version
```
It means in old rhel version, the path `sos_commands/cluster/pcs_status` is also valid, right.

In this case, we'd better keep the spec `pcs_status` compatible with both the new and old rhel version by using `first_file`, e.g.:
```
pcs_status = first_file(["sos_commands/pacemaker/pcs_status", "sos_commands/cluster/pcs_status"]),
```

crackcodecamp:
Hi, @xiangce Yah its need to keep both of the paths. I will update the PR according to that.

crackcodecamp:
@xiangceI have updated PR with adding new path into pcs_status.

crackcodecamp:
@bfahr Thanks for merging.

PR #1320, opened 2018-08-03
Dockerfile for adding insights-core layer to content-server on Openshift

mhuth:


mhuth:
@kylape need to add this webhook as well:
https://api.insights.openshift.com/oapi/v1/namespaces/advisor-ci/buildconfigs/content-server-layer4-
core/webhooks/h74iXXORe6nh3KaNpZ03/github

PR #1324, opened 2018-08-06
test_validate_remove_file: Clean up /tmp/remove.cfg post testing

### LINKED ISSUE #1248: test_validate_remove_file doesn't clean up /tmp/remove.cfg
`test_validate_remove_file` in `insights/tests/client/test_utilities.py` doesn't clean up `/tmp/remove.cfg` after unit testing.

sourabhtk37:
Closes #1248 

Signed-off-by: T K Sourab <sourabhtk37@gmail.com>

### DIFF
--- a/insights/tests/client/test_utilities.py
+++ b/insights/tests/client/test_utilities.py
@@ -54,7 +54,8 @@ def test_generate_machine_id():
     assert machine_id_regex.group(0) is not None
     with open('/tmp/testmachineid', 'r') as _file:
         machine_id = _file.read()
-    assert util.generate_machine_id(destination_file='/tmp/testmachineid') == machine_id
+    assert util.generate_machine_id(
+        destination_file='/tmp/testmachineid') == machine_id
     os.remove('/tmp/testmachineid')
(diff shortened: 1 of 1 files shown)


insights/tests/client/test_utilities.py
```diff
@@ -89,6 +91,10 @@ def test_validate_remove_file():
     os.chmod(tf, 0o600)
     assert util.validate_remove_file(remove_file=tf) is not False
 
+    if os.path.isfile(tf):
```
csams: Use something like the [pytest setup and teardown for functions](https://docs.pytest.org/en/latest/xunit_setup.html#method-and-function-level-setup-teardown) so the file gets cleaned up even if the assert fires.

sourabhtk37: Make sense. Made changes for that. Thanks.

kylape:
ok to test

psachin:
@sourabhtk37 With the recent changes in [utilities.validate_remove_file](https://github.com/RedHatInsights/insights-core/blob/master/insights/client/utilities.py#L182) please confirm if the tests are still valid.

PR #1325, opened 2018-08-06
Added description for running tests and generating docs in quickstart…

sourabhtk37:
… contributor setup.

Signed-off-by: T K Sourab <sourabhtk37@gmail.com>

### DIFF
--- a/docs/quickstart_insights_core.rst
+++ b/docs/quickstart_insights_core.rst
@@ -167,7 +167,7 @@ using the following commands::
 
 
 After setup
-==========
+===========
 
 You can validate the setup by running the unit tests::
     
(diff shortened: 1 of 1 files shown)


docs/quickstart_insights_core.rst
```diff
@@ -165,6 +165,21 @@ using the following commands::
     [userone@hostone project_dir/insights-core]$ git pull upstream master
     [userone@hostone project_dir/insights-core]$ git push origin master
 
+
+After setup
+==========
```
bfahr: @sourabhtk37 you need to add one more `=` here, the underline has to be the same length as the heading.

sourabhtk37: Fixed. Thanks.

PR #1327, opened 2018-08-06
check directory before writing a file (BZ1610497)

gravitypriest:
I previously made a fix to maintain the .registered and .unregistered files in both needed locations. Well, turns out it doesn't work so well if you delete the old directory: https://bugzilla.redhat.com/show_bug.cgi?id=1610497

This change makes it so that existence of the containing directory is checked before writing any files.


gravitypriest:
Fair enough. I just like to err on the side of more logging just in case we ever have to debug anything.

PR #1328, opened 2018-08-06
remove dupe method, fix func call, fix timeout parse

gravitypriest:
The registration refactoring brought about some bugs we didn't catch til today, namely:

- still using `InsightsClient.get_connection()`
- `http_timeout` not being parsed to a float correctly

This commit fixes those, as well as removing the duplicate function `get_registration_information` in favor of `get_registration_status`, to keep name parity with the matching function in `client.py`.

### DIFF
--- a/insights/client/config.py
+++ b/insights/client/config.py
@@ -487,10 +487,20 @@ class InsightsConfig(object):
             pass
         d = dict(parsedconfig.items(constants.app_name))
         for key in d:
-            if key == 'retries':
-                d[key] = parsedconfig.getint(constants.app_name, key)
-            if key in DEFAULT_BOOLS and isinstance(d[key], six.string_types):
-                d[key] = parsedconfig.getboolean(constants.app_name, key)
+            try:
+                if key == 'retries':
+                    d[key] = parsedconfig.getint(constants.app_name, key)
+                if key == 'http_timeout':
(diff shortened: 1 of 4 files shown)


gravitypriest:
That function's still used. `registration_check` is where all the response parsing is done.

insights/client/config.py
```diff
@@ -487,10 +487,20 @@ def load_config_file(self, fname=None):
...
+                    d[key] = parsedconfig.getfloat(constants.app_name, key)
+                if key in DEFAULT_BOOLS and isinstance(
+                        d[key], six.string_types):
+                    d[key] = parsedconfig.getboolean(constants.app_name, key)
+            except ValueError as e:
+                if self._print_errors:
+                    sys.stdout.write(
+                        'ERROR: {0}.\nCould not read configuration file, '
+                        'using defaults\n'.format(e))
+                return
```
kylape: So returning here means it just stops processing the remaining keys?  Seems like it should either halt the whole process or keep going.

gravitypriest: The config object isn't updated with the values from the .conf file until the very end of the function. So if any exception forces a return it won't be changed from the defaults. We've never supported a case where we only parse _some_ of the config vars from the file.

kylape: Makes sense, thanks.

PR #1332, opened 2018-08-07
Explicit filterable datasources

csams:
Allow developers to mark a `RegistryPoint` or datasource as `filterable=True`.

1. Datasources that provide implementations for `RegistryPoint`s inherit their `filterable`, `raw`, and `multi_output` values automatically, and they can't be overridden.
2. Attempting to add a filter to any datasource not marked as filterable will raise an exception.
3. Any rule that depends on a filterable datasource but does not add a filter to it from the rule module will cause an exception during the rule's integration tests.

\#3 will cause many rule integration test failures at the moment. Both \#2 and \#3 can be disabled if the environment variable `INSIGHTS_FILTERS_ENABLED=False`.


insights/specs/__init__.py
```diff
@@ -207,71 +207,71 @@ class Specs(SpecSet):
...
+    neutron_metadata_agent_ini = RegistryPoint(filterable=True)
+    neutron_metadata_agent_log = RegistryPoint(filterable=True)
+    neutron_ovs_agent_log = RegistryPoint(filterable=True)
     neutron_plugin_ini = RegistryPoint()
-    neutron_server_log = RegistryPoint()
+    neutron_server_log = RegistryPoint(filterable=True)
     nfnetlink_queue = RegistryPoint()
-    nfs_exports_d = RegistryPoint(multi_output=True)
-    nfs_exports = RegistryPoint()
+    nfs_exports_d = RegistryPoint(multi_output=True, filterable=True)
```
lonicerae: Any reason we add filters to the `nfs_exports_d` and `nfs_exports` specs?

xiangce: @lonicerae there are filters added for the 2 specs in some security plugins, then we have to add filters in insights-rules to make sure the expected lines will be collected.

xiangce: @csams @lonicerae  - After checking the related rules in security rules, I believe the filters should be removed. And the two specs (`nfs_exports` and `nfs_exports_d`) should be filtered.

https://gitlab.cee.redhat.com/insights-open-source/insights-prodsec-rules/blob/master/prodsec/rules/hardening_nfs.py
https://gitlab.cee.redhat.com/insights-open-source/insights-prodsec-rules/blob/master/prodsec/rules/hardening_nfs_world_access.py

xiangce:
@csams - Is there any chance to make the filter check more intelligent?
For example, for specs `virt_who_conf` and `odbc_ini`, some common filters have been added in the parsers already. 
https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/virt_who_conf.py
https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/odbc.py
But due to the filter-check in this patch, we have to add them again in the plugins repo, such as,

https://github.com/RedHatInsights/insights-plugins/pull/982/files#diff-f4dc8a0c01f2bed8148e1b78ba8187d7R42
https://github.com/RedHatInsights/insights-plugins/pull/981/files#diff-57146b9c64356f5f0f8630579c0311f1R27

Of course, we could also remove them from the parser and add them in the rules' upper `__init__.py`

insights/specs/__init__.py
xiangce: Removed

PR #1336, opened 2018-08-08
New parser for files under /var/log/ovirt-engine/

psachin:
Covers following log files:

- insights.parsers.ovirt_engine_log.ServerLog -- /var/log/ovirt-engine/server.log
- insights.parsers.ovirt_engine_log.UILog -- /var/log/ovirt-engine/ui.log
- insights.parsers.ovirt_engine_log.EngineLog -- /var/log/ovirt-engine/engine.log (replaces `insights.parsers.engine_log`)
- insights.parsers.ovirt_engine_log.ConsoleLog -- /var/log/ovirt-engine/console.log
- insights.parsers.ovirt_engine_log.BootLog -- /var/log/ovirt-engine/boot.log

Checklist:

- [X] Parser
- [X] Specs
- [X] Docs
- [X] Tests


YakovLittle:
test this pleae

esakaiev:
test this please

esakaiev:
test this please

insights/parsers/engine_log.py
```diff
@@ -1,37 +0,0 @@
-"""
```
xiangce: It's not allowed to remove any existing interface from insights-core v3.0, except for fixing bugs.

Your implementation to merge all corresponding engine logs to one parser file sounds good to me too, but you should keep it backward compatible.  E.g. leave an empty entry shell for the EngineLog in this `insights/parsers/engine_log.py` file, such as:

```python
from insights.parsers.ovirt_engine_log import EngineLog
```

insights/parsers/tests/test_engine_log.py
```diff
@@ -1,21 +0,0 @@
-from insights.parsers.engine_log import EngineLog
```
xiangce: Test file could be removed.

insights/specs/__init__.py
```diff
@@ -86,7 +86,6 @@ class Specs(SpecSet):
     dumpdev = RegistryPoint()
     dumpe2fs_h = RegistryPoint(multi_output=True)
     engine_config_all = RegistryPoint()
-    engine_log = RegistryPoint()
```
xiangce: This renaming is also not suggested since it would break some existing tests in some unknown project that depends on this spec.
But in fact, it doesn't impact the rules execution in the production environment, it's acceptable.

psachin: Thanks.

insights/parsers/tests/test_engine_log.py
psachin: Thanks

insights/parsers/engine_log.py
psachin: I like your suggestions to keep the file `engine_log.py` to ensure backward compatibility. I'm re-introducing this file and import statement. I made sure that no rule is using this parser before the changes were incorporated.

psachin: Done

esakaiev:
test this please

insights/parsers/ovirt_engine_log.py
```diff
@@ -0,0 +1,70 @@
...
+    >>> matched_line = '2018-01-17 01:46:17,739+05 WARN  [org.jboss.as.dependency.unsupported] (MSC service thread 1-7) WFLYSRV0019: Deployment "deployment.engine.ear" is using an unsupported module ("org.dom4j") which may be changed or removed in future versions without notice.'
+    >>> server_log.get('WARN')[-1].get('raw_message') == matched_line
+    True
+"""
+
+from insights import LogFileOutput, parser
+from insights.specs import Specs
+
+
+class OvirtEngineLog(LogFileOutput):
```
jhjaggars: I'd prefer to not make this class and just inherit from LogFileOutput in the other parser classes.

bfahr:
@psachin I discussed with the core team this morning and our recommendation is to go ahead and add the new parsers with the recommended changes, leave the existing engine log parser as is, and leave the existing engine_log spec as is.  

We will need to evaluate the best process for deprecating a spec and once that has been established then you'll be able to use that process to deprecate the existing engine log parser and spec.

If you want to go ahead and use the new engine log parser so that your rules depend on it instead of on the old parser you can do that.  It should use the existing engine_log spec.

psachin:
Hi @bfahr @jhjaggars @xiangce 

Thanks for reviewing this parser and discussing it with a core team. As suggested, I implemented the changes and provided info(in the form of comments wherever needed). I used existing specs to parse `engine.log` but commented new specs instead of removing them.

Please review the changes and let me know if they are good to go ahead.

Thanks!!

insights/specs/__init__.py
```diff
@@ -307,6 +307,10 @@ class Specs(SpecSet):
     os_release = RegistryPoint()
     ovirt_engine_confd = RegistryPoint(multi_output=True)
     ovirt_engine_server_log = RegistryPoint()
+    # ovirt_engine_log = RegistryPoint()
+    ovirt_engine_ui_log = RegistryPoint()
```
bfahr: @psachin I don't think we want to have commented out specs here because it would not be clear why they are commented out.  I'm also working with @csams to clarify whether or not we can have two specs collect the same file, and if so how we should mark the old spec as deprecated.

If you want this PR to go ahead and be merged now, we can deal with the spec deprecation in a separate PR, you just need to remove the commented specs.  Otherwise this PR will need to wait until we can determine how to deprecate specs.

insights/specs/default.py
```diff
@@ -452,6 +452,10 @@ def semid(broker):
     ose_node_config = simple_file("/etc/origin/node/node-config.yaml")
     ovirt_engine_confd = glob_file("/etc/ovirt-engine/engine.conf.d/*")
     ovirt_engine_server_log = simple_file("/var/log/ovirt-engine/server.log")
+    # ovirt_engine_log = simple_file("/var/log/ovirt-engine/engine.log")
```
bfahr: Same comment as above.

insights/specs/__init__.py
psachin: Sure

insights/specs/default.py
psachin: Done

psachin:
Hi @bfahr The changes are done. Can you please review again.

PR #1346, opened 2018-08-10
Fix running insights-client --support

Glutexo:
InsightsSupport and InsightsConnection classes’ constructors require passing an InsightsConfig instance. This wasn’t happening; fixed. Running `insights-client --support` doesn’t fail with an error any more and the support data is collected and printed.

~~Marked as _WIP_, since this is a hotfix without test. I‘m going to include some tests for that.~~

Related to #904, fixes a bug reported by @YakovLittle in https://github.com/RedHatInsights/insights-core/issues/904#issuecomment-411769183.


Glutexo:
Removed _WIP_, added some unit tests:

Not testing _InsightsSupport_ internals, that would be out of the scope of this fix. Assuming we trust it for now.

The tests are checking whether _InsightsSupport_ and _InsightsConnection_ are instantiated with _InsightsConfig_ when running the support collection. They also check whether the collection actually happens and that it replaces regular data collection.

Glutexo:
Oh, I got so excited that I forgot to run `flake8`. Fixed.

I aggree that having more tests for the `--support` would be good. I decided to put only those two, because covering the whole support data collection is a big bite and over the scope of a simple fix. That would require writing not only a solid foundation of unit tests, but some higher-level tests checking the whole story too.

I’m still not very satisfied with how the unit tests I write actually look like. But I believe that a significant part comes from the fact that they’re covering an otherwise untested (and thus untrusted) piece of code. As the code will be getting better, so will hopefully do the tests.

kylape:
I'm interested in what you are specifically unsatisfied with in your unit tests.  

Also, I agree that more comprehensive unit tests is beyond the scope of this change.

Glutexo:
Mostly I don't like the excessive mocking/patching that is often necessary to remove all side-effects of running the code. It makes the tests less readable and often requires introducing helper methods. Those are signs that the routines mix more responsibilities and do too much stuff at once, so it’s not necessarily the tests’ fault.

I am also a little afraid whether the unit tests aren’t too brittle. Like [here](https://github.com/RedHatInsights/insights-core/pull/1346/files#diff-001761a1a0e097d4f38dbeb7a55f24e7R17) verifying that a “private” method gets called – it makes sense, because the method does important stuff and established the connection which is what we are testing – on the other hand the test will break with a first refactoring of the _InsightsSupport_ class.

But still I thing we are on the right track, moving from _nothing_ to _at least something_.

kylape:
Yeah the tests are brittle, but right now it just makes sure the code can just _execute_.  So I'm basically agreeing with all your points :)

PR #1348, opened 2018-08-10
Remove check for commit on build_client_egg

dLobatog:
If we are making changes on the insights-client repo itself,
unfortunately we cannot build the package because this script is
checking for changes in the insights-core repo - and we may have done
none.

It's also quite inconvenient when you're testing changes to
insights-client locally.


YakovLittle:
test this please

YakovLittle:
test this please

kylape:
The code you're wanting to remove was originally added to avoid releasing uncommitted code from development boxes.  But the real answer is to stop releasing eggs from development boxes, or at least clone a clean repo when doing a release.

PR #1355, opened 2018-08-12
Make cpu_vulns multi-output

kgrant-rh:
This spec should have been multi-output as well.


esakaiev:
test this please

kgrant-rh:
@esakaiev I don't believe this change has anything to do with the client failures in jenkins. I've tested changing the spec to multi_output with plugins in prodsec.rules.

PR #1360, opened 2018-08-13
FileListing Parser Performance Improvements

csams:
Removed regular expression matching in favor of manual parsing.

With a test file of 4603 lines comprising 383 directories parsed 100 times per test with timings averaged:

regex based: 0.0188 seconds
manual parse: 0.0128 seconds

Manual parsing takes ~2/3 the time as regex.


insights/core/ls_parser.py
```diff
@@ -0,0 +1,135 @@
+import logging
```
bfahr: @csams it would be really nice if you could add some docstrings to this module since its the core of the ls command parsing code.

bfahr: Other than the docs, the changes look good to me.

csams: @bfahr I need to add handling for human readable sizes.

PR #1361, opened 2018-08-13
Refactor formatting and rendering. Add color coding and response formatting

lhuett:


insights/formats/text.py
```diff
@@ -2,7 +2,11 @@
 from pprint import pprint
 from insights import dr, datasource, rule
 from insights.core.context import ExecutionContext
-from insights.formats import Formatter
+from insights.formats import Formatter, render
+from colorama import Fore, Style
+
+
+_counts = {'skip': 0, 'pass': 0, 'rule': 0, 'metadata': 0, 'metadata_key': 0}
```
csams: I'd put this in the class `__init__` and just call it `counts`.

insights/formats/text.py
```diff
@@ -2,7 +2,11 @@
 from pprint import pprint
 from insights import dr, datasource, rule
 from insights.core.context import ExecutionContext
-from insights.formats import Formatter
+from insights.formats import Formatter, render
+from colorama import Fore, Style
```
csams: Maybe we should do something like this for the import so it works even if colorama isn't installed:
```python
from __future__ import print_function                                           
import six                                                                      
try:                                                                            
    from colorama import Fore, Style                                            
except ImportError:                                                             
    print("Install colorama for console colors.")                               
                                                                                
    class Default(type):                                                        
        def __getattr__(*args):                                                 
            return ""                                                           
                                                                                
    class Fore(six.with_metaclass(Default)):                                    
        pass                                                                    
                                                                                
    class Style(six.with_metaclass(Default)):                                   
        pass
```

insights/formats/text.py
lhuett: Yep, that makes sense. Then I guess other formatters could use it if they wanted to.

insights/formats/text.py
lhuett: Ok, so should the option of installing colorama to get colors be at the users discretion and the colorama dependency in setup.py be removed?

insights/formats/__init__.py
```diff
@@ -7,6 +7,9 @@
 _FORMATTERS = {}
 
 
+counts = {'skip': 0, 'pass': 0, 'rule': 0, 'metadata': 0, 'metadata_key': 0}
```
csams: This should go in the formatter class that's using it. Something like
```python
class HumanReadableFormat(Formatter):
...
    def __init__(self, args):
        ...
        self.counts = {'skip': 0, 'pass': 0, 'rule': 0, 'metadata': 0, 'metadata_key': 0}
```

PR #1363, opened 2018-08-13
Remove useless self.target_name in data collector

dLobatog:
Data collector's self.target_name was removed in
613f2cb2c2d6ffe6a10dc8fda903624bfd3587da, hence any reference to it will
fail.

To test, try to run:
```
sudo BYPASS_GPG=True insights-client --no-gpg --no-upload --run-these openshift --register --offline --verbose
```
before and after this change. 

Signed-off-by: Daniel Lobato Garcia <me@daniellobato.me>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ develop = set([
 ])
 
 docs = set([
-    'Sphinx',
+    'Sphinx==1.7.9',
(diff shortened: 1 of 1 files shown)


YakovLittle:
test this please

PR #1365, opened 2018-08-14
--from-file reads uploader.json properly

dLobatog:
Currently if we pass the 'from-file' option to insights-client, it's
unable to run.

For example, try running this:

```bash
sudo BYPASS_GPG=True insights-client --no-gpg  --verbose --offline
--from-file compliance.json
```

where compliance.json is:

```json
{
    "files": [
        {
            "file": "/var/lib/insights/latest-compliance-results.xml",
            "pattern": []
        },
        {
            "file": "/var/lib/insights/latest-compliance-results.html",
            "pattern": []
        }
    ],
    "version": "2017-08-31T17:10:37.456370",
    "sig": "asdasd",
    "uploader.json": "/etc/insights-client/.cache.json",
    "file": "/home/dlobatog/workspace/insights/insights-client/compliance.json"
}
```

uploader.json is read as a string instead of reading the file,
therefore it does not read the rules properly. It should be reading
the file instead as shown in this commit


insights/client/collection_rules.py
```diff
@@ -187,7 +187,8 @@ def get_conf(self, update, stdin_config=None):
             sig_fp.write(stdin_config["sig"])
             sig_fp.flush()
             if not self.gpg or self.validate_gpg_sig(rules_fp.name, sig_fp.name):
-                return json.loads(stdin_config["uploader.json"]), rm_conf
+                with open(str(stdin_config["uploader.json"])) as data_file:
+                    return json.loads(data_file.read()), rm_conf
```
jhjaggars: should probably just do 
```
json.load(data_file), rm_conf

**Comment by kylape** [comment]
It looks like elsewhere in the code it is expected that value of the `uploader.json` config key is actually the content instead of the filename:

```
        if not self.gpg or self.validate_gpg_sig(rules_fp.name, sig_fp.name):
>           with open(str(stdin_config["uploader.json"])) as data_file:
E           IOError: [Errno 2] No such file or directory: '{"key": "value"}'
```

While I admit `--from-file` is pretty misleading when you don't expect the value of that argument to not be a file, it seems like changing this would break some kind of backwards compatibility.  This may be what we should do, but it unfortunately won't be as trivial as it first seemed.

@dLobatog with that in mind, is this a change you still want to pursue?

kylape:
we are taking --from-file out of the client altogether.

Issue #1261, opened 2018-07-11
Split setuptools develop extras bundle


Glutexo:
#1251 introduces installing the _develop_ bundle by setuptools for test running. This might however result in deploying a package with unsatisfied dependencies. If a package present only in the _develop_ bundle is used in the code, the tests would still pass even if it weren’t declared as a runtime dependency.

The _develop_ extras bundle is currently rather big. I suggest splitting it to more atomic sets: one for testing, one for documentation etc. The _develop_ set can then include those along with other development packages. Not installing unnecessary packages for a test run is generally [a Good Thing](http://catb.org/jargon/html/G/Good-Thing.html).

Glutexo:
Resolved by now merged #1369.

PR #1342, opened 2018-08-09
Flake8

eduardocerqueira:
merged content from PR #1335 and appended virtualenv and pip freeze into Jenkinsfile for troubleshooting.

PR #1335, opened 2018-08-08
[wip] Add flake8 test to RHEL6

Glutexo:
Both RHEL7 stages run flake8 test, the RHEL6 one doesn’t. Flake8 documentation says that it is important to run it on the exact version of Python used. Added this test from Python 2.6 used by RHEL6.

Fixes #1262.

PR #1369, opened 2018-08-15
Split setuptools extras bundle

bfahr:
* Fixes issue #1261
* Adds bundles for docs, testing, and flake8
* Modifies Jenkinsfile to use new bundles

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,7 +11,7 @@ pipeline {
           }
           steps {
             echo "Installing Insights..."
-            sh 'pip install --user -e .[develop]'
+            sh 'pip install --user -e .[testing]'
             echo "Testing with Pytest..."
             sh 'pytest'
           }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,10 +10,12 @@ pipeline {
             }
           }
           steps {
-            echo "Installing Insights..."
-            sh 'pip install --user -e .[testing]'
-            echo "Testing with Pytest..."
-            sh 'pytest'
+            sh """
+                virtualenv .testenv
+                source .testenv/bin/activate
+                pip install -e .[testing]
+                pytest
+            """
           }
         }
         stage('Build RHEL7 Python 2.7') {
@@ -30,11 +32,11 @@ pipeline {
                 pip install -e .[testing]
                 pytest
             """
-            echo "Testing with flake8..."
+            echo "Testing with Linter..."
             sh """
-                virtualenv .f8env
-                source .f8env/bin/activate
-                pip install -e .[flake8]
+                virtualenv .lintenv
+                source .lintenv/bin/activate
+                pip install -e .[linting]
                 flake8
             """
           }
@@ -53,11 +55,11 @@ pipeline {
                 pip install -e .[testing]
                 pytest
             """
-            echo "Testing with flake8..."
+            echo "Testing with Linter..."
             sh """
-                /bin/python36 -m venv .f8env
-                source .f8env/bin/activate
-                pip install -e .[flake8]
+                /bin/python36 -m venv .lintenv
+                source .lintenv/bin/activate
+                pip install -e .[linting]
                 flake8
             """
           }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,12 +10,10 @@ pipeline {
             }
           }
           steps {
-            sh """
-                virtualenv .testenv
-                source .testenv/bin/activate
-                pip install -e .[testing]
-                pytest
-            """
+            echo "Installing Insights..."
+            sh 'pip install --user -e .[develop]'
+            echo "Testing with Pytest..."
+            sh 'pytest'
           }
         }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -11,7 +11,7 @@ pipeline {
           }
           steps {
             echo "Installing Insights..."
-            sh 'pip install --user -e .[develop]'
+            sh 'pip install --user -e .[testing]'
```
Glutexo: Why not use virtualenv here on RHEL6 if we are using it on RHEL7 and even with your new changes even for docs?

Jenkinsfile
```diff
@@ -38,12 +46,20 @@ pipeline {
...
+                /bin/python36 -m venv .testenv
+                source .testenv/bin/activate
+                pip install -e .[testing]
+                pytest
+            """
+            echo "Testing with flake8..."
+            sh """
+                /bin/python36 -m venv .f8env
+                source .f8env/bin/activate
+                pip install -e .[flake8]
```
Glutexo: Just wondering, what is the reasoning behind creating a separate environment for _flake8_ and for _pytest_. I understand and favor splitting the two package groups, but what are the benefits of creating two virtualenvs, each with only one of those?

setup.py
```diff
@@ -43,20 +43,29 @@ def maybe_require(pkg):
...
 ])
 
+testing = set([
+    'coverage==4.3.4',
+    'pytest==3.0.6',
+    'pytest-cov==2.4.0',
+    'mock==2.0.0',
+])
+
+flake8 = set([
```
Glutexo: I’d give this group some more generic name, something like _linting_. The other groups are not named by the specific packages too.

Moreover as [you informed](
https://github.com/RedHatInsights/insights-core/pull/1335#issuecomment-412646635), there might be issues with running Flake8 with Python2.6. @eduardocerqueira suggested switching to a different linter that’d be compatible. If this happens, the name would be obsolete, but a more generic name could stay the same.

Jenkinsfile
bfahr: @Glutexo I was actually waiting to see what happens with #1342.  I'm also not sure if `virtualenv` in installed on the docker image that is used for the Python 2.6 build.  I do agree that it should be consistent with the other sections.

Jenkinsfile
bfahr: @Glutexo The primary reason for multiple virtualenvs is that its the only way to ensure the isolated set of python modules is installed for that part of the run.  If we combine `pytest` and `flake8` then we could just use 1 section with a single virtualenv.

setup.py
bfahr: @Glutexo that makes sense to me too, I will switch to `linting`.

Jenkinsfile
Glutexo: Let’s just try. :)

Not sure about the presence of _virtualenv_ there in the image, but yesterday I tried to create a virtual environment on my RHEL6 box and it installed and ran (almost) just fine. I say almost, because there is a problem with dependencies (see #1373). It looks like the image has the correct versions already installed, but when you create a plain virtual environment, wrong versions are currently picked and the installation would fail.

I’ll make a quick test and if it fails, I‘d keep this as it is for now and unify it after #1373 gets fixed.

Glutexo: So just as expected. Creating a virtual environment causes the installation to fail, just as #1373 describes. Let’s just go without venv for now.

https://jenkins-insights-ci.1b13.insights.openshiftapps.com/blue/organizations/jenkins/insights-core/detail/PR-1335/9/pipeline#step-7-log-20

Jenkinsfile
Glutexo: You’re right. This way you don’t only test the actual _tests_ themselves, but the environments too. Okay. Thanks. 👍

Glutexo:
@bfahr Just wondering, will the testing on RHEL6 work now? You reverted the script so it doesn’t use virtualenv, which is good. There is however still `pip install --user -e .[develop]` used even though the `develop` package no longer contains testing tools.

Glutexo:
Hm… it [looks](https://jenkins-insights-ci.1b13.insights.openshiftapps.com/blue/organizations/jenkins/insights-core/detail/bf_split_setuptools/3/pipeline/8) like it works. But I think that is because _pytest_ is already included in the image and that the `pip install --user -e .[develop]` is rather useless, because now after the bundle split it doesn’t install any testing dependencies at all.

YakovLittle:
2018-08-20 18:18:28,525    DEBUG insights.client.connection Testing: https://cert-api.access.redhat.com/r/insights/uploads/
2018-08-20 18:18:28,608     INFO insights.client.connection HTTP Status Code: 503
2018-08-20 18:18:28,608     INFO insights.client.connection HTTP Status Text: Service Unavailable

YakovLittle:
test this please

bfahr:
@Glutexo good catch, I had tried virtualenv and confirmed it wouldn't work and reverted to the version prior to this PR.  That is fixed now. Thanks.

PR #1375, opened 2018-08-16
add option to set display name independent of register

gravitypriest:


insights/client/connection.py
```diff
@@ -713,3 +713,30 @@ def upload_archive(self, data_collected, duration):
             self.config.account_number = None
         logger.debug("Upload duration: %s", upload.elapsed)
         return upload
+
+    def set_display_name(self, display_name):
+        machine_id = generate_machine_id()
```
kylape: This is probably outside the scope of this PR, but calling `generate_machine_id` to get the already-existing machine-id is confusing.

gravitypriest: yeah agreed. it's been like that forever, it's a "if it exists, return it; if it doesn't make one" kind of function but maybe it needs a better nomenclature

PR #1384, opened 2018-08-20
use blacklist for pre-commands

gravitypriest:


YakovLittle:
test this please

insights/client/data_collector.py
```diff
@@ -109,6 +110,11 @@ def _parse_command_spec(self, spec, precmds):
             precmd_alias = spec['pre_command']
             try:
                 precmd = precmds[precmd_alias]
+
+                if set.intersection(set(shlex.split(precmd)),
+                                    constants.command_blacklist):
+                    raise RuntimeError("Command Blacklist")
```
bfahr: @gravitypriest would you want to add the intersection set as part of the exception message?

gravitypriest: Sure, that seems reasonable. I'll update.

PR #1386, opened 2018-08-21
Limit maximum pyOpenSSL version

Glutexo:
The [most recent version](
https://pypi.org/project/pyOpenSSL/18.0.0/) of [pyOpenSSL](https://pypi.org/project/pyOpenSSL/) fails to install on Python 2.6. Its cryptography dependency cannot be satisfied. Forced installation of [the last version](https://pypi.org/project/pyOpenSSL/17.5.0/) that still supports 2.6.

To use the downgraded [pyOpenSSL](
https://pypi.org/project/pyOpenSSL/) in Jenkins pipeline I updated [Jenkinsfile](https://github.com/RedHatInsights/insights-core/pull/1386/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3) to use a virtual environment just like other pipeline branches do. This is necessary to do at once with the downgrade, because the global environment in the image prevents proper installation of [pyOpenSSL](https://pypi.org/project/pyOpenSSL/).

~~This PR should help #1369.~~ (Merged.) Fixes #1373.

No _pytest_ tests fail after the downgrade neither on Python 2.6, 2.7 nor 3.

PR #1388, opened 2018-08-21
Do not install docs tools on Python 2.6

### LINKED ISSUE #1387: Do not install docs tools on Python 2.6
With the tools currently used it is not possible to generate documentation with Python 2.6. [nbsphinx](
https://pypi.org/project/nbsphinx/) it is not compatible with this legacy version of Python and does not even install. This was found out in #1385.

The build tools already take that into account: [build.sh](https://github.com/RedHatInsights/insights-core/blob/8510d6c072bed1a0c9823006ae3c6c9ada30791c/build.sh#L11), [shippable.yml](https://github.com/RedHatInsights/insights-core/blob/372f9228f37c7e118ffc70cdcc4c8fd8b9ddc6cd/shippable.yml#L17). Currently we can add a condition to the [docs packages](https://github.com/RedHatInsights/insights-core/pull/1369/files#diff-2eeaed663bd0d25b7e608891384b7298R51) not to install on Python ≤2.7. This would unblock `pip install -e .[develop]` before #1369 is merged. It might be also nice to print some warning.

After splitting the [develop bundle](https://github.com/RedHatInsights/insights-core/blob/372f9228f37c7e118ffc70cdcc4c8fd8b9ddc6cd/setup.py#L42) I think it would be good to check for the Python version and refuse to install the [docs bundle](https://github.com/RedHatInsights/insights-core/pull/1369/files#diff-2eeaed663bd0d25b7e608891384b7298R51) completely.

For the former it should be possible to use [Environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers). Unfortunately I am currently not sure how to do the warning or how to prevent install when running `pip install -e .`, so no specific suggestion for now.

It would be also nice to mention somewhere in the readme and/or another places where Python 2.6 support is mentioned, that this does not apply for docs.

Glutexo:
The [docs tools](
https://github.com/RedHatInsights/insights-core/pull/1369/files#diff-2eeaed663bd0d25b7e608891384b7298R51) cannot be installed on Python 2.6. The [nbsphinx](
https://pypi.org/project/nbsphinx/) package is not compatible with this legacy version of Python. Building docs is not necessary even for development. Added a [condition](https://github.com/Glutexo/insights-core/blob/no_docs_on_python26/setup.py#L51) to not install those on an unsupported system. That finally prevents `pip install -e .[develop]` from failing. Rejoice!

No warning is issued. Even though it is possible to do so. PIP [discourages](https://stackoverflow.com/a/44617404/1307676) it, it might not work when the wheel is already built in the cache, wouldn’t be printed if ran without `-v` (verbose) etc. It would be good to add some note to the documentation: to the guide to building docs and to the proclaimed Python 2.6 support.

To test this, you have to [limit](https://github.com/Glutexo/insights-core/blob/4272741dff67ebe23cbbdcfd8126b6e25d5ad631/setup.py#L42) pyOpenSSL version to [17.5.0](
https://pypi.org/project/pyOpenSSL/17.5.0/) or simply apply #1386.

Together with #1386 this will help #1342 to run `pip install -e .[develop]` in a clean environment even before #1369 is merged.

Fixes #1387.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,13 +48,13 @@ develop = set([
     'pytest==3.0.6',
     'pytest-cov==2.4.0',
     'mock==2.0.0',
-    'Sphinx',
-    'nbsphinx==0.3.1',
-    'sphinx_rtd_theme',
+    'Sphinx;python_version>="2.7"',
+    'nbsphinx==0.3.1;python_version>="2.7"',
+    'sphinx_rtd_theme;python_version>="2.7"',
     'futures==3.0.5',
     'requests==2.13.0',
     'wheel',
-    'ipython<6',
+    'ipython<6;python_version>="2.7"',
     'colorama',
 ])
 


Glutexo:
Thanks to #1369 this is no longer necessary. Closing.

Issue #1383, opened 2018-08-20
Initialize release in evaluators.py


brantleyr:
`release` needs to initialized to `None` in evaluators.py:

```
response = evaluator_class(broker=broker).process(eval_groups)\n  File \"insights/core/evaluators.py\", line 55, in process\n    return self.get_response()\n  File \"insights/core/evaluators.py\", line 79, in get_response\n    return self.format_response(r)\n  File \"insights/core/evaluators.py\", line 138, in format_response\n    if self.release:\nAttributeError: 'InsightsEvaluator' object has no attribute 'release'\n"
```

An archive may come in that is non-RHEL.

brantleyr:
https://github.com/RedHatInsights/insights-core/pull/1389

PR #1389, opened 2018-08-21
Fix release initialization.

brantleyr:
Needs to be set if we are checking/setting it conditionally elsewhere.


brantleyr:
Resolves https://github.com/RedHatInsights/insights-core/issues/1383

PR #1400, opened 2018-08-27
Updated neutron_conf parser to support filter

ppandey-redhat:
Updated default.py to support OSP13 log files path for neutron_conf parser
Updated neutron_conf.py to support filter


lhuett:
This is approved, @xiangce this is ready to be merged.

PR #1401, opened 2018-08-27
get hostname archive path from uploader.json

gravitypriest:
Changing `/bin/hostname` to `/bin/hostname -f` broke the soscleaner obfuscation... so from now on to avoid the code being out of step let's auto-determine what the hostname path is from what's in uploader.json.


gravitypriest:
if `startswith('/bin/hostname')` is too strict of a check I'm willing to change it

Issue #1380, opened 2018-08-18
Client Test Attempts to  Removes non-user Insights App File


bfahr:
When I run the insights-core pytests on RHEL 6.10 I get the following error.  This file exists on my test VM but I don't believe any tests should be attempting to remove this file.

```
=========== FAILURES ======================
____________ test_delete_registered_file __________

    @patch('insights.client.utilities.constants.registered_files',
           ['/tmp/insights-client.registered',
            '/tmp/redhat-access-insights.registered'])
    def test_delete_registered_file():
>       util.write_registered_file()

insights/tests/client/test_utilities.py:113: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
insights/client/utilities.py:56: in write_registered_file
    delete_unregistered_file()
insights/client/utilities.py:91: in delete_unregistered_file
    write_to_disk(f, delete=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

filename = '/etc/insights-client/.unregistered', delete = True, content = '2018-08-18T09:47:58.402808'

    def write_to_disk(filename, delete=False, content=get_time()):
        """
        Write filename out to disk
        """
        if not os.path.exists(os.path.dirname(filename)):
            return
        if delete:
            if os.path.lexists(filename):
>               os.remove(filename)
E               OSError: [Errno 13] Permission denied: '/etc/insights-client/.unregistered'

insights/client/utilities.py:102: OSError
===== 1 failed, 1032 passed, 7 skipped in 5.04 seconds ====
```

PR #1406, opened 2018-08-29
complete patching of dotfiles

gravitypriest:
`write_(un)registered_file()` will delete the opposite dotfile but these weren't mocked out in testing. mock em out


kylape:
This fixes #1380

PR #1416, opened 2018-08-31
Added parser for file /var/log/ceph/ceph.log

ppandey-redhat:
New parser for parsing file /var/log/ceph/ceph.log.

### DIFF
--- a/insights/parsers/ceph_log.py
+++ b/insights/parsers/ceph_log.py
@@ -1,0 +1,39 @@
+"""
+CephLog - file ``/var/log/ceph/ceph.log``
+=========================================
+"""
+
+from .. import LogFileOutput, parser
+from insights.specs import Specs
+
+
+@parser(Specs.ceph_log)
+class CephLog(LogFileOutput):
+    """
+      Parse the ``/var/log/ceph/ceph.log`` file.
+
+      .. note::
+          Please refer to its super-class :class:`insights.core.LogFileOutput` for more
+          details.
+
+      Sample log lines::
+
+          2017-05-31 13:01:44.034376 mon.0 192.xx.xx.xx:6789/0 742585 : cluster [INF] pgmap v5133969: 320 pgs: 3 active+clean+scrubbing+deep, 317 active+clean; 898 GB data, 1828 GB used, 48447 GB / 50275 GB avail; 2027 kB/s rd, 20215 kB/s wr, 711 op/s
+          2017-05-31 13:01:45.041760 mon.0 192.xx.xx.xx:6789/0 742586 : cluster [INF] pgmap v5133970: 320 pgs: 3 active+clean+scrubbing+deep, 317 active+clean; 898 GB data, 1828 GB used, 48447 GB / 50275 GB avail; 1606 kB/s rd, 17354 kB/s wr, 718 op/s
+          2017-05-31 13:01:46.933829 osd.22 192.xx.xx.xx:6814/42154 172581 : cluster [WRN] 44 slow requests, 2 included below; oldest blocked for > 49.982746 secs
+          2017-05-31 13:01:46.933946 osd.22 192.xx.xx.xx:6814/42154 172582 : cluster [WRN] slow request 30.602517 seconds old, received at 2017-05-31 13:01:06.330484: osd_op(client.3395798.0:2855671 1.54392173 gnocchi_06c8214c-afae-4e64-8a4a-a466c4f257dc_1244160000.0_median_86400.0_v3 [write 26253~9] snapc 0=[] ondisk+write+known_if_redirected e487) currently waiting for subops from 23
+          2017-05-31 13:01:46.933955 osd.22 192.xx.xx.xx:6814/42154 172583 : cluster [WRN] slow request 30.530961 seconds old, received at 2017-05-31 13:01:06.402041: osd_op(client.324182.0:46141816 1.e637a4b3 measure [omap-rm-keys 0~107] snapc 0=[] ondisk+write+skiprwlocks+known_if_redirected e487) currently waiting for subops from 23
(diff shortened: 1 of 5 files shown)


insights/parsers/ceph_log.py
```diff
@@ -0,0 +1,39 @@
+"""
+CephLog - file ``/var/log/ceph/ceph.log``
+=========================================
+"""
+
+from .. import LogFileOutput, parser
```
psachin: Please avoid using relative imports.

insights/parsers/ceph_log.py
```diff
@@ -0,0 +1,39 @@
...
+          2017-05-31 13:01:48.057187 mon.0 192.xx.xx.xx:6789/0 742590 : cluster [INF] pgmap v5133972: 320 pgs: 3 active+clean+scrubbing+deep, 317 active+clean; 898 GB data, 1828 GB used, 48447 GB / 50275 GB avail; 2373 kB/s rd, 5138 kB/s wr, 354 op/s
+          2017-05-31 13:01:49.064950 mon.0 192.xx.xx.xx:6789/0 742598 : cluster [INF] pgmap v5133973: 320 pgs: 3 active+clean+scrubbing+deep, 317 active+clean; 898 GB data, 1828 GB used, 48447 GB / 50275 GB avail; 4187 kB/s rd, 10266 kB/s wr, 714 op/s
+          2017-05-31 13:01:50.069437 mon.0 192.xx.xx.xx:6789/0 742599 : cluster [INF] pgmap v5133974: 320 pgs: 3 active+clean+scrubbing+deep, 317 active+clean; 898 GB data, 1828 GB used, 48447 GB / 50275 GB avail; 470 MB/s rd, 11461 kB/s wr, 786 op/s
+
+      Examples:
+          >>> len(ceph_log.get("[WRN] slow request")) == 2
+          True
+          >>> from datetime import datetime
+          >>> len(list(ceph_log.get_after(datetime(2017, 5, 31, 13, 1, 46))))
```
psachin: This is great but we hardly use hard-coded datetime in rule logic. The rule check for certain pattern, and use the timestamp to search for lines using `.get_after()`. How about we directly provide timestamp as a `datetime` object? Something like https://github.com/RedHatInsights/insights-core/pull/1411/files

insights/specs/__init__.py
```diff
@@ -22,6 +22,7 @@ class Specs(SpecSet):
     ceph_config_show = RegistryPoint(multi_output=True)
     ceph_df_detail = RegistryPoint()
     ceph_health_detail = RegistryPoint()
+    ceph_log = RegistryPoint(multi_output=True, filterable=True)
```
psachin: Do we really need `multi_output` here? AFAIK this is used when file path is in multiple places, mostly suitable to conf files but not strictly.

insights/parsers/ceph_log.py
ppandey-redhat: @psachin 

Can you please elaborate a bit more. Are you suggesting to add a function to parse each line and return a msg_info including **timestamp** if available so that this info can be used effectively?
If I am not correct, could you please point the expected requirement/code in https://github.com/RedHatInsights/insights-core/pull/1411/files ? 

Sample code:

    def _parse_line(self, line):
        msg_info = {'raw_message': line}
        line_splits = line.split()
        dt = ' '.join(line_splits[0:2]).split('.')[0]
        if datetime.strptime(dt, self.time_format):
                msg_info['timestamp'] = datetime.strptime(dt, self.time_format)
                msg_info['ceph_component'] = line_splits[2]
                msg_info['level'] = line_splits[7].strip('[]')
                msg_info['message'] = ' '.join(line_splits[8:])
        return msg_info

psachin: @ppandey-redhat Please ignore this. It was just a suggestion.

ppandey-redhat: @psachin 
Acknowledged, but even if it was a suggestion for the betterment and ease of use, I would have liked to incorporate the suggested changes in the parser. 
In the latest commit, no modifications are made for this comment.

insights/specs/__init__.py
ppandey-redhat: @psachin 
Acknowledged. 'multi_output' is not required here. Changes done in latest commit.

insights/parsers/ceph_log.py
ppandey-redhat: @psachin 
Acknowledged. Changes done in latest commit.

YakovLittle:
test this please

insights/parsers/ceph_log.py
ppandey-redhat: Acknowledged.

insights/parsers/ceph_log.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+from insights import LogFileOutput, parser
+from insights.specs import Specs
+
+
+@parser(Specs.ceph_log)
+class CephLog(LogFileOutput):
+    """
+      Parse the ``/var/log/ceph/ceph.log`` file.
```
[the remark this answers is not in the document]
ppandey-redhat: Acknowledged. Parser updated in the latest commit.

xiangce:
Looks good to me, and seems there are more than one log files under this dir `/var/log/ceph`, e.g.:
~~~
[root@mon1 log]# ls /var/log/ceph/*.log
ceph.audit.log    ceph.log     ceph-mon.mon1.log
~~~
If parsers for other logs are needed in the future, better place them in this python file as well, just like the [ovirt_engine_log.py](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/ovirt_engine_log.py)

BTW, please create a JIRA to add the spec.

ppandey-redhat:
@xiangce 

JIRA for specs addition is created (INSPEC-92) and plugin PR (1047) dependent on this parser is also created.

PR #1418, opened 2018-08-31
Limit time of requests issued before failure

Glutexo:
If the API servers are not reachable, the first thing that fails is the egg update. To reach there, several requests are made to set up the right configuration. Most of those have no timeout set though. That results to a long waits when connection blocked by a firewall.

This is the first step to solve the reported bug that client hangs too long when connection is blocked by a firewall. In the long run it would be better to transparently use the timeout from config for every request. Solution to the failure on update will come in a separate PR. @kylape suggested not crashing on this, more specifically not crashing at all if it doesn’t prevent the app flow.

The functional tests are currently testing only one possible configuration. Improvements to this will come in a separate PR.

### DIFF
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -127,10 +127,10 @@ class InsightsClient(object):
         if current_etag and not force:
             logger.debug('Requesting new file with etag %s', current_etag)
             etag_headers = {'If-None-Match': current_etag}
-            response = self.session.get(url, headers=etag_headers)
+            response = self.session.get(url, headers=etag_headers, timeout=self.config.http_timeout)
         else:
             logger.debug('Found no etag or forcing fetch')
-            response = self.session.get(url)
+            response = self.session.get(url, timeout=self.config.http_timeout)
 
         # Debug information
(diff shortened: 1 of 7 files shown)


YakovLittle:
test this please

YakovLittle:
test this please

esakaiev:
test this please

kylape:
@Glutexo the only thing I can comment on is the use of `insights_client()` in `test_fetch.py`.  Could you add the `@pytest.fixture` decorator on the `insights_client`?  Then you could pass `insights_client` in as a parameter rather than having to instantiate it in the function.

Glutexo:
You‘re right. I somehow forgot that there are fixtures. Unfortunately I already did this in other tests. At least here it will be right. Thanks!

Glutexo:
Rebased so this PR becomes immediately mergeable. Improvements to [_test_net_comm.py_](
https://github.com/RedHatInsights/insights-core/pull/1418/files#diff-059148a32698b34cd2ca269bfddc48cc) will come in a follow-up PR.

PR #1420, opened 2018-09-03
Fix Satellite 5 auto configuration

Glutexo:
A [_set_auto_configuration_](https://github.com/RedHatInsights/insights-core/blob/5979ee1438cb75b46614c57c165c36daf269b753/insights/client/auto_config.py#L53) function call was [missing](https://github.com/RedHatInsights/insights-core/compare/master...Glutexo:fix_satellite5_autoconfig?expand=1#diff-2ff859dc8d708efc4e54514c0a483be0L196) a _config_ argument. That caused a runtime error if auto configuration [falls back](https://github.com/RedHatInsights/insights-core/blob/5979ee1438cb75b46614c57c165c36daf269b753/insights/client/auto_config.py#L212) to Satellite 5.

The test is not very nice – uses a lot of patching and virtually tests many things that are out of its scope. That‘s because the method is very non-atomic, does a lot of stuff that is not well mockable and thus is only difficultly testable. Please 🐻 with that for now. It’s a nice candidate for refactoring.

### DIFF
--- a/insights/client/auto_config.py
+++ b/insights/client/auto_config.py
@@ -193,7 +193,7 @@ def _try_satellite5_configuration(config):
                 else:
                     proxy = proxy + proxy_host_port
                     logger.debug("RHN Proxy: %s", proxy)
-            set_auto_configuration(hostname, rhn_ca, proxy)
+            set_auto_configuration(config, hostname, rhn_ca, proxy)
         else:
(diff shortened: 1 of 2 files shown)


Glutexo:
Jenkins is failing, but I believe that it’s not my fault.

Asking @gravitypriest for a review.

PR #1423, opened 2018-09-05
Restore exit handler

gravitypriest:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1597308

### DIFF
--- a/insights/client/insights_spec.py
+++ b/insights/client/insights_spec.py
@@ -51,7 +51,8 @@ class InsightsCommand(InsightsSpec):
         '''
         # all commands should timeout after a long interval so the client does not hang
         # prepend native nix 'timeout' implementation
-        timeout_command = 'timeout %s %s' % (self.config.cmd_timeout, self.command)
+        timeout_command = 'timeout -s KILL %s %s' % (
+            self.config.cmd_timeout, self.command)
 
         # ensure consistent locale for collected command output
         cmd_env = {'LC_ALL': 'C'}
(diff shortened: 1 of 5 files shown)


insights/client/insights_spec.py
```diff
@@ -44,7 +44,8 @@ def get_output(self):
         '''
         # all commands should timeout after a long interval so the client does not hang
         # prepend native nix 'timeout' implementation
-        timeout_command = 'timeout %s %s' % (self.config.cmd_timeout, self.command)
+        timeout_command = 'timeout -s KILL %s %s' % (
```
kylape: Did you consider using `-k` instead of starting off with the `KILL`?

gravitypriest: I did but I thought that might affect our timeout interval.

gravitypriest:
Changing the command timeout seems prudent. Currently if even one command takes that long, the process already hits our systemd/cron timeout.

insights/client/insights_spec.py
kylape: imho if we lover the command timeout, that shouldn't be a big concern.  we could even just do `-k 5` to give a small buffer for graceful termination.

kylape:
What do you think for a default command timeout?  60 seconds?

gravitypriest:
I'm in favor of `-k`. As for timeout, I'm not really sure what would be appropriate. Evidently there are some machines out there taking a long-ass time with some commands. Maybe 2 mins to be safe?

Makes me think we should really start including the elapsed time in the upload headers again.

kylape:
Sure 2m works for me.  It would be great if we included elapsed time in the logs.

PR #1427, opened 2018-09-11
Blocked: Add parser for ls_pgsql_data

JoySnow:


xiangce:
This is a workaround to try to handle the `Log Date issue`. Since the resolution for this issue is under discussing, and this workaround is not as good as expected, I would close it at first.

PR #1428, opened 2018-09-11
New Parser For /etc/ansible/hosts

akshay196:
Parser for file /etc/ansible/hosts parses hosts, hosts groups and variables
from the inventory file. This file is used for OpenShift container platform
configuration.

Checklist
- [x] Parser
- [x] Tests
- [x] Specs changes
- [x] Docs


akshay196:
@xiangce Thank you for suggestion. Yes, It would be better if we parse using `ansible-inventory` command instead of parsing `/etc/ansible/hosts` file.

Currently we don't need this parser as we don't have any rule that uses this. So, I am closing this PR as of now. When needed we can reopen this PR and would like to work on this.

PR #1429, opened 2018-09-11
Add parser for multipath_conf_initramfs

JoySnow:


insights/parsers/multipath_conf.py
```diff
@@ -90,7 +100,7 @@ class MultipathConf(Parser, LegacyItemAccess):
         }
 
     Examples:
-        >>> conf = shared[MultipathConf]
+        >>> conf = shared[MultipathConfParser]
```
xiangce: Since the `MultipathConfParser` is a base class without any spec, which means the base class itself will never hit, I think better move the examples to the sub-class.

insights/parsers/multipath_conf.py
```diff
@@ -130,7 +129,46 @@ def _create_parser(cls):
         return my_conf
 
     def parse_content(self, content):
-        self.data = MultipathConf._create_parser().parseString("\n".join(content))[0].asDict()
+        self.data = MultipathConfParser._create_parser().parseString("\n".join(content))[0].asDict()
```
xiangce: AFAIK, the output of the command `/bin/lsinitrd -f /etc/multipath.conf` could be empty when there is no such file in the `initramfs`, I think we'd better skip this case by ignoring an Exception for empty content, how do you think?

insights/specs/insights_archive.py
```diff
@@ -121,6 +121,7 @@ class InsightsArchiveSpecs(Specs):
     mount = simple_file("insights_commands/mount")
     multicast_querier = simple_file("insights_commands/find_.sys.devices.virtual.net._-name_multicast_querier_-print_-exec_cat")
     multipath__v4__ll = simple_file("insights_commands/multipath_-v4_-ll")
+    multipath_conf_initramfs = simple_file("insights_commands/lsinitrd_-f_.etc.multipath.conf")
```
xiangce: Better move it forward to `multipath__v4__ll`

Issue #1432, opened 2018-09-12
utf-8 UnicodeDecodeError in 'var/log/messages' parsing


JoySnow:
A example of this broken:
```
(insights-cli-venv) ╭[localhost] ~/Work/insights-cli-venv
╰> insights-cli  --component-modules diag_insights_rules.plugins.storage.lun_assignment_changed --verbose  -- ~/Downloads/sosreport-example.tar.xz
...
INFO     Trying insights.parsers.messages.Messages
DEBUG    [['grep', '-F', 'kernel', '/tmp/insights-PRsYcB/sosreport-example/var/log/messages']]
WARNING  Traceback (most recent call last):
  File "/home/joy/Work/insights-core/insights/core/dr.py", line 689, in run
    result = DELEGATES[component].process(broker)
  File "/home/joy/Work/insights-core/insights/core/dr.py", line 501, in process
    return self.invoke(broker)
  File "/home/joy/Work/insights-core/insights/core/plugins.py", line 51, in invoke
    return self.component(dep_value)
  File "/home/joy/Work/insights-core/insights/core/__init__.py", line 82, in __init__
    self._handle_content(context)
  File "/home/joy/Work/insights-core/insights/core/__init__.py", line 85, in _handle_content
    self.parse_content(context.content)
  File "/home/joy/Work/insights-core/insights/core/spec_factory.py", line 104, in content
    self._content = self.load()
  File "/home/joy/Work/insights-core/insights/core/spec_factory.py", line 178, in load
    rc, out = self.ctx.shell_out(cmd, keep_rc=True, env=SAFE_ENV)
  File "/home/joy/Work/insights-core/insights/core/context.py", line 142, in shell_out
    raw = self.check_output(cmd, timeout=timeout, keep_rc=keep_rc, env=env)
  File "/home/joy/Work/insights-core/insights/core/context.py", line 138, in check_output
    keep_rc=keep_rc, env=env)
  File "/home/joy/Work/insights-core/insights/util/subproc.py", line 117, in call
    return rc, output.decode(output_encoding) if output_encoding else output
  File "/home/joy/Work/insights-cli-venv/lib64/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xd2 in position 1407487: invalid continuation byte
```
Solution: 
[...]

PR #1433, opened 2018-09-12
Fix the UnicodeDecodeError in call() of util/subproc.py

JoySnow:
Resolve https://github.com/RedHatInsights/insights-core/issues/1432


xiangce:
Looks good to me

xiangce:
The resolution is from https://docs.python.org/3/howto/unicode.html#the-unicode-type

PR #1442, opened 2018-09-19
New parsers and combiner for `ipcs -m` commands

xiangce:
- Move the IpcsS and IpcsSI to ipcs.py
- Mark the original IpcsS and IpcsSI in ipcs_sem.py as deprecated


insights/specs/default.py
```diff
@@ -281,6 +281,8 @@ def httpd_cmd(broker):
     ip_route_show_table_all = simple_command("/sbin/ip route show table all")
     ip_s_link = simple_command("/sbin/ip -s link")
     ipaupgrade_log = simple_file("/var/log/ipaupgrade.log")
+    ipcs_m = simple_command("/usr/bin/ipcs -s")
+    ipcs_m_p = simple_command("/usr/bin/ipcs -s -p")
```
lhuett: I think these need to be ...
```
ipcs_m = simple_command("/usr/bin/ipcs -m")
ipcs_m_p = simple_command("/usr/bin/ipcs -m -p")
```

xiangce: Thanks, @lhuett. I fixed them.

Issue #1441, opened 2018-09-18
Incorrect configuration file paths for specs related to OpenStack


pratikgadiya12:
Configuration file path added to the OpenStack components related specs in insights-core seems to be incorrect.

Ex:
[https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/default.py#L399](https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/default.py#L399)
[https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/default.py#L417](https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/default.py#L417)

and pretty much all the specs path where we have config-data as a part of path for *.conf files

pratikgadiya12:
To re-verify the same I have deployed an OSP 13 cluster and found that changes made in the file path mentioned in the repository are not reflected in containers of the cluster even after restart

Incorrect Path: /var/lib/config-data/neutron/etc/neutron/neutron.conf
Correct Path: /var/lib/config-data/puppet-generated/neutron/etc/neutron/neutron.conf

PR #1445, opened 2018-09-19
Fixed OSP specs path in insights-core

pratikgadiya12:
https://github.com/RedHatInsights/insights-core/issues/1441


pratikgadiya12:
The file paths mentioned for OSP conf file parsers under specs in insights-core seems to be incorrect because those paths are not used to update the configuration within the containers from that of controller node.

I re-tested the same by deploying an OSP 13 environment and hence the patch.

**Example**:
Neutron Conf:
Specs: https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/default.py#L399
Incorrect Path: `/var/lib/config-data/neutron/etc/neutron/neutron.conf`
Correct Path: `/var/lib/config-data/puppet-generated/nova/etc/nova/nova.conf`

pratikgadiya12:
I will wait for everyone's view on this since it will affect OSP<=12 and OSP>12

xiangce:
@pratikgadiya12 - As discussed, have you checked the paths in OSP 12? Since from OSP12,  some of the components (except neutron) were deployed in containers.

pratikgadiya12:
@xiangce : yes, I have checked that

Issue #1303, opened 2018-07-28
Add xfs_info datasource to default.py


csams:
An xfs_info datasource exists in `sos_archive.py`. Should one exist in `default.py`, and how should it be implemented if so?

ritwik12:
@csams I can see xfs_info already present [here](https://github.com/RedHatInsights/insights-core/blob/9ece3f37358987ed6d245585e065237aedcce4cc/insights/specs/default.py#L713) in default.py. Am I getting confused with something here?

csams:
Hey @ritwik12, yep, it looks like that was added back in September. It's an interesting issue because it's the first case where a datasource depends on a parser. That'll have implications for other things soon, but we can open separate issues as they come up. Thanks!

Fixed in 4f07378d0

ritwik12:
Fixed in #1446

ritwik12:
@csams That looks great, your welcome.

PR #1446, opened 2018-09-19
Added xfs_mounts datasource and xfs_info spec

lhuett:
Added the following to default.py
Added xfs_mounts datasource using the Mount parser to retrieve 'xfs' mount points
@datasource(Mount)
    def xfs_mounts(broker):
        mnt = broker[Mount]
        mps = mnt.search(mount_type='xfs')
        return [m.mount_point for m in mps if mps]
Added xfs_info spec. Iterates through mount points from xfs_mounts datasource to get xfs info for each.
    xfs_info = foreach_execute(xfs_mounts, "/usr/sbin/xfs_info %s")


ritwik12:
Solves #1303

PR #1436, opened 2018-09-14
make remove.conf location customizable

gravitypriest:
we documented it for so long, we might as well make it a real option

PR #1447, opened 2018-09-20
Fix remove file usage in collection tests

Glutexo:
[_test_get_rm_conf.test_no_file_](https://github.com/RedHatInsights/insights-core/blob/ce519eca9785a56495a2cd50e84917ab3f4aa058/insights/tests/client/collection_rules/test_get_rm_conf.py#L44) did not work properly: it tested the default value of the remote_file constant instead of the [patched](https://github.com/RedHatInsights/insights-core/blob/ce519eca9785a56495a2cd50e84917ab3f4aa058/insights/tests/client/collection_rules/test_get_rm_conf.py#L7) one. Fixed this to be consistend with other tests.

[_test_collect.patch_isfile_](https://github.com/RedHatInsights/insights-core/blob/ce519eca9785a56495a2cd50e84917ab3f4aa058/insights/tests/client/test_collect.py#L114) was patching the _isfile_ method and the [_collection_remove_file_](https://github.com/RedHatInsights/insights-core/blob/ce519eca9785a56495a2cd50e84917ab3f4aa058/insights/client/constants.py#L24) constant at once, which was not very transparent and violated SRP. [Split](https://github.com/RedHatInsights/insights-core/compare/master...Glutexo:fix_collection_tests?expand=1#diff-bc977d9fd95882baacf1c6411537a40bR115) into two methods.

[Changed](https://github.com/RedHatInsights/insights-core/compare/master...Glutexo:fix_collection_tests?expand=1#diff-bc977d9fd95882baacf1c6411537a40bR16) mocked [removed files](https://github.com/RedHatInsights/insights-core/blob/ce519eca9785a56495a2cd50e84917ab3f4aa058/insights/tests/client/test_collect.py#L15) in [_test_collect_](https://github.com/RedHatInsights/insights-core/blob/ce519eca9785a56495a2cd50e84917ab3f4aa058/insights/tests/client/test_collect.py) so their names are not so confusing. It looked like configuration file names, but they are files to be removed. Also [renamed](https://github.com/RedHatInsights/insights-core/compare/master...Glutexo:fix_collection_tests?expand=1#diff-bc977d9fd95882baacf1c6411537a40bR16) the variables to make their purpose clearer.

This is related to @gravitypriest’s PR #1436. I found those issues when reviewing. If it gets merged before this one, I’ll resolve the conflicts.

### DIFF
--- a/insights/tests/client/collection_rules/test_get_rm_conf.py
+++ b/insights/tests/client/collection_rules/test_get_rm_conf.py
@@ -4,8 +4,8 @@ from .helpers import insights_upload_conf
 from mock.mock import patch
 
 
-remove_file = '/etc/insights-client/remove.conf'
-remove_files = ["/etc/remove.conf", "/tmp/remove.conf"]
+conf_remove_file = '/tmp/remove.conf'
+removed_files = ["/etc/some_file", "/tmp/another_file"]
 
 
 def patch_isfile(isfile):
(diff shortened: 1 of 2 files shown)


Glutexo:
There were some conflict, because #1436 got merged. Rebased and resolved.

PR #1450, opened 2018-09-21
Add parser for 'ceph insights' command

noahdesu:
The current set of Ceph-related parsers are fed by various CLI-based data sources (e.g. `ceph version` or `ceph osd dump`). This creates a tight dependency between insights parsers and the output formats of  Ceph CLI commands. The second issue is that we may want to provide certain types of data to insights that aren't necessarily relevant for exposing through CLI commands.

The merged PR here https://github.com/ceph/ceph/pull/23497 introduces a single Ceph CLI command designed specifically as a single data source for insights (i.e. `/usr/bin/ceph insights`).

This PR introduces a parser for this new data source. This renders most of the other Ceph-related data sources redundant. However, those parsers are specs are left alone since I suspect they will need to remain in the tree for older versions of Ceph deployments.

Follow on PRs will add new combiners to unify these data sources.

### DIFF
--- a/insights/parsers/ceph_insights.py
+++ b/insights/parsers/ceph_insights.py
@@ -1,0 +1,88 @@
+"""
+ceph_insights - command ``ceph insights``
+=========================================
+"""
+import json
+import re
+import six
+from .. import Parser, parser
+from insights.specs import Specs
+
+@parser(Specs.ceph_insights)
+class CephInsights(Parser):
+    """
+    Parse the output of the ``ceph insights`` command.
+
+    Attributes:
+        version (dict): version information of the Ceph cluster.
+        data (dict): a dictionary of the parsed output.
+
+    The ``data`` attribute is a dictionary containing the parsed output of the
+    ``ceph insights`` command. The following are available in ``data``:
+
+    * ``crashes`` - summary of daemon crashes for the past 24 hours
+    * ``health`` - the current and historical (past 24 hours) health checks
+    * ``config`` - cluster and daemon configuration settings
+    * ``osd_dump`` - osd and pool information
+    * ``df`` - storage usage statistics
+    * ``osd_tree`` - osd topology
+    * ``fs_map`` - file system map
+    * ``crush_map`` - the CRUSH map
+    * ``mon_map`` - monitor map
+    * ``service_map`` - service map
+    * ``manager_map`` - manager map
+    * ``mon_status`` - monitor status
+    * ``pg_summary`` - placement group summary
+    * ``osd_metadata`` - per-OSD metadata
+    * ``version`` - ceph software version
+    * ``errors`` - any errors encountered collecting this data
+
+    The ``version`` attribute contains a normalized view of ``self.data["version"]``.
+
+    Examples:
+        >>> ceph_insights.version["release"] == 14
+        True
+        >>> ceph_insights.version["major"] == 0
+        True
+        >>> ceph_insights.version["minor"] == 0
+        True
+        >>> isinstance(ceph_insights.data["crashes"], dict)
+        True
+        >>> isinstance(ceph_insights.data["health"], dict)
+        True
+    """
+    IGNORE_RE = [
+         "\*\*\* DEVELOPER MODE",
+         "\d+-\d+-\d+.+WARNING: all dangerous"
+    ]
+
+    def _sanitize_content(self, content):
+        """Remove lines matching IGNORE_RE at start of content"""
+        slice_point = 0
+        ignore_re = re.compile('|'.join(CephInsights.IGNORE_RE))
+        for line in content:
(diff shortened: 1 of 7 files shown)


bfahr:
@noahdesu you have some linting errors in your changes.  You can run `flake8` from the insights-core root directory to find those.

bfahr:
@noahdesu no need to squash commits, we normally do that when merging.  Squashing sometimes makes it difficult to review history of the PR.

insights/parsers/ceph_insights.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    IGNORE_RE = [
+         "\*\*\* DEVELOPER MODE",
+         "\d+-\d+-\d+.+WARNING: all dangerous"
+    ]
+
+    bad_lines = CommandParser.bad_lines + [
+        "module 'insights' is not enabled",
+        "no valid command found"
+    ]
```
bfahr: @noahdesu this is the first time anyone has extended `bad_lines` in this manner.  Rather that permanently tie the internal implementation to a list of bad lines, would you mind to implement a new class method `add_bad_lines` in the `CommandParser` class that can be invoked in the sub class body (replacing this assignment) to add a list of subclass lines to the CommandParser?

csams: @noahdesu Also want to rename `CommandParser.bad_lines` to `CommandParser.__bad_lines` so subclasses don't accidentally shadow it.

noahdesu: @csams @bfahr do you want this to be a class variable, rather than an instance variable? in the former case changes to bad_lines will be seen by all other command parser instance. caveat: perhaps there is a python trick / pattern I'm not aware of?

csams: @noahdesu actually, the `CommandParser`'s logic should be for lines that might indicate failure for *any* command, not just a particular one. Since the lines in this case are specific to the command, the parser itself should just check for them in its `parse_content` function in a way that makes sense to it. We should still adjust `CommandParser.bad_lines` to `CommandParser.__bad_lines` so subclasses don't accidentally clobber it, but I don't think we should provide a general API for this. The `CommandParser` exists only to prevent every parser from needing to deal with a collection bug in the wild that mistakes error messages from the shell as output from commands.

insights/specs/default.py
```diff
@@ -89,6 +89,7 @@ def tomcat_base(broker):
     ceph_config_show = foreach_execute(ceph_socket_files, "/usr/bin/ceph daemon %s config show")
     ceph_df_detail = simple_command("/usr/bin/ceph df detail -f json-pretty")
     ceph_health_detail = simple_command("/usr/bin/ceph health detail -f json-pretty")
+    ceph_insights = simple_command("/usr/bin/ceph insights")
```
csams: Let's wait until #1455 gets merged before adding the datasource implementation. We can leave the `RegistryPoint`, but we want to protect this command from running on every machine. It'll need to depend on a separate component that figures out whether it should run or not so we we don't stampede the cluster.

For development I can send you an example of how to use a datasource implementation that hooks into a `RegistryPoint` without needing to be in core.

insights/parsers/ceph_insights.py
noahdesu: I'm fine with moving the check into the Ceph parser, but the semantics are exactly the same except for the strings being search for, so it'd be nice to reuse that. Another option would be to add an optional argument to CommandParser constructor. Something like:

```python
  class CommandParser(Parser):
    def __init__(self, context, extra_bad_lines=[]):
      self.validate_lines(context.content, self.__badlines + extra_bad_lines)

    @staticmethod
    def validate_lines(results, bad_lines):
       ....
```

insights/specs/default.py
noahdesu: Sounds great, I'll keep an eye on #1455. I'm going to dig into the instructions you sent for running end-to-end examples on Friday or Monday, so I could look at both at the same time.

insights/parsers/ceph_insights.py
csams: @noahdesu I didn't want to make it part of the API for subclasses since we want to remove the check from the base class when the old collectors with the bug are deprecated, but we can remove the check in a smarter way while allowing subclasses to reuse the validation code.

Go ahead with the constructor keyword. Just protect the concatenation so we're not creating a new list on every call if `extra_bad_lines` is empty.

Does the ceph command have a nonzero return code in the cases you're protecting against?

noahdesu: pushed changes for this. if it looks ok i'll let it sit for now until #1455 is done.

bfahr:
@noahdesu the last commit you made has some issues.

noahdesu:
@bfahr thanks. i updated with the fixes.

Issue #1373, opened 2018-08-16
Limit pyOpenSSL version to 17.5.0


Glutexo:
The [pyOpenSSL](https://pypi.org/project/pyOpenSSL/) package currently [required](https://github.com/RedHatInsights/insights-core/blob/95d79585f06a99dc6c6e910bad0b2d896a9b9d79/setup.py#L42) does not support Python 2.6 in its most recent version [18.0.0](https://pypi.org/project/pyOpenSSL/18.0.0/). However only a one step previous version [17.5.0](https://pypi.org/project/pyOpenSSL/17.5.0/) still supports it and installs just fine.

If the tests don’t fail and there are no other issues (security, since it’s SSL?), we should note this specific version in the [dependency list](https://github.com/RedHatInsights/insights-core/blob/95d79585f06a99dc6c6e910bad0b2d896a9b9d79/setup.py#L42). Otherwise the package is not installable on systems with this legacy version of Python (RHEL6).

Glutexo:
Added new information to the description:

* Version limit must not have lower boundary.
* It’s necessary to limit version of [_pycparser_](https://pypi.org/project/pycparser/).

PR #1451, opened 2018-09-24
Run pytest in virtualenv on RHEL6

Glutexo:
A Jenkins _pytest_ [task](https://github.com/RedHatInsights/insights-core/compare/master...Glutexo:rhel6_virtualenv?expand=1#diff-58231b16fdee45a03a4ee3cf94a9f2c3L19) for RHEL 6 did not use a virtual environment. That was because of [broken dependencies](https://github.com/RedHatInsights/insights-core/blob/4f07378d044c63559fadd709fbe7809ee5fe667c/setup.py#L42) (#1373) in [setup.py](https://github.com/RedHatInsights/insights-core/blob/4f07378d044c63559fadd709fbe7809ee5fe667c/setup.py). With those fixed, it’s possible to use a virtualenv on RHEL 6 the same way it is used on [RHEL 7](https://github.com/RedHatInsights/insights-core/blob/4f07378d044c63559fadd709fbe7809ee5fe667c/Jenkinsfile#L26).

Temporarily added [_pycparser_](https://github.com/RedHatInsights/insights-core/pull/1451/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R24) and [_pyOpenSSL_](https://github.com/RedHatInsights/insights-core/pull/1451/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R25) dependencies manually. It satisfies those currently [set](https://github.com/RedHatInsights/insights-core/blob/4f07378d044c63559fadd709fbe7809ee5fe667c/setup.py#L42) in [_setup.py_](https://github.com/RedHatInsights/insights-core/blob/4f07378d044c63559fadd709fbe7809ee5fe667c/setup.py), but makes it work on RHEL 6 even before #1386 gets merged.


Glutexo:
Tests pass! 🍏

Glutexo:
@eduardocerqueira or @bfahr may be able to review this.

Jenkinsfile
```diff
@@ -17,10 +17,15 @@ pipeline {
...
           steps {
-            echo "Installing Insights..."
-            sh 'pip install --user -e .[testing]'
             echo "Testing with Pytest..."
-            sh 'pytest'
+            sh """
+                virtualenv .testenv
+                source .testenv/bin/activate
+                pip install "pycparser<=2.18"
+                pip install "pyOpenSSL<=17.5.0"
```
bfahr: @Glutexo this may not be an apples to apples comparison, but when I try this on an up-to-date RHEL 6.10 vm I get this error:
```
pip install "pyOpenSSL<=17.5.0"
You are using pip version 6.0.8, however version 18.0 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting pyOpenSSL<=17.5.0
  Using cached https://files.pythonhosted.org/packages/79/db/7c0cfe4aa8341a5fab4638952520d8db6ab85ff84505e12c00ea311c3516/pyOpenSSL-17.5.0-py2.py3-none-any.whl
Collecting six>=1.5.2 (from pyOpenSSL<=17.5.0)
  Using cached https://files.pythonhosted.org/packages/67/4b/141a581104b1f6397bfa78ac9d43d8ad29a7ca43ea90a2d863fe3056e86a/six-1.11.0-py2.py3-none-any.whl
Collecting cryptography>=2.1.4 (from pyOpenSSL<=17.5.0)
  Using cached https://files.pythonhosted.org/packages/22/21/233e38f74188db94e8451ef6385754a98f3cad9b59bedf3a8e8b14988be4/cryptography-2.3.1.tar.gz
      File "<string>", line 20, in <module>
      File "/tmp/pip-build-H9mTpv/cryptography/setup.py", line 28, in <module>
        "cryptography requires setuptools 18.5 or newer, please upgrade to a "
    RuntimeError: cryptography requires setuptools 18.5 or newer, please upgrade to a newer version of setuptools
    Complete output from command python setup.py egg_info:
    
      File "<string>", line 20, in <module>
    
      File "/tmp/pip-build-H9mTpv/cryptography/setup.py", line 28, in <module>
    
        "cryptography requires setuptools 18.5 or newer, please upgrade to a "
    
    RuntimeError: cryptography requires setuptools 18.5 or newer, please upgrade to a newer version of setuptools
    
    ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-H9mTpv/cryptography
```

If the goal is to make this work in our specific test environment then the above error is not as important as making this work on a stock RHEL vm.

Glutexo: I intended this PR only for the Jenkins pipeline. If there is a problem with cryptography (and thus pyOpenSSL) dependencies, it should better come to #1396. The [_insights-client_](https://github.com/RedHatInsights/insights-client/) package has _setuptools_ in its RPM [dependencies](https://github.com/RedHatInsights/insights-client/blob/a931d13279158838d9527f61a357a7d0d8d92511/etc/insights-client.spec#L25), so I guess this is why the error doesn’t happen in the real world.

bfahr: @Glutexo thanks, I just wanted to confirm.

Glutexo: Some things I figured out:

* When I create a new virtual environment with `virtualenv venv`, it has `setuptools<37` installed in there even though I haven’t got the `python-setuptools` package installed.
* Even after manually removing _setuptools_ from the virtual environment, both _pycparser_ and _pyOpenSSL_ install just fine.

I would like to ask you to check that a little deeper, better in the context of #1396 than by doing these manual steps. If you can confirm this, I can try to spin up a fresh RHEL 6 instance to see whether _insights-core_ installs without relying on what the _insights-client_ RPM brings along.

Glutexo:
Who could merge this… @eduardocerqueira, @kylape?

PR #1453, opened 2018-09-25
Enhance parser nginx_conf

wushiqinlou:


wushiqinlou:
Previously, the max depth is 7, however, I found that nested location can be configured in location, so the max depth should be 11.

csams:
Arbitrary configuration depths are handled already in the [ConfTree version of the http parser and combiner](https://github.com/RedHatInsights/insights-core/blob/master/insights/combiners/httpd_conf.py#L369). Please use that combiner instead, and feel free to reach out if you need help with the conversion.

wushiqinlou:
@csams - Thanks for your info. The arch of nginx configuration content is like "{} in {}" , so the depth from my comment is about {} depth in configuration content, it is not about configuration file depths.

csams:
@wushiqinlou no problem! Either way, we should use the [combiner based on configtree](https://github.com/RedHatInsights/insights-core/blob/master/insights/combiners/httpd_conf.py#L369). Are you having problems with it?

wushiqinlou:
@csams - Thanks for your info, it is OK to me now, if any issue, I will submit a issue.

xiangce:
@wushiqinlou - just changed the "_listdepth_XXX" functions to a recursive function, please have a check.

wushiqinlou:
@xiangce - The recursion is OK to me.

PR #1455, opened 2018-09-26
Core3 collector without insights-client integration

csams:
This PR includes a collection module that produces an archive from `datasource` components. The module can be invoked with the `collect` function in python or by running `insights-collect` from the command line. The archive it produces is recognized and processed by the rest of the framework in the same way as other archives.

Both invocations take a manifest with collection parameters and descriptions of which components to enable and persist to the archive.

This PR also allows `simple_command` and `foreach_execute` datasources to specify a list of environment variables to inherit when executing. This is sometimes necessary when calling `oc` in openshift, for example.


insights/specs/default.py
```diff
@@ -428,23 +430,41 @@ def semid(broker):
...
+    oc_get_project = simple_command("oc get project -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
+    oc_get_pv = simple_command("oc get pv -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
+    oc_get_pvc = simple_command("oc get pvc -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
+    oc_get_rc = simple_command("oc get rc -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
+    oc_get_role = simple_command("oc get role -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
+    oc_get_rolebinding = simple_command("oc get rolebinding -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
+    oc_get_route = simple_command("oc get route -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
+    oc_get_service = simple_command("oc get service -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
+    oc_get_configmap = simple_command("oc get configmap -o yaml --all-namespaces", deps=[is_openshift_master], inherit_env=["KUBECONFIG"])
```
bfahr: @csams could you create an `oc_simple_command` derived from `simple_command` to encapsulate all of the common stuff here and just pass args for spec specific stuff?  Can we create a hierarchy like that in specs?

bfahr: This same question applies to commands like `ls` as well.

csams: I think so. We'd have to define it within the `DefaultSpecs` class so we can make sure they all have the same deps. I'll look into it.

csams:
Don't merge this PR yet please. I have some changes to make for deserialization and large command output.

esakaiev:
test this please

csams:
This PR was made irrelevant by https://github.com/RedHatInsights/insights-core/pull/1598.

PR #1456, opened 2018-09-26
Fixes an error when script executes insights run

bfahr:
* When a script executes insights.run directly and the -d option is used
then a exception will occur in _find_context when trying to determine
the context of a report function, when it assumes a class context.
* Adding a check for a class object fixes the error in
insights.formats.text._find_context
* duplicate by running `./sample_script.py -d`

### DIFF
--- a/insights/formats/text.py
+++ b/insights/formats/text.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 import six
 import sys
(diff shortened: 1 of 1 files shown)


YakovLittle:
test this please

YakovLittle:
test this please

PR #1459, opened 2018-09-27
Fix setting display name

gravitypriest:
Previously, by design, setting display name would immediately exit the client.

An unfortunate side effect of this is that, because the client treats all config (env, conf file, cmd line) identically, having display_name permanently set in the config file will always cause the client to exit.

This patch will check against the command line args to see whether display_name came from there or not. Then, it will only exit when setting display-name as a command line switch. Otherwise it will continue with collection and upload as normal.

In addition, the PUT to update display name is no longer made if the name is unchanged, and some messaging is improved.

### DIFF
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -411,7 +411,9 @@ class InsightsClient(object):
     @_net
     def set_display_name(self, display_name):
         '''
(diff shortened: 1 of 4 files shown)


brantleyr:
:+1:

PR #1464, opened 2018-10-02
RHEL 6 timeout has no -k option

gravitypriest:
uhhhhh whoops


insights/client/insights_spec.py
```diff
@@ -50,7 +50,7 @@ def get_output(self):
 
         # all commands should timeout after a long interval so the client does not hang
         # prepend native nix 'timeout' implementation
-        timeout_command = 'timeout -k 5 %s %s' % (
+        timeout_command = 'timeout -s KILL %s %s' % (
```
bfahr: @gravitypriest should it be `timeout -s KILL 5 %s %s`?

gravitypriest:
The 5 is an argument to -k, i.e., if it doesn't respond to the timeout
signal, send KILL after 5 seconds

bfahr:
On RHEL6 this is what I get:
```
$ timeout -s KILL sleep 10
timeout: invalid time interval `sleep'
Try `timeout --help' for more information.

$ timeout -s KILL 5 sleep 10
Killed
```

gravitypriest:
In this case the command being timed out would be `sleep 10`, which goes into the second `%s` in the command string.

Timeout interval goes into the first `%s`, which is separate from the original `-k 5`.

Thus the full thing would be `timeout -s KILL 5 sleep 10`, as you saw.

bfahr:
@gravitypriest makes sense to me, thanks for confirming.

PR #1465, opened 2018-10-05
Reformat output missing requirement

wushiqinlou:
Previously, the output of missing_requirement is mess when executing insights-run command with "-m" option. According to my investigation, there are types in the missing requirement, one is that requested file is not collected, another is that requested file is collected however the content is not correct. This PR tries to reformat the output and make it more clear for user.


wushiqinlou:
Due to issue https://github.com/RedHatInsights/insights-core/issues/1466, output may not be inappropriate.

PR #1467, opened 2018-10-05
Handle read timeouts on registration check

Glutexo:
Requests made during a registration check (done on running [`insights-client --support`](https://github.com/RedHatInsights/insights-core/blob/2079beb80e88ce4cfaabaa846e65285d19a0377c/insights/client/phase/v1.py#L102)) cause crash if read (not connection) timeout occurs. Fixed by catching these read timeouts the same way as other connection errors.

There are only unit tests, checking that the timeout errors get caught just as the connection errors do. It would be possible and plausible to make a functional test that would check the whole run. That would require mocking out the HTTP requests. I have begun working on that for #1418, but it has been postponed. With that done it will be possible to add functional tests for this PR too.


Glutexo:
Does @YakovLittle want to do a review, or is it more something for @gravitypriest or @kylape?

Glutexo:
Closed by accident. Reopening.

Glutexo:
This may be still relevant. Is it worth checking? @gravitypriest @subpop

insights/client/connection.py
```diff
@@ -245,7 +245,7 @@ def _test_urls(self, url, method):
                 else:
                     logger.info("Connection failed")
                     return False
-            except requests.ConnectionError as exc:
+            except (requests.Timeout, requests.ConnectionError) as exc:
```
subpop, line 248: Does this list-like syntax catch the exception if it's either a `requests.Timeout` or a `requests.ConnectionError`?

subpop:
I agree. These changes seem reasonable to me. The more error-handling the better.

gravitypriest (approved):

gravitypriest:
Looks good but I'll just comment that this will eventually get swallowed up in the connection.py changes once those are finished and tested (the new stuff also handles timeouts)

PR #1487, opened 2018-10-17
Split MD5CheckSum parser into prelink version and normal version

lonicerae:
- Fix md5chk_files spec. Don't check libfreeblpriv3.so, libsoftokn3.so and pem file as prelink has been deprecated on RHEL7
 - Check extra timezone files
 - Keep prelink_orig_md5 spec disabled as it needs extra work under current framework

### DIFF
--- a/insights/parsers/md5check.py
+++ b/insights/parsers/md5check.py
@@ -7,19 +7,20 @@ Module for processing output of the ``prelink -y --md5`` or ``md5sum`` command.
 The name and md5 checksums of specified files are stored as lists. Each
 entry will be processed by the shared parser.
 
-This MD5CheckSum class provide the following properties:
+This MD5CheckSum class provide the following attributes:
 
-* ``filename``: the file name which derived from the output.
-* ``md5sum``: the checksums for the file.
+* ``files``: the files which derived from the output.
+* ``checksums``: the checksums for the files.
 """
+
 from .. import parser, CommandParser
 from insights.specs import Specs
+from insights.parsers import ParseException
 
 
-@parser(Specs.prelink_orig_md5)
 class MD5CheckSum(CommandParser):
     """
-    Class to parse ``prelink -y --md5`` or ``md5sum`` command information.
+    Common class to parse ``prelink -y --md5`` or ``md5sum`` command information.
 
     The output of these two commands contains two fields, the first is the
     md5 checksum and the second is the file name.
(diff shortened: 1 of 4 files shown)


insights/parsers/md5check.py
```diff
@@ -7,61 +7,84 @@
...
     Sample input for ``prelink -y --md5`` or ``md5sum`` command::
 
         d1e6613cfb62d3f111db7bdda39ac821  /usr/lib64/libsoftokn3.so
 
-    Examples::
-
-        >>> collection = shared[MD5CheckSum]
-        >>> len(collection) # All MD5 checksums in a list
+    Examples:
+        >>> md5info = shared[MD5CheckSum]
```
xiangce: `shared` is no longer used by insights-core 3.0, please update the example test correspondingly. See https://github.com/RedHatInsights/insights-core/blob/7b8762c9a377c45f4b022ee7e0c64fccff72293d/insights/parsers/ps.py#L181

insights/parsers/md5check.py
```diff
@@ -7,61 +7,84 @@
...
+        >>> md5info['/usr/lib64/libsoftokn3.so']
         d1e6613cfb62d3f111db7bdda39ac821
 
+    Attributes:
+        data (dict): All MD5 checksums are stored in this dictionary with the
+                     filename as the key and the checksum as the value.
+        files (list): Return the all file names in the checksum results.
+        checksums (list): Return the checksums in the checksum results.
     """
-    @property
```
xiangce: It's not allowed to remove any `existing Interfaces` from insights-core 3.0 except for bugfixing. If they are no longer used, mark it as deprecated, see https://github.com/RedHatInsights/insights-core/blob/7b8762c9a377c45f4b022ee7e0c64fccff72293d/insights/parsers/systemd/config.py#L174

insights/parsers/md5check.py
```diff
@@ -7,61 +7,84 @@
...
+        >>> "d1e6613cfb62d3f111db7bdda39ac821" in md5info.checksums
+        True
+        >>> md5info['/usr/lib64/libsoftokn3.so']
         d1e6613cfb62d3f111db7bdda39ac821
 
+    Attributes:
+        data (dict): All MD5 checksums are stored in this dictionary with the
+                     filename as the key and the checksum as the value.
+        files (list): Return the all file names in the checksum results.
+        checksums (list): Return the checksums in the checksum results.
```
xiangce: Is this list must be used? it's a list contains MD5 checksums only, but without the mapping file.

How about adding a method, e.g. `get_checksums(file_name)`?

insights/parsers/md5check.py
```diff
@@ -7,61 +7,84 @@
...
-        >>> record.md5sum
+        >>> "/usr/lib64/libsoftokn3.so" in md5info.files
+        True
+        >>> "d1e6613cfb62d3f111db7bdda39ac821" in md5info.checksums
+        True
+        >>> md5info['/usr/lib64/libsoftokn3.so']
         d1e6613cfb62d3f111db7bdda39ac821
 
+    Attributes:
+        data (dict): All MD5 checksums are stored in this dictionary with the
```
xiangce: In the case that `data` is a simple `dict`, it's good to mix the `LegacyItemAccess`. See https://github.com/RedHatInsights/insights-core/blob/7b8762c9a377c45f4b022ee7e0c64fccff72293d/insights/parsers/systemctl_show.py#L45

And this method could be saved https://github.com/RedHatInsights/insights-core/pull/1487/files#diff-87e157f602bb2411a500e065eca62608R70

insights/parsers/md5check.py
```diff
@@ -7,61 +7,84 @@
...
 from .. import parser, CommandParser
 from insights.specs import Specs
+from insights.parsers import ParseException
 
 
-@parser(Specs.prelink_orig_md5)
 class MD5CheckSum(CommandParser):
     """
-    Class to parse ``prelink -y --md5`` or ``md5sum`` command information.
+    Common class to parse ``prelink -y --md5`` or ``md5sum`` command information.
```
xiangce: Should the `prelink -y --md5` be removed as well?

insights/parsers/md5check.py
lonicerae: Got it.

insights/parsers/md5check.py
lonicerae: Per our discussion we will remove it due to reconstruction and this property will be no longer usable.

insights/parsers/md5check.py
lonicerae: Removed unnecessary attribute.

insights/parsers/md5check.py
lonicerae: Yes, it's good.

insights/parsers/md5check.py
lonicerae: I would think currently not, because the `prelink` command is not usable due to framework's limitation and we indeed have plugin depends on this parser.

insights/parsers/md5check.py
xiangce: This change will break some tests of the insights-rules, please correct these rules based on this branch, then we can move forward to merge this PR.

Issue #1496, opened 2018-10-19
`list.sort()` doesn't work as expected in python3


xiangce:
~~~
telemetry/rules/tests/integration.py::test_integration[telemetry.rules.tests.webservers.test_httpd_mod_jk_not_work.integration_tests#test_hit_no_3-00000] Traceback (mo
st recent call last):
  File "/Users/liuxc/work/insights/insights-core/insights/core/dr.py", line 692, in run
    result = DELEGATES[component].process(broker)
  File "/Users/liuxc/work/insights/insights-core/insights/core/dr.py", line 501, in process
    return self.invoke(broker)
  File "/Users/liuxc/work/insights/insights-core/insights/core/dr.py", line 481, in invoke
    return self.component(*args)
  File "/Users/liuxc/work/insights/insights-core/insights/combiners/httpd_conf.py", line 131, in __init__
    config_files_data.sort()
TypeError: '<' not supported between instances of 'dict' and 'dict'
~~~

xiangce:
Should use `HttpdConfTree` instead.

csams:
@xiangce should this issue be closed?

xiangce:
See #1497

PR #1497, opened 2018-10-19
Deprecate HttpdConf and HttpdConfAll

xiangce:
For #1496

### DIFF
--- a/insights/combiners/httpd_conf.py
+++ b/insights/combiners/httpd_conf.py
@@ -29,6 +29,9 @@ from insights.util import deprecated
 @combiner(HttpdConf)
 class HttpdConfAll(object):
     """
+    .. note::
+        This combiner class is deprecated, please use :py:class:`HttpdConfTree` instead.
+
     A combiner for parsing all httpd configurations. It parses all sources and makes a composition
     to store actual loaded values of the settings as well as information about parsed configuration
     files and raw values.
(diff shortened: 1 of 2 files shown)


insights/combiners/httpd_conf.py
```diff
@@ -105,6 +57,8 @@ class HttpdConfAll(object):
     ConfigData = namedtuple('ConfigData', ['file_name', 'file_path', 'full_data_dict'])
 
     def __init__(self, httpd_conf):
+        deprecated(HttpdConfAll, "Import HttpdConfTree instead.")
```
bfahr: @xiangce since there is no way to see this deprecation in the documentation, would you mind to also add a note to the class docstring that this is deprecated?

insights/parsers/httpd_conf.py
```diff
@@ -109,6 +110,7 @@ class HttpdConf(LegacyItemAccess, Parser):
     """
 
     def __init__(self, *args, **kwargs):
+        deprecated(HttpdConf, "Import HttpdConfTree from insights.combiner.httpd_conf instead.")
```
bfahr: Same comment as above, there is no way to know this is deprecated until you look at the code.

insights/combiners/httpd_conf.py
xiangce: Thanks, @bfahr, I added the note in the latest commit.

PR #1499, opened 2018-10-19
WIP Lh callouts

lhuett:
New datasource for call-outs to remote web resources

### DIFF
--- a/insights/core/plugins.py
+++ b/insights/core/plugins.py
@@ -79,6 +79,11 @@ class metadata(parser):
 
 
 class combiner(dr.ComponentType):
+    """ ComponentType for a component for remote web resources. """
+    pass
+
+
+class remoteresource(dr.ComponentType):
     """ ComponentType for a component that composes other components. """
     pass
 
--- a/insights/core/remote_resource.py
+++ b/insights/core/remote_resource.py
@@ -1,0 +1,70 @@
+import requests
+import requests_cache
+
+
+class RemoteResource(object):
+    """
+    RemoteResource class for accessing external Web resources.
+
+    Functions:
+        get(): Calls `get()` function of the RemoteRersource superclass and returns the response payload.
+
+    Each line is parsed into a dictionary with the following keys:
+
+    Examples:
+        >>> from insights.core.remote_resource import RemoteResource
+        >>> rr = RemoteResource()
+        >>> rtn = rr.get("http://google.com")
+        >>> print (rtn.content)
+    """
+
+    timeout = 10
+
+    @classmethod
+    def get(cls, url, params={}, headers={}):
+        """
+        Returns the response payload from the request to the given URL.
+        Args:
+            url (str): The URL for the WEB API that the request is being made too.
+            params (dict): Dictionary containing the query string parameters
+            headers (dict): Dictionary containing anny HTTP headers needed.
+        Returns:
+            InstalledRpm: Installed RPM with highest version
+        """
+
+        return requests.get(url, params=params, headers=headers, timeout=cls.timeout, verify=False)
+
+
+class CachedRemoteResource(RemoteResource):
+    """
+    RemoteResource subclass for accessing external Web resources using caching.
+
+    Functions:
+        get(): Calls `set_cache()` to initialize the cache then calls the `get()` function of the
+        RemoteResource superclass and returns the response payload
+        set_cache(): Called by the `get()` function to initialize the cache for the Web request
+
+    Examples:
+        >>> from insights.core.remote_resource import CachedRemoteResource
+        >>> crr = CachedRemoteResource()
+        >>> rtn = crr.get("http://google.com")
+        >>> print (rtn.content)
+
+    """
+
+    expire_after = 180
+    old_data_on_error = True
+    __backend = "memory"
+
+    def __init__(self):
+        """
+        Initializes the cache for the web requests and supplies a `get()` function which calls the `get()` function in
+        the RemoteResource superclass
+
+        """
+
+        try:
+            requests_cache.install_cache('core', old_data_on_error=self.old_data_on_error,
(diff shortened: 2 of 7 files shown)


insights/core/remote_resource.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class RemoteResource(object):
+    """
+    RemoteResource class for accessing external Web resources.
+
+    Functions:
+        get(): Calls `get()` function of the RemoteRersource superclass and returns the response payload.
+
+    Each line is parsed into a dictionary with the following keys:
```
bfahr: @lhuett did you mean to add some doc here?

setup.py
```diff
@@ -23,6 +23,7 @@
 runtime = set([
     'pyyaml>=3.10,<=3.13',
     'six',
+    'requests_cache',
```
bfahr: Since this will pull in `requests` and we have a specific version requirement in the `develop` target, you should probably move that into `runtime`.

insights/core/remote_resource.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class CachedRemoteResource(RemoteResource):
+    """
+    RemoteResource subclass for accessing external Web resources using caching.
+
+    Functions:
+        get(): Calls `set_cache()` to initialize the cache then calls the `get()` function of the
+        RemoteResource superclass and returns the response payload
+        set_cache(): Called by the `get()` function to initialize the cache for the Web request
```
bfahr: Normally these would just be documented as docstrings in each method. If you want to have them here as well, your indentation needs to be fixed so that the methods don't all wrap as one line.

insights/core/remote_resource.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Examples:
+        >>> from insights.core.remote_resource import RemoteResource
+        >>> rr = RemoteResource()
+        >>> rtn = rr.get("http://google.com")
+        >>> print (rtn.content)
+    """
+
+    timeout = 10
+
```
bfahr: This class attribute needs a docstring, or needs to be documented in the class docstring.

insights/core/remote_resource.py
```diff
@@ -0,0 +1,70 @@
...
+    @classmethod
+    def get(cls, url, params={}, headers={}):
+        """
+        Returns the response payload from the request to the given URL.
+        Args:
```
bfahr: Need a space above this to correctly render.  Same with `Returns` below.

insights/core/remote_resource.py
```diff
@@ -0,0 +1,70 @@
...
+        >>> crr = CachedRemoteResource()
+        >>> rtn = crr.get("http://google.com")
+        >>> print (rtn.content)
+
+    """
+
+    expire_after = 180
+    old_data_on_error = True
+    __backend = "memory"
+
```
bfahr: Class attributes need docstrings or description included in the class docstring.

insights/core/remote_resource.py
```diff
@@ -0,0 +1,70 @@
...
+        Initializes the cache for the web requests and supplies a `get()` function which calls the `get()` function in
+        the RemoteResource superclass
+
+        """
+
+        try:
+            requests_cache.install_cache('core', old_data_on_error=self.old_data_on_error,
+                                         backend=self.__backend, expire_after=self.expire_after)
+        except Exception as ex:
+            raise Exception("Error initializing requests_cache", ex)
```
bfahr: You should probably include a `Raises:` section in your class docstring since you are explicitly raising this exception.

insights/tests/mock_web_server.py
```diff
@@ -0,0 +1,100 @@
...
+    Examples:
+
+        >>> from insights.tests.mock_web_server import TestMockServer, RequestHandler
+
+        >>> crr = CachedRemoteResource()
+        >>> sv = TestMockServer()
+        >>> sv.setup_class()
+        >>> sv.get_server_port()
+        35917
+        >>> curl -X GET http://localhost:40713/mock/
```
bfahr: Should this curl command be going to port 35917?

lhuett: Yep, corrected

insights/core/remote_resource.py
lhuett: Cleaned this up

setup.py
lhuett: Moved the 'requests==2.13.0' def to runtime

insights/core/remote_resource.py
lhuett: Removed "Functions:" from class def

insights/core/remote_resource.py
lhuett: Added

insights/core/remote_resource.py
lhuett: Added space

insights/core/remote_resource.py
lhuett: Added

insights/core/remote_resource.py
lhuett: Added for all but __backend

insights/core/remote_resource.py
csams: Doesn't the cache name need to be configurable as well? And maybe use the full name of the current class if a cache name isn't specified? It seems like subclasses will override the cache policies of other subclasses otherwise.

insights/core/plugins.py
```diff
@@ -79,6 +79,11 @@ class metadata(parser):
 
 
 class combiner(dr.ComponentType):
+    """ ComponentType for a component for remote web resources. """
```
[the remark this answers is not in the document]
lhuett: Well that is a bonehead mistake, thanks for catching that.

PR #1500, opened 2018-10-22
improve parsing for number config vars

gravitypriest:
Discovered an issue where ENV-specified http_timeout wasn't parsed as a float and caused exceptions where it was being used during requests. Made the parsing a little more robust.

### DIFF
--- a/insights/client/config.py
+++ b/insights/client/config.py
@@ -462,10 +462,7 @@ class InsightsConfig(object):
             if k in insights_env_opts:
                 v = insights_env_opts[k]
                 try:
-                    if k == 'http_timeout':
-                        insights_env_opts[k] = float(v)
-                    else:
-                        insights_env_opts[k] = int(v)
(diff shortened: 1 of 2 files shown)


insights/client/config.py
```diff
@@ -435,6 +435,18 @@ def _boolify(v):
...
                                  k.upper() not in ignore)
+
+        for k in ['retries', 'cmd_timeout', 'http_timeout']:
+            if k in insights_env_opts:
+                v = insights_env_opts[k]
+                try:
+                    if k == 'http_timeout':
+                        insights_env_opts[k] = float(v)
+                    else:
+                        insights_env_opts[k] = int(v)
```
kylape: What do you think of  `insights_env_opts[k] = float(v) if k == 'http_timeout' else int(v)`?

gravitypriest: much cleaner :+1:

PR #1503, opened 2018-10-23
passenger-status

TripleJQK:


insights/parsers/passenger_status.py
```diff
@@ -0,0 +1,161 @@
...
+    Examples:
+        >>> passenger_status.data["Version"]
+        '4.0.18'
+        >>> 'rack_default' in passenger_status.data
+        True
+        >>> len(passenger_status.data['foreman_default']['p_list'])
+        3
+
+    Raises:
+        ParseException: When there is no usefull data or the input content is
```
xiangce: nitpick: a typo? useful

insights/parsers/passenger_status.py
```diff
@@ -0,0 +1,161 @@
...
+        empty,  or does contain the header line.
```
xiangce: And a tab indent should be added in the second line to display it smooth in html docstring.

PR #1521, opened 2018-11-05
new parser for QpidStatG

TripleJQK:


insights/parsers/qpid_stat.py
```diff
@@ -193,3 +195,123 @@ def search(self, **kwargs):
...
+        >>> enqueues = qpid_stat_g.search(Statistic__contains='enqueues')  # Keyword search
+        >>> type(enqueues) == type([])
+        True
+        >>> len(enqueues)
+        4
+        >>> enqueues[0] == qpid_stat_g.data[2]  # List contains matching items
+        True
+    """
+
+    def __iter__(self):
```
xiangce: There is some duplicate code in the current there parsers which are a bit similar. Could you please to make a base class for them to save the duplicate codes?

insights/parsers/qpid_stat.py
```diff
@@ -193,3 +195,123 @@ def search(self, **kwargs):
...
+        return iter(self.data)
+
+    def parse_content(self, content):
+        index = 0
```
xiangce: A much simple way to get the index could be as below, And I think an `Exception` should be raised when no such line is found.
```
index = [i for i, l in enumerate(content) if l.startswith("Aggregate Broker Statistics")]
if not index:
   raise ParseException("Incorrect Content..BLABLA....")
index = index[0]
```

insights/parsers/qpid_stat.py
```diff
@@ -71,39 +106,21 @@ class QpidStatQ(CommandParser):
...
 
-    def __iter__(self):
-        return iter(self.data)
-
     def parse_content(self, content):
+        content = super(QpidStatQ, self).parse_content(content)
+
         self.data = parse_fixed_table(
-            [line for line in content if '========' not in line],
+            [line for line in content],
```
bfahr: @TripleJQK  `content` is already a list here so no need to create the new list.

insights/parsers/qpid_stat.py
```diff
@@ -71,39 +106,21 @@ class QpidStatQ(CommandParser):
...
             heading_ignore=['queue']
         )
         self.by_queue = dict(
             (q['queue'], q)
             for q in self.data
         )
```
bfahr: For this class and `QpidStatU` I think you could just pull all of `parse_content` into the base class and just use `pass` for these two classes.  Then just completely overload `parse_content` in `QpidStatG`.

xiangce: @bfahr - we have made changes as per your suggestions, please check it again. Thanks.

PR #1522, opened 2018-11-05
Callouts datasource using cachecontrol

lhuett:


insights/core/remote_resource.py
```diff
@@ -0,0 +1,142 @@
...
+    def get(self, url, params={}, headers={}, auth=(), verify=False):
+        """
+        Returns the response payload from the request to the given URL.
+
+        Args:
+            url (str): The URL for the WEB API that the request is being made too.
+            params (dict): Dictionary containing the query string parameters.
+            headers (dict): HTTP Headers that may be needed forthe request.
+            auth (tuple): User ID and password for Basic Auth
+            verify (str/bool): Value must be path to the Cert Bundle if verifying the SSL certificate
```
bfahr: @lhuett would it be better to use `certificate_path` instead of `verify` here and just have it be either `None` or a `str` value?

insights/core/remote_resource.py
```diff
@@ -0,0 +1,142 @@
...
+        self.session = session or requests.Session()
+
+    def get(self, url, params={}, headers={}, auth=(), verify=False):
+        """
+        Returns the response payload from the request to the given URL.
+
+        Args:
+            url (str): The URL for the WEB API that the request is being made too.
+            params (dict): Dictionary containing the query string parameters.
+            headers (dict): HTTP Headers that may be needed forthe request.
```
bfahr: spelling: `for the`

insights/core/remote_resource.py
```diff
@@ -0,0 +1,142 @@
...
+            auth (tuple): User ID and password for Basic Auth
+            verify (str/bool): Value must be path to the Cert Bundle if verifying the SSL certificate
+             or boolean False to ignore verifying the SSL certificate.
+
+        Returns:
+            Response:(Response): Response object from requests.get api request
```
bfahr: Is the returned value a `requests` response object?  For some reason in the docs this is linking to an internal Response object when it renders.  This applies to all of the Response returns.  Normally this would be in the format `type: description` for return values.

insights/core/remote_resource.py
```diff
@@ -0,0 +1,142 @@
...
+                            timeout=self.timeout)
+
+
+class CachedRemoteResource(RemoteResource):
+    """
+    RemoteResource subclass that sets up caching for subsequent Web resource requests.
+
+    Attributes:
+        expire_after (float): Amount of time in seconds that the cache will expire
+        backend (str): Type of storage for cache `DictCache1, `FileCache" or `RedisCache`
```
bfahr: Looks like you're mixing ticks and quotes here.

insights/core/remote_resource.py
```diff
@@ -0,0 +1,142 @@
...
+        redis_host (str): Hpstname of redis instance if `RedisCache` backend is specified
```
bfahr: spelling: `Hostname`

insights/core/remote_resource.py
```diff
@@ -0,0 +1,142 @@
...
+        redis_port (int): Port used to contact the redis instance if `RedisCache` backend is specified
+        file_cache_path (string): Path to where file cache will be stored if `FileCache` backend is specified
+        session (object): Requests session object
```
bfahr: Is this still an attribute?

insights/core/remote_resource.py
lhuett: Corrected

insights/core/remote_resource.py
lhuett: Corrected

insights/core/remote_resource.py
lhuett: Corrected

insights/core/remote_resource.py
lhuett: Removed. Made the change in the code review but forgot to remove from the doc

PR #1526, opened 2018-11-07
proxy string check fix

gravitypriest:
convert the whole function to unicode while we're here


gravitypriest:
resolves at least part of https://bugzilla.redhat.com/show_bug.cgi?id=1645491

PR #1531, opened 2018-11-08
Adding canonical_facts module

jhjaggars:
The canonical_facts module provides an easy way to gather identifying
information from the host or host archive.

This module will be used by both the client and the validation service to extract host identifying information from archives and hosts.


insights/specs/__init__.py
```diff
@@ -235,6 +238,7 @@ class Specs(SpecSet):
     lvs_noheadings = RegistryPoint()
     lvs_noheadings_all = RegistryPoint()
     lvs = RegistryPoint()
+    mac_addresses = RegistryPoint()
```
csams: This needs to have `multi_output=True` since it's a glob below.

bfahr:
@jhjaggars other than @csams comment, it looks good.

Issue #1518, opened 2018-10-31
Working without parsers


ritwik12:
We need to write some rules related to gluster and there aren't any parsers for that. Just a quick thought, if we can use insights using SOS report without parsers. Is there a way to do that?

csams:
What kind of work are you doing specifically? What kinds of clusters/sos reports?

ritwik12:
It's gluster not cluster. Sorry for the confusion.

ritwik12:
We want to parse gluster volume info from SOS reports. Is there a way to call those values without using parser?

csams:
Hey @ritwik12, if the files you're wanting aren't already included for sos reports in core, just create a file like this:
```python
# import whatever you need from the spec_factory
from insights.core.spec_factory import simple_file, glob_file, SpecSet

# remember that paths are relative to the common root of the sos report
class MySpecs(SpecSet):
    foo = simple_file("/path/to/foo/in/sosreport")
    bars = glob_file("/path/to/bars/in/sosreport/*")
    etc_stuff = glob_file("/etc/stuff/*")
```

You can use those specs directly in a rule like this:
```python
from insights import rule, make_fail
from myspecs import MySpecs

@rule(MySpecs.foo, MySpecs.bar, MySpecs.etc_stuff)
def report(foo, bars, etc):
    foo_content = foo.content  # lines of foo
    # do some stuff...

    # remember globs return lists of things they collect
    for bar in bars:
        bar_content = bar.content  # lines of a bar
        # do some other stuff...

    # more rule logic...

    return make_fail("BOOM", baz="baz")
```

With all of that said, what files and/or commands are you wanting to use? It might be easy to create a better level of abstraction for them than the raw content, and I can also show you how to write their specs so that your rule can run directly on the gluster node as well as against a sos report without having to change.

ritwik12:
Thanks a lot for all your help @csams. Sorry for the late reply, I was busy with festivals here :P. I guess the gluster work related to this is already in process after your help here. @chawlanikhil24 had already sent a Merge Request for the same internally. Is there anything else that we need help for gluster specifically? @chawlanikhil24

chawlanikhil24:
This issue can be closed now. :)

PR #1532, opened 2018-11-12
Added parser for extracting gluster vol info

chawlanikhil24:
To get the information of gluster volumes created or running on
the machines, this parser has been created , which reads the gluster_vol_info
file in sosreport and then structure it in a key-val pairs for further use.

This is created in reference to issue #1518

### DIFF
--- a/insights/specs/insights_archive.py
+++ b/insights/specs/insights_archive.py
@@ -9,6 +9,7 @@ glob_file = partial(glob_file, context=HostArchiveContext)
 
 class InsightsArchiveSpecs(Specs):
 
+    all_installed_rpms = glob_file("insights_commands/rpm_-qa*")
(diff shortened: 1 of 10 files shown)


insights/specs/insights_archive.py
```diff
@@ -59,11 +60,11 @@ class InsightsArchiveSpecs(Specs):
     grub1_config_perms = simple_file("insights_commands/ls_-l_.boot.grub.grub.conf")
     grub_config_perms = simple_file("insights_commands/ls_-l_.boot.grub2.grub.cfg")
     grubby_default_index = simple_file("insights_commands/grubby_--default-index")
+    gluster_v_info = simple_file("sos_commands/gluster/gluster_volume_info")
```
xiangce: For `insights_archive`, the simple_file should be `simple_file("insights_commands/gluster_volume_info")`

PR #1537, opened 2018-11-12
add new patterns section for better regex matching

gravitypriest:
todo: add unit tests


kylape:
Can you explain the ConfigParser limitations that make you reopen the file and parse the content manually?

kylape:
Looks like there are already unit tests that need to be updated for the newly added code

PR #1548, opened 2018-11-14
Add method to get count of processes with given string

smalleni:
This commit adds a method to the Ps Parser that returns the number
of occurencies of commands with a given string. This is very much needed
and useful when writing rules for OpenStack because we have some performance
considerations wehn using a large number of worker processes for services.
For example, the recommendation is to cap the woker count at
ming(processorcount/2, 12) to avoid issues with database conenctions etc, so
we need the parser to provide this data in order to write rules.

### DIFF
--- a/insights/parsers/ps.py
+++ b/insights/parsers/ps.py
@@ -130,6 +130,18 @@ class Ps(CommandParser):
         """
         return any(proc in row[self.command_name] for row in self.data)
 
+    def number_occurences(self, proc):
+        """
+        Returns the number of occurencies of commands that contain given text
+
+        Returns:
+            int: The number of occurencies of commands with given text
+
+        .. note:: 'proc' can match anywhere in the command path, name or
+           arguments.
+        """
+        return len([True for row in self.data if proc in row[self.command_name]])
+
     def search(self, **kwargs):
         """
         Search the process list for matching rows based on key-value pairs.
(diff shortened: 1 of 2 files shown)


insights/parsers/ps.py
```diff
@@ -130,6 +130,22 @@ def fuzzy_match(self, proc):
...
+
+        .. note:: 'proc' can match anywhere in the command path, name or
+           arguments.
+        """
+        proc_count = 0
+        for row in self.data:
+            if proc in row[self.command_name]:
+                proc_count += 1
+        return proc_count
+
```
bfahr: @smalleni you could accomplish what you need with the existing`search` method using something like this in your own code:
```
number_of_occurences = len(ps_parser_obj.search(COMMAND_NAME__contains='text'))
```
Would this be sufficient instead of writing a new method in the parser?

smalleni: @bfahr Thanks for the review.  I guess, so. Just figured it would be nice to have this because I plan on implementing a lot of OpenStack rules and we will be relying on this extensively.

bfahr: @smalleni in that case, I'd recommend a simplification of the function to something like this:
```python
return len([True for row in self.data if proc in row[self.command_name]])
```
Please keep in mind that the data is filtered for some `PS` parsers and will not include all of the commands present on the system.  The unfiltered `PS` parsers do not include command arguments so search results may not be accurate if you are looking for command args.

smalleni: @bfahr thanks for the suggestion, amending the PR.

PR #1549, opened 2018-11-14
Allow blacklist of patterns and keyword substitution. Refactor subprocessing and spec helpers.

csams:
This is part of getting core3 ready for client collection.

* Extend the blacklist capability to patterns and keyword substitution the current client does. Regexes patterns against lines in a file are not yet supported.
* New functions are added to insights.core.blacklist for adding patterns and keywords. Requires grep and sed to be installed on the target machine.
* Update the subproc module to simplify use.
* Allow specs to pass a list of environment variables to inherit from the parent process. Useful for inheriting kubernetes, openshift, etc.
* Introduces a new context for new style archives, although the collection implementation isn't in yet.

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -298,8 +298,7 @@ class CommandOutputProvider(ContentProvider):
         return command
 
     def create_env(self):
-        env = {}
-        env.update(SAFE_ENV)
+        env = dict(SAFE_ENV)
         for e in self.inherit_env:
             if e in os.environ:
                 env[e] = os.environ[e]
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -7,20 +7,21 @@ import traceback
 
 from collections import defaultdict
 from glob import glob
+from subprocess import call
 
 from insights.core import blacklist, dr
 from insights.core.filters import get_filters
 from insights.core.context import ExecutionContext, FSRoots, HostContext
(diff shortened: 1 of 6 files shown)


insights/core/spec_factory.py
```diff
@@ -233,20 +274,42 @@ def validate(self):
...
+            keywords = blacklist.get_disallowed_keywords()
+            if keywords:
+                sed = ["sed"]
+                for kw in keywords:
+                    sed.extend(["-e", "s/%s/keyword/g" % kw.replace("/", "\\/")])
+                command.append(sed)
+        return command
+
+    def create_env(self):
+        env = {}
```
jhjaggars:
```suggestion
        env = dict(SAVE_ENV)
```

insights/core/spec_factory.py
```diff
@@ -233,20 +274,42 @@ def validate(self):
...
+        env.update(SAFE_ENV)
```
jhjaggars:
```suggestion
```

insights/util/streams.py
```diff
@@ -49,7 +49,7 @@ def stream(command, stdin=None, env=os.environ, timeout=None):
 
     cmd = which(command[0])
     if cmd is None:
-        raise Exception("Couldn't execute: %s" % command)
+        raise Exception("Command not in PATH: %s" % command)
```
jhjaggars: Maybe ad the PATH to the message?

Issue #1473, opened 2018-10-09
Evolution of Rules to Share Code


csams:
## Problem
Rules with identical data dependencies and logic can’t share code because each rule can return only one response.

## Proposed Solution
Create a new `make_composite(results, exceptions)` response type where results is a list of `make_response`, `make_pass`, or `make_fail` instances and exceptions is a list of any exceptions raised during rule processing.

## Impact
* A rule that returns `make_composite` should accumulate results and exceptions in its internal loop.
* The rule component type will need to validate the new response type and register any exceptions from it with the broker.
* Integration test infrastructure will need to handle lists of responses in addition to the current single response. This might “just work”, but we need to check.
* The evaluators in `insights.core.evaluators` need to unpack the new `make_composite` and serialize the response list.
* The `HumanReadableFormat` will need to appropriately tally and display the different kinds of responses.
* The UI and reporting systems will need to handle multiple error keys from the same rule in the same json response. This might "just work".

miclark:
Closing due to inactivity.

PR #1550, opened 2018-11-14
Add make_composite to core/plugins.py

ritwik12:
I want to give a headstart to #1473 and I guess this will be a good way of implementing it in an abstract way and if it is fine then we can move to writing concrete code for this enhancement i.e `Evolution of Rules to Share Code`


bfahr:
Closing this PR due to inactivity.

PR #1557, opened 2018-11-22
Add parser avc_hash_stats

shzhou12:


insights/parsers/avc_hash_stats.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    def parse_content(self, content):
+        hash_stats_dict = {}
+        for line in get_active_lines(content):
+            key, value = map(lambda x: x.strip(), line.split(':'))
+            if key not in hash_stats_dict:
+                hash_stats_dict[key] = value
+        self.hash_stats_dict = hash_stats_dict
+        self.entries = int(hash_stats_dict['entries']) if hash_stats_dict.get('entries') else None
```
xiangce: In the condition check, better use `in` instead of `get()` which is a bit heavy than `in`
```
self.entries = int(data['entries']) if 'entries' in data else None
```

insights/parsers/avc_hash_stats.py
```diff
@@ -0,0 +1,56 @@
...
+        self.buckets_used, self.buckets = map(lambda x: int(x.strip()),
```
xiangce: the same to above.

insights/parsers/avc_hash_stats.py
```diff
@@ -0,0 +1,56 @@
...
+                                              hash_stats_dict['buckets used'].split('/')) \
+            if hash_stats_dict.get('buckets used') else [None, None]
+        self.longest_chain = int(hash_stats_dict['longest chain']) if hash_stats_dict.get('longest chain') else None
```
xiangce: the same

insights/parsers/avc_hash_stats.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Class ``AvcHashStats`` parses the content of the ``/sys/fs/selinux/avc/hash_stats``.
+
+    Attributes:
+        entries (int): It is used to show the count of avc hash entries.
+        buckets (int): It is used to show the total count of buckets.
+        buckets_used (int): It is used to show the count of used buckets.
+        longest_chain (int): It is used to show the longest chain.
+
+    A small sample of the content of this file looks like::
```
xiangce: nitpick: small -> typical?

lhuett:
This looks good as long the comments by as @xiangce have been addressed.

PR #1558, opened 2018-11-22
Add parser proc_stat

shzhou12:


insights/parsers/proc_stat.py
```diff
@@ -0,0 +1,113 @@
...
+            key, value = line.split(None, 1)
+            if key not in proc_stat_dict:
+                proc_stat_dict[key] = value
+        self.data = proc_stat_dict
+
+    @property
+    def cpu_percentage(self):
+        if 'cpu' in self.data:
+            cpu_list = list(map(lambda x: int(x), self.data['cpu'].split()))
+            cpu_total = float(sum(cpu_list[:8]))
```
xiangce: Is the number `8` correct for all cases, I mean better add a corresponding explanation (comment) for the number you choose (like `3` and `4` in below line.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/s2-proc-stat

insights/parsers/proc_stat.py
```diff
@@ -0,0 +1,113 @@
...
+            cpu_idle = float(sum([cpu_list[3], cpu_list[4]]))
+            cpu_pct = (cpu_total - cpu_idle) * 100 / cpu_total
+            return "{0:.2f}".format(cpu_pct) + '%'
+        return None
+
+    @property
+    def intr(self):
```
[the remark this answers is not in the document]
shzhou12: I think the total counts of interrupt is more commonly used, so I changed to return the total counts.

insights/parsers/proc_stat.py
```diff
@@ -0,0 +1,131 @@
...
+
+        if 'cpu' in self.data:
+            cpu_list = list(map(lambda x: int(x), self.data['cpu'].split()))
+            if len(cpu_list) < 7:
+                raise ParseException("Error: Invalid cpu length", self.data['cpu'])
+            cpu_total = float(sum(cpu_list))
+            cpu_idle = float(cpu_list[3])
+            cpu_pct = (cpu_total - cpu_idle) * 100 / cpu_total
+            return "{0:.2f}".format(cpu_pct) + '%'
+        return None
```
bfahr: @shzhou12 since the data necessary to compute these results are known at the time the data is parsed you could calculate these once in `parse_content` instead of each time this property is invoked.  Having the `ParseException` raised here doesn't really make sense anyway, it should be raised in `parse_content`.

insights/parsers/proc_stat.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @property
+    def intr_total(self):
+        return int(self.data['intr'].split(None, 1)[0]) if 'intr' in self.data else None
```
bfahr: Same comment as above applies to all of these properties.  You could compute them once and store them as attributes of the class.

insights/parsers/tests/test_proc_stat.py
```diff
@@ -0,0 +1,39 @@
...
+softirq 11867930 1 3501158 6 4705528 368244 0 79 2021509 0 1271405
+""".strip()
+
+
+def test_proc_stat():
+    proc_stat = ProcStat(context_wrap(PROC_STAT))
+    assert proc_stat.btime == '1542179825'
+    assert proc_stat.softirq_total == 11867930
+    assert proc_stat.cpu_percentage == '6.73%'
+
```
bfahr: Since you are computing each of the attributes in the class, you should have a test for each of the attributes to ensure that all of your code is tested.

PR #1571, opened 2018-11-29
Update connection.py

anazmy:
This modification provides two changes in `_test_openssl` function:

- Change the SSL chain test - with a proxy in the picture - to send the proxy a connect request to destination's port 443 `CONNECT {0}:443`. Some proxies - for example Zscaler - will forward traffic to port 80 if no port was defined, which breaks the SSL chain test.

-  Change the proxy response check to the `CONNECT` request to be case insensitive, as some proxies - for example Zscaler - respond with "Connection Established", note the capital "E" in established, which also breaks the SSL chain check.


lphiri:
While this will likely work for production  - it will potentially break in development environments when a port other than 443 is used for SSL. The bullet proof approach would be to check which port is defined in base_url  and only default to 443 if none is defined- its definitely not a blocker though.

PR #1572, opened 2018-11-29
Output HTML pytest reports to S3 bucket

SteveHNH:
Removes the need to have anonymous access to Jenkins. These reports will be available to read from S3 so that contributors can see what tests failed.

This is an initial pass and may be updated to be more clear as we go.


bfahr:
@SteveHNH are you still working on this?  I wasn't sure if you were waiting on a review.

bfahr:
@SteveHNH are you still working on this?

bfahr:
@SteveHNH is there still a need for this PR?

PR #1573, opened 2018-11-30
Parser to parse output of 'id -u nova' & 'id -u nova_migration'

psachin:

### DIFF
--- a/insights/parsers/nova_user_ids.py
+++ b/insights/parsers/nova_user_ids.py
@@ -15,14 +15,13 @@ NovaMigrationUID - command ``id -u nova_migration``
 ---------------------------------------------------
 """
 from insights import CommandParser, parser
-from insights.parsers import SkipException
+from insights.parsers import ParseException, SkipException
 from insights.specs import Specs
 
 
 @parser(Specs.nova_uid)
 class NovaUID(CommandParser):
     '''Parse output of ``id -u nova`` and get the ``uid`` (int).
-    Return ``None`` if nova user not found.
 
     Typical output of the ``id -u nova`` command is::
 
@@ -34,25 +33,27 @@ class NovaUID(CommandParser):
 
         >>> nova_uid.data
         162
-        >>> nova_user_not_found.data is None
-        True
 
     Attributes:
         data: ``int`` if nova user exist otherwise ``None``.
 
+    Raises:
+        SkipException: If output is empty.
+        ParseException: If nova user not found,
     '''
     def parse_content(self, content):
         self.data = None
         if not content:
-            raise SkipException("No such user.")
-        if len(content) == 1 and "no such user" not in content[0]:
+            raise SkipException()
+        if len(content) == 1:
+            if "no such user" in content[0]:
+                raise ParseException('No such user.')
             self.data = int(content[0])
 
 
 @parser(Specs.nova_migration_uid)
-class NovaMigrationUID(CommandParser):
+class NovaMigrationUID(NovaUID):
     '''Parse output of ``id -u nova_migration`` and get the ``uid`` (int).
-    Return ``None`` if nova_migration user not found.
 
     Typical output of the ``id -u nova_migration`` command is::
 
@@ -64,16 +65,13 @@ class NovaMigrationUID(CommandParser):
 
         >>> nova_migration_uid.data
         153
-        >>> nova_migration_user_not_found.data is None
-        True
 
     Attributes:
         data: ``int`` if nova_migration user exist otherwise ``None``.
 
+    Raises:
+        SkipException: If output is empty.
+        ParseException: If nova_migration user not found.
+
     '''
-    def parse_content(self, content):
-        self.data = None
-        if not content:
-            raise SkipException("No such user.")
-        if len(content) == 1 and "no such user" not in content[0]:
-            self.data = int(content[0])
+    pass
--- a/insights/parsers/nova_user_ids.py
+++ b/insights/parsers/nova_user_ids.py
@@ -15,7 +15,7 @@ NovaMigrationUID - command ``id -u nova_migration``
 ---------------------------------------------------
 """
 from insights import CommandParser, parser
-from insights.parsers import ParseException, SkipException
+from insights.parsers import SkipException
 from insights.specs import Specs
 
 
@@ -35,19 +35,16 @@ class NovaUID(CommandParser):
         162
 
     Attributes:
-        data: ``int`` if nova user exist otherwise ``None``.
+        data: ``int`` if 'nova user' exist otherwise ``None``.
 
     Raises:
-        SkipException: If output is empty.
-        ParseException: If nova user not found,
+        SkipException: If 'nova' user not found or output is empty.
     '''
     def parse_content(self, content):
         self.data = None
-        if not content:
+        if not content or len(content) == 1 and ("no such user" in content[0]):
             raise SkipException()
-        if len(content) == 1:
-            if "no such user" in content[0]:
-                raise ParseException('No such user.')
+        elif len(content) == 1 and ("no such user" not in content[0]):
             self.data = int(content[0])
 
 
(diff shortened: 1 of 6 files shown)


insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Attributes:
+        data: ``int`` if nova user exist otherwise ``None``.
+
+    '''
+    def parse_content(self, content):
+        self.data = None
+        if not content:
+            raise ParseException("Input content is empty.")
+        if len(content) == 1 and "no such user" not in content[0]:
```
xiangce: How about raising a `SkipException` when `no such user`?

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Attributes:
+        data: ``int`` if nova_migration user exist otherwise ``None``.
+
+    '''
+    def parse_content(self, content):
+        self.data = None
+        if not content:
+            raise ParseException("Input content is empty.")
+        if len(content) == 1 and "no such user" not in content[0]:
```
xiangce: How about raising a `SkipException` when `no such user`, to avoid generating a parser object which you have to check if it's None in rules?

psachin: @xiangce As of now I don't see much difference in the functionality of `ParseException` and `SkipException`. However yous suggestion is valid and `SkipException` is more suitable in the current scenario. I'll implement the changes.

bfahr:
@psachin this parser requires a new spec review.

insights/parsers/nova_user_ids.py
psachin: Done

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,79 @@
...
+``CommandParser`` to get uids of the user ``nova`` and
+``nova_migration``.
+
+Parsers included in this module are:
+
+NovaUID - command ``id -u nova``
+--------------------------------
+
+NovaMigrationUID - command ``id -u nova_migration``
+----------------------------------------------------
```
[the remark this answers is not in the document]
psachin: Done

psachin:
New specs proposal: **INSPEC-129**

insights/parsers/nova_user_ids.py
bfahr: @psachin `SkipException` is appropriate when the parser recognizes that there is no data to process.  `ParseException` should be raised when an error occurs in the data that is unexpected.  `ParseException` is logged and should be investigated for errors in insights-core or data issues.  `SkipException` is ignored by the engine.

If it is important to your rule to know that there is no `nova` user on the system then you should write a separate parser that looks explicitly for that condition.  

In this case I agree with @xiangce that you could include the check for "no such user" as part of your logic to raise the `SkipException`, for example:

```
if not content or (len(content) == 1 and "no such user" in content[0]):
    raise SkipException()
```
Please apply this to the other parser as well.

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,79 @@
...
+        True
+
+    Attributes:
+        data: ``int`` if nova user exist otherwise ``None``.
+
+    '''
+    def parse_content(self, content):
+        self.data = None
+        if not content:
+            raise SkipException("No such user.")
```
xiangce: @psachin - I do not think this logic (empty output means "No such user") is correct.

You could `Skip` the `empty`, but you shouldn't state it‘s `No such user`.

psachin:
@bfahr @xiangce 

I think 'no user found' is useful enough and should not be skipped where as an empty output can be safely skipped. Based on this I've used `ParseException` to handle 'no user found' & `SkipException` for empty output. 

Thanks you both for providing suggestions on this parser.

psachin:
As discussed there will be large number of non openstack environments with no `nova` or `nova_migration` user. It is safe to skip this warning('no such user') instead of logging it using `ParseException`. Hence using `SkipException()` for both empty output and 'no such user' warning.

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Attributes:
+        data: ``int`` if 'nova user' exist otherwise ``None``.
+
+    Raises:
+        SkipException: If 'nova' user not found or output is empty.
+    '''
+    def parse_content(self, content):
+        self.data = None
+        if not content or len(content) == 1 and ("no such user" in content[0]):
```
xiangce: I think the brackets should surround both the `len(content) == 1` and  the `"no such user" in content[0]`, but not the `"no such user" in content[0]` only:
```python
if not content or (len(content) == 1 and "no such user" in content[0])
```

In my opinion, how about
```python
if not content or "no such user" in content[0]):
  raise SkipException()
elif not content[0].isdight() or len(content) > 1:
  raise ParseException('Incorrect output: {}', content)
self.data = int(content[0])
```
which checks multiple lines (it's not anticipated, right?) as well

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,72 @@
...
+            raise SkipException()
+        elif len(content) == 1 and ("no such user" not in content[0]):
```
xiangce: After above safe checks, this line can be saved. The `content` is good enough here, I think, it can be parsed directly.

psachin: I agree the multiple lines will never happen and that is an overcheck but it won't make any difference.

insights/parsers/nova_user_ids.py
psachin: Done

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Raises:
+        SkipException: If 'nova' user not found or output is empty.
+    '''
+    def parse_content(self, content):
+        self.data = None
+        if not content or (len(content) == 1 and "no such user" in content[0]):
+            raise SkipException()
+        elif len(content) == 1 and ("no such user" not in content[0]):
+            self.data = int(content[0])
```
bfahr: Perhaps it would be useful to have an `else` case that raises a `ParserException` here.  It would avoid a situation where the `id` command changes to have multiple lines of output and `self.data` is `None`. For example `len(content) > 1`.  `ParseException` would be appropriate for the unexpected condition where you expect only 1 line of output and you get multiple lines of output.

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,72 @@
...
+
+    However the id number may vary.
+
+    Examples:
+
+        >>> nova_migration_uid.data
+        153
+
+    Attributes:
+        data: ``int`` if 'nova_migration' user exist otherwise ``None``.
```
bfahr: Same comment as above. It should not be `None`.

insights/parsers/nova_user_ids.py
psachin: @bfahr I included `ParseException` in case of multi-line output. I think this is a bit of over testing at the moment(unless `id` command actually has multi-line output).

Let me know if the changes look ok.

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,72 @@
...
+
+    However the id number may vary.
+
+    Examples:
+
+        >>> nova_uid.data
+        162
+
+    Attributes:
+        data: ``int`` if 'nova user' exist otherwise ``None``.
```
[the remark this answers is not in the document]
psachin: It was `None` in case of multi-line output but now will return an exception. Thanks.

insights/parsers/nova_user_ids.py
```diff
@@ -0,0 +1,75 @@
...
+        SkipException: If 'nova' user not found or output is empty.
+        ParseException: If multi-line output. Output of such kind is not yet expected from the command `id`.
+    '''
+    def parse_content(self, content):
+        if not content or (len(content) == 1 and "no such user" in content[0]):
+            raise SkipException()
+        elif len(content) == 1 and ("no such user" not in content[0]) and content[0].isdigit():
+            self.data = int(content[0])
+        elif len(content) > 1:
+            raise ParseException("Unable to parse user ID")
```
bfahr: @psachin I believe this should just be an `else:`.  If len(content) == 1 and content[0] is not a number that should still be a ParseException shouldn't it?  You shouldn't get a uid value that is not a number should you?

psachin: @bfahr  Yes. For the last one the o/p should not be a number(hopefully). I agree with your suggestion and better raise an exception in `else`. Thanks

PR #1581, opened 2018-12-04
Add --diagnosis option

gravitypriest:
Replacement for --to-json for platform implementation

To meet the spec, replace optparse with argparse in config.py but maintain behavior


kylape:
Why did optparse need to be replaced in order to meet the spec?

gravitypriest:
FMS wanted the --diagnosis option to be able to be specified with or
without a value, which isn't possible with optparse.

PR #1586, opened 2018-12-05
Add component, type, and key to rule results in evaluators' json response.

csams:
This branch adds to each rule response the rule's full name (package and function name) as "component", the response type (rule, pass, fingerprint, etc.) as "type", and the first argument of `make_*` as "key".

Old doc:
```json
  "reports": [
    {
      "rule_id": "stand_alone|TOO_MANY_HOSTS",
      "details": {
        "num": 2,
        "type": "rule",
        "error_key": "TOO_MANY_HOSTS"
      }
    }
  ]
```

New doc:
```json
  "reports": [
    {
      "type": "rule",
      "component": "examples.rules.stand_alone.report",
      "rule_id": "stand_alone|TOO_MANY_HOSTS",
      "key": "TOO_MANY_HOSTS",
      "details": {
        "num": 2,
        "type": "rule",
        "error_key": "TOO_MANY_HOSTS"
      }
    }
  ]
```

The fully qualified name lowers the chance of collision between rules (especially across projects) and when combined with "key" is a better response identifier in other systems that store hits or content.

This change lifts metadata about the response out of the details section but leaves the existing entries in place for backward compatibility.


bfahr:
@csams this looks good to me.

PR #1589, opened 2018-12-06
Network namespace list

vishwanathjadhav:
Added parser for specs `/bin/ls /var/run/netns`


insights/parsers/net_namespace.py
```diff
@@ -0,0 +1,50 @@
...
+        self.netns_list = []
+        for line in get_active_lines(content):
+            netns = line.split(" ")
+            if netns and '' in netns:
+                self.netns_list = [x for x in netns if x]
+            elif netns:
+                self.netns_list.append(netns[0])
+
+    @property
+    def get_netns(self):
```
xiangce: Seems this `get_netns` is redundant. We can get the list via `netns_list` directly.

Better remove the `get_netns` and add corresponding `Attributes` docstring for `netnt_list`

insights/parsers/net_namespace.py
```diff
@@ -0,0 +1,50 @@
...
+from insights import Parser, parser, get_active_lines
+from insights.specs import Specs
+
+
+@parser(Specs.namespace)
+class NetworkNamespace(Parser):
+    def parse_content(self, content):
+        self.netns_list = []
+        for line in get_active_lines(content):
+            netns = line.split(" ")
```
xiangce: Better use `split()` instead of `split(" ")`

insights/parsers/net_namespace.py
```diff
@@ -0,0 +1,50 @@
...
+            if netns and '' in netns:
```
xiangce: Why check `'' in netns` here? By using "split()", this can be avoided.

insights/parsers/net_namespace.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+from insights import Parser, parser, get_active_lines
+from insights.specs import Specs
+
+
+@parser(Specs.namespace)
+class NetworkNamespace(Parser):
+    def parse_content(self, content):
+        self.netns_list = []
```
xiangce: Better raise a `SkipException` when the content is `empty`

insights/parsers/net_namespace.py
vishwanathjadhav: Done

insights/parsers/net_namespace.py
vishwanathjadhav: Done

insights/parsers/net_namespace.py
vishwanathjadhav: Done

insights/parsers/net_namespace.py
vishwanathjadhav: Done

insights/parsers/net_namespace.py
```diff
@@ -0,0 +1,54 @@
...
+@parser(Specs.namespace)
+class NetworkNamespace(Parser):
+    def parse_content(self, content):
+        self._netns_list = []
+        if not content:
+            raise SkipException('Nothing to parse.')
+
+        for line in get_active_lines(content):
+            netns = line.split()
+            if netns and len(netns) > 1:
```
xiangce: Why divide the checks to these two situations( which are overlapped with each other) ?

I think `self._netns_list.extend(line.split())` is enough.

vishwanathjadhav: Done

PR #1596, opened 2018-12-12
parser for ssh client config files

TripleJQK:


insights/parsers/ssh_client_config.py
```diff
@@ -0,0 +1,147 @@
...
+            ForwardX11Trusted yes
+        # Send locale-related environment variables
+            SendEnv LANG LC_CTYPE LC_NUMERIC LC_TIME LC_COLLATE LC_MONETARY LC_MESSAGES
+            SendEnv LC_PAPER LC_NAME LC_ADDRESS LC_TELEPHONE LC_MEASUREMENT
+            SendEnv LC_IDENTIFICATION LC_ALL LANGUAGE
+            SendEnv XMODIFIERS
+
+        Host proxytest
+            HostName 192.168.122.2
+
```
bfahr: @TripleJQK could you add the two attributes to the docstring for these two classes?  I think it would help with the understanding.

insights/parsers/ssh_client_config.py
```diff
@@ -0,0 +1,147 @@
...
+class SshClientConfig(Parser):
+    """Base class for sshclient config file."""
+
+    def __init__(self, context):
+        self.global_lines = []
+        self.host_lines = {}
+        super(SshClientConfig, self).__init__(context)
+
+    def _split_content(self, content):
+        index = [i for i, l in enumerate(get_active_lines(content)) if l.startswith('Host ')]
```
xiangce: Every time calling the `get_active_lines`, it will iterate the whole `content` and do the same thing to each line. To get a good efficiency, better assign its result to a local variable just before the first calling.

insights/parsers/ssh_client_config.py
```diff
@@ -0,0 +1,147 @@
...
+        if not index:
+            self._global_content = get_active_lines(content)
+            self._host_content = []
+        else:
+            if index:
```
xiangce: I believe this `if index` can be removed

insights/parsers/ssh_client_config.py
```diff
@@ -0,0 +1,147 @@
...
+                self._global_content = get_active_lines(content)[0:index[0]]
+                self._host_content = get_active_lines(content)[index[0]:]
+
+    KeyValue = namedtuple('KeyValue', ['keyword', 'value', 'line'])
+
+    def parse_content(self, content):
+        self._split_content(content)
+        for line in self._global_content:
```
xiangce: Just a suggestion to save all the private attributes (`self._*`):
```python
index_list = [i for i, l in enumerate(content) if l.startswith('Host ')]
index = index_list[0] if index_list else len(content) 
for line in content[:index] <-- replace the self._global_content
  ...
for line in content[index:] <-- replace the self._host_content
  ...
```

PR #1599, opened 2018-12-17
Ifcfg static route

ghost:


xiangce (commented):
Would you please to move this parser to [sysconfig](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/sysconfig.py)? Since they have the same content format and under the `/etc/sysconfig/`

rasrivas-redhat:
@xiangce 
I have updated the code in the latest PR as suggested by you, kindly review it when you are free

insights/parsers/sysconfig.py
```diff
@@ -506,3 +509,52 @@ class VirtWhoSysconfig(SysconfigOptions):
         'A TEST'
     """
     pass
+
+
+@parser(Specs.ifcfg_static_route)
+class IfCFGStaticRoute(LegacyItemAccess, Parser):
```
xiangce: By inheriting it from `SysconfigOptions`, you do not implement any further operations to this Parser.

insights/parsers/sysconfig.py
```diff
@@ -506,3 +509,52 @@ class VirtWhoSysconfig(SysconfigOptions):
...
+        >>> sorted(conn_info.data.items())
+        [('ADDRESS0', '10.65.223.0'), ('GATEWAY0', '10.65.223.1'), ('NETMASK0', '255.255.254.0')]
+
+    Properties:
+        static_route (str): static route file name derived from file path
+        data (dict): all static route connection details
+    """
+    def __init__(self, context):
+        self.data = {}
+        self.static_route = context.path.split("network-scripts/route-", 1)[1]
```
xiangce: After changing its superclass to `SysconfigOptions`, the `__init__` and `parse_content` can be removed.

And the `self.static_route` could be:
```python
    @property
    def static_route(self):
        """(str): static route file name derived from file path"""
        return self.file_name.split("-", 1)[1]
```

insights/parsers/sysconfig.py
```diff
@@ -506,3 +509,52 @@ class VirtWhoSysconfig(SysconfigOptions):
...
+
+        >>> conn_info.static_route
+        'test-net'
+        >>> conn_info.data['ADDRESS0']
+        '10.65.223.0'
+        >>> sorted(conn_info.data.items())
+        [('ADDRESS0', '10.65.223.0'), ('GATEWAY0', '10.65.223.1'), ('NETMASK0', '255.255.254.0')]
+
+    Properties:
+        static_route (str): static route file name derived from file path
```
xiangce: And then this `Properties` section can be removed.

xiangce (changes requested):

insights/parsers/sysconfig.py
```diff
@@ -506,3 +509,52 @@ class VirtWhoSysconfig(SysconfigOptions):
...
+
+        ADDRESS0=10.65.223.0
+        NETMASK0=255.255.254.0
+        GATEWAY0=10.65.223.1
+
+    Examples:
+
+        >>> conn_info.static_route
+        'test-net'
+        >>> conn_info.data['ADDRESS0']
```
xiangce: Since the base class has been mixed with `LegacyItemAccess` which provides feature to access the items via the obj directly, better avoid to use `data` in the examples.

rasrivas-redhat:
@xiangce thank you for the feedback, I will update the changes as requested.

insights/parsers/sysconfig.py
rasrivas-redhat: acknowledged, updated the code in the latest PR

insights/parsers/sysconfig.py
rasrivas-redhat: acknowledged, updated the code in the latest PR

insights/parsers/sysconfig.py
rasrivas-redhat: acknowledged, updated the code in the latest PR

insights/parsers/sysconfig.py
rasrivas-redhat: acknowledged, updated the code in the latest PR

insights/parsers/sysconfig.py
rasrivas-redhat: @xiangce Can we use "return self.file_name.split("route-", 1)[1]", so that we can only get route files, please suggest

insights/parsers/sysconfig.py
```diff
@@ -506,3 +509,39 @@ class VirtWhoSysconfig(SysconfigOptions):
...
+
+    Attributes:
+        data (dict): all the static route cofiguration details
+        static_route_name (str): static route name
+
+    """
+    def parse_content(self, content):
+        self.data = dict()
+        self.static_route_name = self.file_name.split("route-", 1)[1]
+        self.data[self.static_route_name] = content
```
xiangce: I believe the `self.data` should be the parsed content, but not the RAW `content`.

You may want to achieve that by calling the `parser_content` of the superclass as below:

```python
self.static_route_name = self.file_name.split("route-", 1)[1]
super(IfCFGStaticRoute, self).parse_content(content)
```

xiangce (approved):
Approved pending the issue of Jenkins being fixed

PR #1602, opened 2018-12-17
Added parser tlb_dynamic_lb

vishwanathjadhav:


insights/parsers/bond_dynamic_lb.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@parser(Specs.bond_dynamic_lb)
+class BondDynamicLB(Parser):
+    """
+    Models the ``/sys/class/net/bond[0-9]*/bonding/tlb_dynamic_lb`` file.
+
+    0 - Hash based load balancing.
+
+    1 - Load based load balancing.
```
xiangce: better add docstring for the two Exceptions here, see https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/grubby.py#L34

insights/parsers/bond_dynamic_lb.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    def parse_content(self, content):
+        self._dynamic_lb_status = None
+        self._bond_name = self.file_path.rsplit("/")[-3]
+        if not content:
```
xiangce: Better move the safety check to the very begging of the function.

PR #1603, opened 2018-12-18
Parser for proc_ovs_vswitchd_limits

ppandey-redhat:
Updated `proc_limits` parser to get limits for process `ovs-vswitchd`.


xiangce:
`INFO:` @datasource with `pre-command` is used for this spec, no need to add it to uploader.json.

PR #1604, opened 2018-12-18
[IfCFG] Detect blank line in interface file

psachin:


insights/parsers/ifcfg.py
```diff
@@ -126,3 +127,10 @@ def bonding_mode(self):
         if m in bond_mode_map:
             return bond_mode_map[m]
         return None
+
+    @property
```
xiangce: Every time when calling this `property` the `in` checking operation in line-136 will be executed, which results in a not so good efficiency.

```return self._has_empty_line```

File: insights/parsers/ifcfg.py [hunk 2]
```diff
@@ -126,3 +127,10 @@ def bonding_mode(self):
...
+    def has_empty_line(self):
+        """
+        (bool) `True` if the file has empty line else `False`.
+        """
+        return '' in self.context.content
```
xiangce: Could a line which only contains spaces, like `   ` can be treated as a blank line and has the same effect to empty line``?

Once it is. Better check them as well.

`any(l.strip() == '' for l in context)`

Otherwise, please ignore this comment.

insights/parsers/ifcfg.py
```diff
@@ -79,6 +79,7 @@ class IfCFG(LegacyItemAccess, CommandParser):
 
     def __init__(self, context):
         super(IfCFG, self).__init__(context)
+        self.context = context
```
xiangce: Better change this to
```self._has_empty_line = any(l.strip() == '' for l in context)```

File: insights/parsers/ifcfg.py (hunk 2 above)
Comment (psachin) [reply insights/parsers/ifcfg.py]: Could a line which only contains spaces, like can be treated as a blank line and has the same effect to empty line``?
> No

File: insights/parsers/ifcfg.py (hunk 3 above)
Comment (psachin) [reply insights/parsers/ifcfg.py]: @xiangce and remove the `@property` ?

Comment (psachin) [reply insights/parsers/ifcfg.py]: The test will fail

```bash
>       assert r._has_empty_line is False
E       AttributeError: 'IfCFG' object has no attribute '_has_empty_line'

```

xiangce: I don't think it will fail after you added the assignment e.g. `self._has_empty_line =` to the `__init__`

psachin: Done. Please check

insights/parsers/ifcfg.py
```diff
@@ -79,8 +79,10 @@ class IfCFG(LegacyItemAccess, CommandParser):
...
```
xiangce: You might forget to remove this line

psachin: Oops!! Good catch. Removed now.

PR #1610, opened 2018-12-20
Check if all_files and common_path is empty

bsquizz:
Related to https://github.com/RedHatInsights/insights-pup/issues/18

In the above exception, common_path was an empty string.

We'll now check to see if all_files is an empty list, or if common_path is an empty string, and raise `InvalidArchive` if so.

I don't think we actually need to catch any exceptions, since:
* `os.walk()` (used in `get_all_files`) just returns an empty list if there's no dirs to walk (or it doesn't have read permissions on dirs),
* `os.path.commonprefix` returns an empty strings if it can't find a common prefix
* `os.path.dirname` returns an empty string if one was passed to it.


kylape:
Based off of my past experience with debugging the code that determines the root directory of a given tarfile, it is incredibly brittle and there are tons of edge cases.  I'm afraid that having an empty common prefix is actually a valid archive format.  @csams or @jhjaggars can you confirm?

jhjaggars:
It _was_ a valid archive when we supported the uber-archive style.

kylape:
If we are confident that it's not a valid archive now, then I think we'd be okay to merge it.  We can always roll back if we start rejecting valid archives due to this change.

PR #1619, opened 2019-01-08
Specs ip_route_show_table_all in sos_archive

akshay196:
Add `ip_route_show_table_all` spec in sos archive


esakaiev:
test this please

PR #1622, opened 2019-01-14
Network initscripts slave config issue

ghost:


insights/parsers/ifcfg.py
```diff
@@ -97,7 +97,7 @@ def parse_content(self, content):
             # of the key is before dequoting and stripping.
             if key == "BONDING_OPTS":
                 self.data["raw_bonding_value"] = value
-            if key != "DEVICE":
+            if key not in ["DEVICE", "MASTER"]:
```
xiangce: A more reasonable patch for this update could be,  I think, adding a "raw_master_value", just like the "raw_bonding_value".

For `DEVICE`, adding a "raw_device_value" is better too, but to keep compatible we have to reserve the `DEVICE` as it was.

```python
if key in ["DEVICE", "MASTER", "BONDING_OPTS"]:
   self.data["raw_{0}_value".format(key.split('_')[0].lower())] = value
if key != "DEVICE":
   value = value.strip().strip(QUOTES)
```

How do you think?

rasrivas-redhat: @xiangce acknowledged, updated the code in the latest PR.

xiangce (approved):

PR #1624, opened 2019-01-15
updated the code for team_slave in ifconfig file

### LINKED ISSUE #1626: ifcfg parser to support for MASTER and TEAM_MASTER keys slave type
We need to update the ifcfg parser to support TEAMING and BONDING slave type in the configuration file, so, that we can use MASTER' and 'TEAM_MASTER' keys in raw format.

For ex-  `obj['MASTER']="'bond0'"` or `obj['TEAM_MASTER']="'team0'"`

ghost:
Fix #1626

### DIFF
--- a/insights/parsers/tests/test_ifcfg.py
+++ b/insights/parsers/tests/test_ifcfg.py
@@ -121,6 +121,15 @@ IPV6INIT=no
 MASTER="bond0"
 """.strip()
 
+IFCFG_TEST_RAW_TEAM_MASTER_VALUE = """
+TYPE=Ethernet
+DEVICE="eth0"
+BOOTPROTO=none
+ONBOOT=yes
+TEAM_MASTER="team0"
+MTU=9000
+""".strip()
+
 IFCFG_TEST_RAW_BONDING_VALUE = """
 DEVICE="bond0"
 IPADDR=10.11.96.172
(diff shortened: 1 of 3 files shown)


insights/parsers/tests/test_ifcfg.py
```diff
@@ -147,6 +156,8 @@
 
 IFCFG_PATH_NAMED_BOND_MASTER = "etc/sysconfig/network-scripts/ifcfg-eth0"
 
+IFCFG_PATH_NAMED_TEAM_MASTER = "etc/sysconfig/network-scripts/ifcfg-eth0"
```
xiangce: The value of this `IFCFG_PATH_NAMED_TEAM_MASTER` is the same to `IFCFG_PATH_NAMED_BOND_MASTER`. Better just reserve one and rename it to e.g. `IFCFG_PATH_1`

xiangce (changes requested):
Hi @rasrivas-redhat 

What's the reason that you keep the RAW value of the "TEAM_MASTER" ?
Please open an issue for it [here](https://github.com/RedHatInsights/insights-core/issues) at first.

BTW, please explain the reason for keeping the RAW value for "MASTER" as well in the same issue. Thanks.

rasrivas-redhat:
@xiangce 
First, we updated the code for "Bonding slave type" which we mentioned as MASTER="xxx" in config file but for "teaming slave type" we need to mention as TEAM_MASTER="xxx" in config file.

insights/parsers/tests/test_ifcfg.py
```diff
@@ -147,6 +156,10 @@
...
```
xiangce: You need to remove the `IFCFG_PATH_NAMED_BOND_MASTER` and `IFCFG_PATH_NAMED_BOND_MASTER`

rasrivas-redhat: @xiangce removed the `IFCFG_PATH_NAMED_TEAM_MASTER` and `IFCFG_PATH_NAMED_BOND_MASTER` and updated the test case for `IFCFG_PATH_ETH1` and `IFCFG_PATH_ETH2` path which are used in test case

bfahr (approved):
Approved pending @xiangce approval.

xiangce (approved):

PR #1625, opened 2019-01-15
Small tweaks to remove.conf

gravitypriest:
First of all, an actual bug: due to how configparser works it was double-escaping slashes in the remove.conf file -- e.g. if you specified `/etc/yum.repos.d/()*.*\\.repo`, it would end up as `/etc/yum.repos.d/()*.*\\\\.repo` once parsed, which then would fail the comparison against uploader.json. This was fixed by adding `.decode('string-escape')` for PY2, or `.encode('utf-8').decode('unicode-escape')` for PY3.

Second, I added a small feature that would allow you to hide files *after* wildcard matching. That is, you may specify anything that a wildcard might include in order to hide it. For example, if you specified `/etc/yum.repos.d/redhat.repo` in remove.conf, it would be compared to the list of files after expanding `/etc/yum.repos.d/()*.*\\.repo` and then ignored.


kylape:
Couple of notes:

1. While the fix seems reasonable, it would be nice if the user didn't have to specify complex commands for exclusion.  Since symbolic names are available, I suggest we allow users to specify the symbolic name as an alternative to the entire command/file.

2. Please provide a unit test that parses a raw remove.conf to exercise the code that splits the comma delimited lines.

kylape:
1. assign local variable so the compound `if` can be collapsed to one line
2. use fixture teardown function to delete temp file at end of raw config test

PR #1632, opened 2019-01-18
Parser to collect file metadata from `/rhev/data-center` on the RHEV Host

psachin:


esakaiev:
test this please

bfahr:
@psachin #1679 has been merged that provides the underlying infrastructure for the new datasource.  Please update this PR to use the new datasource.

insights/parsers/rhev_data_center.py
```diff
@@ -0,0 +1,46 @@
...
+        >>> assert len(rhev_dc.data) == 3
+        >>> assert len(rhev_dc.bad_image_ownership) == 1
+        >>> assert rhev_dc.bad_image_ownership[0]['path'] == '/rhev/data-center/mnt/host1.example.com:_nfsshare_data/a384bf5d-db92-421e-926d-bfb99a6b4b28/images/b7d6cc07-d1f1-44b3-b3c0-7067ec7056a3/4d6e5dea-995f-4a4e-b487-0f70361f6137'
+        >>> assert rhev_dc.bad_image_ownership[0]['name'] == 'root'
+        >>> assert rhev_dc.bad_image_ownership[0]['group'] == 'root'
+    """
+    def parse_content(self, content):
+        self.data = json.loads(''.join(content))
+
+    @property
```
xiangce: When the method is decorated with `@property`, it will work as an attribute of the object. 
It's not recommended to compute the results in each calling of the property.
Since the results are known at the time the `self.data` is parsed, you could calculate them once in parse_content instead of each time this property is invoked.

insights/parsers/rhev_data_center.py
```diff
@@ -0,0 +1,46 @@
...
+    def bad_image_ownership(self):
+        """Returns full path of RHEV VM images in the Data Center not having correct file ownership.
+
+        Returns:
+            (list): [{'name': 'root', 'group': 'root', 'path': '/rhev/data-center/mnt/example.com:_nfsshare_data/a384bf5d/images/b7d6cc07/4d6e5dea'}]
```
xiangce: This is just an example of the `return value`. Better change it to a more common description, otherwise, it looks like this interface always returns this fixed value.

psachin: Sure

insights/parsers/rhev_data_center.py
psachin: @xiangce Do you mean I should move the logic of `bad_image_ownership` to `parse_content`?

xiangce: Yes. @psachin

psachin: @xiangce Done

insights/parsers/rhev_data_center.py
```diff
@@ -0,0 +1,42 @@
...
+==================================================
+"""
+import json
+from re import compile as re_compile
+from insights import Parser, parser
+from insights.specs import Specs
+
+
+@parser(Specs.rhev_data_center)
+class RhevDataCenter(Parser):
```
xiangce: Better change to use "CommandParser" as the base, since I have found the `python_-m_insights.tools.cat ...` command sometimes fails as below on some host:

~~~
~]# cat insights-vm37-39.gsslab.pek2.redhat.com-20190222010448/insights_commands/python_-m_insights.tools.cat_--no-header_rhev_data_center
/usr/bin/python: No module named insights.tools
~~~

This is an issue of insight-client, but I think we'd better make sure the parser robust enough to work.

insights/parsers/rhev_data_center.py
```diff
@@ -0,0 +1,42 @@
...
+    Examples:
+        >>> assert len(rhev_dc.data) == 4
+        >>> assert len(rhev_dc.incorrect_volume_ownership) == 1
+        >>> assert rhev_dc.incorrect_volume_ownership[0]['path'] == '/rhev/data-center/mnt/host1.example.com:_nfsshare_data/a384bf5d-db92-421e-926d-bfb99a6b4b28/images/b7d6cc07-d1f1-44b3-b3c0-7067ec7056a3/4d6e5dea-995f-4a4e-b487-0f70361f6137'
+        >>> assert rhev_dc.incorrect_volume_ownership[0]['name'] == 'root'
+        >>> assert rhev_dc.incorrect_volume_ownership[0]['group'] == 'root'
+
+    """
+    def parse_content(self, content):
+        self.data = json.loads(''.join(content))
```
xiangce: Better skip empty content by raising a `SkipException`

insights/parsers/rhev_data_center.py
xiangce: I just opened an issue for it: https://github.com/RedHatInsights/insights-core/issues/1697

psachin: ack

psachin:
```
[12:13]<psachin> xiangce: Regarding your suggestion on using the
		 CommandParser, I have see the issue with tools.cat too. How
		 does using CommandParser solves that?
[12:14]<xiangce> psachin: I just submitted a new PR, trying to fix it
       <xiangce> psachin:
		 https://github.com/RedHatInsights/insights-core/pull/1698
       <psachin> xiangce: Adding to this, the new DS is a python was to get
		 data, will CommandParser help as we are not using any
		 command?
[12:15]<psachin> xiangce: Can I stick to Parser until this is fixed?
[12:16]<xiangce> psachin: I think your Parser can go on
```

xiangce:
~~~
144820 xiangc+| psachin: the new DS will also need to run the command "python -m insights.tools....." to get the "running result" of
              | special spec. But once the "python -m" command fails, the error message outputted by the command will still be collected.
145001 psachi+| xiangce: Yes. Every specs is a command.
145143 xiangc+| psachin: Oh, my fault
145156      ➥ | psachin: I mean your PR can go on
145158 psachi+| xiangce: Huh?
145209 psachi+| xiangce: Okay.
145210 xiangc+| psachin: You'd better change to ComamndParser
145221 psachi+| xiangce: Again?
145341 xiangc+| psachin: Or you skip the "/usr/bin/python: No module named insights.tools" manaully in the parse_content
145355 psachi+| xiangce: Can we wait for your PR review and get some extra advice on my parser?
145449 xiangc+| psachin: No problem, if you are not in a hurry to merge your new plugin
145617 psachi+| xiangce: Well, I'm in hurry, but at the same time I don't want to re-visit the parser/plugin for changes.
~~~

insights/parsers/rhev_data_center.py
psachin: Done

insights/parsers/rhev_data_center.py
bfahr: @psachin #1698 has been merged so you can go ahead and use CommandParser.

xiangce: @bfahr - I think it's because that here the `content` is a `list` of `dict`s, but not a `dict`. The [JSONParser](https://github.com/RedHatInsights/insights-core/blob/9cabaa5159651bd3efe8625fb821e225cc985a65/insights/core/__init__.py#L701) seems can only parse `dict`.

bfahr: @psachin I'm not sure I understand what you are saying.  You are using the json module to parse it, `self.data = json.loads(''.join(content))`, that is what the `JSONParser` does.

xiangce: @bfahr - Sorry for the poor explanation above. I have met this problem in the [sapcontrol.py](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/sapcontrol.py) The root cause should be the [JSONParser](https://github.com/RedHatInsights/insights-core/blob/9cabaa5159651bd3efe8625fb821e225cc985a65/insights/core/__init__.py#L701) is mixed with the [LegacyItemAccess](https://github.com/RedHatInsights/insights-core/blob/18d9502188929b68fb3cb6c3a31e84ba4824e578/insights/core/__init__.py#L388) which requires the `self.data` a dictionary (otherwise, the `get` and `[]` operation will fail):

https://github.com/RedHatInsights/insights-core/blob/18d9502188929b68fb3cb6c3a31e84ba4824e578/insights/core/__init__.py#L706

Any thoughts?

psachin: @xiangce As we know that the datasource return type is  json, I think I should use a JSONParser as suggested by @bfahr

xiangce: @psachin - I'm afraid you cannot do that. Unless you modify the DS to return a `dict` instead of a `list of dict`. Since it's not allowed to remove the `LegacyItemAccess` from its base-classes, this will break the tests of many plugins.

bfahr: @psachin @xiangce sorry for the delayed response, I thought I had already responded to this and I guess I didn't click the button.  I agree with @xiangce in this case, the implementation like [sapcontrol.py](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/sapcontrol.py) seems to be a better solution and less confusing.

psachin: @bfahr @xiangce We agree to use `CommandParser` ?

psachin: The latest commit uses the `CommandParser`

xiangce: CommandParser is OK, since the `python -m` is also a command.

PR #1647, opened 2019-01-23
Enhance combiner "redhat_release"

xiangce:
- add a utility 'rhel' attr
- add a new class parser `RedHatRelease`
  Better keep them both


esakaiev:
test this please

insights/combiners/redhat_release.py
```diff
@@ -25,14 +16,15 @@
 from insights.parsers.redhat_release import RedhatRelease as rht_release
 from insights.parsers.uname import Uname
 from insights.core.serde import serializer, deserializer
+from insights.parsers import ParseException
 
-Release = namedtuple("Release", field_names=["major", "minor"])
+Release = namedtuple("Release", field_names=["major", "minor", "rhel"])
```
bfahr: @xiangce this will break backward compatibility with the existing combiner.  You could add a `rhel` method to the existing combiner to return this value, and then use a new named tuple to represent the 3 tuple for the new combiner.

xiangce: @bfahr - I add the default value for each filed in the namedtuple, and removed the code that used the `Release` in the rules' code. Is it OK, please have a check.

bfahr: @xiangce the issue is that the current combiner returns the tuple Release(major,minor), and we need to maintain that interface for backward compatibility.  You need to leave the existing Release tuple as it is, and create a new tuple for the new combiner, if it is needed.

insights/combiners/redhat_release.py
```diff
@@ -42,23 +40,94 @@ def deserialize(_type, obj, root=None):
...
 
     if rh_release:
-        return Release(rh_release.major, rh_release.minor)
+        return Release(rh_release.major, rh_release.minor, rh_release.version)
+
+    raise ParseException("Unabled to determine release.")
+
+
+@combiner([Uname, rht_release])
+class RedHatRelease(object):
```
bfahr: Although we don't currently use them, this class also needs a serializer/deserializer.  See this code for an example of these for a class: https://github.com/RedHatInsights/insights-core/blob/master/insights/tests/test_serde.py#L32

insights/combiners/redhat_release.py
xiangce: @bfahr - Done, please check the latest commit.

insights/combiners/redhat_release.py
xiangce: @bfahr - Done. However I didn't involve the three `rhel[6-8]` attributes in the serializer and deserializer, please check if this is OK. Thanks.

insights/combiners/redhat_release.py
```diff
@@ -61,4 +66,72 @@ def redhat_release(rh_release, un):
...
+
+        self.rhel6 = self.rhel if self.major == 6 else None
+        self.rhel7 = self.rhel if self.major == 7 else None
+        self.rhel8 = self.rhel if self.major == 8 else None
+
+
+# No need to serialize/deserialize the `rhel#` attributes
+@serializer(RedHatRelease)
+def serialize_RedHatRelease(obj, root=None):
+    return {"major": obj.major, "minor": obj.minor, "rhel": obj.rhel}
```
csams: All attributes of the combiner instance that could be accessed by a rule need to be serialized and deserialized or there will be attribute errors since the deserializer bypasses `__init__`. Need to add rhel[6-8].

xiangce: I got it. Thanks for the explanation. @csams

insights/combiners/redhat_release.py
```diff
@@ -42,17 +35,29 @@ def deserialize(_type, obj, root=None):
 
 @combiner([rht_release, Uname])
 def redhat_release(rh_release, un):
-    """Check uname and redhat-release for rhel major/minor version.
+    """
```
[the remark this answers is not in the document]
xiangce: Done

xiangce:
test this please

PR #1649, opened 2019-01-24
Add RHEL8 string release that can be used for rule tests

matysek:


insights/tests/__init__.py
```diff
@@ -246,6 +246,7 @@ def __repr__(self):
 RHEL5 = "Red Hat Enterprise Linux Server release 5.11 (Tikanga)"
 RHEL6 = "Red Hat Enterprise Linux Server release 6.5 (Santiago)"
 RHEL7 = "Red Hat Enterprise Linux Server release 7.0 (Maipo)"
+RHEL8 = "Red Hat Enterprise Linux release 8.0 Beta (Ootpa)"
 
 
 def redhat_release(major, minor=""):
```
bfahr: @matysek should you go ahead and change the `redhat_release` function to support RHEL8 as well?  My concern is being consistent in this module.

matysek: @bfahr I updated also the redhat_release function as requested

Issue #1648, opened 2019-01-23
Investigate collecting `/etc/dnf/modules.d` and other dnf related information


csams:
We currently collect configuration information for yum but don't have anything in place for `dnf`. We should look into that.

tkasparek:
Prime example of this is the ```/etc/dnf/modules.d/``` directory which contains modularity related metadata.

miclark:
Fixed by #1656

PR #1656, opened 2019-01-28
Gather modularity related information

tkasparek:
Newer versions of YUM (based on DNF) are storing
additional metadata related to modularity functionality
in the /etc/dnf/modules.d/ directory.
These files are needed in order to properly calculate
package updates.

Signed-off-by: Tomas Kasparek <tkasparek@redhat.com>

### DIFF
--- a/docs/shared_parsers_catalog/dnf_modules.rst
+++ b/docs/shared_parsers_catalog/dnf_modules.rst
@@ -1,0 +1,3 @@
+.. automodule:: insights.parsers.dnf_modules
+    :members:
+    :show-inheritance:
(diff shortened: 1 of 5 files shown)


docs/shared_parsers_catalog/dnf_modules.srt
```diff
@@ -0,0 +1,3 @@
+.. automodule:: insights.parsers.dnf_modules
```
bfahr: @tkasparek this file is named incorrectly, it should be `.rst` instead of `.srt`.  Once you rename it please check your documentation as described here to make sure that it renders correctly.  https://insights-core.readthedocs.io/en/latest/docs_guidelines.html#testing-your-docstring

bfahr:
Related to #1648

docs/shared_parsers_catalog/dnf_modules.srt
tkasparek: Should be resolved now.

PR #1658, opened 2019-01-29
New parser for SapControlSystemUpdateList

xiangce:


insights/specs/default.py
```diff
@@ -583,29 +621,46 @@ def package_and_java(broker):
...
+
+        Returns:
+            (list): List of SID.
+
+        """
+        return list(set(sn[0] for sn in broker[DefaultSpecs.sap_sid_nr]))
+
+    sap_hdb_version = foreach_execute(sap_sid, "/usr/bin/sudo -iu %sadm HDB version", keep_rc=True)
     sap_host_profile = simple_file("/usr/sap/hostctrl/exe/host_profile")
+    sapcontrol_getsystemupdatelist = foreach_execute(sap_sid_nr, "/usr/bin/sudo -iu %sadm sapcontrol -nr %s -function GetSystemUpdateList", keep_rc=True)
```
csams: This isn't going to work correctly with the simple_file definition in insights_archive. The cat tool will shove all of the outputs into a single file. The definition in insights_archive will read it with simple file, but the parser won't handle it properly.

If you just ran the parser on the live system and let it get results of the foreach_command directly, that's different from what it would get from the simple_file in insights_archive. The single file will cause one parser to be created that's expected to understand how to break the output apart. The direct evaluation will create a separate parser per thing generated by the foreach_execute.

xiangce: The current parser in this PR was for the output of a single command. Yes, I'm sure it needs to be updated once we decided to use the "insights-cat sapcontrol_getsystemupdatelist" command to collect these "foreach_*" specs.

One of my concern is, how to handle the metadata in insights archives and sosreport archives correctly at the same time. E.g. the `ethtool` or `ifcfg` related commands, in sosreport, they are separated in different files (as per `iface`s). If refactor it with `insigths-cat`, the parser has to support the two different format?

xiangce: @csams, please check it again.

esakaiev:
test this please

insights/specs/default.py
```diff
@@ -632,29 +633,67 @@ def rhev_data_center(broker):
...
+        Returns:
+            (list): List of SID.
+
+        """
+        return list(set(sn[0] for sn in broker[DefaultSpecs.sap_sid_nr]))
+
+    sap_hdb_version = foreach_execute(sap_sid, "/usr/bin/sudo -iu %sadm HDB version", keep_rc=True)
     sap_host_profile = simple_file("/usr/sap/hostctrl/exe/host_profile")
+
+    def sapcontrol_getsystemupdatelist(broker):
```
csams: Needs at `@datasource`

insights/specs/default.py
```diff
@@ -632,29 +633,67 @@ def rhev_data_center(broker):
...
+        import json
+        rks_cmd = "/usr/bin/sudo -iu {0}adm sapcontrol -nr {1} -function GetSystemUpdateList"
+        relative_path = "sudo_-iu_*adm_sapcontrol_-nr_*_-function_GetSystemUpdateList"
+        header = "hostname, instanceNr, status, starttime, endtime, dispstatus"
+        line_set = set()
+        results = list()
+        for sid, nr in broker[DefaultSpecs.sap_sid_nr]:
+            cmd = rks_cmd.format(sid, nr).split()
+            ret = Popen(cmd, stdout=PIPE, stderr=PIPE)
```
csams: Get the context from the broker and use the `shell_out` function on it instead of using subprocess directly. This will give us a hook to help enforce filtering and command blacklisting.

insights/specs/default.py
xiangce: Done

insights/specs/default.py
xiangce: Done. Thanks a lot.

PR #1662, opened 2019-01-30
Ignore bad lines in df

xiangce:


insights/parsers/df.py
```diff
@@ -154,6 +154,7 @@ def __iter__(self):
             yield row
 
     def parse_content(self, content):
+        content = [l for l in content if self.bad_lines[0] not in l.lower()]
```
bfahr: @xiangce doesn't this create a dependency on the base class that `self.bad_lines` could change and break this parser.  Do you think it would be better if you defined the bad line filter string in this class?

xiangce: Yes, I agree with you. Please check the latest commit, Thanks.

PR #1667, opened 2019-02-01
Fix ceph version - add 'downstream_release' field

mykaul:
1. Fix the confusion between 'release' and 'version' according to N-V-R convention (note - this WILL BREAK existing rules)
2. Add 'downstream_release' field, to distinguish between (for example) 3.1.0 and 3.1.1.

### DIFF
--- a/insights/parsers/ceph_version.py
+++ b/insights/parsers/ceph_version.py
@@ -18,7 +18,7 @@ Note:
 
 Example:
     >>> ceph_ver = shared[CephVersion]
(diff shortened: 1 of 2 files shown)


mykaul:
Whoops, forgot to change the regexp. Let me fix that.

xiangce:
Please fix the flake8 error:

insights/parsers/tests/test_ceph_version.py:16:1: E302 expected 2 blank lines, found 1

PR #1676, opened 2019-02-09
SCTP eps parser

vishwanathjadhav:
Written a parser to parse endpoint details of SCTP sockets, eps details are present in `/proc/net/sctp/eps`.

### DIFF
--- a/insights/parsers/sctp_eps.py
+++ b/insights/parsers/sctp_eps.py
@@ -22,8 +22,8 @@ Examples:
     True
     >>> sorted(sctp_info.sctp_local_ips) == sorted(['10.0.0.102', '10.0.0.70', '172.31.1.2'])
     True
-    >>> sctp_info.search(local_port="11165")
-    [{'endpoints': 'ffff88017e0a0200', 'socket': 'ffff880299f7fa00', 'sk_type': '2', 'sk_state': '10', 'hash_bkt': '29', 'local_port': '11165', 'uid': '200', 'inode': '299689357', 'local_addr': ['10.0.0.102', '10.0.0.70']}]
+    >>> sorted(sctp_info.search(local_port="11165")) == sorted([{'endpoints': 'ffff88017e0a0200', 'socket': 'ffff880299f7fa00', 'sk_type': '2', 'sk_state': '10', 'hash_bkt': '29', 'local_port': '11165', 'uid': '200', 'inode': '299689357', 'local_addr': ['10.0.0.102', '10.0.0.70']}])
+    True
     >>> len(sctp_info.search(local_port="11165")) == 1
     True
     >>> len(sctp_info.search(endpoints="ffff88017e0a0200")) == 1
--- a/insights/parsers/sctp_eps.py
+++ b/insights/parsers/sctp_eps.py
@@ -1,0 +1,68 @@
+"""
+SCTPEps - file ``/proc/net/sctp/eps ``
+======================================
+
+This provides detail information about SCTP endpoints.
+Which includes Endpoints, Socket, Socket type, Socket State,
+hash bucket, bind port, UID, socket inodes, Local IP
+address.
+
+Typical content of ``eps`` file is::
+    ENDPT            SOCK             STY SST HBKT LPORT   UID INODE     LADDRS
+    ffff88017e0a0200 ffff880299f7fa00 2   10  29   11165   200 299689357 10.0.0.102 10.0.0.70 
+    ffff880612e81c00 ffff8803c28a1b00 2   10  30   11166   200 273361203 10.0.0.102 10.0.0.70 
+    ffff88061fba9800 ffff88061f8a3180 2   10  31   11167   200 273361145 10.0.0.102 10.0.0.70 
+    ffff88031e6f1a00 ffff88031dbdb180 2   10  32   11168   200 273365974 10.0.0.102 10.0.0.70
+
+Output data is stored in the dictionary format
+
+Examples:
+    >>> type(sctp_info)
+    <class 'insights.parsers.sctp_eps.SCTPEps'>
+"""
+
+from insights import Parser, parser, get_active_lines
+from insights.specs import Specs
+from insights.parsers import SkipException, ParseException
+
+
+@parser(Specs.sctp_eps)
+class SCTPEps(Parser):
+    """
+    This parser parses the content of ``/proc/net/sctp/eps`` file.
+    Currently we need "Bind Port" and "Local address"
+    """
+    
+    def parse_content(self, content):
+        if (not content) or (not self.file_path):
+            raise SkipException("No Contents")
+        
+        COLUMN_IDX = {
+            'ENDPT': 'endpoints',
+            'SOCK': 'socket',
+            'STY': 'sk_type',
+            'SST': 'sk_state',
+            'HBKT': 'hash_bkt',
+            'LPORT': 'local_port',
+            'UID': 'uid',
+            'INODE': 'inode',
+            'LADDRS': 'local_addr'
+        }
+        
+        self.data = []
+        exp_column = sorted(COLUMN_IDX.keys())
+        for line in content:
+            row = {}
+            line = line.strip()
+            line = line.split()
+            if ("LPORT" in line):
+                if len(line) == len(exp_column):
+                    columns = line
+                else:
+                    raise ParseException("Unrecognised Column Name".format(b=raws))
+            else:
+                for idx, val in enumerate(columns):
+                    key = COLUMN_IDX[val]
+                    row[key]=line[idx]
+                self.data.append(row)
+
--- a/insights/parsers/sctp_eps.py
+++ b/insights/parsers/sctp_eps.py
@@ -18,14 +18,21 @@ Output data is stored in the dictionary format
 Examples:
     >>> type(sctp_info)
     <class 'insights.parsers.sctp_eps.SCTPEps'>
-    >>> sorted(sctp_info.get_ports()) == sorted(['11165', '11166'])
+    >>> sorted(sctp_info.sctp_ports) == sorted(['11165', '11166'])
     True
-    >>> sorted(sctp_info.get_local_ips()) == sorted(['10.0.0.102', '10.0.0.70', '172.31.1.2'])
+    >>> sorted(sctp_info.sctp_local_ips) == sorted(['10.0.0.102', '10.0.0.70', '172.31.1.2'])
+    True
+    >>> sctp_info.search(local_port="11165")
+    [{'endpoints': 'ffff88017e0a0200', 'socket': 'ffff880299f7fa00', 'sk_type': '2', 'sk_state': '10', 'hash_bkt': '29', 'local_port': '11165', 'uid': '200', 'inode': '299689357', 'local_addr': ['10.0.0.102', '10.0.0.70']}]
+    >>> len(sctp_info.search(local_port="11165")) == 1
+    True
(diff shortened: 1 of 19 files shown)


insights/parsers/sctp_eps.py
```diff
@@ -0,0 +1,109 @@
...
+                        row[key] = []
+                        while (idx != len(line)):
+                            row[key].append(line[idx])
+                            idx = idx + 1
+                    else:
+                        key = COLUMN_IDX[val]
+                        row[key] = line[idx]
+                self.data.append(row)
+
+    def get_ports(self):
```
bfahr: @vishwanathjadhav since `self.data` has already been computed in `parse_content` its not necessary to recompute sctp_ports every time this method is called.  Have you considered just creating an attribute `ports` or `sctp_ports` in `parse_content` and just using that value instead of this method.

insights/parsers/sctp_eps.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        (list): This function returns a list of SCTP ports if SCTP
+                endpoints are created, else `[]`.
+        """
+        sctp_ports = []
+        for eps in self.data:
+            sctp_ports.append(eps['local_port'])
+        return sctp_ports
+
+    def get_local_ips(self):
```
bfahr: Same comment as above.

insights/parsers/sctp_eps.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        (list): This function returns a list of all local ip addresses
+                if SCTP endpoints are created, else `[]`.
+        """
+        sctp_ips = set([])
+        for eps in self.data:
+            sctp_ips = set.union(sctp_ips, set(eps['local_addr']))
+        return list(sctp_ips)
+
+    def get_eps_ips(self):
```
bfahr: Same comment as above, and also your docstring indicates that `None` is the default results, it actually an empty dict `{}`.

insights/parsers/sctp_eps.py
```diff
@@ -0,0 +1,109 @@
...
+            'STY': 'sk_type',
+            'SST': 'sk_state',
+            'HBKT': 'hash_bkt',
+            'LPORT': 'local_port',
+            'UID': 'uid',
+            'INODE': 'inode',
+            'LADDRS': 'local_addr'
+        }
+
+        self.data = []
```
bfahr: Have you considered using `parse_fixed_table` for this data? https://github.com/RedHatInsights/insights-core/blob/ac47281f9a83bb762982b14fb3b0824eb70d0c04/insights/parsers/__init__.py#L281

vishwanathjadhav: @bfahr - We can not use this function as the number of columns are variable, Local address could be of one or more than one. Also, I have tested this functionality, it does not suit
Sample Input:
~~~
SCTP_EPS_DETAILS = """ 
 ENDPT     SOCK   STY SST HBKT LPORT   UID INODE LADDRS
ffff880422ab2600 ffff8804238c6ac0 2   10  28   3868    500 61868 10.0.200.114 10.0.201.114 2010:0010:0000:0200:0000:0000:0000:0114 2010:0010:0000:0201:0000:0000:0000:0114
""".STRIP()
~~~
Sample Output:
~~~
[{'ENDPT': 'ffff880422', 'SOCK': 'ab2600', 'STY': 'ffff', 'SST': '8804', 'HBKT': '238c6', 'LPORT': 'ac0 2', 'UID': '10', 'INODE': '28   3', 'LADDRS': '868    500 61868 10.0.200.114 10.0.201.114 2010:0010:0000:0200:0000:0000:0000:0114 2010:0010:0000:0201:0000:0000:0000:0114'}]
~~~

insights/parsers/sctp_eps.py
vishwanathjadhav: Done

insights/parsers/sctp_eps.py
bfahr: That makes sense, thanks for checking.

insights/parsers/tests/test_sctp_eps.py
```diff
@@ -0,0 +1,61 @@
...
+    assert sorted(sctp_info.sctp_local_ips) == sorted(['10.0.0.102', '10.0.0.70', '172.31.1.2', '192.168.11.2'])
+    assert sctp_info.get_eps_ips == {'ffff88017e0a0200': ['10.0.0.102', '10.0.0.70'],
+                                       'ffff880612e81c00': ['10.0.0.102', '10.0.0.70', '172.31.1.2'],
+                                       'ffff88061fba9800': ['10.0.0.102', '10.0.0.70'],
+                                       'ffff88031e6f1a00': ['10.0.0.102', '10.0.0.70', '192.168.11.2']}
+    assert len(sctp_info.search(local_port='11165')) == 1
+
+    with pytest.raises(ParseException) as exc:
+        sctp_obj = SCTPEps(context_wrap(SCTP_EPS_DETAILS_NO))
+        assert not sctp_obj.sctp_local_ips
```
bfahr: If the exception is thrown this line will not execute.  If the exception is not thrown then the pytest.raises will throw an assertion, so I don't think this is necessary.  Were you trying to check something else with this assert?

insights/parsers/tests/test_sctp_eps.py
```diff
@@ -0,0 +1,61 @@
...
+    assert 'Contents are not compatible to this parser' in str(exc)
+
+    with pytest.raises(SkipException) as exc:
+        sctp_obj = SCTPEps(context_wrap(SCTP_EPS_DETAILS_NO_2))
+        assert not sctp_obj.sctp_local_ips
```
bfahr: Same comment as above.

insights/parsers/sctp_eps.py
```diff
@@ -0,0 +1,130 @@
...
+                    columns = line
+                else:
+                    raise ParseException("Contents are not compatible to this parser".format(row))
+            else:
+                for idx, val in enumerate(columns):
+                    if val == "ENDPT":
+                        # Save endpoint
+                        _eps = line[idx]
+                        self._sctp_eps_ips[_eps] = []
+                    if val == "LADDRS":
```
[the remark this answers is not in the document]
vishwanathjadhav: `line` is separately processed within `if` to collect eps id, it still needs to get through  `if elif` which will further process the line to create final `self.data`

insights/parsers/tests/test_sctp_eps.py
vishwanathjadhav: Just trying to remove `flake8` warnings `Local variable 'sctp_obj' is assigned to but never used`.

insights/parsers/sctp_eps.py
```diff
@@ -0,0 +1,130 @@
...
+    True
+    >>> sorted(sctp_info.sctp_local_ips) == sorted(['10.0.0.102', '10.0.0.70', '172.31.1.2'])
+    True
+    >>> sorted(sctp_info.search(local_port="11165")) == sorted([{'endpoints': 'ffff88017e0a0200', 'socket': 'ffff880299f7fa00', 'sk_type': '2', 'sk_state': '10', 'hash_bkt': '29', 'local_port': '11165', 'uid': '200', 'inode': '299689357', 'local_addr': ['10.0.0.102', '10.0.0.70']}])
+    True
+    >>> len(sctp_info.search(local_port="11165")) == 1
+    True
+    >>> len(sctp_info.search(endpoints="ffff88017e0a0200")) == 1
+    True
+"""
```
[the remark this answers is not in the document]
vishwanathjadhav: Done

insights/parsers/sctp_eps.py
```diff
@@ -0,0 +1,130 @@
...
+        (dict): This function returns a dict of all endpoints and corresponding
+                local ip addresses used by SCTP endpoints if SCTP endpoints are
+                created, else `None`.
+        """
+        return self._sctp_eps_ips if self._sctp_eps_ips else None
+
+    def search(self, **args):
+        """
+        (list): This function return a list of all endpoints when args search
+                matches, when args search do not match then it returns `[]` list.
```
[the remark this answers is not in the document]
vishwanathjadhav: I think that's because of double astrics in the above line.

insights/parsers/tests/test_sctp_eps.py
vishwanathjadhav: Done

insights/parsers/tests/test_sctp.py
```diff
@@ -0,0 +1,61 @@
...
+    assert sorted(sctp_info.sctp_local_ips) == sorted(['10.0.0.102', '10.0.0.70', '172.31.1.2', '192.168.11.2'])
+    assert sctp_info.sctp_eps_ips == {'ffff88017e0a0200': ['10.0.0.102', '10.0.0.70'],
+                                       'ffff880612e81c00': ['10.0.0.102', '10.0.0.70', '172.31.1.2'],
+                                       'ffff88061fba9800': ['10.0.0.102', '10.0.0.70'],
+                                       'ffff88031e6f1a00': ['10.0.0.102', '10.0.0.70', '192.168.11.2']}
+    assert len(sctp_info.search(local_port='11165')) == 1
+
+    with pytest.raises(ParseException) as exc:
+        sctp_obj = SCTPEps(context_wrap(SCTP_EPS_DETAILS_NO))
+        sctp_obj.sctp_eps_ips   # Just added ro remove flake8 warnings
```
bfahr: If you are going to check something here it should be `assert sctp_obj is None`, since nothing should be returned from the `__init__` method in this case.  If its not `None` then there is a problem.

insights/parsers/tests/test_sctp.py
```diff
@@ -0,0 +1,61 @@
...
+    assert 'Contents are not compatible to this parser' in str(exc)
+
+    with pytest.raises(SkipException) as exc:
+        sctp_obj = SCTPEps(context_wrap(SCTP_EPS_DETAILS_NO_2))
+        sctp_obj.sctp_eps_ips   # Just added to remove flake8 warnings
```
bfahr: Same as above comment.

insights/parsers/tests/test_sctp.py
vishwanathjadhav: Done

insights/parsers/tests/test_sctp.py
vishwanathjadhav: Done

insights/parsers/tests/test_sctp.py
```diff
@@ -0,0 +1,61 @@
...
+    assert len(sctp_info.search(local_port='11165')) == 1
+
+    with pytest.raises(ParseException) as exc:
+        sctp_obj = SCTPEps(context_wrap(SCTP_EPS_DETAILS_NO))
+        assert sctp_obj.sctp_eps_ips is None   # Just added ro remove flake8 warnings
+    assert 'Contents are not compatible to this parser' in str(exc)
+
+    with pytest.raises(SkipException) as exc:
+        sctp_obj = SCTPEps(context_wrap(SCTP_EPS_DETAILS_NO_2))
+        assert sctp_obj.sctp_eps_ips is None   # Just added to remove flake8 warnings
```
bfahr: @vishwanathjadhav this assert and the assert on line 47 should be to check if the object returned from SCTPEps() is `None`.  This assert is a bit confusing in that its not checking for the right thing.  If `sctp_obj.sctp_eps_ips` is empty, it would actually be an empty dict, not `None`.  The better test here would be `assert sctp_obj is None` because that is actually the expected condition that no parser is created.

PR #1677, opened 2019-02-11
Enhance lspci parser

akshay196:
- This includes
  * Add lspci files from sosreport in the `sos_archive` specs.
  * Modify lspci parser to make it compatible with output of `lspci -nnvv`, `lspci -nvv` and `lspci -k`

Signed-off-by: Akshay Gaikwad <akgaikwa@redhat.com>

### DIFF
--- a/insights/parsers/lspci.py
+++ b/insights/parsers/lspci.py
@@ -5,6 +5,7 @@ LsPci - Command ``lspci -k``
 To parse the PCI device information gathered from the ``/sbin/lspci -k`` command.
 
 """
+import re
 
 from .. import LogFileOutput, parser, CommandParser, get_active_lines
 from insights.specs import Specs
@@ -68,15 +69,19 @@ class LsPci(CommandParser, LogFileOutput):
             scanner(self)
         # Parse kernel driver lines
         self.data = {}
-        bus_device_function = ""
+        bus_device_function = None
+        bus_device_function_re = re.compile(r'[0-9a-f]+:[0-9a-f]+.[0-9a-f]+')
+
+        fields = ["Subsystem", "Kernel driver in use", "Kernel modules"]
+
         for line in get_active_lines(content):
-            parts = line.split(None)
-            if parts[0] and parts[0] not in ['Subsystem:', 'Kernel'] \
-                    and len(parts[0].split(':')) == 2 and len(parts[0].split('.')) == 2:
+            parts = line.split()
+
+            if bus_device_function_re.match(parts[0]):
                 bus_device_function = parts[0]
                 device_details = ' '.join(map(str, parts[1:]))
                 self.data[bus_device_function] = {'Dev_Details': device_details.lstrip()}
-            elif bus_device_function:
+            elif bus_device_function and (line.split(":")[0].strip() in fields):
                 parts = line.split(':')
                 self.data[bus_device_function][parts[0]] = parts[1].lstrip()
 
--- a/insights/parsers/tests/test_lspci.py
+++ b/insights/parsers/tests/test_lspci.py
@@ -204,8 +204,18 @@ LSPCI_DRIVER_DOC = """
         Kernel modules: ixgbe
 """.strip()
 
+LSPCI_OUTPUT_WITH_BAD_LINES = """
+03:00.0 Network controller: Intel Corporation Wireless 7260 (rev bb)
+        Subsystem: Intel Corporation Dual Band Wireless-AC 7260
+
+        Kernel driver in use: iwlwifi
+# a comment
+        Kernel modules: iwlwifi
+""".strip()
+
 
 def test_lspci():
+    LsPci.token_scan('centrino', 'Centrino')
     lspci_0 = LsPci(context_wrap(LSPCI_0))
     assert [i['raw_message'] for i in lspci_0.get("Intel Corporation")] == INTEL.splitlines()
     assert len(lspci_0.get("Network controller")) == 1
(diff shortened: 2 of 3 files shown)


insights/parsers/lspci.py
```diff
@@ -69,14 +69,20 @@ def parse_content(self, content):
...
+
+        fields = ["Subsystem", "Kernel driver in use", "Kernel modules"]
+
         for line in get_active_lines(content):
             parts = line.split(None)
-            if parts[0] and parts[0] not in ['Subsystem:', 'Kernel'] \
+            if not parts:
+                continue
+
+            if parts[0] and parts[0] not in fields \
```
bfahr: @akshay196 I believe you can simplify this logic given that you have established that `parts` must have something in it in order to get here.  Also won't `parts[0]` just have the first word ('Subsystem:' and 'Kernel') due to the split on line 76? So `parts[0] not in fields` is not really useful here is it since it will always be `True`?

akshay196: @bfahr I have simplified this logic. Also `parts[0] not in fields` is removed which is not really necessary here. We are checking in fields in elif part later in the code. 

Thanks for pointing.

insights/parsers/lspci.py
```diff
@@ -69,14 +70,20 @@ def parse_content(self, content):
         # Parse kernel driver lines
         self.data = {}
         bus_device_function = ""
```
psachin: Better assign `None` than an empty string

insights/parsers/lspci.py
```diff
@@ -69,14 +70,20 @@ def parse_content(self, content):
...
+        bus_device_function_re = re.compile(r'[0-9a-f]+:[0-9a-f]+.[0-9a-f]+')
+
+        fields = ["Subsystem", "Kernel driver in use", "Kernel modules"]
+
         for line in get_active_lines(content):
             parts = line.split(None)
-            if parts[0] and parts[0] not in ['Subsystem:', 'Kernel'] \
-                    and len(parts[0].split(':')) == 2 and len(parts[0].split('.')) == 2:
+            if not parts:
```
psachin: @akshay196 Is this check required? because the `get_active_lines()` should handle the blank and the commented lines.

psachin: Please see below diff altogether which should get 100% coverage

```diff
diff --git a/insights/parsers/lspci.py b/insights/parsers/lspci.py
index f54d9aaf6edf..bde7eff145eb 100644
--- a/insights/parsers/lspci.py
+++ b/insights/parsers/lspci.py
@@ -69,15 +69,13 @@ class LsPci(CommandParser, LogFileOutput):
             scanner(self)
         # Parse kernel driver lines
         self.data = {}
-        bus_device_function = ""
+        bus_device_function = None
         bus_device_function_re = re.compile(r'[0-9a-f]+:[0-9a-f]+.[0-9a-f]+')
 
         fields = ["Subsystem", "Kernel driver in use", "Kernel modules"]
 
         for line in get_active_lines(content):
             parts = line.split(None)
-            if not parts:
-                continue
 
             if parts[0] and bus_device_function_re.match(parts[0]):
                 bus_device_function = parts[0]
diff --git a/insights/parsers/tests/test_lspci.py b/insights/parsers/tests/test_lspci.py
index 7347b1c9b0f2..125b0d6cd2d7 100644
--- a/insights/parsers/tests/test_lspci.py
+++ b/insights/parsers/tests/test_lspci.py
@@ -205,7 +205,18 @@ LSPCI_DRIVER_DOC = """
 """.strip()
 
 
+LSPCI_OUTPUT_WITH_BAD_LINES = """
+03:00.0 Network controller: Intel Corporation Wireless 7260 (rev bb)
+        Subsystem: Intel Corporation Dual Band Wireless-AC 7260
+
+        Kernel driver in use: iwlwifi
+# a comment
+        Kernel modules: iwlwifi
+""".strip()
+
+
 def test_lspci():
+    LsPci.token_scan('centrino', 'Centrino')
     lspci_0 = LsPci(context_wrap(LSPCI_0))
     assert [i['raw_message'] for i in lspci_0.get("Intel Corporation")] == INTEL.splitlines()
     assert len(lspci_0.get("Network controller")) == 1
@@ -237,6 +248,10 @@ def test_lspci_driver():
     assert len(lspci_obj.data) == 0
     assert len(lspci_obj.pci_dev_list) == 0
 
+    output = LsPci(context_wrap(LSPCI_OUTPUT_WITH_BAD_LINES))
+    assert len(output.data) == 1
+    assert len(output.pci_dev_list) == 1
+
 
 def test_doc_examples():
     env = {
```

insights/parsers/lspci.py
```diff
@@ -69,14 +70,20 @@ def parse_content(self, content):
...
+                continue
+
+            if parts[0] and bus_device_function_re.match(parts[0]):
```
bfahr: @akshay196 its not necessary to check `parts[0]` here since its already been checked and confirmed its not empty.

insights/parsers/lspci.py
akshay196: Done

insights/parsers/lspci.py
akshay196: Done.
@psachin Thanks for suggestions.

insights/parsers/lspci.py
akshay196: Done

psachin:
@bfahr Can you merge this please.

PR #1681, opened 2019-02-14
Add Ceph version from RHOSP 10 environment

psachin:
I noticed that in RHOSP 10 environment the output of command `ceph -v` is

  ceph version 10.2.5 (c461ee19ecbc0c5c330aca20f7392c9a00730367)

which is not listed in the `community_to_release_map` and hence cause one of the
playbook(insights-playbooks PR#1783) to fail locally during breakit. This change includes the ceph version as listed in latest RHOSP 10 Controller node.


psachin:
Including @mykaul for suggestions

mykaul:
LGTM - I don't think we need a unit test for it, but it's OK.

PR #1682, opened 2019-02-14
[ceph_version] Add latest released and latest update fields.

mykaul:
To be used later to ensure a customer is running the latest release version
(would be a warning) or the latest update of a specific release train (
again, a warning) I've added the facility and data for the release versions.

Signed-off-by: Yaniv Kaul <ykaul@redhat.com>


insights/parsers/tests/test_ceph_version.py
```diff
@@ -51,9 +51,12 @@ def test_ceph_version():
...
+    assert ceph_version8.latest_release == "This is not the latest released version"
+    assert ceph_version8.latest_update == "This is not the latest released update"
 
     ceph_version9 = CephVersion(context_wrap(CV9))
     assert ceph_version9.version == "3.1.1"
     assert ceph_version9.major == '3'
     assert ceph_version9.minor == "1"
-    assert ceph_version9.downstream_release == "1"
+    assert ceph_version9.latest_release == "This is not the latest released version"
+    assert ceph_version9.latest_update == "yes"
```
psachin: @mykaul Can we use Boolean here? It will be more reliable than comparing a string

```diff
diff --git a/insights/parsers/ceph_version.py b/insights/parsers/ceph_version.py
index 6f3ce855b056..beccf8312fcd 100644
--- a/insights/parsers/ceph_version.py
+++ b/insights/parsers/ceph_version.py
@@ -36,7 +36,7 @@ community_to_release_map = {
     "0.94.5-9": {'version': "1.3.2", 'major': '1.3', 'minor': '2', 'downstream_release': 'NA'},
     "0.94.9-3": {'version': "1.3.3", 'major': '1.3', 'minor': '3', 'downstream_release': 'NA'},
     "0.94.9-9": {'version': "1.3.3", 'major': '1.3', 'minor': '3', 'downstream_release': 'async 2'},
-    "0.94.10-2": {'version': "1.3.4", 'major': '1.3', 'minor': '4', 'downstream_release': 'NA', 'latest_update': 'yes'},
+    "0.94.10-2": {'version': "1.3.4", 'major': '1.3', 'minor': '4', 'downstream_release': 'NA', 'latest_update': True},
     "10.2.2-38": {'version': "2.0", 'major': '2', 'minor': '0', 'downstream_release': '0'},
     "10.2.3-13": {'version': "2.1", 'major': '2', 'minor': '1', 'downstream_release': '0'},
     "10.2.5-37": {'version': "2.2", 'major': '2', 'minor': '2', 'downstream_release': '0'},
@@ -46,14 +46,14 @@ community_to_release_map = {
     "10.2.10-16": {'version': "2.5", 'major': '2', 'minor': '5', 'downstream_release': '0'},
     "10.2.10-28": {'version': "2.5.1", 'major': '2', 'minor': '5', 'downstream_release': '1'},
     "10.2.10-40": {'version': "2.5.2", 'major': '2', 'minor': '5', 'downstream_release': '2'},
-    "10.2.10-43": {'version': "2.5.3", 'major': '2', 'minor': '5', 'downstream_release': '3', 'latest_update': 'yes'},
+    "10.2.10-43": {'version': "2.5.3", 'major': '2', 'minor': '5', 'downstream_release': '3', 'latest_update': True},
     "12.2.4-6": {'version': "3.0.2", 'major': '3', 'minor': '0', 'downstream_release': '2'},
     "12.2.4-10": {'version': "3.0.3", 'major': '3', 'minor': '0', 'downstream_release': '3'},
     "12.2.4-30": {'version': "3.0.4", 'major': '3', 'minor': '0', 'downstream_release': '4'},
     "12.2.4-42": {'version': "3.0.5", 'major': '3', 'minor': '0', 'downstream_release': '5'},
     "12.2.5-42": {'version': "3.1", 'major': '3', 'minor': '1', 'downstream_release': '0'},
-    "12.2.5-59": {'version': "3.1.1", 'major': '3', 'minor': '1', 'downstream_release': '1', 'latest_update': 'yes'},
-    "12.2.8-52": {'version': "3.2", 'major': '3', 'minor': '2', 'downstream_release': '0', 'latest_release': 'yes', 'latest_update': 'yes'}
+    "12.2.5-59": {'version': "3.1.1", 'major': '3', 'minor': '1', 'downstream_release': '1', 'latest_update': True},
+    "12.2.8-52": {'version': "3.2", 'major': '3', 'minor': '2', 'downstream_release': '0', 'latest_release': True, 'latest_update': True}
 }
 
 
@@ -101,5 +101,5 @@ class CephVersion(CommandParser):
         self.major = release_data['major']
         self.minor = release_data['minor']
         self.downstream_release = release_data['downstream_release']
-        self.latest_release = release_data.get('latest_release', 'This is not the latest released version')
-        self.latest_update = release_data.get('latest_update', 'This is not the latest released update')
+        self.latest_release = release_data.get('latest_release', False)
+        self.latest_update = release_data.get('latest_update', False)
diff --git a/insights/parsers/tests/test_ceph_version.py b/insights/parsers/tests/test_ceph_version.py
index d8747b0eb11b..8d946c3868c7 100644
--- a/insights/parsers/tests/test_ceph_version.py
+++ b/insights/parsers/tests/test_ceph_version.py
@@ -51,12 +51,12 @@ def test_ceph_version():
     assert ceph_version8.major == '2'
     assert ceph_version8.minor == "3"
     assert ceph_version8.downstream_release == "0"
-    assert ceph_version8.latest_release == "This is not the latest released version"
-    assert ceph_version8.latest_update == "This is not the latest released update"
+    assert ceph_version8.latest_release is False
+    assert ceph_version8.latest_update is False
 
     ceph_version9 = CephVersion(context_wrap(CV9))
     assert ceph_version9.version == "3.1.1"
     assert ceph_version9.major == '3'
     assert ceph_version9.minor == "1"
-    assert ceph_version9.latest_release == "This is not the latest released version"
-    assert ceph_version9.latest_update == "yes"
+    assert ceph_version9.latest_release is False
+    assert ceph_version9.latest_update is True
```

mykaul: We could use a boolean, but I did not want to, as I wanted to be flexible and provide a string that we can change, especially if it is NOT the latest release or update. For example, 'This is not the latest release and contains severe security vulnerabilities, please upgrade to XXX'.

If I reduce it to a boolean, I need to have all this logic back in the test, and not in the data. 
WDYT?

esakaiev:
test this please

insights/parsers/tests/test_ceph_version.py
csams: Hey @mykaul, let's separate the concern of what's wrong with a given release into a rule instead of the parser itself. Likewise for semantics like latest release and latest update. They'll be easier to track and update as a rule and maybe even automatically keep up-to-date by taking advantage of data feeds we can't depend on at the parser level.

In general, parsers should try to be as context-free as possible and just parse and represent the data. That way we don't have to remember to update core for every release of something we have a parser for unless the format of the raw data changes.

mykaul:
@csams - I initially thought similar to you - let's have this in a rule, but then I thought the rule ('Is this latest release' or 'Is this latest update') should always be the same - and the data is what's actually changing.

But I understand - I can of course make those rules instead. I suspect my line of thought was exactly the opposite ;-)

bfahr:
@mykaul are you still working on this PR?

mykaul:
No, gave up - I'll close it. I'll write, someday, a rule for this.

PR #1685, opened 2019-02-15
parser for vmcore-dmesg.txt

PetrLukac:
Simple parser needed for new diag rules. Specs for vmcore_dmesg already exists but wasn't used.


insights/parsers/vmcore_dmesg.py
```diff
@@ -0,0 +1,43 @@
+"""
+Crash log ``vmcore-dmesg.txt``
+==============================
+
+This module provides ``vmcore-dmesg.txt`` crash logs that are stored in ``/var/crash/[host]-YYYY-MM-DD-HH:MM:SS/``.
```
xiangce: To make the parser can be triggered, it's necessary to add this to the `insights.spec` module: `insights/specs/__init__.py`, `insights/specs/default.py` and `insights/specs/insights_archive.py` and/or `insights/specs/sos_archive.py`

insights/parsers/vmcore_dmesg.py
```diff
@@ -0,0 +1,43 @@
...
+    [  345.728570] PGD 0
+    [  345.729279] Oops: 0000 [#1] SMP
+    [  345.729950] Modules linked in: raid1 dm_raid raid456 async_raid6_recov async_memcpy async_pq raid6_pq async_xor xor async_tx
+    [  345.734752]  drm_kms_helper syscopyarea sysfillrect sysimgblt fb_sys_fops ttm drm bfa ahci crct10dif_pclmul crct10dif_common
+    [  345.737334] CPU: 6 PID: 952 Comm: systemd-udevd Kdump: loaded Not tainted 3.10.0-862.9.1.el7.x86_64 #1
+    [  345.738198] Hardware name: Dell Inc. PowerEdge R730xd/072T6D, BIOS 2.4.3 01/17/2017
+    [  345.739334] task: ffff9f888d030fd0 ti: ffff9f86b76b4000 task.ti: ffff9f86b76b4000
+
+The class inherites LogFileOutput and doesn't implement new methods/attributes.
+vmcore_dmesg defined in Specs allowes multioutput parsing and doesn't allow filtering.
```
xiangce: In normal, we need to filter these log files, since their size might be very large. Would you please to explain why is it not allowed to be filtered?

insights/parsers/vmcore_dmesg.py
xiangce: If you want to add it to "insights_archive.py" and ensure it can be collected by 'insights-client' tool, the `uploader.json` file in [insights-core-assets](https://github.com/RedHatInsights/insights-core-assets) repo should also be updated correspondingly. 
And a Jira Request should be opened at first as per the Insights Configuration Management Processes, to let the CEE/Engr make a Spec Review.

PetrLukac: Specs are already defined:
https://github.com/RedHatInsights/insights-core/blob/e17e542dd3010f1dc1293379f351b825c058a6d7/insights/specs/default.py#L817
https://github.com/RedHatInsights/insights-core/blob/5f536d47492e84cd130bd578f177749482e67316/insights/specs/__init__.py#L510
`insights/specs/insights_archive.py` and `insights/specs/sos_archive.py` don't seem to be necessary for this parser

insights/parsers/vmcore_dmesg.py
PetrLukac: Filtering seems rasonable, so it was added.
https://github.com/RedHatInsights/insights-core/blob/5f536d47492e84cd130bd578f177749482e67316/insights/specs/__init__.py#L510

bfahr: @xiangce after discussion with @PetrLukac this morning my understanding is that this will be used for sosreport analysis, and will not be collected by the Insights client.

PR #1703, opened 2019-02-25
Add new spec in qemu_xml parser

akshay196:
This includes:
* New parser ``VarQemuXML`` to parse XML files
  under "/var/run/libvirt/qemu/" directory.

* New parser is based on exising ``QemuXML`` parser.

### DIFF
--- a/insights/parsers/qemu_xml.py
+++ b/insights/parsers/qemu_xml.py
@@ -13,9 +13,27 @@ VarQemuXML - file ``/var/run/libvirt/qemu/*.xml``
 from .. import XMLParser, parser
 from insights.specs import Specs
 
+class BaseQemuXML(XMLParser):
+    """Base class for parsing Qemu XML files. It uses ``XMLParser`` mixin
+    class.
+
+    """
+    def parse_dom(self):
+        if self.dom is None:
+            return {}
+        else:
+            domain = {}
+            for child in self.dom:
+                if len(child) == 0:
+                    domain[child.tag] = child.text
+                else:
+                    domain[child.tag] = [c.items() for c in child]
+
+            return domain
+
 
 @parser(Specs.qemu_xml)
-class QemuXML(XMLParser):
+class QemuXML(BaseQemuXML):
     """This class parses xml files under ``/etc/libvirt/qemu/`` using
     ``XMLParser`` base parser.
 
@@ -148,26 +166,13 @@ class QemuXML(XMLParser):
         >>> memnode[1].get('mode') == 'strict'
         True
     """
-    def parse_dom(self):
-        if self.dom is None:
-            return
-        else:
-            domain = {}
-            for child in self.dom:
-                if len(child) == 0:
-                    domain[child.tag] = child.text
-                else:
-                    domain[child.tag] = [c.items() for c in child]
-
-            return domain
-
     @property
     def vm_name(self):
         return self.data.get('name', None)
 
 
 @parser(Specs.var_qemu_xml)
-class VarQemuXML(QemuXML):
+class VarQemuXML(BaseQemuXML):
     """This class parses xml files under ``/var/run/libvirt/qemu/`` using
     ``QemuXML`` base parser.
 
@@ -289,19 +294,6 @@ class VarQemuXML(QemuXML):
           </domain>
         </domstatus>
     """
-    def parse_dom(self):
-        if self.dom is None:
-            return
-        else:
-            domstatus = {}
-            for child in self.dom:
-                if len(child) == 0:
-                    domstatus[child.tag] = child.text
-                else:
-                    domstatus[child.tag] = [c.items() for c in child]
-
-            return domstatus
-
     @property
     def vm_name(self):
         if self.get_elements("./domain/name"):
--- a/insights/parsers/qemu_xml.py
+++ b/insights/parsers/qemu_xml.py
@@ -1,6 +1,14 @@
 """
-QemuXML file ``/etc/libvirt/qemu/*.xml``
-----------------------------------------
+QemuXML file ``/etc/libvirt/qemu/*.xml`` and ``/var/run/libvirt/qemu/*.xml``
+----------------------------------------------------------------------------
+
+Parsers provided by this module are:
+
+QemuXML - file ``/etc/libvirt/qemu/*.xml``
+------------------------------------------
+
+VarQemuXML - file ``/var/run/libvirt/qemu/*.xml``
+-------------------------------------------------
 """
 from .. import XMLParser, parser
 from insights.specs import Specs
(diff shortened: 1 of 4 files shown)


psachin:
@akshay196 Can you please override `parse_dom` so that we can have values in the `data` attrib similar to `QemuXML`

insights/parsers/qemu_xml.py
```diff
@@ -156,3 +164,145 @@ def parse_dom(self):
...
+            for child in self.dom:
+                if len(child) == 0:
+                    domstatus[child.tag] = child.text
+                else:
+                    domstatus[child.tag] = [c.items() for c in child]
+
+            return domstatus
+
+    @property
+    def vm_name(self):
```
xiangce: The same content as https://github.com/RedHatInsights/insights-core/pull/1632#discussion_r258813658

insights/parsers/qemu_xml.py
```diff
@@ -156,3 +164,145 @@ def parse_dom(self):
...
+            <seclabel type='dynamic' model='dac' relabel='yes'>
+              <label>+107:+107</label>
+              <imagelabel>+107:+107</imagelabel>
+            </seclabel>
+          </domain>
+        </domstatus>
+    """
+    def parse_dom(self):
+        if self.dom is None:
+            return
```
xiangce: Here an empty `{}` should be returned.

insights/parsers/qemu_xml.py
```diff
@@ -156,3 +164,145 @@ def parse_dom(self):
...
+              </input>
+            </devices>
+            <seclabel type='dynamic' model='dac' relabel='yes'>
+              <label>+107:+107</label>
+              <imagelabel>+107:+107</imagelabel>
+            </seclabel>
+          </domain>
+        </domstatus>
+    """
+    def parse_dom(self):
```
xiangce: The `parse_dom` is the same as the `QemuXML`, better make a base class and inherit the `QemuXML` and `VarQemuXML` from it.

akshay196: Done. (https://github.com/RedHatInsights/insights-core/pull/1703/commits/b4c0f79920c518606117b5d13bd064438baab474)

insights/parsers/qemu_xml.py
akshay196: Done

insights/parsers/qemu_xml.py
akshay196: Done (https://github.com/RedHatInsights/insights-core/pull/1703/commits/34f662efd44797ea0773d3b1606e7e855132612d)

insights/parsers/qemu_xml.py
```diff
@@ -1,13 +1,51 @@
...
+    def parse_content(self, content):
+        super(BaseQemuXML, self).parse_content(content)
+
+        # Setting vm_name attribute
+        self.vm_name = None
+        ret = self.get_elements('./domain/name') if self.dom else None
+        if ret:
+            self.vm_name = ret[0].text
+        else:
+            self.vm_name = self.data.get('name')
```
xiangce: The getting method of the `self.vm_name` depends on the content of parsed by the sub-classes, the above logic should be moved to the corresponding sub-class.

insights/parsers/qemu_xml.py
```diff
@@ -140,19 +178,130 @@ class QemuXML(XMLParser):
...
 
-    @property
-    def vm_name(self):
-        return self.data.get('name', None)
+@parser(Specs.var_qemu_xml)
+class VarQemuXML(BaseQemuXML):
+    """This class parses xml files under ``/var/run/libvirt/qemu/`` using
+    ``QemuXML`` base parser.
+
+    Sample file::
```
xiangce: Please simplify the content of the `Sample file::`, since these `pydoc` will be printed on the user's console, please make it as shorter as it is.

insights/parsers/qemu_xml.py
```diff
@@ -1,13 +1,51 @@
 """
-QemuXML file ``/etc/libvirt/qemu/*.xml``
-----------------------------------------
+QemuXML file ``/etc/libvirt/qemu/*.xml`` and ``/var/run/libvirt/qemu/*.xml``
+----------------------------------------------------------------------------
```
bfahr: This should be a line of `===` chars.  If you look at the rendered output it creates three headings at the same level instead of 1 major heading and two subheadings.

bfahr: This should be a line of `===` chars. If you look at the rendered output it creates three headings at the same level instead of 1 major heading and two subheadings.

insights/parsers/qemu_xml.py
```diff
@@ -140,19 +178,130 @@ class QemuXML(XMLParser):
...
 
-            return domain
 
-    @property
-    def vm_name(self):
-        return self.data.get('name', None)
+@parser(Specs.var_qemu_xml)
+class VarQemuXML(BaseQemuXML):
+    """This class parses xml files under ``/var/run/libvirt/qemu/`` using
+    ``QemuXML`` base parser.
```
bfahr: This docstring is correct, but the other class docstring needs to be udpated to indicate that it uses `BaseQemuXML` and not `XMLParser`.

insights/parsers/qemu_xml.py
```diff
@@ -1,13 +1,51 @@
...
+
+        # Setting vm_name attribute
+        self.vm_name = None
+        ret = self.get_elements('./domain/name') if self.dom else None
+        if ret:
+            self.vm_name = ret[0].text
+        else:
+            self.vm_name = self.data.get('name')
+
+    def parse_dom(self):
```
bfahr: Please add a docstring to this method so that it doesn't use the default docstring from the base class.

insights/parsers/qemu_xml.py
```diff
@@ -1,13 +1,51 @@
...
+class BaseQemuXML(XMLParser):
+    """Base class for parsing Qemu XML files. It uses ``XMLParser`` mixin
+    class.
+
+    """
+    def parse_content(self, content):
+        super(BaseQemuXML, self).parse_content(content)
+
+        # Setting vm_name attribute
+        self.vm_name = None
```
bfahr: Please document the `vm_name` class attribute.

insights/parsers/qemu_xml.py
xiangce: Just realized that I may be wrong. Both the 2 methods `self.get_elements('./domain/name')[0].text` and `self.data.get('name')` work for the 2 different `.xml` files, so this logic looks goo.

insights/parsers/qemu_xml.py
akshay196: Done

insights/parsers/qemu_xml.py
akshay196: Done

insights/parsers/qemu_xml.py
akshay196: Done

insights/parsers/qemu_xml.py
akshay196: Done

xiangce:
@akshay196 - You'd better resolve the conflicts in `insights/specs/sos_archive.py` before merging.

akshay196:
@xiangce Conflict resolved.

akshay196:
@psachin @bfahr @xiangce Thanks for reviewing.

PR #1705, opened 2019-02-26
Add datasource and parser for detecting httpd openfile on nfs v4

chenlizhong:


chenlizhong:
@kylape 

Revert the format of default.py

insights/parsers/tests/test_httpd_open_nfs.py
```diff
@@ -0,0 +1,25 @@
...
+from insights.parsers.httpd_open_nfs import HttpdOnNFSFilesCount
+from insights.tests import context_wrap
+import doctest
+
+http_nfs = """
+{"http_ids": [1787, 2399], "nfs_mounts": ["/data", "/www"], "open_nfs_files": 1000}
+""".strip()
+
+
+class TestHttpOpenNFS():
```
xiangce: It's not suggested to use a class to apply the unit test in insights-3.0, please remove it.

insights/parsers/httpd_open_nfs.py
```diff
@@ -0,0 +1,45 @@
...
+    Examples:
+        >>> httpon_nfs.data.get("http_ids") == [1787,2399]
+        True
+        >>> httpon_nfs.data.get("nfs_mounts") == ["/data", "/www"]
+        True
+        >>> httpon_nfs.data.get("open_nfs_files") == 1000
+        True
+
+    Attributes:
+        data (list): dict with keys "http_ids", "nfs_mounts" and "open_nfs_files"
```
xiangce: I believe it should be `data (dict):`

xiangce: Maybe you want to add another 3 attributes to return the 3 keys in the `dict` for fetching convenience and hide this `self.data`.

insights/specs/default.py
```diff
@@ -311,6 +311,27 @@ def httpd_cmd(broker):
...
+    @datasource(Mount)
+    def httpd_on_nfs(broker):
+        mnt = broker[Mount]
+        mps = mnt.search(mount_type='nfs4')
+        # get nfs 4.0 mount points
+        nfs_mounts = [m.mount_point for m in mps if m['mount_options'].get("vers") == "4.0"]
+        if nfs_mounts:
+            # get all httpd ps
+            out = broker[HostContext].shell_out("pgrep httpd")
+            httpd_pids = out.splits("\n")
```
xiangce: This `splits` is not necessary, since out is already `splitlines`-ed:


https://github.com/RedHatInsights/insights-core/blob/a442128998c39413f674a995dd81d0e493805704/insights/core/context.py#L149


Related traceback:
~~~
  File "/root/insights_dev/insights-core/insights/specs/default.py", line 324, in httpd_on_nfs
    httpd_pids = out.splits("\n")
AttributeError: 'list' object has no attribute 'splits'
~~~

insights/specs/default.py
```diff
@@ -311,6 +311,27 @@ def httpd_cmd(broker):
...
+            if httpd_pids:
+                open_nfs_files = 0
+                lsof_cmds = ["lsof -p {}".format(pid) for pid in httpd_pids if pid]
+                # maybe there are thousands open files
+                httpd_open_files = broker[HostContext].shell_out(lsof_cmds)
+                for line in httpd_open_files.split("\n"):
```
xiangce: Same comment as above.

xiangce:
Hacked a bit on the DS and tested it on a Satellite server:
```python
-        mps = mnt.search(mount_type='nfs4')
+        mps = mnt.search(mount_type='xfs')
         # get nfs 4.0 mount points
-        nfs_mounts = [m.mount_point for m in mps if m['mount_options'].get("vers") == "4.0"]
+        nfs_mounts = [m.mount_point for m in mps if m['mount_options'].get("rw")]
```
It works well:
~~~
insights-core]# insights-cat --no-header insights.specs.default.DefaultSpecs.httpd_on_nfs
{'http_ids': [u'1131', u'31481', u'31482', u'31483', u'31484', u'31485', u'31486', u'31487', u'31506', u'31507', u'31508', u'31509', u'31510', u'31511', u'31512', u'31513'], 'open_nfs_files': 103, 'nfs_mounts': [u'/', u'/boot']}
~~~

insights/parsers/httpd_open_nfs.py
```diff
@@ -0,0 +1,49 @@
...
+        True
+        >>> httpon_nfs.open_nfs_files == 1000
+        True
+
+
+    Attributes:
+        data (dict): dict with keys "http_ids", "nfs_mounts" and "open_nfs_files"
+    """
+
+    def parse_content(self, content):
```
xiangce: Would you please to think about using the `JSONParser` as the base class?

insights/parsers/httpd_open_nfs.py
```diff
@@ -0,0 +1,49 @@
...
+        if not content:
+            raise SkipException("Empty output from httpd_on_nfs datasource.")
+        try:
+            self.data = json.loads("".join(content))
+            self.http_ids = self.data.get("http_ids")
```
xiangce: These new defined attributes should be explained in the `Attributes` section of the docstring.

insights/parsers/httpd_open_nfs.py
chenlizhong: Yes

insights/specs/default.py
```diff
@@ -311,6 +311,26 @@ def httpd_cmd(broker):
...
+            if httpd_pids:
+                open_nfs_files = 0
+                lsof_cmds = ["lsof -p {}".format(pid) for pid in httpd_pids if pid]
+                # maybe there are thousands open files
+                httpd_open_files = broker[HostContext].shell_out(lsof_cmds)
+                for line in httpd_open_files:
+                    items = line.split()
+                    if len(items) > 8 and items[8].startswith(tuple(nfs_mounts)):
+                        open_nfs_files += 1
+                return {"http_ids": httpd_pids, "nfs_mounts": nfs_mounts, "open_nfs_files": open_nfs_files}
```
csams: Datasources defined with `@datasource` that will be invoked with `python -m insights.tools.cat` should return a [DatasourceProvider](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/spec_factory.py#L137) instance so they'll output correctly.

chenlizhong:
@csams @kylape 

Updated code. Please review again.

insights/specs/default.py
```diff
@@ -729,7 +752,6 @@ def block(broker):
     ss = simple_command("/usr/sbin/ss -tupna")
     ssh_config = simple_file("/etc/ssh/ssh_config")
     ssh_foreman_config = simple_file("/usr/share/foreman/.ssh/ssh_config")
-    ssh_foreman_proxy_config = simple_file("/usr/share/foreman-proxy/.ssh/ssh_config")
```
xiangce: This line should not be removed

chenlizhong: Done

PR #1707, opened 2019-02-26
Add VDSMImportLog parser

akshay196:
Parser to parse ``/var/log/vdsm/import/import-*.log`` files.

File contains logs related to VM imports from a KVM host, a VMWare
provider, or a Xen host, including import failure information.


insights/parsers/vdsm_log.py
```diff
@@ -156,3 +156,59 @@ def parse_lines(self, lines):
...
+        >>> len(log)
+        1
+        >>> log[0].get('raw_message', None)
+        '[    0.2] preparing for copy'
+        >>> log[0].get('vm_uuid', None)              # file: import-1f9efdf5-2584-4a2a-8f85-c3b6f5dac4e0-20180130T154807.log
+        '1f9efdf5-2584-4a2a-8f85-c3b6f5dac4e0'
+        >>> log[0].get('timestamp', None)
+        datetime.datetime(2018, 1, 30, 15, 48, 07)
+    """
+    time_format = '%Y%m%dT%H%M%S'
```
xiangce: This `self.time_format` is used for the `get_after` function inherited from the base class. It should be the format of the timestamp printed in the lines of the log. Please remove it and re-implement the `get_after` by referring to the [DmesgLineList's `get_after`](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/dmesg.py#L65)

insights/parsers/vdsm_log.py
```diff
@@ -156,3 +156,59 @@ def parse_lines(self, lines):
...
+    def _parse_line(self, line):
+        """Parse line into fields.
+
+        Dictionary with following keys:
+
+           raw_message (str): line
+           vm_uuid (str): UUID of imported VM.
+           timestamp (datetime): Timestamp that import began.
+        """
+        line_info = {'raw_message': line, 'vm_uuid': self.vm_uuid}
```
xiangce: The `vm_uuid` is a property of a `VDSMImportLog` object, but not a `key` of each line. Better just keep it as an attribute of this class instead of adding a duplicated `key` and `value` pair to each line.

insights/parsers/vdsm_log.py
```diff
@@ -156,3 +156,59 @@ def parse_lines(self, lines):
...
+
+        try:
+            line_info['timestamp'] = datetime.strptime(self.datetime, self.time_format)
```
xiangce: Same content as above for this `timestamp` key.

akshay196: @xiangce `[ 546.9]` is not time-stamp. For the log `[ 546.9] Finishing off`, 546.9 is elapsed time duration.

Let's not include `timestamp` key for each log line. I will update code to keep timestamp attribute to class instead of adding it in each line.

insights/parsers/vdsm_log.py
akshay196: @xiangce Correct. I have removed `self.time_format` in update commit. Also as `timestamp` attribute is specific to file and not for each line, we not need to implement `get_after`, right?

insights/parsers/vdsm_log.py
akshay196: Done

insights/parsers/vdsm_log.py
xiangce: It's also a kind of `timestamp` to record the duration time from the time the log file is created.

insights/parsers/vdsm_log.py
xiangce: If you dod not implement it, the version defined in the base class will be inherited. It's still a callable method of this sub-class. And ERROR will be thrown when calling it. As I said you need to re-implement it as per the [DmesgLineList's `get_after`](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/dmesg.py#L65), otherwise, re-implement it by a single `pass` and explain it in the docstring.

akshay196: Done. (https://github.com/RedHatInsights/insights-core/pull/1707/commits/9251595d1e8a46b10ef6b7eaa2fcf6155c7b5f48)

insights/parsers/vdsm_log.py
```diff
@@ -156,3 +156,91 @@ def parse_lines(self, lines):
...
+        """
+        super(VDSMImportLog, self).parse_content(content)
+        splited_file_name = self.file_name.split('-')
+        self.vm_uuid = '-'.join(splited_file_name[1:-1])
+        _datetime = splited_file_name[-1].replace('.log', '')
+
+        try:
+            self.file_datetime = datetime.strptime(_datetime, '%Y%m%dT%H%M%S')
+        except:
+            pass
```
xiangce: Here, the pass should be `self.file_datetime = None` OR raise a `ParseException`, since once `Exception` is thrown, there will be no such attribute.

akshay196: Done

insights/parsers/vdsm_log.py
```diff
@@ -156,3 +156,91 @@ def parse_lines(self, lines):
...
+        '1f9efdf5-2584-4a2a-8f85-c3b6f5dac4e0'
+        >>> vdsm_import_logs.file_datetime
+        datetime.datetime(2018, 1, 30, 15, 48, 07)
+    """
+    _line_re = re.compile(r'^(?:\[\s+(?P<timestamp>\d+\.\d+)\]\s+)?(?P<message>.*)$')
+
+    def parse_content(self, content):
+        """Parse ``import-@UUID-@datetime.log`` log file.
+
+        Attributes:
```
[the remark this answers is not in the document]
akshay196: Done

akshay196:
@xiangce Thank you for reviewing and merging.

PR #1709, opened 2019-02-26
Mangle egg less

jhjaggars:
This change changes the egg construction script in a way that includes all the functional bits of core including parsers and combiners as well as the tools.

The script still removes the archive/ and tests to keep the artifact small.  The script currently produces artifacts that are around 250k, my changes increase that to 750k.  The upstream egg is around 3 megabytes for comparison.


jhjaggars:
I'm not sure how I managed to build this PR against such an old commit. closing and trying again.

Issue #1712, opened 2019-02-27
insights/util/subproc.py", line 100: An unexpected keyword argument 'STDOUT'


xiangce:
~~~
  File "/root/insights_dev/insights-core/insights/util/subproc.py", line 100, in _build_pipes
    stdout = Popen(arg, bufsize=self.bufsize, stdin=stdout, stderr=STDOUT, STDOUT=PIPE, env=self.env).stdout
TypeError: __init__() got an unexpected keyword argument 'STDOUT'
~~~

xiangce:
Fixed in #1713

PR #1713, opened 2019-02-27
Fix a typo of stdout in subproc.py

xiangce:
- for #1712


bfahr:
@csams I believe this is a type unless you know of a reason it should not be changed.

PR #1728, opened 2019-03-06
Provide common base class for our plugin types

### LINKED ISSUE #1726: exceptions logged at warn level when calling get_canonical_facts
When calling [get_canonical_facts](https://github.com/RedHatInsights/insights-core/blob/master/insights/util/canonical_facts.py#L74), some exceptions which usually get logged at debug level are getting logged at warn level from the [main component evaluation loop](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/dr.py#L964). This is because some datasource failures are delayed until the first time they're accessed, which could be in the context of another component invocation (parser, etc.)

Here's an example of what's happening when the client attempts to collect canonical facts on platforms that aren't registered:

```
 File "/var/lib/insights/newest.egg/insights/core/dr.py", line 952, in run
 result = DELEGATES[component].process(broker)
 File "/var/lib/insights/newest.egg/insights/core/dr.py", line 673, in process
 return self.invoke(broker)
 File "/var/lib/insights/newest.egg/insights/core/plugins.py", line 90, in invoke
 return self.component(dep_value)
 File "/var/lib/insights/newest.egg/insights/core/__init__.py", line 82, in __init__
 self._handle_content(context)
 File "/var/lib/insights/newest.egg/insights/core/__init__.py", line 85, in _handle_content
 self.parse_content(context.content)
 File "/var/lib/insights/newest.egg/insights/core/spec_factory.py", line 110, in content
 self._content = self.load()
 File "/var/lib/insights/newest.egg/insights/core/spec_factory.py", line 331, in load
 timeout=self.timeout, env=self.create_env())
 File "/var/lib/insights/newest.egg/insights/core/context.py", line 142, in shell_out
 raw = self.check_output(cmd, timeout=timeout, keep_rc=keep_rc, env=env)
 File "/var/lib/insights/newest.egg/insights/core/context.py", line 137, in check_output
 keep_rc=keep_rc, env=env)
 File "/var/lib/insights/newest.egg/insights/util/subproc.py", line 208, in call
 res = p(keep_rc=keep_rc)
 File "/var/lib/insights/newest.egg/insights/util/subproc.py", line 121, in __call__
 raise CalledProcessError(rc, self.cmds[0], output)
CalledProcessError: (1, ['timeout', '-s', '9', '30', '/usr/bin/subscription-manager', 'identity'], "This system is not yet registered. Try 'subscription-manager register --help' for more information.\n")
```
It's just a warning from the logger, but it's an eyesore and not relevant for that particular collection.

csams:
Fixes #1726.

Provides a common base class for our custom plugin types (parser, rule, combiner, etc.) so that common exceptions raised by datasources are caught and logged at the appropriate level.

### DIFF
--- a/insights/core/plugins.py
+++ b/insights/core/plugins.py
@@ -43,10 +43,39 @@ class ContentException(dr.SkipComponent):
     pass
 
 
-component = dr.ComponentType
+class PluginType(dr.ComponentType):
+    """
+    PluginType is the base class of plugin types like datasource, rule, etc.
+    It provides a default invoke method that catches exceptions we don't
+    want bubbling to the top of the evaluation loop. These exceptions are
+    commonly raised by datasource components but could be in the context of any
+    component since most datasource runtime errors are lazy.
+
+    It's possible for a datasource to "succeed" and return an object but for an
+    exception to be raised when the parser tries to access the content of that
+    object. For example, when a command datasource is evaluated, it only checks
+    that the command exists and is executable. Invocation of the command itself
+    is delayed until the parser asks for its value. This helps with performance
+    and memory consumption.
+    """
+    def invoke(self, broker):
+        try:
+            return super(PluginType, self).invoke(broker)
+        except ContentException as ce:
+            log.debug(ce)
+            broker.add_exception(self.component, ce, traceback.format_exc())
+            raise dr.SkipComponent()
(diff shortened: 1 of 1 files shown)


gravitypriest:
Looks like this fixes the RHSM traceback when the system is unregistered :+1:

PR #1736, opened 2019-03-13
Handle invalid characters from command line

Rezney:
When accidentally passing in an invalid character e.g. in case of
copy-pasting from web and getting an em dash instead of -- in front
of "help" option we fail with traceback.

Let's print an user friendly error instead.

### DIFF
--- a/insights/__init__.py
+++ b/insights/__init__.py
@@ -338,6 +338,13 @@ def run(component=None, root=None, print_summary=False,
         else:
             raise
 
+    except UnicodeEncodeError as e:
+        if "codec can't encode character" in str(e):
+            msg = "Invalid character provided in command option/argument: {}"
+            log.error(msg.format(e))
+        else:
+            raise
+
 
 def main(argv=sys.argv[1:]):
     if "" not in sys.path:


bfahr:
@rezney nice catch.  What do think about checking for this condition earlier in the execution? Perhaps checking argv earlier in the execution chain?

Rezney:
@bfahr Thank you very much for reviewing my PR. Definitely agree with catching this earlier, I was not aware of the "command_parser" as it was introduced after I got the bits.

Could you please elaborate on your solution a bit? From the output it looks like even "run" subcommand is taken as argument and not recognized. Or did you mean to simply iterate through sys.argv and check if it can be encoded?

If I understand it correctly when "help" is not recognized as option it is passed into "_run" as "root" parameter. So it can be checked if the path, be it file or directory exists as we will get from "_magic.file(name)":

`cannot open `/tmp/insights-core/—help' (No such file or directory)`

But my knowledge of the code is "None" so I will leave it up to you :)

bfahr:
@Rezney I had to pull out the functionality that I referred to above.  I tried this again using unicode em/en dash, without your fix and was not able to duplicate the stack trace.  Are you able to reproduce the stack trace?  Here's what I get now:

```
$ insights-run –plugins insights.plugins
Install colorama if console colors are preferred.
usage: insights-run [-h] [-p PLUGINS] [-c CONFIG] [-i INVENTORY] [-v]
                    [-f FORMAT] [-s] [-D] [--context CONTEXT] [-m] [-t] [-d]
                    [-F]
                    [archive]
insights-run: error: unrecognized arguments: insights.plugins
```

I think we need to refactor the argparsing for all of the commands to consolidate this logic and make `insights-command` work like git, i.e., `insights command`.  Its possible that this issue might come back up in the future.

bfahr:
I did try again with `insights-run –h` and it does create the stack trace.  Probably because argparse processes the help command itself, and all other args are parsed by the insights code.

bfahr:
So bottom line is this is a problem where the `em dash h` is being treated as an archive and when insights tries to extract it, the exception is being thrown.  I've opened and issue #1832 for this bug.

Rezney:
@bfahr Thanks for the info. Please feel free to close this PR as a more robust solution / refactoring is needed.

bfahr:
@Rezney thanks, I'll go ahead and close this PR.

PR #1747, opened 2019-03-15
Add parser ls_origin_local_volumes_pods

shzhou12:


xiangce:
A related issue https://github.com/RedHatInsights/insights-core/issues/1748 is open, we'd better solve the issue at first.

shzhou12:
Test result of the spec:

~~~
(3.6) [root@master insights-core]# insights-cat --no-header ls_origin_local_volumes_pods
["5946c1f644096161a1242b3de0ee5875", "6ea3d5cd-d34e-11e8-a142-001a4a160152", "77d6d959-d34f-11e8-a142-001a4a160152", "7ad952a0-d34e-11e8-a142-001a4a160152", "7b63e8aa-d34e-11e8-a142-001a4a160152", "7d1f9443-d34f-11e8-a142-001a4a160152", "8e879171c85e221fb0a023e3f10ca276", "b6b60cca-d34f-11e8-a142-001a4a160152", "bc70730f-d34f-11e8-a142-001a4a160152", "dcf2fe412f6a174b0e1f360c2e0eb0a7", "ef66562d-d34f-11e8-a142-001a4a160152"]
~~~

insights/specs/default.py
```diff
@@ -418,6 +418,16 @@ def semid(broker):
     ls_etc = simple_command("/bin/ls -lanR /etc")
     ls_lib_firmware = simple_command("/bin/ls -lanR /lib/firmware")
     ls_ocp_cni_openshift_sdn = simple_command("/bin/ls -l /var/lib/cni/networks/openshift-sdn")
+    ls_origin_local_volumes_pods_base = listdir("/var/lib/origin/openshift.local.volumes/pods")
+
+    @datasource(ls_origin_local_volumes_pods_base)
```
xiangce: I think we have discussed it yesterday. This datasource is not needed. Use the `listdir` directly.

insights/parsers/ls_origin_local_volumes_pods.py
```diff
@@ -0,0 +1,67 @@
...
+        content = ''.join(content)
+        try:
+            self.data = json.loads(content)
+        except Exception:
+            raise ParseException("Incorrect content: '{0}'".format(content))
+
+        if not self.data:
+            raise SkipException("Empty or useless output.")
+
+        self.pods = self.data
```
xiangce: Is the `self.pods` necessary. Use `self.data` directly.

insights/specs/default.py
```diff
@@ -418,6 +418,7 @@ def semid(broker):
     ls_etc = simple_command("/bin/ls -lanR /etc")
     ls_lib_firmware = simple_command("/bin/ls -lanR /lib/firmware")
     ls_ocp_cni_openshift_sdn = simple_command("/bin/ls -l /var/lib/cni/networks/openshift-sdn")
+    ls_origin_local_volumes_pods = listdir("/var/lib/origin/openshift.local.volumes/pods")
```
xiangce: @csams - Would you please to have a confirm on this? if it's OK to use the `listdir` as a datasource directly? (instead of using the DatasourceProvider) It's tested OK on RHEL host, but I'm not sure if it works well with the client host.

bfahr: @xiangce the `listdir` spec is a datasource (see [__init__()](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/spec_factory.py#L664)) that returns a list of names with no attributes, so if that's all the parser needs then it should work.

insights/specs/insights_archive.py
```diff
@@ -104,6 +104,7 @@ class InsightsArchiveSpecs(Specs):
     ls_docker_volumes = simple_file("insights_commands/ls_-lanR_.var.lib.docker.volumes")
     ls_etc = simple_file("insights_commands/ls_-lanR_.etc")
     ls_ocp_cni_openshift_sdn = simple_file("insights_commands/ls_-l_.var.lib.cni.networks.openshift-sdn")
+    ls_origin_local_volumes_pods = simple_file("insights_commands/python_-m_insights.tools.cat_--no-header_ls_origin_local_volumes_pods")
```
bfahr: @shzhou12 this spec should be to reflect the mangled name of the simple command above.

insights/specs/default.py
shzhou12: @bfahr I updated the code according to your comments. I choose to use `simple_command("ls -l /var/lib/origin/openshift.local.volumes/pods")` instead, because it contains more information (OpenShift pods created time).

PR #1749, opened 2019-03-15
Reverting changes from PR #1727

bfahr:
* The previous PR introduced some issues with calling the run function
from ipython and python code
* All changes are being reverted


xiangce:
@bfahr - I have a question about this reverting, since there are several docs depends on it.

Will this feature `# insights [space] run/info/cat` be added again in the future?

PR #1750, opened 2019-03-15
initialize formatters var

gravitypriest:
before fix, the following exception appears when running the `canonical_facts` module:
```
UnboundLocalError: local variable 'formatters' referenced before assignment
```


bfahr:
@gravitypriest I just noticed this, thanks for submitting.

Issue #1697, opened 2019-02-22
python_-m_insights.tools.cat reports: "/bin/python: No module named tools"


xiangce:
Old client:
~~~
root@vm37-39 ~]# cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.2 (Maipo)
[root@vm37-39 ~]# rpm -qa | grep insights
redhat-access-insights-1.0.6-0.el7.noarch
[root@vm37-39 ~]# cat insights-vm37-39.gsslab.pek2.redhat.com-20190222010448/insights_commands/python_-m_insights.tools.cat_--no-header_rhev_data_center
/usr/bin/python: No module named insights.tools
~~~
New version client:
~~~
[root@liuxc-rhel-client-test ~]# cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.5 (Maipo)
[root@liuxc-rhel-client-test ~]# rpm -qa | grep insights
insights-client-3.0.3-9.el7_5.noarch
[root@liuxc-rhel-client-test ~]# cat insights-liuxc-rhel-client-test-20190222012531/insights_commands/python_-m_insights.tools.cat_--no-header_rhev_data_center
/bin/python: No module named tools
~~~

xiangce:
I just reinstalled the client tool and this issue still there. @csams @bfahr - Would please take a look at it?

bfahr:
The fix in #1698 will catch the error, but there's still a problem with the client egg build that needs to be fixed.  The path to the modules was added in #1639.  The problem in the client egg build is here: https://github.com/RedHatInsights/insights-core/blob/9cabaa5159651bd3efe8625fb821e225cc985a65/build_client_egg.sh#L15

jhjaggars:
https://github.com/RedHatInsights/insights-core/commit/a442128998c39413f674a995dd81d0e493805704 should resolve this.

xiangce:
Thank you @jhjaggars, for the updating. The `/bin/python: No module named tools` issue is gone, but a new issue comes:
~~~
[root@liuxc-insights-dev archives]# cat insights-liuxc-insights-dev-20190305222857/insights_commands/python_-m_insights.tools.cat_--no-header_*
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
[...]

PR #1753, opened 2019-03-19
Change sapcontrol to use foreach_execute

xiangce:
Reason to change this spec:
- due to issue #1697, the spec cannot be collected yet
- foreach_collect of client-v3 works well for this spec

API is not changed.


xiangce:
The related spec is removed from uploader.json by https://github.com/RedHatInsights/insights-core-assets/pull/116

PR #1762, opened 2019-03-26
Add spec ip_addr for sosreport

wushiqinlou:


wushiqinlou:
To get ip_addr info when executing insights-run or python module against sosreport, the spec from this PR is necessary.

insights/specs/sos_archive.py
```diff
@@ -52,6 +52,7 @@ class SosSpecs(Specs):
     hostname = first_of([simple_file("sos_commands/general/hostname_-f"), simple_file("sos_commands/general/hostname")])
     httpd_conf_sos = glob_file(["/conf/httpd/conf/httpd.conf", "/conf/httpd/conf.d/*.conf"])
     installed_rpms = simple_file("installed-rpms")
+    ip_addr = simple_file("sos_commands/networking/ip_-d_address")
```
xiangce: I saw there're at least two different files can be used for this spec in sosreport archives:
* sos_commands/networking/ip_address
* sos_commands/networking/ip_-d_address

Please have a check

PR #1771, opened 2019-03-28
Add spec to get disk type

smalleni:
This helps us know if a disk is rotational or SSD.

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -180,6 +180,7 @@ class DefaultSpecs(Specs):
     dirsrv = simple_file("/etc/sysconfig/dirsrv")
     dirsrv_access = glob_file("var/log/dirsrv/*/access")
     dirsrv_errors = glob_file("var/log/dirsrv/*/errors")
+    disk_type = glob_file("/sys/block/*/queue/rotational")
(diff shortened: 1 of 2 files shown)


insights/specs/default.py
```diff
@@ -180,6 +180,7 @@ def is_ceph_monitor(broker):
     dirsrv = simple_file("/etc/sysconfig/dirsrv")
     dirsrv_access = glob_file("var/log/dirsrv/*/access")
     dirsrv_errors = glob_file("var/log/dirsrv/*/errors")
+    disk_type = glob_file("/sys/block/*/queue/rotational")
```
bfahr: @smalleni you also need to add a RegistryPoint for the `disk_type` spec to the file `insights/specs/__init__.py`.

bfahr:
@smalleni are you adding a parser for this new spec?

smalleni:
@bfahr Not right now, Is it required to go along with the spec in this PR?

bfahr:
@smalleni have you had a chance to work on a parser related to your spec changes?

smalleni:
@bfahr I think we can close this as the same info was available through lsblk. Doesn't look like this is needed.

bfahr:
Closing PR as data is available from other sources.

PR #1775, opened 2019-03-29
Add cluster_prop_dict property for PCSConfig

chenlizhong:


insights/parsers/pcs_config.py
```diff
@@ -136,3 +138,15 @@ def parse_content(self, content):
...
+    def cluster_props_dict(self):
+        """
+        (dict): This function returns a dict of all cluster property and value
+        """
+        prop_result = {}
+        if "Cluster Properties" in self.data and self.data.get("Cluster Properties"):
+            for property_line in self.data.get("Cluster Properties"):
+                key, _, value = property_line.partition(":")
+                prop_result[key.strip()] = value.strip()
+        return prop_result
```
bfahr: @chenlizhong would it make sense to compute this as an attribute, once, in `parse_content`?

xiangce: @chenlizhong - better to make it as an attribute, or move the code into the `parse_content` to avoid the computation in every calling of it.

xiangce: @chenlizhong - if you forget to `git add` your updates to the last commit? Seems nothing is changing in that commit.

insights/parsers/pcs_config.py
```diff
@@ -66,6 +66,8 @@
     ['cluster-infrastructure: corosync', 'cluster-name: cluster-1', 'dc-version: 1.1.13-10.el7_2.4-44eb2dd', 'have-watchdog: false', 'no-quorum-policy: ignore', 'stonith-enable: true', 'stonith-enabled: false']
     >>> pcs_config.get("Colocation Constraints")
     ['clone-1 with clone-x (score:INFINITY) (id:clone-INFINITY)', 'clone-2 with clone-x (score:INFINITY) (id:clone-INFINITY)']
+    >>> 'have-watchdog' in pcs_config.cluster_props_dict
```
xiangce: nitpick: since the value of `have-watchdog` is `false` in the sample output, would it be better to change this example line to show the actual value instead of checking if it's `contain`ed?

```python
>>> pcs_config.cluster_props_dict['have-watchdog']
'false'
```

insights/parsers/pcs_config.py
```diff
@@ -78,6 +80,7 @@ class PCSConfig(CommandParser):
     """Class to process the output of ``pcs config`` command.
 
     Attributes:
+        cluster_props_dict (dict): A dictionary containing all cluster properties key, value
```
xiangce: Would you think the name `cluster_ properties` (used in #1796) better than this one?

PR #1777, opened 2019-03-29
Enhance spec for pssa_files

wushiqinlou:


wushiqinlou:
@csams - Got it.  I added some codes to handle tar files in pssa repo, and it works fine. Could you please approve the modification for `hostname`?

PR #1779, opened 2019-04-01
enhacement of code for teamdctl parser

ghost:


insights/specs/insights_archive.py
```diff
@@ -234,6 +234,8 @@ class InsightsArchiveSpecs(Specs):
     systemctl_qdrouterd = simple_file("insights_commands/systemctl_show_qdrouterd")
     systemctl_smartpdc = simple_file("insights_commands/systemctl_show_smart_proxy_dynflow_core")
     systool_b_scsi_v = simple_file("insights_commands/systool_-b_scsi_-v")
+    teamdctl_config_dump = glob_file("insights_commands/usr/bin/teamdctl_*_config_dump")
+    teamdctl_state_dump = glob_file("insights_commands/usr/bin/teamdctl_*_state_dump")
```
bfahr: @rasrivas-redhat you need to remove the root path, `/usr/bin` from these two specs, it is not included in the mangled archive spec.

bfahr (changes requested):

insights/specs/insights_archive.py
rasrivas-redhat: @bfahr - I have updated the code, by mistakenly it happened.

xiangce (approved):

PR #1781, opened 2019-04-01
Fix exception when formatting output as yaml

bfahr:
* When formatting output as yaml, the yaml parser tries to serialize
some of the skip info and fails. It seems to fail on LogFileOutput
parsers only.   Adding a Yaml Representer for ScanMeta appears to fix the
problem.


xiangce:
It works well on my test env: the exception is gone when executing `insights-run` against an archive which would hit a rule depends on any `LogFileOutput` parsers.

All tests of python2.7 and python3.6 work well too. The Jenkins failed due to `not enough space`, ignore them.

PR #1782, opened 2019-04-03
Update spec for krb5

wushiqinlou:


xiangce:
@wushiqinlou - this modification requires all the files under `/etc/krb5.conf.d/` have the same format with the file `/etc/krb5.conf`, if you confirmed that, please modify the related specs in the `uploader.json` as well.

wushiqinlou:
https://github.com/RedHatInsights/insights-core-assets/pull/126

PR #1786, opened 2019-04-03
implement tower venv rule

chrismeyersfsu:
* Adds a spec to pip freeze all tower virtualenvs
* Adds a spec to read Ansible Tower version file
* Adds a parser to process pip freeze
* Adds a parser to process tower version file

### DIFF
--- a/insights/parsers/pip_freeze.py
+++ b/insights/parsers/pip_freeze.py
@@ -1,0 +1,62 @@
+"""
+pip freeze - Command
+====================
+
+This module provides the python packages installed in a virtualenv, gathered
+from the ``pip freeze`` command.
+
+The parser breaks out each python package into <pkg_name> and <version>, stored
+in two parallel arrays.
+
+Typical output of pip freeze is::
+    requests==2.21.0
+    scandir==1.10.0
+    simplegeneric==0.8.1
+    six==1.12.0
+    traitlets==4.3.2
+    typing==3.6.6
+    urllib3==1.24.1
+    wcwidth==0.1.7
+
+Attributes:
+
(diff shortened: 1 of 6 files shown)


insights/specs/default.py
```diff
@@ -810,6 +810,9 @@ def tomcat_home_base(broker):
 
     tomcat_vdc_targeted = foreach_execute(tomcat_home_base, "/bin/grep -R -s 'VirtualDirContext' --include '*.xml' %s")
     tomcat_vdc_fallback = simple_command("/usr/bin/find /usr/share -maxdepth 1 -name 'tomcat*' -exec /bin/grep -R -s 'VirtualDirContext' --include '*.xml' '{}' +")
+    tower_virtualenvs = listdir(os.path.join(BASE_DIR, 'venv'))
```
csams: BASE_DIR is undefined

insights/parsers/pip_freeze.py
```diff
@@ -0,0 +1,62 @@
...
+@parser(Specs.tower_pip_freeze)
+class PipFreeze(Parser):
+    def parse_content(self, content):
+
+        # For testing. I couldn't get at this class to mock the args
+        venv_dir = 'foobar'
+        if hasattr(self, 'args') and self.args:
+            venv_dir = self.args
+
+        self.venv_dir = os.path.join(BASE_DIR, venv_dir)
```
csams: We might have to do some hacking here to extract `venv_dir` from `self.file_path` if `self.args` isn't defined.

xiangce:
Closing it, please open it in another new PR when necessary.

PR #1745, opened 2019-03-14
Parser for parsing command 'ip netns exec <namespace> lsof i'

ppandey-redhat:
Parser for parsing output of command `ip netns exec <namespace > lsof -i`

PR #1788, opened 2019-04-04
Updated parser ip_netns_exec_namespace_lsof to add specs for insights_archive

ppandey-redhat:
Parser for parsing output of command `ip netns exec <namespace > lsof -i`.
Merged Parser PR: https://github.com/RedHatInsights/insights-core/pull/1745

This PR adds spec in insights_archive for parser **ip_netns_exec_namespace_lsof**.


ppandey-redhat:
> @ppandey-redhat - this update looks good, but unfortunately, this kind of specs that used `foreach_execute` cannot be collected yet. So please don't expect the related rules depend on this Spec could be hit.

@xiangce - Yes, I agree with you that the specs which use `foreach_execute` cannot be collected yet. So, the dependent rules will not hit. I have raised this PR to include specs in `insights_archive` so that the rule can be tested using `insights-run` command with a customized archive/sos-report.

PR #1796, opened 2019-04-05
Added new cluster_properties property to pcs_config parser to return a dict

lhuett:
- Added new cluster_properties property to pcs_config parser to return a
  dict. 
- Updated examples and tests.


insights/parsers/pcs_config.py
```diff
@@ -136,3 +138,16 @@ def parse_content(self, content):
 
     def get(self, key):
         return self.data.get(key)
+
+    @property
+    def cluster_properties(self):
```
xiangce: Same content as https://github.com/RedHatInsights/insights-core/pull/1775#discussion_r271077063.

It would be good to compute it once in `parse_content` for saving duplicated computation in each of calling of this API.

xiangce: @lhuett - Seems @chenlizhong has moved the `computation` into the `paser_content` already, if the #1775 works to you, would you mind to close this PR?

lhuett: Sure thing, thanks for the heads up @xiangce.

PR #1806, opened 2019-04-09
Add parser ceph_osd_tree_text and combiner ceph_osd_tree

shzhou12:


insights/parsers/ceph_osd_tree_text.py
```diff
@@ -0,0 +1,114 @@
...
+                prev_line = row_num - 1
+                parents.append(prev_line)
+                nodes[parents[-1]]['children'].insert(0, int(node['id']))
+            elif shift == len(parents):
+                nodes[parents[-1]]['children'].insert(0, int(node['id']))
+            elif shift < len(parents):
+                parents.pop()
+                nodes[parents[-1]]['children'].insert(0, int(node['id']))
+            # update type name
+            if node['name'].startswith('osd.'):
```
xiangce: as per our discussion, the `osd` is not the only type, right?

PR #1809, opened 2019-04-09
uuid should be a string

jhjaggars:


insights/util/canonical_facts.py
```diff
@@ -16,7 +16,7 @@ def valid_uuid_or_None(s):
     if not s:
         return None
     try:
-        return uuid.UUID(s)
+        return uuid.UUID(s).hex
```
SteveHNH: This will produce it without the hyphens. Not sure if that matters, but we need to do `str(uuid.UUID(s))` if we want the hyphens in there.

PR #1810, opened 2019-04-09
only call branch_info when connecting through sat

gravitypriest:


insights/tests/client/auto_config/test_branch_info_call.py
```diff
@@ -0,0 +1,22 @@
...
+    '''
+    config = Mock(base_url=None, upload_url=None, legacy_upload=False, insecure_connection=False)
+    set_auto_configuration(config, 'test.com:443/redhat_access', 'some_cert', None)
+    connection.return_value.get_branch_info.assert_called_once()
+
+
+@patch("insights.client.auto_config.InsightsConnection")
+def test_rhsm_branch_info_not_called(connection):
+    '''
+    When ca cert is None, means we're on direct RHSM. get_bgranch_info should not be called.
```
Jason-RH: small typo - 'get_bgranch_info'

insights/client/auto_config.py
```diff
@@ -66,6 +66,11 @@ def set_auto_configuration(config, hostname, ca_cert, proxy):
         saved_proxy = config.proxy
         config.proxy = proxy
     config.base_url = hostname + '/r/insights'
+    if ca_cert is None:
```
lphiri: I think it would have been clearer to pass in an explicit 'is_satellite_managed' flag  since strictly speaking  direct connected systems also have a CA cert, its just that we handle it differently.  I am ok with shipping this as is though.

gravitypriest: agreed

PR #1811, opened 2019-04-09
Adding log messages when patterns and keywords are included in remove…

Jason-RH:
….conf to resolve BZ-1684429


Jason-RH:
@gravitypriest Look reasonable to you?

PR #1814, opened 2019-04-11
Skip the bad_lines for CommandParser

xiangce:
- raising SkipException instead of ContentException


bfahr:
@xiangce after discussing with @miclark and @csams this morning I don't believe we want to make this change.  It would impact our ability to detect issues with spec and collection.  We can however change how we report on the issues collected to make research of the collection issue more effective.

xiangce:
@bfahr - Thanks for the update. I closed it.

PR #1815, opened 2019-04-11
Adding command to get current PAGE_SIZE

chawlanikhil24:
This is a required paser to get the current memory PAGE_SIZE.


insights/parsers/getconf.py
```diff
@@ -0,0 +1,21 @@
+"""
+getconf PAGE_SIZE - command ``/usr/sbin/getconf PAGE_SIZE``
+=============================================
```
xiangce: Better keep the '==' line and the title line in the same length.

insights/parsers/getconf.py
```diff
@@ -0,0 +1,21 @@
...
+
+This very simple parser returns the output of the ``getconf PAGE_SIZE`` command.
+
+    >>> output = shared[getconfpagesize]
```
xiangce: An `Examples:` section header is necessary for these examples. 

See https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

insights/parsers/getconf.py
```diff
@@ -0,0 +1,21 @@
...
+    >>> output['value']
+    '4096'
+"""
+from .. import parser
+from insights.specs import Specs
+
+
+@parser(Specs.getconf_page_size)
+def getconfpagesize(context):
```
xiangce: Better refactor this Parser method to a `Class` and provide an attribute e.g. `value`. (e.g. https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/hostname.py)

insights/specs/insights_archive.py
```diff
@@ -58,6 +58,7 @@ class InsightsArchiveSpecs(Specs):
     fdisk_l = simple_file("insights_commands/fdisk_-l")
     foreman_rake_db_migrate_status = simple_file('insights_commands/foreman-rake_db_migrate_status')
     getcert_list = simple_file("insights_commands/getcert_list")
+    getconf = simple_file("insights_commands/getconf_pagesize")
```
xiangce: Its simple_file should be `insights_commands/getconf_PAGE_SIZE`, you could get its correct format via the `mangle` tool of the core. E.g.

~~~
# source bin/activate
# mangle "/usr/bin/getconf PAGE_SIZE"
insights_commands/getconf_PAGE_SIZE
~~~

insights/specs/default.py
```diff
@@ -256,6 +256,7 @@ def docker_container_ids(broker):
     fstab = simple_file("/etc/fstab")
     galera_cnf = first_file(["/var/lib/config-data/puppet-generated/mysql/etc/my.cnf.d/galera.cnf", "/etc/my.cnf.d/galera.cnf"])
     getcert_list = simple_command("/usr/bin/getcert list")
+    getconf = simple_command("/usr/bin/getconf PAGE_SIZE")
```
xiangce: You may want to name it as `getconf_page_size`

insights/specs/insights_archive.py
xiangce: And you may want to name it as `getconf_page_size`

insights/parsers/getconf_pagesize.py
```diff
@@ -0,0 +1,32 @@
+"""
+getconf PAGE_SIZE - command ``/usr/sbin/getconf PAGE_SIZE``
```
xiangce: Nitpick: Would you mind to change this title to "GetconfPageSize" correspondingly

insights/parsers/getconf_pagesize.py
```diff
@@ -0,0 +1,32 @@
...
+from .. import parser, CommandParser
+from insights.specs import Specs
+
+
+@parser(Specs.getconf_page_size)
+class GetconfPageSize(CommandParser):
+    """Class for parsing 'getconf PAGE_SIZE' command output
+
+    Output: page_size
+    """
```
xiangce: Since the `page_size` is a valuable attribute provided by this parser, better add `Attributes` docstring for it.

For details, please see: https://www.sphinx-doc.org/en/1.8/usage/extensions/example_google.html?highlight=Attributes.
Thanks.

docs/shared_parsers_catalog/getconf_page_size.rst
```diff
@@ -0,0 +1,3 @@
+.. automodule:: insights.parsers.getenforce
```
bfahr: @chawlanikhil24 the module name needs to match your parser module, `insights.parsers.getconf_pagesize`.

insights/parsers/getconf_pagesize.py
bfahr: For example,
```
Attributes:
    page_size (int): integer value of the current configuration page size.
```

insights/parsers/getconf_pagesize.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def parse_content(self, content):
+        if len(content) != 1:
+            msg = "getconf PAGE_SIZE output contains multiple non-empty lines"
+            raise ParseException(msg)
+        raw = content[0].strip()
+        self.page_size = raw
```
bfahr: Would it make sense to go ahead and convert this to an integer value here so that developers don't have to perform this conversion in their rules?

bfahr:
The [spec](https://github.com/RedHatInsights/insights-core-assets/pull/151) for this has been merged

insights/parsers/tests/test_getconf_pagesize.py
```diff
@@ -0,0 +1,25 @@
...
+    print(result)
+
+class Testgetconfpagesize():
+    def test_getconf_PAGESIZE1(self):
+        result = GetconfPageSize(context_wrap(GETCONFPAGESIZE1))
+        assert result.get("page_size") == 4096
+
+    def test_getconf_PAGESIZE2(self):
+        result = GetconfPageSize(context_wrap(GETCONFPAGESIZE2))
+        assert result.get("page_size") == 16384
```
bfahr: You should also add a test that the `ParseException` is raised when `len(content) != 1`.  See [this example](https://github.com/RedHatInsights/insights-core/blob/4ae338b4345191ded4c36852988eaaa1209b9e53/insights/parsers/tests/test_gluster_vol.py#L90).

docs/shared_parsers_catalog/getconf_page_size.rst
```diff
@@ -0,0 +1,3 @@
+.. automodule:: insights.parsers.getconf_pagesize
```
xiangce: The file name should be renamed to `getconf_pagesize.rst` in which the basename should be the same as the parser.

insights/parsers/tests/test_getconf_pagesize.py
```diff
@@ -0,0 +1,20 @@
...
+GETCONFPAGESIZE1 = """
+4096
+""".strip()
+
+GETCONFPAGESIZE2 = """
+16384
+""".strip()
+
+
+class Testgetconfpagesize():
```
xiangce: Would you mind to remove the `class` and change to test it with functions directly?

insights/parsers/getconf_pagesize.py
```diff
@@ -0,0 +1,35 @@
+"""
+GetconfPageSize - command ``/usr/sbin/getconf PAGE_SIZE``
+===========================================================
+
+This very simple parser returns the output of the ``getconf PAGE_SIZE`` command.
+
+Examples:
+
+    >>> pagesize_parsed.page_size
+    '4096'
```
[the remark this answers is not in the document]
xiangce: @chawlanikhil24 - As discussed, better add a test for the doc.

PR #1817, opened 2019-04-11
New parser aws_instance_type

xiangce:


lonicerae:
Just some reference: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

xiangce:
@lhuett - Thank you for the check. I have updated to check the `CloudProvider` only in https://github.com/RedHatInsights/insights-core/pull/1829.

PR #1822, opened 2019-04-12
New parser for checking io_timeout of nvme_core

lonicerae:


insights/parsers/nvme_core_io_timeout.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@parser(Specs.nvme_core_io_timeout)
+class NVMeCoreIOTimeout(Parser):
+    """
+    Class for parsing the content of the ``/sys/module/nvme_core/parameters/io_timeout``.
+
+    A typical sample of the content of this file looks like::
+
+        30
```
xiangce: Better use the same value with the one in `Examples`. (line-31)

lonicerae: Okay, got it.

PR #1823, opened 2019-04-12
Adding Cloud Provider combiner and tests

lhuett:
Uses the following parsers to to get facts to determine cloud provider
- InstallledRpms
- YumRepoList 
- DMIDecode

### DIFF
--- a/insights/combiners/cloud_provider.py
+++ b/insights/combiners/cloud_provider.py
@@ -1,0 +1,170 @@
+"""
+Cloud Provider
+===============
+
+Combiner for Cloud information. It uses the results of
+:class:`InstalledRpms`,
+:class:`YumRepoList` and
+:class:`DMIDecode` parsers
+The combiner uses these parsers determine the Cloud Provider.
+
+Examples:
+    >>> cp.cp_bios_vendor
+    {'aws': 'amazon', 'google': '', 'azure': ''}
+    >>> cp.cp_bios_version
+    {'aws': '4.2.amazon', 'google': '', 'azure': ''}
+    >>> cp.cp_rpms
+    {'aws': ['rh-amazon-rhui-client-2.2.124-1.el7'], 'google': [], 'azure': []}
+    >>> cp.cp_yum
+    {'aws': [], 'google': [], 'azure': ['rhui-microsoft-azure-rhel7-2.2-74']}
+    >>> cp.cp_asset_tag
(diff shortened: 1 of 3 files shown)


insights/combiners/tests/test_cloud_provider.py
```diff
@@ -0,0 +1,422 @@
...
+
+DMIDECODE_FAIL = "# dmidecode 2.11\n# No SMBIOS nor DMI entry point found, sorry.\n"
+
+
+def test_rpm_google():
+    irpms = IRPMS(context_wrap(RPMS_GOOGLE))
+    dmi = DMIDecode(context_wrap(DMIDECODE))
+    yrl = YumRepoList(context_wrap(YUM_REPOLIST_NOT_AZURE))
+    ret = CloudProvider(irpms, dmi, yrl)
+    assert ret.cloud_provider is 'google'
```
csams: Use `==` for equality between strings. `is` does an address comparison, which means it might work or not depending on whether the string is in the internal cache.

insights/combiners/cloud_provider.py
```diff
@@ -0,0 +1,170 @@
...
+from insights.parsers.installed_rpms import InstalledRpms
+from insights.parsers.dmidecode import DMIDecode
+from insights.parsers.yum import YumRepoList
+from collections import namedtuple
+
+
+@combiner(optional=[InstalledRpms, DMIDecode, YumRepoList])
+class CloudProvider(object):
+    """
+    Combiner class to provide cloud vedor facts
```
csams: Let's include the different values that `cp.cloud_provider` can be so users will know which things they can test when using it. Or you might have them as constant module or class attributes or something (`CloudProvider.AWS`, `CloudProvider.GOOGLE`, etc.) so users can do `if cp.cloud_provider == CloudProvider.AWS)`.

(Also small typo with "vendor.")

beav:
tests I'm working on:

- [ ] aws

- [ ] azure

- [ ] google

- [ ] "regular" non-cloud archive

insights/combiners/tests/test_cloud_provider.py
lhuett: Thanks for the input Chris, made suggested changes.

insights/combiners/cloud_provider.py
lhuett: Thanks for the input Chris, made suggested changes.

PR #1828, opened 2019-04-16
Parser for listing /var/opt/mssql

psachin:


psachin:
Jenkins job failed due to ` No space left on device`

insights/parsers/ls_var_opt_mssql.py
```diff
@@ -0,0 +1,30 @@
+"""
+LsVarOptMSSql - command ``ls -laR /var/opt/mssql``
+==================================================
+
+The ``ls -laR /var/opt/mssql`` command provides information for the listing of
+the ``/var/opt/mssql`` directory. See ``FileListing`` class for addtional information.
+
+
+Sample ``ls -laR /var/opt/`` output::
```
lonicerae: Should be `/var/opt/mssql`?

psachin: @lonicerae Looks like I missed the full path. Fixed in the latest commit.

insights/parsers/ls_var_opt_mssql.py
```diff
@@ -0,0 +1,30 @@
...
+    'mssql'
+"""
+from insights.specs import Specs
+from .. import parser, CommandParser
+from .. import FileListing
+
+
+@parser(Specs.ls_var_opt_mssql)
+class LsVarOptMSSql(CommandParser, FileListing):
+    """Parses output of ``ls -laR /var/opt/mssql`` command."""
```
xiangce: Better add a `pass` for empty implementation.

insights/specs/default.py
```diff
@@ -452,6 +452,7 @@ def semid(broker):
     ls_var_lib_mongodb = simple_command("/bin/ls -la /var/lib/mongodb")
     ls_R_var_lib_nova_instances = simple_command("/bin/ls -laR /var/lib/nova/instances")
     ls_var_lib_nova_instances = simple_command("/bin/ls -laRZ /var/lib/nova/instances")
+    ls_var_opt_mssql = simple_command("/bin/ls -laR /var/opt/mssql")
```
xiangce: @psachin - is it necessary to `recursively` list all the items in this dir? I'm afraid the size with `-R` would be too large.

Does the `ls -la /var/opt/mssql` work for you?

psachin:
@xiangce  I plan to raise a JIRA ticket for specs after the parse PR approval.

psachin:
@lonicerae @xiangce This parser has very limited usage and commands like `ls -ld /var/opt/mssql` & `ls -la /var/opt/mssql` cannot be used to derive ownership of the directory(we decided to remove `-R` because it may collect large data). Hence I've used the datasource which is a good choice for this kind of implementation.

insights/specs/default.py
```diff
@@ -672,6 +672,20 @@ def rhev_data_center(broker):
...
+    def ownership_of_var_opt_mssql(broker):
+        """Return the ownership of /var/opt/mssql if it is not mssql:mssql.
+        """
+        import json
+        root = broker[HostContext].root
+        relative_path = "var/opt/mssql"
+        path = os.path.join(root, relative_path)
+        if os.path.isdir(path):
+            name, group = get_owner(path)
+            if (name, group) != ("mssql", "mssql"):
```
chenlizhong: Do not add logic here.  just return name/group in datasource.
Logic is rule thing.

insights/specs/default.py
```diff
@@ -672,6 +672,20 @@ def rhev_data_center(broker):
...
+                return DatasourceProvider(content=json.dumps({"owner": name, "group": group}), relative_path=relative_path)
```
chenlizhong: Best to return dir also.

insights/parsers/var_opt_mssql.py
```diff
@@ -0,0 +1,29 @@
...
+=====================================================================
+"""
+import json
+from insights.specs import Specs
+from insights.parsers import SkipException
+from .. import CommandParser, parser
+
+
+@parser(Specs.ownership_of_var_opt_mssql)
+class BadOwnershipOfVarOptMSSql(CommandParser):
```
chenlizhong: The parser is too specific.  if we check other dir/file ower/group, need repeat the codes.

chenlizhong: Maybe we create one base class for owner/group checking.  Others just inherit it.

xiangce:
@psachin - Since the new `datasource` (`python -m insights.tools.cat`) will execute the python code on the customer's host, which is not suggested, especially when there is another regular method to achieve the goal. In this case, you could just modify the command to `ls -ld /var/opt/mssql`, the necessary data can be collected.

psachin:
Hi @csams This is the parser I wanted to discuss. Thanks for quick discussion that helped me clear the doubt. The missed the 'Note' section under https://insights-core.readthedocs.io/en/latest/api_index.html#insights.core.FileListing

psachin:
@chenlizhong @lonicerae @xiangce Please review again.

xiangce:
My fault, this PR should not be merged before the JIRA request is created. Please create the JIRA asap.

psachin:
JIRA: https://projects.engineering.redhat.com/browse/INSPEC-195

PR #1841, opened 2019-04-17
Enhance combiner redhat release

vishwanathjadhav:
Most of the existing plugins cannot be enhanced for RHEL8 because `RedHatRelease` combiner does not support RHEL-8 kernel. Added RHEL-8 kernel(**4.18.0-70.el8**) in Uname parser's `kernel to RHEL release map`. The kernel version is from following [kernel-rhel8](http://git.app.eng.bos.redhat.com/git/kernel-rhel8.git) releases. This version is temporary and will get change after final RHEL-8GA release.


insights/parsers/uname.py
```diff
@@ -97,7 +97,9 @@
     "3.10.0-514": "7.3",
     "3.10.0-693": "7.4",
     "3.10.0-862": "7.5",
-    "3.10.0-957": "7.6"
+    "3.10.0-957": "7.6",
+    "4.18.0-70": "8.0",
+    "4.18.0-74": "8.0"
```
xiangce: These versions are only for RC# of RHEL-8, which won't be released. Only the version for GA is the target of insights.

Please use the RedHatRelease as a workaround. and close this PR.

vishwanathjadhav: @xiangce - ack

PR #1842, opened 2019-04-17
New parser and spec for bind crypto policy in RHEL8.

jsvob:


insights/parsers/tests/test_crypto_policies_bind.py
```diff
@@ -0,0 +1,38 @@
+# TODO
```
skontar: Remove TODO

jsvob: Ouch :D Thanks!

skontar:
Hi all. Can we please try to speed up the process? This spec is needed for a rule which is being written for the summit demo. Thanks! @bfahr @xiangce

PR #1848, opened 2019-04-19
Add parser ps_aexww

shzhou12:


insights/parsers/ps.py
```diff
@@ -342,3 +342,42 @@ def parse_content(self, content):
...
+        24477 ?        Ssl    6:56 /usr/bin/openshift-router PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin HOSTNAME=node1 RELOAD_INTERVAL=10s
+
+    Examples:
+        >>> type(ps_aexww)
+        <class 'insights.parsers.ps.PsAexww'>
+        >>> ps_aexww.data[5]['ENVIRON']['RELOAD_INTERVAL']
+        '10s'
+    """
+    command_name = 'COMMAND'
+    user_name = 'PID'
```
xiangce: `user_name` is not `PID`.
As per the definition of `user_name`:
~~~
``user_name`` is the name of the first column from the header of ps output,
the subclass must override it correspondingly
~~~

Would you think it's reasonable to rename it as `first_column`?

xiangce: And since there is no `user` info provided in this `ps_aexwww`, you should reimplement the `users` method, to raise an Exception or others.

insights/parsers/ps.py
```diff
@@ -342,3 +342,42 @@ def parse_content(self, content):
...
+            2 ?        S      0:00 [kthreadd]
+            3 ?        S      0:00 [ksoftirqd/0]
+            5 ?        S<     0:00 [kworker/0:0H]
+            7 ?        S      0:01 [migration/0]
+        24477 ?        Ssl    6:56 /usr/bin/openshift-router PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin HOSTNAME=node1 RELOAD_INTERVAL=10s
+
+    Examples:
+        >>> type(ps_aexww)
+        <class 'insights.parsers.ps.PsAexww'>
+        >>> ps_aexww.data[5]['ENVIRON']['RELOAD_INTERVAL']
```
xiangce: How about adding a method to get the `environ` of special command?, E.g

```
ps_aexww.get_environ('openshift-router')
# Maybe you can only support to search the base command name 
```

insights/parsers/ps.py
```diff
@@ -342,3 +342,42 @@ def parse_content(self, content):
...
+        '10s'
+    """
+    command_name = 'COMMAND'
+    user_name = 'PID'
+    max_splits = 4
+
+    def parse_content(self, content):
+        super(PsAexww, self).parse_content(content)
+        for row in self.data:
+            row['ENVIRON'] = {}
```
xiangce:
```
row['ENVIRON'] = optlist_to_dict(row['ARGS'], opt_sep=' ')
```

xiangce:
For `simple_command`, a new `simple_file` spec to `insights_archive.py` is needed.

insights/parsers/ps.py
```diff
@@ -76,7 +77,7 @@ def parse_content(self, content):
                 proc["COMMAND_NAME"] = cmd_name
                 self.cmd_names.add(cmd_name)
                 proc["ARGS"] = cmd.split(" ", 1)[1] if " " in cmd else ""
-                self.services.append((cmd_name, proc[self.user_name], proc[raw_line_key]))
+                self.services.append((cmd_name, proc.get('USER') or proc.get('UID', 'unknow'), proc[raw_line_key]))
```
xiangce: This line should not be updated in this way.

insights/parsers/ps.py
```diff
@@ -107,15 +108,17 @@ def users(self, proc):
...
 
         .. note:: 'proc' must match the entire command and arguments.
         """
         ret = {}
-        for row in self.data:
-            if proc == row[self.command_name]:
-                if row[self.user_name] not in ret:
-                    ret[row[self.user_name]] = []
-                ret[row[self.user_name]].append(row["PID"])
+        if self.first_column in ['USER', 'UID']:
```
xiangce: I don't think it's a good idea to do so in the base class.

insights/parsers/ps.py
```diff
@@ -342,3 +341,53 @@ def parse_content(self, content):
...
+    def get_environ(self, proc):
+        """
+        Searches for the first command matching ``proc`` and returns its
+        environment variable as a dict.
+
+        Returns:
+            dict: Dictionary with environment variable name as key and containing  environment variable value
+            ``{}`` if ``proc`` is not found.
+
+        .. note:: 'proc' must contain the command
```
xiangce: Please update the `node` as per http://www.sphinx-doc.org/en/1.5/markup/para.html

insights/parsers/ps.py
```diff
@@ -342,3 +341,53 @@ def parse_content(self, content):
...
+        """
+        env = {}
+        for row in self.data:
+            if row[self.command_name].startswith(proc.strip() + ' '):
+                if '=' in row['ARGS']:
```
xiangce: Have you checked https://github.com/RedHatInsights/insights-core/pull/1848#discussion_r276935713

insights/parsers/ps.py
xiangce: proc.get('USER', proc.get('UID')

PR #1850, opened 2019-04-22
Use command 'subscription-manager facts --list' to list all facts

psachin:
- and deprecated `VirtUuidFacts` parser.

### DIFF
--- a/insights/parsers/subscription_manager_list.py
+++ b/insights/parsers/subscription_manager_list.py
@@ -22,6 +22,7 @@ import re
 from datetime import datetime
 import six
 from insights.specs import Specs
+from insights.parsers import split_kv_pairs, SkipException
 from .. import parser, CommandParser, LegacyItemAccess
 from . import keyword_search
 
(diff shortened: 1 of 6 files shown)


insights/parsers/subscription_manager_list.py
```diff
@@ -254,3 +256,33 @@ class SubscriptionManagerReposListEnabled(SubscriptionManagerList):
...
+        uname.release: 3.10.0-327.el7.x86_64
+        uname.sysname: Linux
+        uname.version: #1 SMP Thu Oct 29 17:29:29 EDT 2015
+        virt.host_type: virtualbox, kvm
+        virt.is_guest: True
+        virt.uuid: 81897b5e-4df9-9794-8e2a-b496756b5cbc
+
+    Examples::
+
+        >>> facts = shared[SubscriptionManagerFactsList]
```
xiangce: Since `shared` is deprecated.
Would you mind to delete it from the `Examples` and update the `test_subscription_manager_list_docs` as well? And this also needs you to modify the other `Examples` in the file. Thank.s

insights/parsers/subscription_manager_list.py
```diff
@@ -254,3 +256,33 @@ class SubscriptionManagerReposListEnabled(SubscriptionManagerList):
...
+        >>> facts['virt.uuid'] == "81897b5e-4df9-9794-8e2a-b496756b5cbc"
+        True
+        >>> facts['uname.sysname'] == "Linux"
+        True
+    """
+    def parse_content(self, content):
+        self.data = {}
+        for line in content:
```
xiangce: Since the filter is marked as `filterable = True`, that means the content might be empty. Better raise a `SkipExceptoin` if the content is empty.

insights/parsers/subscription_manager_list.py
psachin: Sure

insights/parsers/subscription_manager_list.py
psachin: Sure

PR #1853, opened 2019-04-23
Cleanup jupyter notebooks

csams:
Remove the `Development Examples` and `1.x to 3.x Migration` notebooks. Cleans up the others to use `make_pass` and `make_fail` instead of `make_response` and ensures they all still work.


docs/notebooks/Diagnostic Walkthrough.ipynb
```diff
@@ -47,7 +49,7 @@
     "editable": true
    },
    "source": [
-    "The first import line has the most common components used when creating `insights-core`.  The `rule` decorator marks a function that encodes logic to be applied by the framework and the required or optional components it needs to function.  `@rule` decorated functions use `make_response` to pass back results.  The `run` method executes the system, simplifying usage of `insights-core` for small, standalone scripts and from the python interpreter.\n",
+    "The first import line has the most common components used when creating rules.  The `rule` decorator marks a function that encodes logic to be applied by the framework and the required or optional components it needs to execute.  `@rule` decorated functions use `make_fail` or `make_pass` to return results.  The `run` method executes the system, simplifying usage of `insights-core` for small, standalone scripts and from the python interpreter.\n",
```
bfahr: There is a formatting error about this line in the notebook, would you mind to fix it?
```
We’ll assume we have insights-core already installed following the instructions on the ``README.rst` <https://github.com/RedHatInsights/insights-core/blob/master/README.rst>`__. Next we need to import the necessary modules.
```
The leading double-tick needs to be removed since this is a link.  You can see the results in the output.

csams: Thanks for the catch! I've updated the branch.

PR #1859, opened 2019-04-24
Enhance init_process_cgroup for RHEL 8

huali027:


insights/parsers/init_process_cgroup.py
```diff
@@ -52,5 +52,5 @@ def parse_content(self, content):
         for line in content:
             values = line.split(":")
             self.data[values[1]] = [values[0], values[2]]
-            if "system.slice/docker-" in values[2] and not self.is_container:
+            if ("system.slice/docker-" in values[2] or 'machine.slice/libpod-' in values[2]) and not self.is_container:
```
bfahr: @huali027 couldn't you just drop the `and not self.is_container` here?  You might set it to `True` more than once, but the results will be the same, and the code will be less complex.

huali027: @bfahr, sure, I'll update it and submit a patch later, thanks.

PR #1860, opened 2019-04-24
Added parser for PostgreSQL table record counts

jhutar:
This is kinda ask for guidance on how to proceed. I assume that with his parser, I still need some change so client actually covers that info? Thank you in advance.

### DIFF
--- a/insights/parsers/foreman_tasks_psql_count.py
+++ b/insights/parsers/foreman_tasks_psql_count.py
@@ -1,0 +1,42 @@
+"""
+Foreman Tasks table size
+========================
+
+Module for parsing size of foreman_tasks_tasks table.
+
+This module is supposed to consume this output:
+
+```
+# su - postgres -c "echo 'select count(*) from foreman_tasks_tasks' | psql -d foreman"
+ count 
+-------
+   209
+(1 row)
+```
+
+and provides `count` value.
+
+"""
+
+from insights import Parser, parser, get_active_lines
+from insights.parsers import ParseException, SkipException
+from insights.specs import Specs
+
+
+@parser(Specs.foreman_tasks_psql_count)
+class ForemanTasksPsqlCount(Parser):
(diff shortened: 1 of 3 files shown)


jhutar:
Please note that the trailing whitespace is part of testing data. How to tell to flake8 to ignore it?

mccun934:
Also would be good to check cp_event :

https://bugzilla.redhat.com/show_bug.cgi?id=1591908

bfahr:
@jhutar you can use the comment `# noqa: W291` immediately after the trailing `'''` for the strings to ignore the trailing whitespace if it is required.

jhutar:
Thank you @bfahr ! I did not realized I can use this on multi-line string.

jhutar:
@mccun934 , for Candlepin's cp_event, I have created https://github.com/RedHatInsights/insights-core/pull/1958

bfahr:
@jhutar in order to limit the execution of this to the systems that are running the appropriate version of satellite you should include a datasource in `insights/specs/default.py` that would look something like the following code.  Your parser would depend upon this datasource.

```
    satellite_rpm = simple_command('/bin/rpm -q satellite')
    @datasource(satellite_rpm)
    def foreman_tasks_psql_count(broker):
        sat_rpm = broker[DefaultSpecs.satellite_rpm]
        if sat_rpm.content:
            _, ver, _ = sat_rpm.content[0].split('-')
            # Check for appropriate version of Satellite (may not be needed)
            if ver.startswith('6.4') or ver.startswith('6.5'):
                cmd = """su - postgres -c "echo 'select count(*) from foreman_tasks_tasks' | psql -d foreman" """
                ctx = broker[HostContext]
                results = ctx.shell_out(cmd)
                return DatasourceProvider(content=list(results), relative_path="foreman_tasks_psql_count")
        raise SkipComponent()
```

PR #1862, opened 2019-04-24
Add a parser for openstack server list --all command

asyedham:
A parser is required to parse /usr/bin/openstack server list --all


insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,64 @@
...
+        | 3d62cd7e-41d2-43dd-a5bf-5935bc319fae | controller-0 | ACTIVE | ctlplane=192.168.24.10 | overcloud-full | controller |
+        +--------------------------------------+--------------+--------+------------------------+----------------+------------+
+    """
+
+    def parse_content(self, content):
+        """
+        Parse output content table of command ``/usr/bin/openstack server list --all``.
+        Set each variable as an class attribute.
+        """
+        data = []
```
bfahr: You might want to consider using a combination of the insights-core helper functions [get_active_lines](https://insights-core.readthedocs.io/en/latest/api_index.html#insights.parsers.get_active_lines) and [parse_delimited_table](https://insights-core.readthedocs.io/en/latest/api_index.html#insights.parsers.parse_delimited_table) to parse this table.  For instance you could do something like the following:
```python
# Removes the lines starting with "+----"
table = get_active_lines(content, comment_char='+')
# Parses header line and each row using the "|" separator
servers = parse_delimited_table(table, delim='|')
```
This much will give you a list of dicts that looks something like this:
```python
>>> servers
[{'': '',
  'ID': '410b05bb-59b7-4b4c-88e9-975c811d68da',
  'Name': 'compute-0',
  'Status': 'ACTIVE',
  'Networks': 'ctlplane=192.168.24.20',
  'Image': 'overcloud-full',
  'Flavor': 'compute'},
 ...
]
```
You'd still have to deal with the empty key because of the leading `|`, but it would be less complicated.

insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,64 @@
...
+            d = {
+                "ID": lsp[0].strip(),
+                "Name": lsp[1].strip(),
+                "Status": lsp[2].strip(),
+                "Networks": lsp[3].strip(),
+                "Image": lsp[4].strip(),
+                "Flavor": lsp[5].strip()
+            }
+            data.append(d)
+            self.data = data
```
bfahr: Although `data` is used in a lot of parsers, its not the most accurate attribute name.  Wouldn't `servers` be a more appropriate attribute name here?

insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,64 @@
...
+    Sample command output::
+
+        +--------------------------------------+--------------+--------+------------------------+----------------+------------+
+        | ID                                   | Name         | Status | Networks               | Image          | Flavor     |
+        +--------------------------------------+--------------+--------+------------------------+----------------+------------+
+        | 410b05bb-59b7-4b4c-88e9-975c811d68da | compute-0    | ACTIVE | ctlplane=192.168.24.20 | overcloud-full | compute    |
+        | f891e98b-4df6-4c90-9bf1-39cf8ac900b0 | compute-1    | ACTIVE | ctlplane=192.168.24.9  | overcloud-full | compute    |
+        | 3d62cd7e-41d2-43dd-a5bf-5935bc319fae | controller-0 | ACTIVE | ctlplane=192.168.24.10 | overcloud-full | controller |
+        +--------------------------------------+--------------+--------+------------------------+----------------+------------+
+    """
```
bfahr: Please add **Attributes:** and **Examples:** sections to your docstring to help developers understand how to use your parser.  See [this reference](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google) and the [insights-core documentation guidelines](https://insights-core.readthedocs.io/en/latest/docs_guidelines.html) for example docstrings.

insights/parsers/tests/test_openstack_server_list_all.py
```diff
@@ -0,0 +1,61 @@
...
+    assert len(res.bad_lines) == 1
+    assert res.bad_lines[0] == ' | f891e98b-4df6-4c90-9bf1-39cf8ac900b0 | compute-1    | ACTIVE | ctlplane=192.168.24.9 | overcloud-full | compute      |x'
+
+
+def test_openstack_server_list_all_documentation():
+    failed_count, tests = doctest.testmod(
+        openstack_server_list_all,
+        globs={'output': openstack_server_list_all.OpenstackServerListAll(context_wrap(VALID_TABLE))}
+    )
+    assert failed_count == 0
```
bfahr: This code actually tests the **Examples:** section of your docstring, which you don't currently have.

insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,54 @@
...
+            raise SkipException("Empty content.")
+        if len(content) < 5:
+            raise ParseException("Wrong content in table: '{0}'.".format(content))
+        for _l in content:
+            # Removes the lines starting with "+----"
+            table = get_active_lines(content, comment_char='+')
+
+            # Parses header line and each row using the "|" separator
+            servers = parse_delimited_table(table, delim='|')
+            server_list = [{k: v for k, v in d.items() if k != ''} for d in servers]
```
bfahr: @asyedham Python 2.6 on RHEL6 does not support dictionary comprehensions, you'll need to change this.

asyedham: @bfahr Done. Could you please review

insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,98 @@
...
+    def parse_content(self, content):
+        """
+        Parse output content table of command ``/usr/bin/openstack server list --all``.
+        Set each variable as an class attribute.
+        """
+        if not content:
+            raise SkipException("Empty content.")
+        if len(content) < 5:
+            raise ParseException("Wrong content in table: '{0}'.".format(content))
+        for _l in content:
```
bfahr: @asyedham this for loop isn't necessary, you are parsing the entire content in line 54.

insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,98 @@
...
+        '''
+        returns the list of values for the specified key if the key exists
+        returns ['', '', ''] if the key doesn't exist
+
+        Example:
+            >>> parser_result_server_list.get_list('Name')
+            ['compute-0', 'compute-1', 'controller-0']
+        '''
+        list_ = [i.get(key, '') for i in self.data]
+        self.list_ = list_
```
bfahr: This is really just a transient value and should not be cached in the object since you never use it again.

insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,98 @@
...
+=============================================
+"""
+
+from insights import CommandParser, parser, LegacyItemAccess
+from insights.parsers import ParseException, SkipException, get_active_lines, parse_delimited_table
+from insights.specs import Specs
+
+
+@parser(Specs.openstack_server_list_all)
+class OpenstackServerListAll(LegacyItemAccess, CommandParser):
```
bfahr: Your use of `LegacyItemAccess` here is not really appropriate.  It really should not be used for new parsers, and it is meant to be used with a `self.data` that is a of type `dict` instead of `list` as you have in this parser.

You could simplify this parser and make it easier to understand and use if you stored the output in a `dict` based on `Name` which appears to be the primary access mechanism to this data.  So for instance instead of creating `self.data` which is a list, you could create `self.servers` which is a `dict` indexed by the key in the `Name` column, and containing the existing line dict as the value.  You only need to keep a list if order of the items matters.  

The second simplification would be to remove `LegacyItemAccess` and use a `dict` mixin class to your parser as was implemented in the [Partitions parser](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/partitions.py).  This would allow you to access the `self.servers` dictionary directly from the parser object, like `parser_result['compute-0']`, and you could provide methods to access the other information that would be useful.

insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,98 @@
...
+        '''
+        If the name_value exists in the dict returns the value for the key
+
+        Example:
+        >>> parser_result_server_list.get_from_name('controller-0','Flavor')
+        'controller'
+        '''
+        for i in self.data:
+            if name_value in i.values() and key != 'Name':
+                return i.get(key, '')
```
bfahr: This is confusing because "name" is a key in the dict, but you are actually ignoring "Name" in your search.

insights/parsers/openstack_server_list_all.py
```diff
@@ -0,0 +1,98 @@
...
+from insights.specs import Specs
+
+
+@parser(Specs.openstack_server_list_all)
+class OpenstackServerListAll(LegacyItemAccess, CommandParser):
+    """
+    This parser processes the output of the command `openstack server list --all` and provides
+    the information as a list of dictionary.
+    The LegacyItemAccess class provides some helper functions for dealing with a
+    class having a `data` attribute.
```
bfahr: Whatever you decide to name your attributes, if they are public you should have an `Attributes:` section in your docstring to define them for the developer.  It be better not to use this wording here.

insights/parsers/tests/test_openstack_server_list_all.py
```diff
@@ -0,0 +1,54 @@
...
+    assert parser_result.get_startwith('Name', 'controller') == ['controller-0']
+    assert parser_result.get_startwith('Name', 'compute') == ['compute-0', 'compute-1']
+    assert parser_result.get_from_name('compute-0', 'Flavor') == 'compute'
+    assert parser_result.get_from_name('controller-0', 'Image') == 'overcloud-full'
+
+
+def test_doc():
+    INFO = VALID_TABLE
+    env = {
+            "parser_result_server_list": OpenstackServerListAll(context_wrap(INFO)),
```
bfahr: This is not really a list, its a parser object.  Since this is the name that must be used in the examples, it would be better to use a name that more accurately describes what it is, like `parser_result` or `openstack_servers`.

insights/parsers/tests/test_openstack_server_list_all.py
```diff
@@ -0,0 +1,54 @@
...
+ | f891e98b-4df6-4c90-9bf1-39cf8ac900b0 | compute-1    | ACTIVE | ctlplane=192.168.24.9  | overcloud-full | compute    |
+ | 3d62cd7e-41d2-43dd-a5bf-5935bc319fae | controller-0 | ACTIVE | ctlplane=192.168.24.10 | overcloud-full | controller |
+ +--------------------------------------+--------------+--------+------------------------+----------------+------------+
+ """
+
+
+def test_empty_openstack_server_list():
+    with pytest.raises(SkipException) as e_info:
+        OpenstackServerListAll(context_wrap(""))
+    assert "Empty content." in str(e_info.value)
```
bfahr: You should also add a test for the `ParseException`.

xiangce:
Closing it for now, please open it in another PR if necessary

PR #1863, opened 2019-04-24
[WIP] Enhance rsyslog conf rhel8

vishwanathjadhav:
Enhanced the parser for RHEL-8, The format of the file is changed and need different parsing methods.


insights/parsers/rsyslog_conf.py
```diff
@@ -77,3 +77,130 @@ def __len__(self):
     def __iter__(self):
         for d in self.data:
             yield d
+
+
+@parser(Specs.rsyslog_conf)
```
xiangce: This `RsyslogConf8` used the same spec `Specs.rsyslog_conf` to `RsyslogConf` which means both the `RsyslogConf` and `RsyslogConf8` will parse the  `/etc/rsyslog.conf` at the same time for the same collected `/etc/rsyslog.conf`.  This must not what want.

This spec is quite a typical case to us about how to handle the file/command which is placed in the same path but has different formats in different RHEL versions.

@bfahr @csams - Any thoughts?

BTW, I think both the original `RsyslogConf` and the new `RsyslogConf8` could be refactored to use the [ConfigParser](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/__init__.py#L338) by re-implementing the `parse_doc`. See https://github.com/RedHatInsights/insights-core/blob/master/insights/combiners/httpd_conf.py#L270

insights/parsers/rsyslog_conf.py
```diff
@@ -77,3 +77,130 @@ def __len__(self):
...
+            else:
+                line_split = line.split()
+                self._log_details[line_split[0]] = line_split[1]
+
+    def __len__(self):
+        return len(self.data)
+
+    def __iter__(self):
+        for d in self.data:
+            yield d
```
bfahr: Do you need an iterator for the class?  It seems that you only use self.data to parse the content, and that a developer would not need the `__iter__` or `__len__` methods.

insights/parsers/rsyslog_conf.py
bfahr: @xiangce @vishwanathjadhav you definitely need separate parsers for the same spec, but it does seem like there should be code that could be shared between them.  One thing that you really want to try to do is hit the eject button as early in the parsing process as you can.  For example if RHEL7 is exclusively `$Module` and RHEL8 is exclusively `module(` then the respective parser should raise a `SkipException` as soon as it detects the other version.

@vishwanathjadhav are the two versions completely different or can RHEL8 have a RHEL 7 style config file?

@csams any suggestions how you might want to handle this scenario?

csams: @bfahr @xiangce I think we should extend `@parser` to allow additional dependencies that just control whether or not the parser fires. Then we can have different parsers for the same spec, and we can have little components like `is_rhel8`, `is_rhel7` they can depend on.

That ensures any exceptions the parsers raise are still meaningful and not just the result of trying to parse something they could never understand because the data is from the wrong RHEL.

We also still have the option of putting a combiner in front of them when it makes sense.

Approaches I considered:
- Have multiple specs for the same file. This just moves the problem up a level. They'd have to depend on something like `is_rhelX`, too, and it could mean a different spec and registry point for every single version of something we want to support.
- Allow parsers to depend on other components like `RedHatRelease` which they then use to select the correct parsing function. This has the advantage of providing a single interface to the data for rules to use, but it complicates the parser, and it *forces* the same interface to the data, even if there are major differences across versions.
- Allow parsers to depend on other components, but those dependencies just control whether the parser fires. Parsers don't actually get to inspect them. This is conceptually similar to the second approach, but it leans on the framework to the keep the parser implementations simpler and decoupled from one another, and it allows us to present a single interface through a combiner without forcing it where it doesn't make sense.

csams: I created #1888 to track the `@parser` work.

vishwanathjadhav: @csams - `@parser` to allow additional dependencies like rhel version is best thing because it will help  to switch the parsing method based on RHEL version. RHEL-8 gives more control over logging, so need to add few access methods to the `RsyslogConf` parser which may not be supported on RHEL-7 so we have to create new parser `RsyslogConf8`. On the back-end(rule valid for RHEL-7/8) implementation will look like 

Pseudo Code:
~~~
@condition(RedhatRelease, RsyslogConf, RsyslogConf8)
def check_logs(ver_obj, rsys_obj, rsys8_obj):
        if ver_obj.major == 7:
              return rsys_obj.config_val(xyz)
        else:
              return rsys8_obj.module_details
~~~

vishwanathjadhav:
Covered in new PR: https://github.com/RedHatInsights/insights-core/pull/1952

PR #1866, opened 2019-04-25
Revert "Add parser ps_aexww (#1848)"

shzhou12:
This reverts commit caaf6d6c7020fc7fe9ee5331499c52cfe0a9da87.


xiangce:
After a further discussion with @shzhou12, we both think it's not good to get the ENVs from 'ps aexww' due to PII issue and the complex parsing logic. Since the parser is just newly added and its content collected. We would like to revert this Parser and turn to use the '/proc/<PID>/environ' instead.

How do you think about this revert?

bfahr:
This is related to #1848 and #1866.

PR #1869, opened 2019-04-25
Enhanced parser to fix the bug in krb5 Parser

vishwanathjadhav:
Modified parsing logic to fix the issue. After parsing the configuration file `/etc/krb5.conf` last section was missing in the output.


bfahr:
@vishwanathjadhav if this is a bug as you mention in the title would you please open an issue?

insights/parsers/krb5.py
```diff
@@ -146,9 +146,6 @@ def parse_content(self, content):
                 else:
                     is_squ = True
                     squ_section_name = line.split("=")[0].strip()
-        # Set the value to the last section
```
xiangce: This lines should not be removed since the content in the last section will be lost.

insights/parsers/tests/test_krb5.py
```diff
@@ -98,6 +126,13 @@ def test_krb5configuration():
     assert common_conf_info.includedir == ["/etc/krb5.conf.d/"]
     assert common_conf_info.module == ["/etc/krb5test.conf:residual"]
 
+    common_conf_info = krb5.Krb5Configuration(context_wrap(KRB5CONFIG3, path=KRB5_CONF_PATH))
+    assert len(common_conf_info.sections()) == 4
+    assert common_conf_info.has_section('domain_realm') is True
+    assert sorted(common_conf_info.options('logging')) == sorted(['default', 'kdc', 'admin_server'])
+    assert common_conf_info.has_option('libdefaults', 'dns_lookup_realm') is True
+    assert common_conf_info.has_option('domain_realm', 'example.com') is False
```
xiangce: Better add a test to check `assert common_conf_info.has_option('domain_realm', 'xyz') is True`

Then, you'll find the code should be modified.

vishwanathjadhav: @xiangce - no   `assert common_conf_info.has_option('domain_realm', 'xyz') is True` will be  `assert common_conf_info.has_option('domain_realm', 'xyz') is Fase` becase as per input configuration 
~~~
 99 [domain_realm]
100 # .example.com = EXAMPLE.COM
101 # example.com = EXAMPLE.COM
102 []
103 xyz = abc
~~~

insights/parsers/krb5.py
vishwanathjadhav: @xiangce - its already handled in the code so need of the line

insights/parsers/tests/test_krb5.py
wushiqinlou: @vishwanathjadhav - I have two questions:

1, In real environment, is there empty [] inside section?

2, If it is, why should we ignore the 103 line? Is there any document or any test instance to support this opinion?

vishwanathjadhav: @wushiqinlou - no their won't be such a situation but its added for just a test purpose,  the situation give below is possible.
~~~
 99 [domain_realm]
100 # .example.com = EXAMPLE.COM
101 # example.com = EXAMPLE.COM
~~~

xiangce: This will lead to misunderstanding. Better remove the line-102 and line-103, and just add a check to make sure there is a section named `domain_realm` (and it's content is empty)

vishwanathjadhav: @xiangce - Done, ` just add a check to make sure there is a section named domain_realm (and it's content is empty)`  is already present in line +129.

PR #1872, opened 2019-04-25
Work around to not archive command outputs from commands not found.

Jason-RH:
This removes command output from those commands that are not found from the archive. The reason it is checking for the output from the timeout command is because that is what gets archived. The return code and/or output of the actual command doesn't propagate through.


Jason-RH:
@kylape @jhjaggars jcrafts suggested I add you two on this PR in case you had any additional input. I can provide some more context if the previous comment isn't clear.

gravitypriest:
submitted for the approval of the midnight society: say we changed the check to this?

`if output and not re.search(r'^timeout: failed to run command .+: No such file or directory$', output)`

Jason-RH:
Cool w/ me. Probably best to be as specific as possible here.

PR #1877, opened 2019-04-26
Correct an issue where only the first rpm that matches the rpm pattern

lhuett:
Correct an issue where only the first rpm that matches the rpm pattern
for a provider is stored in the list.
Does not affect provide selection.

### DIFF
--- a/insights/combiners/cloud_provider.py
+++ b/insights/combiners/cloud_provider.py
@@ -122,7 +122,8 @@ class CloudProvider(object):
         if rpms:
             for p in self.__PROVIDERS:
                 for key, val in rpms.packages.items():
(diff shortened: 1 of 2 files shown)


lhuett:
@xiangce Since you were involved with the development of the 'aws_instance_type' spec and the rule I wanted to be sure you reviewed this. If it is something you would feel more comfortable waiting to Merge  after Summit that is fine. This mainly for when we persist data to the Datahub, just want to be sure it is complete.  Made a slight change to be sure all rpms that matched the provider rpm pattern got in the rpm list for that provider. Should not have any effect on cloud_provider selection either way, just the list of matching rpms.

PR #1878, opened 2019-04-28
Add cgroups parser for /proc/cgroups

chenlizhong:


insights/parsers/cgroups.py
```diff
@@ -0,0 +1,105 @@
...
+    Attributes:
+        data (list): A list of the subsystem cgroup information
+
+    """
+
+    def parse_content(self, content):
+        # replace #subsys_name to subsys_name
+        content[0] = content[0].replace("#", '')
+        table = parse_delimited_table(content)
+        self.data = [dict((k.lower(), v) for (k, v) in item.items()) for item in table]
```
xiangce: I think it's OK to assign the result of `parse_delimited_table` to `self.data` directly.

This loop will retrieve all the dicts which have the same keys from the list and apply the `lower()` to them (the `keys` which are the same to all of the dicts) repeatedly. 
If there is any chance that the `keys` could be in `uppercase`, better `lower` them in just one time, e.g. `content[0] = content[0].lower()`, before the `parse_delimited_table`.

PS, do you intend to lower all the `subsys_name` instead of the `keys` like `subsys_name	hierarchy	num_cgroups	enabled`?

insights/parsers/cgroups.py
```diff
@@ -0,0 +1,105 @@
...
+
+"""
+
+from .. import parser, CommandParser
+from insights.specs import Specs
+from . import ParseException, parse_delimited_table
+
+
+@parser(Specs.cgroups)
+class Cgroups(CommandParser):
```
xiangce: BTW, it's not a CommadParser, right?

insights/parsers/cgroups.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Attributes:
+        data (list): A list of the subsystem cgroup information
+
+    """
+
+    def parse_content(self, content):
+        # replace #subsys_name to subsys_name
+        content[0] = content[0].replace("#", '')
+        table = parse_delimited_table(content)
```
xiangce: By using the utilities of the `parse_delimited_table`, the line-53 can be saved:
```python  
parse_delimited_table(content,
                heading_ignore=['subsys_name', 'hierarchy'],
                header_substitute=[('#subsys_name', 'subsys_name')])
```

insights/parsers/cgroups.py
bfahr: @chenlizhong a list doesn't really seem to be the best way to store this info.  Why not use an attributed called `subsystems` containing the values as dict?  Both of your methods have to iterate through the list to check for names.  Also it would be better to use attribute names that better represent the data instead of a generic `data` name.

insights/parsers/cgroups.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Returns:
+
+            value (int): Int value of the specified subsystem cgroups
+        """
+        for subsys_item in self.data:
+            if subsys_item["subsys_name"] == i_subsys_name:
+                return int(subsys_item["num_cgroups"])
+        # raise exception if give the wrong subsys_name
+        raise ParseException("Wrong subsys_name: {0}".format(i_subsys_name))
```
bfahr: If you used a dict of subsystems then you could simplify this logic.  Also a `KeyError` exception would be more appropriate here since this method is called by rules, where a `ParseException` is not really meaningful.

insights/parsers/cgroups.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Returns:
+
+            value (boolean): Return True if the cgroup of specified subsystem is enabled, else return False
+        """
+        for subsys_item in self.data:
+            if subsys_item["subsys_name"] == i_subsys_name:
+                return subsys_item["enabled"] == "1"
+        # raise exception if give the wrong subsys_name
+        raise ParseException("Wrong subsys_name: {0}".format(i_subsys_name))
```
bfahr: Same comment as above.

insights/parsers/tests/test_cgroups.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_cgroups():
+    context = context_wrap(cgroups_content)
+    i_cgroups = Cgroups(context)
+    assert i_cgroups.get_num_cgroups("memory") == 232
+    assert i_cgroups.is_subsys_enabled("memory") is True
+    try:
+        assert i_cgroups.get_num_cgroups("Wrong_memory") == 232
+    except ParseException:
+        assert True
```
bfahr: The better approach to testing exceptions is to use `pytest.raises`.

insights/parsers/cgroups.py
```diff
@@ -0,0 +1,108 @@
...
+
+        Args:
+
+            i_subsys_name (str): specified subsystem name.
+
+        Returns:
+
+            value (int): Int value of the specified subsystem cgroups
+        """
+        if i_subsys_name in self.subsystems:
```
xiangce: How do you think about `return int(self.subsystems[i_subsys_name]["num_cgroups"])` directly without the above `if`-check and the following manually `raise`, which will raise the `KeyError` by itself?

insights/parsers/cgroups.py
```diff
@@ -0,0 +1,108 @@
...
+        Args:
+
+            i_subsys_name (str): specified subsystem name.
+
+        Returns:
+
+            value (boolean): Return True if the cgroup of specified subsystem is enabled, else return False
+        """
+        if i_subsys_name in self.subsystems:
+            return self.subsystems[i_subsys_name]["enabled"] == "1"
```
xiangce: Same comment as above: How do you think about `return self.subsystems[i_subsys_name]["enabled"] == "1"` directly without the check-condition and manually `raise`?

chenlizhong: Both OK. But I prefer the adding `if` as this is prediction error.

insights/parsers/cgroups.py
chenlizhong: Both OK. But I prefer the adding `if` as this is prediction error.

xiangce: Actually, both are good to me too, yours gives a more detailed prompt hint to the users

PR #1880, opened 2019-04-29
New parser for /proc/partitions file

akshay196:
A parser to read content of file `/proc/partitions`, which contains
partitions block allocation information.


insights/parsers/partitions.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Raises:
+        SkipException: When input is empty.
+        ParseException: When input cannot be parsed.
+    """
+    def parse_content(self, content):
+        if not content:
+            raise SkipException('Empty content')
+
+        self.data = []
```
xiangce: The `parse_delimited_table` can replace all the following logics.
~~~python
        self.data = parse_delimited_table(content,
                heading_ignore=['major', 'minor'],
                header_substitute=[('#blocks', 'blocks')])
~~~

But you might lose the feature given by the `nametuple`

insights/parsers/partitions.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+from collections import namedtuple
+
+from insights import Parser, parser, LegacyItemAccess
+from insights.parsers import SkipException, ParseException
+from insights.specs import Specs
+
+
+@parser(Specs.partitions)
+class Partitions(Parser, LegacyItemAccess):
```
xiangce: The `LegacyItemAccess` only works when `self.data` is a dictionary, but here it's a list of dicts.

insights/parsers/partitions.py
akshay196: `parse_delimited_table` is cleaner approach and saves lines in the parser. I have modify parser to use `parse_delimited_table` method.
@xiangce  Thanks

insights/parsers/partitions.py
akshay196: Done

xiangce:
@akshay196 - please modify the code for python2.6 test

insights/parsers/partitions.py
```diff
@@ -0,0 +1,60 @@
...
+       '22708224'
+
+    Raises:
+        SkipException: When input is empty.
+    """
+    def parse_content(self, content):
+        if not content:
+            raise SkipException('Empty content')
+
+        self.data = parse_delimited_table(content,
```
bfahr: @akshay196 the `data` attribute is an old name that would be better not to use because is too generic.  What about using the name `partitions` or something more descriptive of the data that is being provided by the parser?

Also would it be better for the attribute to be a `dict` of partitions so that it could be indexed by name?  You could present it as both a list and a dict if that is how you anticipate it being accessed.  For instance you could create an attribute of partition names and implement `__getitem__` to index by partition name and `__contains__` to allow checking using the `in` operator.

insights/parsers/partitions.py
```diff
@@ -0,0 +1,60 @@
...
+
+        Args:
+            name (str): Name of the partition.
+
+        Returns:
+            string: Returns `blocks` if `name` present. Otherwise None.
+        """
+        for partition in self.data:
+            if partition.get('name', '') == name:
+                return partition.get('blocks', '')
```
bfahr: This is a little confusing to me, you are implicitly returning `None` if the partition is not in the list, but you are returning an empty string if the name is present and the `blocks` field is not present?  Keep in mind how a developer will use this function, the way it is written it will require a check for both `''` and `is not None`. You might want to make the return values consistent.

You might also want to consider raising a `KeyError` here if the name is not present to avoid any ambiguity about what is happening and be consistent with the python std modules.

As mentioned in the previous comment, having a dict of names would simplify this logic as well.

insights/parsers/partitions.py
akshay196: Done. Updated code now. 
@bfahr Thanks for suggestion.

insights/parsers/partitions.py
akshay196: This function is not really necessary now. It is now easy to get `blocks` field for a partition.
```python
partitions_obj['sdb'].get('blocks')
```

akshay196:
@bfahr Done with the changes as suggested. Please take another look.

insights/parsers/partitions.py
```diff
@@ -0,0 +1,81 @@
...
+
+        Returns:
+            dict: The information for the given partition name.
+        """
+        return self.partitions[item]
+
+    def __contains__(self, item):
+        return item in self.partitions
+
+    def __iter__(self):
```
xiangce: This is not so good to iterate the `dict` in this way, which covers the default `.items` of a `dict`.
And just learned from Chris: you can directly mix this Parser with `dict` like:
```python
class Partitions(Parser, dict):
```

And re-implement the `__init__` function as:

```python
        super(Partitions, self).__init__(*args, **kwargs)
        self.update(self.partitions)
```

and then the `Partitions` can be operated as a `Python dict`,  all the self-defined `__getitem__`, `__contains__` and `__iter__` can be removed.

akshay196: Great. This idea is fantastic. @xiangce Thanks for sharing. I have updated code accordingly.

akshay196:
@xiangce @bfahr Thank you reviewing and merging. :)

Issue #1886, opened 2019-05-02
wrong specs for heat api log 


asyedham:
when we check the heat-api.log in infrared setup for OSP13 it shows error info like follow:
[heat-admin@controller-0 ~]$ tail -f /var/log/containers/heat/heat-api.log
tail: cannot open ‘/var/log/containers/heat/heat-api.log’ for reading: No such file or directory
tail: no files remaining
``
Similar issue can also be observed in devstack . The file path is '/var/log/%APACHE_NAME%/heat_api.log' which specified in apache-heat-api.template is https://github.com/openstack/heat/blob/master/devstack/lib/heat#L302

koalakangaroo:
@ppandey-redhat @psachin - can you check this issue?

psachin:
@asyedham 

1. You are checking heat on the controller node. AFAIK, the heat related processes and logs are expected on the OpenStack undercloud node.[1] 

2. I don't find any reference docs to validate that the file name was changed on RHOSP 13(`/var/log/heat/heat-api.log` to `/var/log/heat/heat_api.log`). Do you see the path as `/var/log/heat/heat_api.log` on your Controller node?

If your concern is that the logs are not collected by the insights-client, the reason will be **1.** as well as what explained by @xiangce in the [comment](https://github.com/RedHatInsights/insights-core/pull/1887#pullrequestreview-233768078)


[1] https://github.com/openstack/heat/blob/master/doc/source/configuration/logs.rst

venkataanil:
I have a OSP 14 setup in which undercloud openstack services are deployed using containers. Here I can see heat api log as heat_api.log and not as heat-api.log. May be for container based deployments, change the log path to /var/log/containers/heat/heat_api.log ?

(undercloud) [stack@undercloud-0 ~]$ ls /var/log/containers/heat/
heat_api_cfn.log  heat_api_cfn.log.1  heat_api_cfn.log.2.gz  heat_api.log  heat_api.log.1  heat_api.log.2.gz  heat_api.log.3.gz  heat_api.log.4.gz  heat-engine.log  heat-engine.log.1  heat-engine.log.2.gz

asyedham:
@psachin [2] yeah we can also check the heat_api.log on controller. I have checked it on my OSP 13 setup.
 [heat-admin@controller-0 ~]$ vi /var/log/containers/heat/heat_api.log
[...]

PR #1887, opened 2019-05-02
Correct the specs for heat api log

asyedham:


bfahr:
@asyedham this appears to be a bug, could you please open an issue for this?

asyedham:
> @asyedham this appears to be a bug, could you please open an issue for this?

I have raised a issue https://github.com/RedHatInsights/insights-core/issues/1886

asyedham:
@xiangce I have created a JIRA request for spec change and also raised a PR https://github.com/RedHatInsights/insights-core-assets/pull/167. Could you please have a look 

Thanks!!

insights/specs/default.py
```diff
@@ -287,7 +287,7 @@ def docker_container_ids(broker):
     hammer_ping = simple_command("/usr/bin/hammer ping")
     hammer_task_list = simple_command("/usr/bin/hammer --csv task list")
     haproxy_cfg = first_file(["/var/lib/config-data/puppet-generated/haproxy/etc/haproxy/haproxy.cfg", "/etc/haproxy/haproxy.cfg"])
-    heat_api_log = first_file(["/var/log/containers/heat/heat-api.log", "/var/log/heat/heat-api.log"])
+    heat_api_log = first_file(["/var/log/containers/heat/heat_api.log", "/var/log/heat/heat_api.log"])
```
xiangce: For OpenStack-10, I think the name is `heat-api.log`, (see 
[1.1.11. Orchestration (heat) Log Files](https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/10/html-single/logging_monitoring_and_troubleshooting_guide/index#orchestration_heat_log_files) and [Orchestration log files](https://docs.openstack.org/mitaka/config-reference/orchestration/logs.html)), the `head_api.log` is from OpenStack13(?) from your investigation result. And Insights should support all versions of OpenStack from 8 to 14. So, both the `heat_api.log` and `heat-api.log` should be collected.

In that case, just like @psachin mentioned in https://github.com/RedHatInsights/insights-core/issues/1886#issuecomment-489515790, you should add the items of `heat_api.log` to the spec `heat_api_log`, but should not remove the original items. e.g.

```python
heat_api_log = first_file(["/var/log/containers/heat/heat-api.log", "/var/log/heat/heat-api.log", "/var/log/containers/heat/heat_api.log", "/var/log/heat/heat_api.log"])
```

I'm not familiar with OpenStack, if there is anything wrong, please correct me freely.

asyedham: @xiangce 
I have OSP 12,13,14 setup in which the openstack services are deployed with containers.  I have checked and heat api logs can be seen as heat_api.log and not heat-api.log. So in containerized deployment it is /var/log/containers/heat/heat_api.log where as for non-containerized deployment such as OSP 10 it is /var/log/heat/heat-api.log. 
So in the spec we can add the containerized path as var/log/containers/heat/heat_api.log as the log path may have changed
heat_api_log = first_file(["/var/log/containers/heat/heat_api.log", "/var/log/heat/heat-api.log"])

xiangce:
It's ready to merge after summit.

Issue #1874, opened 2019-04-26
Krb5Configuration do not list all the sections


vishwanathjadhav:
Krb5Configuration parses fails to parse the last section of the configuration file `/etc/krb5.conf`, when `section_value` is empty(or commented) then section does not get created in the list.
~~~
[domain_realm]
 # .example.com = EXAMPLE.COM
 # example.com = EXAMPLE.COM
~~~

vishwanathjadhav:
PR to fix this issue:
https://github.com/RedHatInsights/insights-core/pull/1869

PR #1889, opened 2019-05-02
Fix parsing of /etc/krb5.conf.d/crypto-policies on RHEL8.

jsvob:
When multiple config files in /etc/krb5.conf.d/ have the same section, the sections are not merged by the combiner. However, they should get merged.


bfahr:
@jsvob @vishwanathjadhav is this related to #1869 and #1874?

jsvob:
@bfahr I had no idea about those and from a cursory look, most probably not.

PR #1908, opened 2019-05-06
Dmesg now can return logs with specified prefix

lonicerae:
- Also fix a bug when parsing dmesg output of RHEL7


lonicerae:
Reference https://github.com/RedHatInsights/insights-core/issues/1909

PR #1912, opened 2019-05-07
Update spec for dirsrv log

wushiqinlou:


wushiqinlou:
https://github.com/RedHatInsights/insights-core-assets/pull/170

bfahr:
@xiangce @wushiqinlou since you are fixing a bug, the spec approval process is not required for this spec change.

PR #1913, opened 2019-05-07
Enhance _start_of_distribution method in installed_rpm module 

chenlizhong:
To fix below kind of value error

ValueError: Cannot compare packages that one has distribution while the other does not xz-5.2.2-1.el7 != xz-5.1.2-12alpha.el7'\|


csams:
I've reworked the rpm comparison in #1921 to match rpm's.

PR #1914, opened 2019-05-07
Rpm vercmp fixes #1839

### LINKED ISSUE #1839: get_max(package_name) returns TypeError for some package lists
When parsing installed package lists from various Insights archives, using get_max function to get newest package version of given package name throws often this error:

```
2019-04-15 20:08:51,852:ERROR:__main__:Unable to parse archive: 'Traceback (most recent call last):
File "/listener/listener/upload_listener.py", line 100, in process_upload\n    parser.parse()\n
File "/listener/listener/archive_parser.py", line 62, in parse\n    pkg = pkglist.get_max(pkg_name)\n
File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/insights/parsers/installed_rpms.py", line 211, in get_max\n   return max(self.packages[package_name])\n
File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/insights/parsers/installed_rpms.py", line 599, in __gt__\n    return isinstance(other, InstalledRpm) and other.__lt__(self)\n
File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/insights/parsers/installed_rpms.py", line 584, in __lt__\n    return self_v < other_v\nTypeError: \'<\' not supported between instances of \'int\' and \'str\''|
```

MichaelMraka:
don't reinvent wheel
use rpm_vercmp for version comparison 

fixes #1839

### DIFF
--- a/insights/parsers/installed_rpms.py
+++ b/insights/parsers/installed_rpms.py
@@ -69,7 +69,6 @@ Examples:
     True
 """
 import json
-import re
 import rpm_vercmp
 from collections import defaultdict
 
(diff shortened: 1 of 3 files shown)


csams:
@MichaelMraka Thanks for the PR!

Unfortunately, since core is packaged and sent to the customers' machines as a standalone egg, we can't easily introduce new dependencies. We'd have to include the library in core's contrib package, which we're trying to get away from.

Also, I'm not immediately comfortable depending on a project that hasn't had much activity. It might be correct, but can we depend on the owner to fix bugs if we find something wrong? How did you come across it? Are you familiar with the maintainer?

I'm leaning toward our own implementation of [what's in rpm](https://github.com/rpm-software-management/rpm/blob/master/lib/rpmvercmp.c), with the caveat that it still might not be reasonable to do what it does when distributions differ, in particular when one has a distribution and the other doesn't.

csams:
I'm working on a PR for #1839 that doesn't require external dependencies and passes all of the [rpm tests](https://raw.githubusercontent.com/rpm-software-management/rpm/master/tests/rpmvercmp.at).

PR #1916, opened 2019-05-07
Add new Ps parser and fix errors in base class

### LINKED ISSUE #1833: New parser for PsAlxww
New parser for PsAlxww added to https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/ps.py and supporting files. This will allow for parsing output for ```ps -alxww```.

The parser should look quite similar to PsEf since the command output is similar. These options should be changed for the new parser: 

```
command_name = "COMMAND"
user_name = "F"
max_splits = 12
```

Here is sample output for ```ps -alxww```:

| F  | UID | PID | PPID | PRI | NI | VSZ | RSS | WCHAN | STAT | TTY | TIME | COMMAND
| :-- | :--: | :--: | :--: | :--: | :--: | :--: | :--: | :--: | :--: | :--: | :--: | --: |
| 4 | 0 | 1 | 0 | 20 | 0 | 193704 | 6952 | ep_pol | Ss | ? | 0:10 | /usr/lib/systemd/systemd |
| 1 | 0 | 2 | 0 | 20 | 0 | 0 | 0 | kthrea | S | ? | 0:00 | [kthreadd]
| 1 | 0 | 3 | 2 | 20 | 0 | 0 | 0 | smpboo | S | ? | 0:00 | [ksoftirqd/0] |

### LINKED ISSUE #1915: Ps class incorrectly parses command name and forces position of "user" and "command" columns.
For ps commands that collect command arguments, the command name was only being captured if the command started with a `/` character.  If a command didn't begin with a `/` then it was not parsed out of the command + args.  For instance, `/usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation` was parsed into command name = `dbus-daemon` and args = `--system --address=systemd: --nofork --nopidfile --systemd-activation`.  But `irqbalance --pid=/var/run/irqbalance.pid` would not be parsed into two separate components because it did not begin with `/`.

The base PS parser forces the `self.user_name` and `self.command_name` columns to be the first and last columns, respectively, of the PS output.  Some PS commands produce output that does not conform to this requirement such as the `ps alxww` command, in which UID is the second column.

This was discovered while working on enhancement #1833.

bfahr:
* Add new Ps parser for ps alxwww command
* Add tests for new parser
* Fixes bug in Ps parser the failed to parse some command names
* Fixes all doctests in ps module
* Enhancement/bug fix to recognize ps output that doesn't have
  user/uid/pid in first column
* Fixes #1915
* Fixes #1833


insights/parsers/tests/test_ps.py
```diff
@@ -29,17 +29,58 @@
 nginx    111435 111434  0 22:32 ?        00:00:00 nginx: worker process
 """
 
+PsAex_TEST_DOC = """
```
xiangce: This test data should not be here?

bfahr: @xiangce thanks for catching that, I will remove it.

bfahr:
[TYCHE-521 New Parser for ps alxwww](https://trello.com/c/9pdIV4bD/27-tyche-521-new-parser-for-ps-alxwww)

bfahr:
Hold off on merging, I was not originally planning to add an Insights collection spec but need to now.  Need to wait for approval.

PR #1918, opened 2019-05-08
Syslog can return relevant logs by specifying the process name

lonicerae:


lonicerae:
Resolves https://github.com/RedHatInsights/insights-core/issues/1919

insights/core/__init__.py
```diff
@@ -1255,6 +1255,24 @@ def _parse_line(self, line):
...
 
+    def get_logs_by_procname(self, proc):
+        """
+        Parameters:
+            proc(str): The process or facility that you're looking for
+
+        Returns:
+            (list): The raw syslog messages produced by that process or facility
+        """
+        _log_re = re.compile(r'^[A-Za-z]{3}\s+\d{1,2}\s\d{2}:\d{2}:\d{2}\s\S+\s(?P<procname>\S+):.*$')
```
xiangce: As you might also complain, `REGEX` is not easy to be maintained and time cost. And the `_parse_line` above has been parsed the `lines` into different parts (e.g. procname, message and timestamp), would you please to change to utilize them instead of the `regex`?

https://github.com/RedHatInsights/insights-core/blob/8dbe0c56991e5519005e84cb254ddc6580aaa34b/insights/core/__init__.py#L1255

insights/tests/test_syslog.py
```diff
@@ -11,6 +11,9 @@
 Apr 22 10:37:32 boy-bona crontab[28951]: (root) LIST (root)
 Apr 22 10:40:01 boy-bona CROND[30677]: (root) CMD (/usr/lib64/sa/sa1 -S DISK 1 1)
 Apr 22 10:41:13 boy-bona crontab[32515]: (root) LIST (root)
+Apr 29 11:33:36 kvmr7u5 ehtest: crontab[12345]: {
+April 29 11:33:36 kvmr7u5 ehtest: crontab[12345]: {
```
xiangce: The format of this 2 lines is different from others (one more`ehtest:` appeared), if you confirmed it's a valid line, I think the original logic of the `_parse_line` cannot handle it correctly. It's a bug and needs to be fixed.

Would you please to take a double confirm? And let me know the result, Thanks a lot.

lonicerae: Actually the format is not changed, because the `ehtest:` is still the hostname, the `crontab[12345]: {` is the message body which comes from other system. The parsing method is still correct.

insights/core/__init__.py
lonicerae: Sure, let me provide another commit.

lonicerae:
The change looks good to me, thanks @xiangce !

PR #1920, opened 2019-05-08
Change parser class in insights.core.plugins to allow multiple args

lhuett:
Change parser class in insights.core.plugins to allow multiple
args to support @components used to control if a parser will 
be fired 
Updated Documentation for @component usage and parser changes
Added @component example to updated examples.rules.stand_alone.py.


docs/api.rst
```diff
@@ -251,7 +251,7 @@ Parsers
 A ``Parser`` takes the raw content of a particular ``Data Source`` such as
 file contents or command output, parses it, and then provides a small API for
 plugins to query.  The parsed data and computed facts available via the API are
-also serialized to be used in downstream processes.
+also serialized to be used in downstream processes. The
```
JoySnow: Seems this ` The` is unexpected.

lhuett: Thanks @JoySnow for pointing that out. Looks like inadvertent typing. 
Corrected.

insights/components/rhel_version.py
```diff
@@ -0,0 +1,95 @@
...
+    '8.0'
+"""
+
+from insights.core.plugins import component
+from insights.parsers.redhat_release import RedhatRelease
+from insights.core.dr import SkipComponent
+
+
+@component(RedhatRelease)
+class Is_Rhel6(object):
```
xiangce: Should we follow the `CamelCase` naming style for classes? e.g IsRhel6, IsRhel8?

insights/components/rhel_version.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    def __init__(self, rhel):
+        if rhel.major == 6:
+            self.rhel_version = rhel.version
+        else:
+            raise SkipComponent('Not RHEL6')
+
+
+@component(RedhatRelease)
+class Is_Rhel7(object):
```
xiangce: Same comment as above

insights/components/rhel_version.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    def __init__(self, rhel):
+        if rhel.major == 7:
+            self.rhel_version = rhel.version
+        else:
+            raise SkipComponent('Not RHEL7')
+
+
+@component(RedhatRelease)
+class Is_Rhel8(object):
```
xiangce: Same comment as above

examples/rules/stand_alone.py
```diff
@@ -31,12 +33,26 @@
...
 class Specs(SpecSet):
     """ Datasources for collection from local host """
     hosts = simple_file("/etc/hosts")
 
 
-@parser(Specs.hosts)
+# If this runs on a server that is not Fedora or with an archive that
+# is not from a Fedora server this parser will not fire since the
+# Is_Fedora component raises SkipComponent if not Fedora
+@parser(Specs.hosts, Is_Fedora)
```
xiangce: One more question just came into mind, seems the `Requires`, `At Least One` or `Optional` feature also works to the @parser (since it's also a@ component).  Would you please add one more example for applying multiple dependencies/components to one @parser? For example, when a @parser works for both rhel-6 and rhel-7 (but except for rhel-8).

insights/components/rhel_version.py
```diff
@@ -0,0 +1,95 @@
...
+Example:
+
+    >>> type(Is_Rhel8)
+    <class 'insights.components.rhel_version.Is_Rhel8'>
+    >>> is_rhel8.rhel_version
+    '8.0'
+"""
+
+from insights.core.plugins import component
+from insights.parsers.redhat_release import RedhatRelease
```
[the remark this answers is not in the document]
lhuett: Thanks Xiangce, changed to use RedHatRelease combiner

insights/components/rhel_version.py
lhuett: Thanks Xiangce, good catch. Changed all three components to be CapWords/Camel Case.
IsRhel6, IsRhel7 and IsRhel8

examples/rules/stand_alone.py
lhuett: Sure thing, I reverted stand_alone.py and created a new module named skip_component.py in the examples. It demonstrates a parser and rule that just looks for Rhel8 and then another that  parser and rule that demonstrates a combination of Rhel6 and Rhel7 dependencies.

PR #1921, opened 2019-05-08
Compare rpms in the same way as the rpm project.

### LINKED ISSUE #1836: The default comparison method of InstalledRpm should only for packages from "Red Hat"
We got the following result from one archive:
~~~
{
  "fixed_liberasurecode": "liberasurecode-1.2.0-2.el7ost",
  "liberasurecode": "liberasurecode-1.1.0-3.ibm.el7",
  "node_type": "Controller",
  "rhosp_version": 10,
  "type": "rule"
}
~~~

The customer has installed the package `"liberasurecode-1.1.0-3.ibm.el7"` which is not from `Red Hat`. It's not good to report the rule to the customer in the case that CU has used customed packages.

Better make sure the default comparison method of  InstalledRpm is for packages from "Red Hat".
[Update] - ^ Do not want to break the existing rules

We could add another new comparison method to compare packages from all kinds of sources.

### LINKED ISSUE #1839: get_max(package_name) returns TypeError for some package lists

csams:
Implements rpm comparison using the [same algorithm as the rpm project](https://github.com/rpm-software-management/rpm/blob/master/lib/rpmvercmp.c).

Passes [all tests in the rpm project](https://raw.githubusercontent.com/rpm-software-management/rpm/master/tests/rpmvercmp.at) including those that document "buggy behavior" and that show "non-ascii characters are considered equal."

Removes distribution check from the rpm comparison since extracting the distribution from the release seems unreliable.

Fixes #1836 and Fixes #1839.


xiangce:
The test of an insights-rule is broken by this PR. It's because the rule has incorrectly accessed the internal attribute `_release_sep` which is removed. The insights-rule should be modified and will let you know when it's done.

xiangce:
@csams, Just FYI, the failed test of the insights-rule has been fixed today.

csams:
@xiangce thank you!

Issue #1924, opened 2019-05-10
redhat_release.RedHatRelease Combiner - Change `ParseException` to `SkipComponent` 


lhuett:
redhat_release.RedHatRelease combiner should raise `SkipComponent` instead of`ParseException` when no rhel version is found.

xiangce:
@lhuett, It's resolved in #1924. I go ahead and close this issue. Thanks.

PR #1927, opened 2019-05-14
Raise SkipException instead of ParseException in redhat_release

xiangce:
- When nothing available, skip the combiner


xiangce:
Resolve #1924

xiangce:
@lhuett - Thanks for the review. SkipComponent makes sense to me. Please check it again.

PR #1930, opened 2019-05-14
Adding a try/except to the request made when trying to update the egg

Jason-RH:
For BZ-1708682. This change will prevent the 'no such option' error from occurring when the the client cannot connect (if base_url was set to a bad value, for example), and will instead log the correct error and terminate the process if a connection cannot be made.


xiangce:
Closing it, please re-open it in another new pr if necessary

PR #1932, opened 2019-05-20
Add spec and parser for neutron ml2 config

smalleni:


insights/parsers/neutron_ml2_conf.py
```diff
@@ -0,0 +1,48 @@
+
+"""
+NeutronMl2Conf - file ``/etc/neutron/plugins/ml2/ml2_conf.ini``
+================================================
```
xiangce: Better keep the title and the `===` line in the same length.

PR #1933, opened 2019-05-20
Add alternative integration test fixture

bfahr:
* Add test fixture run_rule that provides integration test
  functionality like @archive_provider, but runs as a normal
  pytest.
* Add plugin and tests for new fixture.


lhuett:
Looks good Bob.

PR #1931, opened 2019-05-15
Add YumListInstalled Parser

csams:
Adds a new parser, `YumListInstalled`, that shares an interface with `InstalledRpms` but uses the output of `yum list installed` instead of `rpm`. It produces `YumInstalledRpm` instances, which are subclass instances of `InstalledRpm` with a `.repo` attribute to hold the originating yum repo name.

Fixes #1854 , although I still need an addition to the uploader.json.

PR #1934, opened 2019-05-21
Remove the subscription_manager_list_installed from insights_archive.py

xiangce:
- The plugins using this rule are against sosreport archives and it's not added to uploader.json ever


bfahr:
@xiangce @csams I'm setting this to WIP so that we can determine any relationship to PR #1931

bfahr:
Nevermind the WIP, wrong PR.

xiangce:
Removed. Please check it again @bfahr. Thanks.

PR #1937, opened 2019-05-21
Remove subscription_manager_list_consumed from insights_archive.py

xiangce:
- Need to change to use other specs in the related rule
- Will also remove it from the uploader.json


xiangce:
Updated, please have another check @bfahr. Thanks.

Issue #1906, opened 2019-05-05
Enhance parser RabbitMQReport 


JoySnow:
For some input data, this current `RabbitMQReport` parser fails and hides errors.
Enhancement is required.

JoySnow:
Related to https://github.com/RedHatInsights/insights-core/issues/1

chenlizhong:
Call trace:

~~~
WARNING:insights.core.dr:Traceback (most recent call last):
  File "/home/brq/user_a/insights-core/insights/core/dr.py", line 952, in run
    result = DELEGATES[component].process(broker)
  File "/home/brq/user_a/insights-core/insights/core/plugins.py", line 272, in process
    r = self.invoke(broker)
  File "/home/brq/user_a/insights-core/insights/core/plugins.py", line 63, in invoke
    return super(PluginType, self).invoke(broker)
  File "/home/brq/user_a/insights-core/insights/core/dr.py", line 653, in invoke
    return self.component(*args)
  File "rabbitmq_report.py", line 13, in report
    result = create_parser().parseString("\n".join(rabbitmq_report.content)).asDict()
  File "/home/brq/user_a/insights-core/insights/contrib/pyparsing.py", line 1217, in parseString
    raise exc
ParseException: Expected "}" (at char 825), (line:17, col:123)
~~~

PR #1938, opened 2019-05-21
Enhance parser for rabbitmq_report

JoySnow:
1. Fix the issue occurs when handling '\n' in content.
    When the '\n' is read from archive-files, it means letters ('\\', 'n'),
    and when from local test-data variables, it will be treated as 'a-new-line'.
    Base on these, the parsing failed toward archive-files.
    To fix this issue, let 'value_tword' can also cover letter '\'.
  2. Remove the hiding of parsing errors for friendly debug usage.

### DIFF
--- a/insights/parsers/rabbitmq.py
+++ b/insights/parsers/rabbitmq.py
@@ -35,7 +35,7 @@ def erlblock_parser():
 
     key = p.Word(p.alphas + '_')
     value_tnum = p.Word(p.nums + '.')
-    value_tword = p.Word(p.alphanums + '/"-[]:.()')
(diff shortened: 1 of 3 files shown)


xiangce:
@JoySnow - Is it related to #1906?

JoySnow:
@xiangce  Yes. Resolve https://github.com/RedHatInsights/insights-core/issues/1906.

xiangce:
It breaks the test of some rule related to `rabbitmq`, please have a check

JoySnow:
@xiangce  The broken rule related to this parser is enhanced. No other rules are affected in our repos.

Issue #1939, opened 2019-05-21
RFE: Easy way to monitor local status


mvollmer:
Hi,

I am working on adding some Insights support to Cockpit (https://cockpit-project.org/) and I am currently pondering how to monitor the local status, such as whether the machine is registered and when the last succesfull upload was, and whether there were any problems recently.

The obvious approach is to run `insights-client --status --quiet` and check the exit code.  This works well enough. It does require a working network connection and is a bit slow, and we wouldn't want to run it periodically just to keep the UI up-to-date for that reason.  Also, it doesn't give information about the latest upload or recent upload failures.

I had hopes that running `insights-client --status --quiet --offline` would be cheap and not require network access, but it crashes.  Also, it still checks the signatures on the eggs, which is again too slow for polling.

The sneaky approach is to monitor `/etc/insights-client/{.registered,.unregistered,.lastupload}` with inotify. I guess there is no stability guarantee for those files and using them as API is problematic.  (If have noticed that with `legacy_upload=True` they seem to be written on every invocation of the client... True?)

So what if insights-client maintains a little JSON record in `/etc/insights-client/status` that we can watch?  Something simple like
```
{
   registered: true,
   last_upload: <time-in-seconds-since-the-epoch>
}
```
for a start.

For failures, it is probably best to refer to the status and logs of the `insights-client` systemd unit.

I am happy to prototype this as part of my work on Cockpit and open a PR once I have something tangible.

mvollmer:
I opened #1945 to show more concretely what I have in mind.

PR #1941, opened 2019-05-22
client: Include timezone in timestamp files

mvollmer:


mvollmer:
```
insights-core/insights/client/utilities.py:51: in get_time
    return datetime.datetime.now(datetime.timezone.utc).astimezone().isoformat()
E   AttributeError: 'module' object has no attribute 'timezone'
```

Oh, does this have to work with Python 2?

bfahr:
@mvollmer yes, this project currently supports the versions of Python that are supported on Red Hat Enterprise Linux.

mvollmer:
@bfahr, thanks! I am on the fence now whether it is worth including the timezone.... wdyt?

bfahr:
@gravitypriest what do you think?

gravitypriest:
From the look of it, the only time `get_time()` is utilized is to generate timestamps for the `.registered` and `.unregistered` files. These have been deprecated for some time, and are only still utilized for Sat integration -- the content has no function, the file simply needs to "exist" or "not exist." So, IMO timezone information is not important.

mvollmer:
> From the look of it, the only time get_time() is utilized is to generate timestamps for the .registered and .unregistered files. 

For Cockpit integration, we read a timestamp from /etc/insights-client/.lastupload (but also see #1939).

> These have been deprecated for some time, and are only still utilized for Sat integration -- the content has no function, the file simply needs to "exist" or "not exist."

Aha! So the Sat people might also be interested in #1939, right?

mvollmer:
Let's close this.  I don't think we will actually read `.lastupload` from Cockpit ever.  Instead, we will get all local state from the `insights-client.timer` and `insights-client.servic` units.

PR #1943, opened 2019-05-22
[WIP] Adding unregister support to the client for non-legacy cases

Jason-RH:
Adding --unregister support when legacy_upload=False:
- host removed from inventory
- scheduler stopped

Some basic tests have been added, but there is still some work that needs to be done around testing unregister fully.

@gravitypriest let me know what you think, marking as WIP for now.


insights/client/connection.py
```diff
@@ -676,6 +676,28 @@ def unregister(self):
...
+        msg_name = determine_hostname(self.config.display_name)
+        try:
+            logger.debug("Unregistering %s", msg_name)
+            url = self.api_url + "/inventory/v1/hosts/" + results[0]['id']
+            net_logger.info("DELETE %s", url)
+            self.session.delete(url)
+            logger.info(
+                "Successfully unregistered from the Red Hat Insights Service")
+            return True
+        except requests.ConnectionError as e:
```
gravitypriest: We may want to expand this to `(requests.ConnectionError, requests.Timeout)` based on some other feedback we've been getting about request exception stuff.

insights/client/phase/v1.py
```diff
@@ -207,8 +207,12 @@ def post_update(client, config):
     # put this first to avoid conflicts with register
     if config.unregister:
         if reg_check:
-            logger.info('Unregistration is not currently available.')
-            sys.exit(constants.sig_kill_bad)
+            logger.info('Unregistering this host with Insights.')
+            if client.unregister():
+                get_scheduler(config).remove_scheduling()
```
gravitypriest: IMO this should be in client.handle_unregistration() for cleanliness parity with the old function. I can revisit if you disagree.

Jason-RH: So I was originally planning to handle it in handle_unregistration(), but I noticed that the daily scheduling was started in v1(https://github.com/RedHatInsights/insights-core/blob/unregister_host_delete/insights/client/phase/v1.py#L179) so I included it here to be consistent with register. I don't have a strong preference either way though so I don't mind changing it around.

insights/client/connection.py
Jason-RH: :+1:

insights/client/client.py
```diff
@@ -232,14 +232,18 @@ def handle_unregistration(config, pconn):
...
-    logger.info('Unregistration is not currently available.')
-    return False
-    # unreg = False  # pconn.unregister()
-    # if unreg:
-    #     # only set if unreg was successful
-    #     write_unregistered_file()
-    #     get_scheduler(config).remove_scheduling()
-    # return unreg
+
+    check = get_registration_status(config, pconn)
```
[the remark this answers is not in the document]
Jason-RH: Yep that's right. Will update

insights/client/connection.py
```diff
@@ -676,6 +676,28 @@ def unregister(self):
...
+        if self.config.legacy_upload:
+            return self._legacy_unregister()
+
+        results = self._fetch_system_by_machine_id()
+        msg_name = determine_hostname(self.config.display_name)
+        try:
+            logger.debug("Unregistering %s", msg_name)
+            url = self.api_url + "/inventory/v1/hosts/" + results[0]['id']
+            net_logger.info("DELETE %s", url)
+            self.session.delete(url)
```
gravitypriest: Should at least check the HTTP status here to make sure it was successful.

insights/client/phase/v1.py
```diff
@@ -207,8 +207,12 @@ def post_update(client, config):
     # put this first to avoid conflicts with register
     if config.unregister:
         if reg_check:
-            logger.info('Unregistration is not currently available.')
-            sys.exit(constants.sig_kill_bad)
+            logger.info('Unregistering this host with Insights.')
```
gravitypriest: s/with/from. "With" sounds kinda weird

insights/client/connection.py
gravitypriest: Maybe raise requests.ConnectionError so it just jumps into the except block if its a non-2xx code, or whatever this call is supposed to return.

PR #1948, opened 2019-05-23
Fix coding to be utf-8 instead of UTF-8

akshay196:
Old UTF-8 resulted in this error from GNU Emacs:
```
   Warning (mule): Invalid coding system 'UTF-8' is specified
   for the current buffer/file by the :coding tag.
   It is highly recommended to fix it before writing to a file.
```

### DIFF
--- a/insights/tests/test_sysconfig_options.py
+++ b/insights/tests/test_sysconfig_options.py
@@ -1,4 +1,4 @@
-# -*- coding: UTF-8 -*-
+# -*- coding: utf-8 -*-
 from insights.core import SysconfigOptions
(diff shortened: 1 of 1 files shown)


bfahr:
@akshay196 this doesn't seem necessary unless you are changing this file for some other reason.  [PEP 263](https://www.python.org/dev/peps/pep-0263) doesn't specify case.  If there's not some other reason for this change would you mind to close it?  Thx.

akshay196:
@bfahr Yes, you are right. As per PEP 263, both ways are acceptable. Closing this PR.

PR #1949, opened 2019-05-24
Enhacne items in sos_archive

wushiqinlou:


insights/specs/sos_archive.py
```diff
@@ -26,12 +26,12 @@ class SosSpecs(Specs):
     date = first_of([simple_file("sos_commands/general/date"), simple_file("sos_commands/date/date")])
     df__al = first_file(["sos_commands/filesys/df_-al", "sos_commands/filesys/df_-al_-x_autofs"])
     df__alP = simple_file("sos_commands/filesys/df_-alP")
-    df__li = first_file(["sos_commands/filesys/df_-li", "sos_commands/filesys/df_-ali_-x_autofs"])
+    df__li = first_file(["sos_commands/filesys/df_-ali", "sos_commands/filesys/df_-ali_-x_autofs"])
```
lonicerae: After a long discussion with @wushiqinlou , using `df_-ali` to replace `df_-li` can lead to false positive for some plugin, e.g. `filesystem_capacity.py` which has been fixed in `rhbz#1661037`. So it may be better to create another parser for the `ali` option.

wushiqinlou: @wushiqinlou - As new parser can be added only when plugin requests, so I suggest to remove `df_-li` from sos_archive.py and not create parser for `df_ali`.

PR #1952, opened 2019-05-24
Enhance rsyslog conf rhel8 parser

vishwanathjadhav:
Enhanced the parser for RHEL-8, The format of the file is changed and need different parsing methods.


bfahr:
If RanierScript is still supported in RHEL8 then can the new parser handle this?
```
if $programname == 'prog1' then {
   action(type="omfile" file="/var/log/prog1.log")
   if $msg contains 'test' then
     action(type="omfile" file="/var/log/prog1test.log")
   else
     action(type="omfile" file="/var/log/prog1notest.log")
}
```

Also removing `filterable` requires spec change approval.

wushiqinlou:
@vishwanathjadhav @xiangce - According to my test, `RanierScript` works well  in RHEL8. As current plugins using rsyslog checks the content line by line, there is no need to parse this complicated file in detail, `LogFileOutput` may be a good choice.

bfahr:
I like the `LogFileOutput` approach to simplify the parser.  Its also best to keep the existing spec filtered  since there was justification to filter the existing parser.

wushiqinlou:
I have checked with xiangce and vish, we all agree that it is better to parse this file like logfile. I will submit a new PR to handle it and the combiner for /etc/rsyslog.d/*.conf.

xiangce:
As per discussion with @vishwanathjadhav and @wushiqinlou, I'm closing this PR since a new PR #2121 is created for the same parser/combiner

PR #1958, opened 2019-05-27
Added parser for Candlepin PostgreSQL cp_event table recrd count

jhutar:
This is related to https://github.com/RedHatInsights/insights-core/pull/1860 - same product (Satellite 6), but different table (cp_event) in a different database (candlepin).

CC @mccun934


bfahr:
@jhutar this is similar to #1860, see my comment there about adding a new datasource to identify whether the system is running the correct version of satellite before running the command.

PR #1961, opened 2019-05-30
Parser for command 'docker exec nova_libvirt /bin/ls -lanR /root'

ppandey-redhat:
* New parser for parsing command `docker exec nova_libvirt /bin/ls -lanR /root'`.

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -565,6 +565,7 @@ class DefaultSpecs(Specs):
                            ])
     nova_crontab = simple_command("/usr/bin/crontab -l -u nova")
     nova_crontab_container = simple_command("docker exec nova_api_cron /usr/bin/crontab -l -u nova")
+    nova_libvirt_ls_root = simple_command("docker exec nova_libvirt /bin/ls -lanR /root")
     nova_uid = simple_command("/usr/bin/id -u nova")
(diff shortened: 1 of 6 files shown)


insights/specs/default.py
```diff
@@ -565,6 +565,7 @@ def semid(broker):
                            ])
     nova_crontab = simple_command("/usr/bin/crontab -l -u nova")
     nova_crontab_container = simple_command("docker exec nova_api_cron /usr/bin/crontab -l -u nova")
+    nova_libvirt_ls_root = simple_command("docker exec nova_libvirt /bin/ls -lanR /root")
```
bfahr: @ppandey-redhat since this command should only be executed on certain OSP nodes is that a way you could implement a datasource to determine whether the system is an OSP compute node before running this command?

psachin: @bfahr For this and https://github.com/RedHatInsights/insights-core/pull/1964, we can add a dependency component [IsOpenStackCompute](https://github.com/RedHatInsights/insights-core/blob/master/insights/components/openstack.py#L17) for a parser.

psachin: @bfahr I'm closing this PR in favour of https://github.com/RedHatInsights/insights-core/pull/2410 which also addresses the dependency component.

Issue #1928, opened 2019-05-14
Update is_hypervisor check in InstalledRpms


csams:
The [is_hypervisor check](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/installed_rpms.py#L229) in `InstalledRpms` targets only rhel6 and rhel7. Does it need updating for rhel8? Also, it seems like this should this be folded into / removed in favor of the [VirtWhat](https://github.com/RedHatInsights/insights-core/blob/master/insights/combiners/virt_what.py) combiner.

koalakangaroo:
@psachin - please check this.

psachin:
@csams Sadly we can't verify this at the moment because RHEL-8 host support for RHV is only available on RHV-4.4(No GA release date for RHV-4.4 GA is announced). However after the GA, we have to update `is_hypervisor_check` by adding `.el8ev` check(or something similar), but I'm in favor of deprecating `is_hypervisor_check` because:

1. The method is only useful for RHEV rules whereas `VirtWhat` combiner has more generic usage across other platforms including OpenStack.
2. The method is only used by [one rule](https://github.com/RedHatInsights/insights-plugins/blob/master/telemetry/rules/plugins/rhev/rhev_lro_on.py)
3. and we have existing RHEV rules using enhanced [method](https://github.com/RedHatInsights/insights-plugins/blob/master/telemetry/rules/plugins/rhev/__init__.py#L6) to check the hypervisor or `vdsm` package.

xiangce:
@JoySnow - Please have a look

JoySnow:
@psachin  I agree with the deprecation of the usage of `is_hypervisor` in our plugin repo.

psachin:
@csams The plugins are updated to use internal function to check the `vdsm` package. The property function `is_hypervisor` can be deprecated/remove.

PR #1963, opened 2019-05-30
Remove is_hypervisor from InstalledRpms

psachin:
All existing plugins using `is_hypervisor` are updated to use
plugins.rhev.get_vdsm_of_rhv4()

Resolves: #1928


insights/parsers/installed_rpms.py
```diff
@@ -226,13 +226,6 @@ def get_min(self, package_name):
...
             return min(self.packages[package_name])
 
-    @property
-    def is_hypervisor(self):
-        """bool: True if ".el[6|7]ev" exists in "vdsm".release, else False."""
-        rpm = self.get_max("vdsm")
-        return (True if rpm and rpm.release.endswith((".el6ev", ".el7ev")) else
-                False)
-
     # re-export get_max/min with more descriptive names
```
bfahr: @psachin this method should be marked as deprecated instead of being removed, see [Feature Deprecation] (https://insights-core.readthedocs.io/en/latest/api.html#feature-deprecation) for more information.  Please include information on which method its replaced by.

psachin: @bfahr The method which replaces `is_hypervisor` is internal to insights-plugins repo, users not having access to internal rules will not be able to use that method. Besides all existing plugins using `is_hypervisor` are now using that (internal)method, hence I decided to remove the function instead of deprecating it.

Please suggest how we should move ahead.

bfahr: @psachin the normal process is to mark it as deprecated because insights-core is a public project and we don't know about all uses of the the code so we deprecate and then at a later time we'll remove the code.

psachin: @bfahr 

Thanks for the feedback. The method is now deprecated.

miclark: Hey @psachin, since the deprecated function is still in the code, we probably should keep the test code.  Can you add those tests back?

psachin: @miclark Sure.

psachin: @miclark @bfahr 

I used `warnings.warn()` to deprecate `is_hypervisor` because the tests failed with `RecursionError` due to recursive nature of `@property`. Let me know if that's ok or should I raise an issue to come up with an approach to safely deprecate the property methods in the framework.

PR #1964, opened 2019-05-30
Parser for command 'docker exec nova_compute /bin/ls -larn /var/lib/nova'

ppandey-redhat:
New parser for parsing command `docker exec nova_compute /bin/ls -larn /var/lib/nova`.


insights/specs/default.py
```diff
@@ -558,6 +558,7 @@ def semid(broker):
     nmcli_dev_show = simple_command("/usr/bin/nmcli dev show")
     nova_api_log = first_file(["/var/log/containers/nova/nova-api.log", "/var/log/nova/nova-api.log"])
     nova_compute_log = first_file(["/var/log/containers/nova/nova-compute.log", "/var/log/nova/nova-compute.log"])
+    nova_compute_ls_var_lib_nova = simple_command("docker exec nova_compute /bin/ls -larn /var/lib/nova")
```
bfahr: @ppandey-redhat this is similar to the situation in #1961, can you implement a datasource that would limit execution of the `docker` command to only those systems where it is applicable?

psachin: @bfahr I'm closing this PR in favour of #2411 which also addresses the dependency component.

Issue #1794, opened 2019-04-05
`CpuInfo` parser cannot handle ARM and PowerPC


skontar:
PowerPC `cpuinfo` output looks like this:
```
processor	: 0
cpu		: POWER7 (architected), altivec supported
clock		: 3612.000000MHz
revision	: 2.0 (pvr 004a 0200)

processor	: 1
cpu		: POWER7 (architected), altivec supported
clock		: 3612.000000MHz
revision	: 2.0 (pvr 004a 0200)

processor	: 2
cpu		: POWER7 (architected), altivec supported
clock		: 3612.000000MHz
revision	: 2.0 (pvr 004a 0200)

processor	: 3
cpu		: POWER7 (architected), altivec supported
clock		: 3612.000000MHz
revision	: 2.0 (pvr 004a 0200)
```

ARM looks like this:
```
processor       : 0
model name      : ARMv7 Processor rev 5 (v7l)
BogoMIPS        : 38.40
Features        : half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm
CPU implementer : 0x41
CPU architecture: 7
CPU variant     : 0x0
CPU part        : 0xc07
CPU revision    : 5

processor       : 1
model name      : ARMv7 Processor rev 5 (v7l)
BogoMIPS        : 38.40
Features        : half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm
CPU implementer : 0x41
CPU architecture: 7
CPU variant     : 0x0
CPU part        : 0xc07
CPU revision    : 5

processor       : 2
model name      : ARMv7 Processor rev 5 (v7l)
BogoMIPS        : 38.40
Features        : half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm
CPU implementer : 0x41
CPU architecture: 7
CPU variant     : 0x0
CPU part        : 0xc07
CPU revision    : 5

processor       : 3
model name      : ARMv7 Processor rev 5 (v7l)
BogoMIPS        : 38.40
Features        : half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm
CPU implementer : 0x41
CPU architecture: 7
CPU variant     : 0x0
CPU part        : 0xc07
CPU revision    : 5

Hardware        : BCM2709
Revision        : a01041
[...]

PR #1965, opened 2019-05-31
cpuinfo - x86 stepping & other architectures

jsvob:
* fixes https://github.com/RedHatInsights/insights-core/pull/1965
* added x86 stepping
* added other architectures
* it is the same mess as before and the new things are accessible only through `["str indexes"]`, not through properties
* rules have to be changed to support non-x86 archs (no change for rules here - they crashed with the previous cpuinfo parser on non-x86 anyway)
* no change necessary for existing rules that support only x86
* if we want to make a deeper and more systematic refactoring, I'm interested in a discussion as to how to achieve it (I see multiple possible ways, e.g. concerning backwards compatibility or the level of abstraction from the raw cpuinfo data)


xiangce:
Fixed #1794

PR #1971, opened 2019-06-05
Add display_name spec

SteveHNH:
Display name was added to the archive in client 3.0.93. We need to be
able to use this in the platform to gather display name from legacy
archives


SteveHNH:
Turns out we don't need this since we already have this spec in insights_archive.py

Issue #1717, opened 2019-03-04
Enhance parser qemu_xml to parse new RHOSP specific attributes


psachin:
The parser should have new attributes to capture 
1. nova package version
2. flavor name
3. nova user name & uuid
4. nova project name & uuid
5. Product name (OpenStack Compute)
more...

Sample file: 
```
<!--
WARNING: THIS IS AN AUTO-GENERATED FILE. CHANGES TO IT ARE LIKELY TO BE
OVERWRITTEN AND LOST. Changes to this xml configuration should be made using:
  virsh edit instance-000008d6
or other application using the libvirt API.
-->

<domain type='kvm'>
  <name>instance-000008d6</name>
  <uuid>601da52c-450f-4d8e-b502-503431536fa69</uuid>
  <metadata>
    <nova:instance xmlns:nova="http://openstack.org/xmlns/libvirt/nova/1.0">
      <nova:package version="14.0.3-8.el7ost"/>
      <nova:name>VM_AT_01</nova:name>
      <nova:creationTime>2017-10-09 08:51:28</nova:creationTime>
      <nova:flavor name="vpc1-cf1-atos">
        <nova:memory>8096</nova:memory>
        <nova:disk>10</nova:disk>
        <nova:swap>0</nova:swap>
        <nova:ephemeral>0</nova:ephemeral>
        <nova:vcpus>4</nova:vcpus>
      </nova:flavor>
      <nova:owner>
        <nova:user uuid="96e9d2b749ea48fcb5a911e6f0e144f2">vpcdi1-01</nova:user>
        <nova:project uuid="5a50e9d0d19746158958be0c759793fb">vpcdi1</nova:project>
      </nova:owner>
      <nova:root type="image" uuid="1a05a423-dfae-428a-ae54-1614d8024e76"/>
    </nova:instance>
  </metadata>
  <memory unit='KiB'>8290304</memory>
  <currentMemory unit='KiB'>8290304</currentMemory>
  <vcpu placement='static'>4</vcpu>
  <cputune>
    <shares>4096</shares>
    <vcpupin vcpu='0' cpuset='1'/>
    <vcpupin vcpu='1' cpuset='2'/>
    <vcpupin vcpu='2' cpuset='13'/>
    <vcpupin vcpu='3' cpuset='14'/>
    <emulatorpin cpuset='1-2,13-14'/>
  </cputune>
  <numatune>
    <memory mode='strict' nodeset='0-1'/>
[...]

PR #1981, opened 2019-06-10
Enhance QemuXML to include Openstack metadata

psachin:
Resolves: #1717

### DIFF
--- a/insights/components/openstack.py
+++ b/insights/components/openstack.py
@@ -1,0 +1,27 @@
+"""
+IsOpenStackCompute
+==================
+
+The ``IsOpenStackCompute`` component uses ``PsAuxcww`` parser to determine
+OpenStack Compute node. It checks if 'nova-compute' process exist, if not raises
+``SkipComponent`` so that the dependent component will not fire. Can be added as
+a dependency of a parser so that the parser only fires if the
+``IsIsOpenStackCompute`` dependency is met.
+"""
+from insights.core.plugins import component
+from insights.parsers.ps import PsAuxcww
+from insights.core.dr import SkipComponent
+
+
+@component(PsAuxcww)
+class IsOpenStackCompute(object):
+    """The ``IsOpenStackCompute`` component uses ``PsAuxcww`` parser to determine
+    OpenStack Compute node. It checks if ``nova-compute`` process exist, if not
+    raises ``SkipComponent``.
+
+    Raises:
+        SkipComponent: When ``nova-compute`` process does not exist.
+    """
+    def __init__(self, ps):
+        if 'nova-compute' not in ps.running:
+            raise SkipComponent('Not OpenStack Compute node')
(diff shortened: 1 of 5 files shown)


psachin:
@xiangce Got it. The new class `OpenStackInstanceXML` handles the metadata now.

insights/parsers/qemu_xml.py
```diff
@@ -240,3 +237,88 @@ class VarQemuXML(BaseQemuXML):
         </domstatus>
     """
     pass
+
+
+class OpenStackInstanceXML(BaseQemuXML):
```
xiangce: This Parser didn't bond to any spec, it will not be able to get fired. what kind of spec do you want to run this parser? 


perhaps, this https://github.com/RedHatInsights/insights-core/pull/1920/ could help.

psachin: @xiangce The spec was missed. Fixed is latest commit. Thanks!!

xiangce: It's not good to apply the same spec to both QemuXML and the new one OpenStackInstanceXML, better add some depentancies.

bfahr: @psachin I agree with @xiangce regarding the second parser for the same spec.  We normally have a 1:1 ratio of specs to parsers.  What makes this case different?

psachin: @xiangce The parser will be fired only if `insights.components.openstack.IsOpenStackCompute` dependency is met.

insights/components/openstack.py
```diff
@@ -0,0 +1,27 @@
...
+class IsOpenStackCompute(object):
+    """The ``IsOpenStackCompute`` component uses ``PsAuxcww`` parser to determine
+    OpenStack Compute node. It checks if ``nova-compute`` process exist, if not
+    raises ``SkipComponent``.
+
+    Raises:
+        SkipComponent: When ``nova-compute`` process does not exist.
+    """
+    def __init__(self, ps):
+        if ('nova-compute' not in ps.running) and ('nova-conductor' in ps.running):
```
xiangce: I don't think this check follows the above description of the Component.
Think about there is not `'nova-compute'` and `'nova-conductor'` processes, it must not an OpenStackCompute node, right? but in that case, the IsOpenStackCompute component would be `NOT None` which will cause the Parser getting fired.

psachin: @xiangce The logic is aligned with the description but we have to think as a "negation".
The `SkipComponent` will be raised if `nova-compute` is not in the output of `ps` which means the node is not running compute services, hence not a compute node.

To support this, I've covered three tests in `components/tests/test_openstack.py` where:

1. `test_compute_node`: `nova-compute` exist in `ps`
2. `test_controller_node`: The controller node will not have `nova-compute` process running
3. `test_director_node`: The director node in OpenStack acts is all in node(compute + controller). `IsOpenStackCompute` will not throws an exception here.

xiangce: For a normal RHEL-host, when there is no 'nova-compute' or 'nova-conductor' process running, The `IsOpenStackCompute` will be something Truthy and cause this Parser getting fired incorrectly.

psachin: @xiangce Removed unwanted `nova-condutor` check.

PR #1982, opened 2019-06-10
Archive version info

Jason-RH:
Archiving the client version and egg version in a top level file 'version_info' that will be handled by pup.


Jason-RH:
@gravitypriest Also added a line in insights_archive.py so that pup can use Specs.version_info (similar to display_name)

Issue #1978, opened 2019-06-10
There is a bug in parser "PassengerStatus"


huali027:
When it parses the requests in the queue, it splits the info by four spaces(see line 138), however for some production data, it is just three spaces, then it causes the front-end data messy.
The production output example is as below:

```
Version : 4.0.18
Date    : 2019-06-10 03:17:38 +0100
Instance: 14745
----------- General information -----------
Max pool size : 60
Processes     : 7
Requests in top-level queue : 0

----------- Application groups -----------
/usr/share/foreman#default:
  App root: /usr/share/foreman
  Requests in queue: 0
  * PID: 39176   Sessions: 0       Processed: 194     Uptime: 24h 9m 0s
    CPU: 0%      Memory  : 488M    Last used: 20m 24s a
  * PID: 39342   Sessions: 0       Processed: 0       Uptime: 24h 8m 58s
    CPU: 0%      Memory  : 178M    Last used: 24h 8m 5
  * PID: 39377   Sessions: 0       Processed: 0       Uptime: 24h 8m 58s
    CPU: 0%      Memory  : 179M    Last used: 24h 8m 5
  * PID: 39478   Sessions: 0       Processed: 0       Uptime: 24h 8m 57s
    CPU: 0%      Memory  : 178M    Last used: 24h 8m 5
  * PID: 39525   Sessions: 0       Processed: 0       Uptime: 24h 8m 57s
    CPU: 0%      Memory  : 173M    Last used: 24h 8m 5
  * PID: 39614   Sessions: 0       Processed: 0       Uptime: 24h 8m 56s
    CPU: 0%      Memory  : 174M    Last used: 24h 8m 5

/etc/puppet/rack#default:
  App root: /etc/puppet/rack
  Requests in queue: 0
  * PID: 39667   Sessions: 0       Processed: 241     Uptime: 24h 8m 56s
    CPU: 0%      Memory  : 45M     Last used: 20m 24s
```

However, in front-end, we render it like:

```
----------- Application groups -----------
/usr/share/foreman#default:
  App root: /usr/share/foreman
  Requests in queue: 0

  * PID: 32008   Sessions: 0   Sessions: undefined       Processed: 1     Uptime: 7h 21m 58s
    CPU: 0%      Memory: 32008   Sessions: 0    Last used: 7h 21m 5

  * PID: 32138   Sessions: 0   Sessions: undefined       Processed: 0     Uptime: 7h 21m 56s
[...]

PR #1984, opened 2019-06-11
Fix a bug of PassengerStatus

xiangce:
- fix: #1978


huali027:
@xiangce, looks good to me

PR #1987, opened 2019-06-11
Modify deprecated pytest code for test fixture

bfahr:
* The pytest.config global is deprecated in newer versions of pytest
  so remove it now to avoid warnings in projects using newer versions
  of pytest with insights-core.


bfahr:
Closing this PR in favor of #1986.

PR #1989, opened 2019-06-11
[RFC] Add --compliance option to run oscap and upload the report

akofink:
This is currently how I'm running it locally, but the x-rh-identity isn't necessary for production.
```
sudo x-rh-identity=$X_RH_IDENTITY BYPASS_GPG=true insights-client --no-gpg --compliance
Running scan... this may take a while
Uploading Insights data.
Successfully uploaded report for rhel7-insights-client.virbr0.akofink-laptop.
```

### DIFF
--- a/insights/client/apps/compliance/__init__.py
+++ b/insights/client/apps/compliance/__init__.py
@@ -1,0 +1,67 @@
+from glob import glob
+from insights.client.connection import InsightsConnection
+from insights.client.constants import InsightsConstants as constants
+from insights.util.canonical_facts import get_canonical_facts
+from logging import getLogger
+from platform import linux_distribution
+from re import findall
+from subprocess import Popen, PIPE, STDOUT
+from sys import exit
+
+OSCAP_RESULTS_OUTPUT = '/tmp/oscap_results.xml'
+NONCOMPLIANT_STATUS = 2
+COMPLIANCE_CONTENT_TYPE = 'application/vnd.redhat.compliance.something+tgz'
+POLICY_FILE_LOCATION = '/usr/share/xml/scap/ssg/content/'
+logger = getLogger(__name__)
+
+
+class ComplianceClient:
+    def __init__(self, config):
+        self.config = config
+        self.conn = InsightsConnection(config)
+        self.hostname = get_canonical_facts().get('fqdn', '')
+
+    def oscap_scan(self):
+        policies = self.get_policies()
+        if not policies:
+            logger.error("ERROR: System does not exist!\n")
+            exit(constants.sig_kill_bad)
+        profile_ref_id = [policy['attributes']['ref_id'] for policy in policies][0]
+        scap_policy_xml = self.find_scap_policy(profile_ref_id)
+        self.run_scan(profile_ref_id, scap_policy_xml)
+        return OSCAP_RESULTS_OUTPUT, COMPLIANCE_CONTENT_TYPE
+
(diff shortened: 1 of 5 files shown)


gravitypriest:
This looks pretty close to what I envisioned. I'll take a closer look in a little bit. FYI you'll need to sign your commits for the checks to pass though. If you didn't get the email I'll forward it to you.

akofink:
My commits are signed, no?

gravitypriest:
The DCO check is failing so I assumed not. Might want to check with @bfahr

akofink:
Ah, nvm. `-s`, not `-S` :stuck_out_tongue:

akofink:
Updated with tests. I think this is good to go for re-review. I'm a bit confused why the tests are failing for RHEL6.

gravitypriest:
Looks like ambiguous replacement fields were added in Python 2.7 so those won't work on RHEL 6. Changing them them {0} and {1} ought to do the trick.

akofink:
This is ready for re-review. All the tests are in place and passing, and all values come from the config.

akofink:
Could I please get a re-review here?

insights/client/apps/compliance/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self, config):
+        self.config = config
+        self.conn = InsightsConnection(config)
+        self.hostname = get_canonical_facts().get('fqdn', '')
+
+    def oscap_scan(self):
+        policies = self.get_policies()
+        if not policies:
+            logger.error("ERROR: System does not exist!\n")
+            exit(1)
```
gravitypriest: The way the client / core wrapper works, to exit the process properly, you should use `sig_kill_bad` (failure exit code) or `sig_kill_ok` (success exit code) from the `constants.py` file. If you exit with 1 it sees that as a "bad egg" i.e., a traceback happened, so it will try to run again with the next available egg.

insights/client/apps/compliance/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        return filenames[0]
+
+    def run_scan(self, profile_ref_id, policy_xml):
+        logger.info("Running scan... this may take a while")
+        oscap = Popen(["oscap", "xccdf", "eval", "--profile", profile_ref_id, "--results", OSCAP_RESULTS_OUTPUT, policy_xml], stdout=PIPE, stderr=STDOUT)
+        if oscap.wait() and oscap.wait() != NONCOMPLIANT_STATUS:
+            logger.error("ERROR: Scan failed")
+            logger.error(oscap.stderr.read())
+            exit(1)
```
gravitypriest: See comment above

akofink:
Updated with `exit(sig_kill_bad)`

gravitypriest:
QE tests are failing on `from constants import sig_kill_bad`

what you'll want to do is

`from insights.client.constants import InsightsConstants as constants` and use `constants.sig_kill_bad` as we do in other modules.

Constants being inside a class is a little weird and something I want to change but that's how it has to be used for now.

akofink:
Whoops! Updated again :)

dLobatog:
:clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap:  

Thank you both!

PR #1993, opened 2019-06-13
 Add new spec and parser for lsinitrd

JoySnow:


insights/parsers/lsinitrd.py
```diff
@@ -0,0 +1,90 @@
...
+        entries = []
+        _unparsed_lines = []
+
+        for l in content:
+            if l and l[0] in file_types and l[1:4] in perm:
+                entries.append(l)
+            else:
+                _unparsed_lines.append(l)
+
+        d = Directory("", len(entries), entries)
```
xiangce: Better use the `insights.core.ls_parser.parse` instead of the `insights.core.ls_parser.Directory`, and raise a `Exception` when nothing to parse.

```python
if entries:
    self.data =  parse(entries, 'lsinitrd').get('lsinitrd')['entries']
else:
    raise SkipException
```

insights/specs/insights_archive.py
```diff
@@ -128,6 +128,7 @@ class InsightsArchiveSpecs(Specs):
     lsblk_pairs = simple_file("insights_commands/lsblk_-P_-o_NAME_KNAME_MAJ_MIN_FSTYPE_MOUNTPOINT_LABEL_UUID_RA_RO_RM_MODEL_SIZE_STATE_OWNER_GROUP_MODE_ALIGNMENT_MIN-IO_OPT-IO_PHY-SEC_LOG-SEC_ROTA_SCHED_RQ-SIZE_TYPE_DISC-ALN_DISC-GRAN_DISC-MAX_DISC-ZERO")
     lscpu = simple_file("insights_commands/lscpu")
     ls_lib_firmware = simple_file("insights_commands/ls_-lanR_.lib.firmware")
+    lsinitrd = simple_file("insights_commands/lsinitrd")
```
xiangce: Since this file is also collected in the sosreport, better add it to sos_archive.py together.

insights/parsers/tests/test_lsinitrd.py
```diff
@@ -0,0 +1,94 @@
...
+    assert 'dev/kmsg' in [l['name'] for l in lsdev]
+    assert 'dev/null' in [l['name'] for l in lsdev]
+    assert len(d.data) == 10
+    assert len(d.unparsed_lines) == 32
+    assert "udev-rules" in d.unparsed_lines
+
+
+def test_lsinitrd_docs():
+    failed_count, tests = doctest.testmod(
+        lsinitrd,
```
xiangce: This line is useless and can be removed.

insights/specs/insights_archive.py
JoySnow: Added.

insights/parsers/tests/test_lsinitrd.py
JoySnow: Removed.

insights/parsers/lsinitrd.py
JoySnow: `insights.core.ls_parser.parse` is used.

JoySnow:
@xiangce  It's ready for review now.

insights/parsers/lsinitrd.py
```diff
@@ -0,0 +1,91 @@
...
+                entries.append(l)
+            else:
+                _unparsed_lines.append(l)
+        self.unparsed_lines = _unparsed_lines
+
+        try:
+            d = ls_parser.parse(entries, '').get('')
+            self.data = d.get('entries')
+        except:
+            raise SkipException("Parsing failure for lsinitrd command output.")
```
xiangce: It's not OK to skip the exception raised by the `parse` since SkipException will just be thrown without any catch, we might miss some unknown formats.

BTW, it's not a good idea to use the empty `''` string the `dir` name, although it's a totally valid key in python.

PR #1996, opened 2019-06-13
Adding a RegistryPoint for version_info in Specs

Jason-RH:
@bfahr @csams Wasn't sure who to add on this PR. A previous PR (https://github.com/RedHatInsights/insights-core/commit/5819271cec0d56bd5ee5c11cfe8e9d6dfd433772#diff-2abbfc805ae6cd76d67de634a9fbb604R252) added a simple_file to InsightsArchiveSpecs, but I neglected to add a similar RegistryPoint to Specs.


bfahr:
@Jason-RH `version_info` doesn't have a path in the archive spec which means it will be in the root directory of the archive?  I assume that the client includes it in the archive in that location explicitly since there is not a default spec to collect it?

bfahr:
I see it in your previous commit.

PR #2005, opened 2019-06-18
Replace xml.etree.ElementTree with defusedxml.ElementTree to protect against malicious xml

lhuett:
Replaced 'xml.etree.ElementTree as ET' with
'defusedxml.ElementTree as ET' for secure xml parsing.


bfahr:
This is a [recommendation by the Python projec](https://docs.python.org/2/library/xml.html#xml-vulnerabilities)t to protect against multiple security issues.

Issue #1413, opened 2018-08-30
Fix command specs containing globs


csams:
For security reasons we don't execute datasources in a shell, so path globbing with commands doesn't work. That means that specs like [this](https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/default.py#L614) are invalid.
```
sysctl_conf_initramfs = simple_command("/bin/lsinitrd /boot/initramfs-*kdump.img -f /etc/sysctl.conf /etc/sysctl.d/*.conf")
```

bfahr:
There are two other specs that also have the same problem.
```
mlx4_port = simple_command("/usr/bin/find /sys/bus/pci/devices/*/mlx4_port[0-9] -print -exec cat {} \;")
md5chk_files = simple_command("/usr/bin/md5sum /dev/null /etc/pki/{product,product-default}/69.pem")
```

PR #2008, opened 2019-06-19
Update md5check to fix regex in spec

bfahr:
* Change spec to foreach_execute to fix regex problem identified
  in #1413
* Change spec ref to multi_output type
* Add combiner and tests combine parsers, similar to original parser
* Remove non-functional spec prelink_orig_md5 and associated parser
  PrelinkMD5


xiangce:
Just resolved the conflicts in the latest commit.

PR #2012, opened 2019-06-20
pyyaml dep set to version 5.1.1

Josca:
- let's use current version of `pyyaml` if possible, there is security issue in old version, see [CVE-2017-18342](https://nvd.nist.gov/vuln/detail/CVE-2017-18342).
- we can use old version only if it's necessary (python version < 2.7)
- it requires only small change in `setup.py` file

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from setuptools import setup, find_packages
 
 __here__ = os.path.dirname(os.path.abspath(__file__))
@@ -25,7 +26,6 @@ entry_points = {
 }
 
 runtime = set([
-    'pyyaml>=3.10,<=3.13',
     'six',
     'requests',
     'redis',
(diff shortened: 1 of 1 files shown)


bfahr:
@Josca what is the reasoning for this change?

Josca:
@bfahr we would like to upgrade pyyaml version in [vulnerability-engine](https://github.com/RedHatInsights/vulnerability-engine) because of security issue notified by GitHub. So I created PR to test if tests of insights-core will pass with the last version of pyyaml. Engine uses insights-core package so it uses old version of pyyaml.

Josca:
This will resolve security issue #2013

Josca:
Hi, according to [sources](https://github.com/RedHatInsights/insights-core/blob/insights-core-3.0.107-1033/setup.py), this change should be included in 3.0.107 version. However, when I install this version from pypi (`pip install insights-core==3.0.107`), pyyaml==3.* is installed :angry: . So I downloaded sources from pypi (`pip download insights-core==3.0.107`) and really this update is not included. Why?

Issue #1812, opened 2019-04-09
New parser for nftables


matysek:
nftables is the successor for iptables. In RHEL8 it is the default firewall interface. On RHEL8 even firewalld will use nftables as its backend. It is necessary to start collecting nftables configuration.

For configuring nftables there is a CLI tool `nft`. To print all defined firewall rules there is command `nft list ruleset`. Output on RHEL8 looks like this:

FirewallD is disabled:
```
# nft list ruleset
table ip filter {
	chain INPUT {
		type filter hook input priority 0; policy accept;
		ct state related,established counter packets 176 bytes 21662 accept
		meta l4proto icmp counter packets 0 bytes 0 accept
		iifname "lo" counter packets 0 bytes 0 accept
		meta l4proto tcp ct state new tcp dport 22 counter packets 3 bytes 180 accept
		counter packets 0 bytes 0 reject with icmp type host-prohibited
	}

	chain FORWARD {
		type filter hook forward priority 0; policy accept;
		counter packets 0 bytes 0 reject with icmp type host-prohibited
	}

	chain OUTPUT {
		type filter hook output priority 0; policy accept;
	}
}
```

FirewallD enabled:
```
# nft list ruleset
table ip filter {
	chain INPUT {
		type filter hook input priority 0; policy accept;
	}

	chain FORWARD {
		type filter hook forward priority 0; policy accept;
	}

	chain OUTPUT {
		type filter hook output priority 0; policy accept;
	}
}
table ip6 filter {
	chain INPUT {
		type filter hook input priority 0; policy accept;
	}

	chain FORWARD {
		type filter hook forward priority 0; policy accept;
	}

	chain OUTPUT {
		type filter hook output priority 0; policy accept;
	}
}
table bridge filter {
	chain INPUT {
		type filter hook input priority -200; policy accept;
	}

	chain FORWARD {
		type filter hook forward priority -200; policy accept;
	}

	chain OUTPUT {
		type filter hook output priority -200; policy accept;
	}
}
table ip security {
	chain INPUT {
		type filter hook input priority 150; policy accept;
	}

	chain FORWARD {
		type filter hook forward priority 150; policy accept;
	}

	chain OUTPUT {
		type filter hook output priority 150; policy accept;
	}
}
table ip raw {
	chain PREROUTING {
		type filter hook prerouting priority -300; policy accept;
	}

	chain OUTPUT {
		type filter hook output priority -300; policy accept;
	}
}
table ip mangle {
[...]

PR #2022, opened 2019-06-25
Parser nft list ruleset

vishwanathjadhav:
Added a parser for the RHEL8 specs `/usr/sbin/nft list ruleset`.
Insights_core_issue: https://github.com/RedHatInsights/insights-core/issues/1812


insights/parsers/nftables.py
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+    Raises:
+        SkipException: When content is empty.
+        ParseException: When content can not be parsed.
+
+    Examples:
+        >>> type(nftables_obj)
+        <class 'insights.parsers.nftables.NFTListRules'>
+
```
bfahr: @vishwanathjadhav can you include a couple of example of how to access the data?

insights/parsers/nftables.py
```diff
@@ -0,0 +1,198 @@
...
+            elif line.endswith('{') and nst_cnt == 1 and idx_key:
+                idx_key_2 = line.split('{')[0].strip()
+                self.data[idx_key][idx_key_2] = []
+                nst_cnt += 1
+            elif line.endswith('}') and nst_cnt == 2 and idx_key_2:
+                idx_key_2 = ''
+                nst_cnt -= 1
+            elif line.endswith('}') and nst_cnt == 1 and idx_key:
+                idx_key = ''
+                nst_cnt -= 1
```
bfahr: @vishwanathjadhav this may just be my lack of knowledge about `nft` but could you just look for `startswith` table | chain to determine the level here?  What about using `self.tables` instead of `self.data` which is not very descriptive, to represent the type of data actually being stored?

insights/parsers/nftables.py
```diff
@@ -0,0 +1,198 @@
...
+        \t\t	ip saddr 192.168.122.0/24 ip daddr 255.255.255.255 counter packets 0 bytes 0 return
+        \t\t	meta l4proto tcp ip saddr 192.168.122.0/24 ip daddr != 192.168.122.0/24 counter packets 0 bytes 0 masquerade to :1024-65535
+        \t\t	meta l4proto udp ip saddr 192.168.122.0/24 ip daddr != 192.168.122.0/24 counter packets 0 bytes 0 masquerade to :1024-65535
+        \t\t	ip saddr 192.168.122.0/24 ip daddr != 192.168.122.0/24 counter packets 0 bytes 0 masquerade
+        \t\t}
+        \t\t
+        \t\tchain OUTPUT {
+        \t\t	type nat hook output priority -100; policy accept;
+        \t\t}
+        }
```
bfahr: The sample data could probably be shorter and still get the idea across.  Also the `\t` are a little distracting.

insights/specs/default.py
```diff
@@ -557,6 +557,7 @@ def semid(broker):
     nfnetlink_queue = simple_file("/proc/net/netfilter/nfnetlink_queue")
     nfs_exports = simple_file("/etc/exports")
     nfs_exports_d = glob_file("/etc/exports.d/*.exports")
+    nft_ruleset = simple_command("/usr/sbin/nft list ruleset")
```
bfahr: You need to include an archive spec for this command spec so that the information can be read from an insights archive.

xiangce:
Since @vishwanathjadhav is in the DevConf@Bangalore, I just update this PR as per Bob 's feedback. @bfahr please have a check again when free.

bfahr:
@vishwanathjadhav this needs to go through the spec approval process.

bfahr:
@vishwanathjadhav have you submitted a request to approve this spec?  If so I couldn't find it.

bfahr:
@vishwanathjadhav @xiangce I apologize, I think I may have pushed you in the wrong direction, the better solution for this parser implementation would be as a ConfigParser, see the [CorosyncConf parser](https://github.com/RedHatInsights/insights-core/blob/dcf07ef24d8776fe2ca61a955cce7ad74bb156d4/insights/parsers/corosync.py#L94) as an example.

Issue #1793, opened 2019-04-05
`modinfo` on all loaded modules


skontar:
If there is a way how to call `modinfo` on all loaded modules, it should be implemented.

The detection script uses the following construction to get all non-retpoline modules:
`lsmod | awk '{ if (NR > 1 && system("modinfo " $1 " | grep -q -e \"retpoline:[ ]*Y\"")) print $1 }'`

Storing `modinfo` for every module returned from `lsmod` would be preferable.

miclark:
This should be covered when we move to "core 3" collection.

jsvob:
@miclark Is there a way I can maybe try to help out with this? I already know the basics of creating a new traditional spec, a new parser, a new combiner :) But this is new, I know nothing about it, and in the future, I will need to know how to develop core 3 collection anyway, so why not learn it now:)

jsvob:
FTR, I've received guidance on Slack. Core collection will not happen for some time so this is not urgent.

jsvob:
@skontar  So IIUC, once core3 collection is in production, we start to get modinfo data.

skontar:
@jsvob 🏅

bfahr:
@skontar @jsvob was this satisified by #2039 ?

jsvob:
@bfahr yes

PR #2039, opened 2019-07-02
Modinfo core collection.

jsvob:
https://github.com/RedHatInsights/insights-core/issues/1793


insights/combiners/modinfo.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@combiner(ModInfoEach)
+class ModInfo:
+    """
+    Combiner for accessing all the modinfo outputs.
+
+    Examples:
+        >>> type(modinfo_obj)
+        <class 'insights.combiners.modinfo.ModInfo'>
```
xiangce: In the test of python2, here `<type instance>` is expected, to the correct expected, better inherit the combiner class from `object` explicitly, e.g `class ModInfo(object):` Or `class ModInfo(dict):` just like the comment above.

matysek: @xiangce do we still need to retain python2 compatibility for insights-core? I though we are py3 only here.

jsvob: @matysek : @xiangce has kindly explained on today's mtg that insights-core will be run in environments with python 2.6 and 2.7, so we need to have that compatibility. I didn't realize it before.

xiangce:
@jsvob - This PR is ready to merge after the following flake8 issues being fixed:

```
+ flake8
./insights/combiners/tests/test_modinfo.py:15:9: F841 local variable 'x' is assigned to but never used
```

PR #2040, opened 2019-07-02
New parser and combiner for dmesg

jobselko:


jobselko:
https://github.com/RedHatInsights/insights-core-assets/pull/216

insights/specs/default.py
```diff
@@ -211,6 +211,7 @@ def is_ceph_monitor(broker):
     dirsrv_errors = glob_file("var/log/dirsrv/*/errors*")
     display_java = simple_command("/usr/sbin/alternatives --display java")
     dmesg = simple_command("/bin/dmesg")
+    dmesg_log = simple_file("/var/log/dmesg")
```
xiangce: Seems the output of `/bin/dmesg` contains the content of file `/var/log/dmesg`, are there any specific rule depends on the `/var/log/dmesg` file but not the output of `/bin/dmesg`?

If the `/bin/dmesg` does not work to you, it's OK to add the new parser, but in that case, you need to implement the parser in a python file.

jobselko: The `dmesg` command shows the current content of the kernel ring buffer while the `/var/log/dmesg` file is filled directly after boot. Therefore, the outputs could be different.

I am working on the new dmesg combiner for the dmesg command/file.

Issue #2041, opened 2019-07-03
'grubenv' does not updated as per grubby


xiangce:
After executing the command:
```
grubby --args "noapic" --update-kernel=ALL
```

The `/boot/grub2/grubenv` doesn't change, even after a reboot.

Need to change to use the `/boot/loader/entries/*.conf`

Issue #1983, opened 2019-06-11
Refine the logic of parsers and combiners in "grub_conf.py"


xiangce:
The code of the existing parsers in [parsers/grub_conf.py](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/grub_conf.py) and combiners in [combiners/grub_conf.py](https://github.com/RedHatInsights/insights-core/blob/master/insights/combiners/grub_conf.py) need a tune-up

Issue #1650, opened 2019-01-24
GRUB2 on RHEL8 - add support for blscfg


jsvob:
* GRUB2 in RHEL8 is configured in such a way that it uses blscfg and entries in /boot/loader instead of menuentry entries in grub.cfg as is in RHEL7.
* The current grub_conf parser doesn't work with blscfg files.

References:

* https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault
* https://www.thegeekdiary.com/modifying-etc-default-grub-rhel-7-v-s-rhel-8/?PageSpeed=noscript

psachin:
I'll take this task

xiangce:
@psachin - have you started working on this issue? If not, I will work on it.

Duplicate issue:  #1972

xiangce:
As discussed in the meeting, I'll take this issue.

jsvob:
It still doesn't work as expected. Steps to reproduce:

1. RHEL8 machine with the default configuration
2. grubby --update-kernel=ALL --args=selinux=0
3. reboot
4. sestatus shows disabled
5. /boot/grub2/grubenv (or the EFI variant on EFI systems) doesn't contain the selinux=0 argument:

```

# GRUB Environment Block
saved_entry=0
kernelopts=root=/dev/mapper/rhel-root ro crashkernel=auto resume=/dev/mapper/rhel-swap rd.lvm.lv=rhel/root rd.lvm.lv=rhel/swap rhgb quiet 
boot_success=0
boot_indeterminate=0
##########################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################

```

6. Only this file (depending on the particular version) contains the argument added by grubby:  `/boot/loader/entries/7b72675be917477dbfea43edd54377ba-4.18.0-80.el8.x86_64.conf`
7. Again, the reason is that insights don't parse blscfg configuration and don't see it.

Thank you

xiangce:
@jsvob - Thanks for checking. I have opened #2041 for this problem too. And I have raised #2043 to fix it by collecting the files under the `/boot/loader/entries/` just like you mentioned above.

Will let you know when the #2043 is merged.

jsvob:
@xiangce thank you!

PR #2043, opened 2019-07-04
Update GrubConf for RHEL 8

xiangce:
- fix: #1983 
- fix: #2041
- fix: #1650


insights/specs/default.py
```diff
@@ -305,8 +306,6 @@ def docker_container_ids(broker):
     grub1_config_perms = simple_command("/bin/ls -l /boot/grub/grub.conf")  # RHEL6
     grub2_cfg = simple_file("/boot/grub2/grub.cfg")
     grub2_efi_cfg = simple_file("boot/efi/EFI/redhat/grub.cfg")
-    grub2_efi_grubenv = simple_file("/boot/efi/EFI/redhat/grubenv")
-    grub2_grubenv = simple_file("/boot/grub2/grubenv")
     grubby_default_index = simple_command("/usr/sbin/grubby --default-index")  # only RHEL7 and updwards
```
bfahr: @xiangce do the above 2 specs work correctly, it seems from the multiple issues that maybe the parser did not work.  If they were working correctly then we should deprecate the specs and parsers instead of removing them.  If they did not work correctly then it is OK to remove them.

xiangce: @bfahr - Yes, these two specs are just added for a new rule. They didn't work as excepted and was removed from the rule. Thanks.

insights/specs/default.py
```diff
@@ -105,6 +105,7 @@ def is_aws(broker):
     blkid = simple_command("/sbin/blkid -c /dev/null")
     bond = glob_file("/proc/net/bonding/bond*")
     bond_dynamic_lb = glob_file("/sys/class/net/bond[0-9]*/bonding/tlb_dynamic_lb")
+    boot_loader_entries = glob_file("/boot/loader/entries/*.conf")
```
jsvob: @xiangce @bfahr  It seems to me that `boot_loader_entries` is not collected via uploader jsons and as a result, the requirements of the GrubConf combiner are not met and the combiner and anything that depends on it doesn't run.

It seems so when I look at https://github.com/RedHatInsights/insights-core-assets/tree/uploader_json and when I try to generate local tar.gz archives and run them in a freshly built dev environment with the fresh insights-core 3.0.

xiangce: @jsvob - Since the related PR to add the spec is still not merged, see: https://github.com/RedHatInsights/insights-core-assets/pull/223

Issue #1976, opened 2019-06-10
Please add parser:OVSvsctlListBridge for sosreport


zerodayz:
Please add parser:OVSvsctlListBridge for sosreport

https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/ovs_vsctl_list_bridge.py

miclark:
This appears to require the addition of a new sosreport spec.  @zerodayz, do you have the location of the file in a sosreport?

zerodayz:
@miclark We have got `/sosreport/sos_commands/openvswitch/ovs-vsctl_list_bridge_br-int` and others such as `/sosreport/sos_commands/openvswitch/ovs-vsctl_list_bridge_br-ex` These files combined are making `ovs-vsctl list bridge` output

xiangce:
For insights-archive, there is a spec to collect the output of the command `ovs-vsctl list bridge` which will output the information about all `bridges`. But in sosreport-archive, this command has been separated into individual files for each `bridge`.


To support both the separated files in sosreport and single file in insights-archive, the following steps are required:

1. Add a new parser OVSvsctlListBridgeSos to just parse the separated files in sosreport, which stored parsed result in a list.
   - Since there are existing rules used the `OVSvsctlListBridge`, better do not apply it to the sos-archive separated file directly.
2. Add a combiner to combine the parsed result of both `OVSvsctlListBridge` and `OVSvsctlListBridgeSos`

xiangce:
I'll take this issue

zerodayz:
@xiangce  We might not need to do it, if the sosreport PR will be merged. Requested adding `ovs-vsctl list bridge` to the sosreport archive

xiangce:
@zerodayz - cool, that's great. Then, I will fix this issue until  https://github.com/sosreport/sos/pull/1698 being merged

miclark:
Having sosreport/sos#1698 merged will not solve the problem since users are not required to update sosreport.  I think the combiner still makes sense.

xiangce:
Ack, makes sense to me. Will do the corresponding updates

PR #2045, opened 2019-07-08
Rabbitmq docker bundle

zerodayz:
Adding RabbitMQ docker bundle check for OSP containers


insights/specs/sos_archive.py
```diff
@@ -98,7 +98,11 @@ class SosSpecs(Specs):
         simple_file("sos_commands/foreman/foreman-debug/qpid_stat_subscriptions"),
         simple_file("sos_commands/foreman/foreman-debug/qpid-stat-u")
     ])
-    rabbitmq_report = simple_file("sos_commands/rabbitmq/rabbitmqctl_report")
+    rabbitmq_report = first_of([
```
xiangce: Since the `glob_file` means a couple of files which matches the wildcard string will get this parser being fired, this update will change the structure of the parser result, from a single Parser object to a list of Parser objects, which will break all the existing rules depend on this Parser.

If there any difference (e.g. the usage in rules) between this Parser against containers or against normal-hosts, but to avoid breaking the existing rules, seems we'd better add a new spec&parser for the new one against containers.

Or, is it possible to add a new command to the `sos` package to collect a single `rabbitmqctl_report` for one container? Just like: https://github.com/RedHatInsights/insights-core/issues/1976

zerodayz: Hi @xiangce The * is there because the file changes per server, it can be any number. Do you have anything to catch that in insights?

xiangce: @zerodayz - per the discussion, I have opened an issue #2049 for this PR. And I will submit another PR for the parser and close this one. Thanks.

PR #2047, opened 2019-07-08
[Enhancement] DockerList: Return data as None

psachin:
Return the value of '.data' as None instead of '.no_data' as True

The attrib `no_data` is not very useful in the rules. It is better to assert by
checking the value of '.data' instead of checking '.no_data'


insights/parsers/docker_list.py
```diff
@@ -85,7 +85,7 @@ def parse_content(self, content):
         """
         self.rows = []
         if len(content) < 2:
-            self.no_data = True
+            self.data = None
```
xiangce: Why not raise a `SkipException` directly?

psachin: @xiangce I did not wanted to change the original implementation, hence did not use `SkipException`. Do you want to use `SkipException`?

xiangce: Generally, it's recommended to keep the original implementation for any update. But in that case, the original code was made a long time ago and at that time `SkipException` is not introduced to the framework yet.

Since you want to do `enhancement` to it, better involve the relatively `new` features which could also bring external benefits.

psachin: Got your point!

Issue #2051, opened 2019-07-10
NmcliConnShow parser fails when there is warning in cmd output


vishwanathjadhav:
`NmcliConnShow` unable to process the command output when there is version miss match warning is present in the output. To handle this warning content validation check is mandetory.

Sample Output:

~~~
Warning: nmcli (1.0.0) and NetworkManager (1.0.6) versions don't match. Use --nocheck to suppress the warning.
NAME           UUID                                  TYPE            DEVICE
enp0s8         00cb8299-feb9-55b6-a378-3fdc720e0bc6  802-3-ethernet  --
enp0s3         bfb4760c-96ce-4a29-9f2e-7427051da943  802-3-ethernet  enp0s3"
~~~

vishwanathjadhav:
This issue has been fixed in https://github.com/RedHatInsights/insights-core/pull/2052

bfahr:
PR has been merged.

PR #2052, opened 2019-07-10
NmcliConnShow parser fails when there is warning in cmd output 

vishwanathjadhav:
The issue: https://github.com/RedHatInsights/insights-core/issues/2051

### DIFF
--- a/insights/parsers/nmcli.py
+++ b/insights/parsers/nmcli.py
@@ -178,19 +178,7 @@ class NmcliConnShow(CommandParser):
 
     """
     def parse_content(self, content):
-        table_columns = set(["NAME", "UUID", "TYPE", "DEVICE"])
-        first_row = set(content[0].split())
-        if table_columns.intersection(first_row) == first_row:
-            self.data = parse_delimited_table(content)
-        elif len(content) > 1:
-            second_row = set(content[1].split())
-            if table_columns.intersection(second_row) == second_row:
-                self.data = parse_delimited_table(content[1:])
-            else:
-                raise SkipException("Invalid Contents!")
-        else:
-            raise SkipException("Invalid Contents!")
-
+        self.data = parse_delimited_table(content, heading_ignore=["NAME", "UUID", "TYPE", "DEVICE"])
         self._disconnected_connection = []
         for all_connection in self.data:
             if all_connection['DEVICE'] == "--":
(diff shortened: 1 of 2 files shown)


insights/parsers/nmcli.py
```diff
@@ -178,7 +178,19 @@ class NmcliConnShow(CommandParser):
 
     """
     def parse_content(self, content):
-        self.data = parse_delimited_table(content)
+        table_columns = set(["NAME", "UUID", "TYPE", "DEVICE"])
```
xiangce: The following logic can be simplified by utilizing the parameters of the `parse_delimited_table`
See:
https://github.com/RedHatInsights/insights-core/blob/8715825b19890cea227c19f6efbb162b5fc80de8/insights/parsers/__init__.py#L396

vishwanathjadhav: Done

insights/parsers/tests/test_nmcli.py
```diff
@@ -91,11 +91,6 @@
 enp0s3         bfb4760c-96ce-4a29-9f2e-7427051da943  802-3-ethernet  enp0s3"
 """.strip()
 
-NMCLI_SHOW_ERROR_2 = """
```
xiangce: I think this is a good abnormal test case which should be handled correctly since you have already imagined this kind of situation.

Better Skip it, e.g. when the `self.data` is empty. how do you think?

vishwanathjadhav: @xiangce  - Done, Added exception check, if data is invalid then it directly throw an error hence added  `try:  <code>  except:<code>` block.

insights/parsers/nmcli.py
```diff
@@ -178,7 +178,10 @@ class NmcliConnShow(CommandParser):
 
     """
     def parse_content(self, content):
-        self.data = parse_delimited_table(content)
+        try:
+            self.data = parse_delimited_table(content, heading_ignore=["NAME", "UUID", "TYPE", "DEVICE"])
+        except:
+            raise SkipException("Inavalid Contents!")
```
bfahr: @vishwanathjadhav a minor spelling error, I believe this should be "Invalid".

vishwanathjadhav: Done

Issue #2056, opened 2019-07-12
JSONParser does not handle JSON texts consisting of an array


fcami:
As per https://tools.ietf.org/html/rfc7159.html#section-2 a JSON text can be many things including an array [] or an object {}.

The JSONParser implementation at:

https://github.com/RedHatInsights/insights-core/blob/8c633203ba72de49b70d0612e45398ed036197d4/insights/core/__init__.py#L717

incorrectly expects a JSON text to be an object (e.g. between {}) as it calls dict() on it.
See: https://docs.python.org/3/library/json.html#encoders-and-decoders

If the JSON text parsed is an array instead a backtrace occurs:

```
WARNING:insights.core.dr:Traceback (most recent call last):
  File "insights-core/insights/core/dr.py", line 954, in run
    result = DELEGATES[component].process(broker)
  File "insights-core/insights/core/dr.py", line 673, in process
    return self.invoke(broker)
  File "insights-core/insights/core/plugins.py", line 123, in invoke
    return self.component(dep_value)
  File "insights-core/insights/core/__init__.py", line 86, in __init__
    self._handle_content(context)
  File "insights-core/insights/core/__init__.py", line 89, in _handle_content
    self.parse_content(context.content)
  File "insights-core/insights/core/__init__.py", line 724, in parse_content
    self.doc = from_dict(self.data)
  File "insights-core/insights/configtree/__init__.py", line 369, in from_dict
    return Root(children=inner(dct))
  File "insights-core/insights/configtree/__init__.py", line 352, in inner
    for name, v in d.items():
AttributeError: 'list' object has no attribute 'items'
```

fcami:
hi @bfahr @csams could you please have a look at the PR?

miclark:
Assigning to @fcami since he seems to be working it.

bfahr:
Fixed by #2057

PR #2057, opened 2019-07-12
JSONParser: protect call to from_dict (#2056)

fcami:
JSONParser currently expects to parse JSON objects only.
Calling from_dict will fail if the JSON text is an array.
Protect the call to from_dict and do nothing if the text is
an array.

### DIFF
--- a/insights/tests/test_json_parser.py
+++ b/insights/tests/test_json_parser.py
@@ -7,10 +7,17 @@ from insights.tests import context_wrap
 class MyJsonParser(JSONParser):
     pass
 
+json_test_strings = {
+    '{"a": "1", "b": "2"}': {'a': '1', 'b': '2'},
+    '[{"a": "1", "b": "2"},{"a": "3", "b": "4"}]':
+        [{'a': '1', 'b': '2'}, {'a': '3', 'b': '4'}]
+}
+
 
 def test_json_parser_success():
-    ctx = context_wrap('{"a": "2"}')
-    assert MyJsonParser(ctx)
+    for jsonstr in json_test_strings:
+        ctx = context_wrap(jsonstr)
+        assert MyJsonParser(ctx).data == json_test_strings[jsonstr]
 
 
 def test_json_parser_failure():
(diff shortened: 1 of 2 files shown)


fcami:
Parsing json texts consisting of arrays is mandatory to parse FreeIPA HealthCheck logs and output:
https://github.com/freeipa/freeipa-healthcheck

csams:
Thanks for the PR! The YAMLParser has the same issue.

The object that's normally assigned to doc isn't a dictionary but an object from `insights.parsr.query` that supports different ways to query a deeply nested structure that can often be created from a dictionary.

I'd rather just remove the `self.doc =  from_dict` assignments altogether from both classes if no current rulesets use actually use it (I couldn't find any).

fcami:
hi @csams thanks for the review!
I've updated the PR for JSONParser only and will do another one for YAMLParser tomorrow.

insights/tests/test_json_parser.py
```diff
@@ -8,11 +8,16 @@ class MyJsonParser(JSONParser):
...
 
 
+def test_json_parser_array_success():
+    ctx = context_wrap('["foo", {"bar":["baz", null, 1.0, 2]}]')
+    assert MyJsonParser(ctx)
+
+
 def test_json_parser_failure():
     ctx = context_wrap("boom")
     with pytest.raises(ParseException) as ex:
```
bfahr: @fcami would you mind to add some additional tests here to make sure not only that the parser returns an object, but that the object contains the correct values.  Those tests should have been in the origin parser, but somehow slipped by.  Also please add tests to confirm that the `LegacyItemAccess` mixin, which assumes that `self.data` is a dictionary will work with a `list` result.

fcami: hi @bfahr I've pushed tests designed the way you wanted for JSONParser.
However wrt LegacyItemAccess I really don't have the knowledge of how it should work.

fcami:
hi @bfahr @csams I've implemented the requested changes to the test, could you please have a 2nd look? thanks

Issue #2069, opened 2019-07-16
ProcMounts doesn't handle paths contain spaces, e.g."/run/media/VMware Tools"


xiangce:
At first, need to figure out how the path with spaces is represented in `/proc/mounts`?

then take corresponding enhancement for it.

ppandey-redhat:
@xiangce 

In `/proc/mounts`,  spaces are encoded using `\040 (octal value)` similar to they are handled in `fstab`. Since the various options in /etc/fstab are separated by spaces, it is not possible to put spaces in the path of the mount point. This limitation is easily bypassed by typing "\ 040" instead of a space in the configuration file.
Ref: https://unix.stackexchange.com/questions/317476/shell-code-to-check-if-a-device-or-file-with-spaces-in-the-path-is-mounted

Example: `/dev/sda1 /mnt/first\040disk ext3 options 0 0`

Issue #2059, opened 2019-07-15
Update the Mount parser and specs depends on '/proc/mounts'


xiangce:
1. Update the existing parser `Mount` to make it a general base class for `/bin/mount` and `/proc/mounts`
2. The `dumpe2fs_h` spec depends on the content of `/proc/mounts` which is just added as a new Spec.
Better re-implement the `dumpdev` as per the new parser.

PR #2070, opened 2019-07-16
Update mount

xiangce:
- Remove the `regex` and use `split` instead
- Refine the related classes
- Keep compatible with the original version
- fix: #2059
- fix: #2069

### DIFF
--- a/insights/specs/insights_archive.py
+++ b/insights/specs/insights_archive.py
@@ -46,7 +46,6 @@ class InsightsArchiveSpecs(Specs):
     docker_info = simple_file("insights_commands/docker_info")
     docker_list_containers = simple_file("insights_commands/docker_ps_--all_--no-trunc")
     docker_list_images = simple_file("insights_commands/docker_images_--all_--no-trunc_--digests")
-    dumpdev = simple_file("insights_commands/awk_.ext_234_._print_1_.proc.mounts")
     engine_config_all = simple_file("insights_commands/engine-config_--all")
     ethtool = glob_file("insights_commands/ethtool_*", ignore="ethtool_-.*")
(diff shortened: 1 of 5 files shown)


insights/specs/insights_archive.py
```diff
@@ -46,7 +46,6 @@ class InsightsArchiveSpecs(Specs):
     docker_info = simple_file("insights_commands/docker_info")
     docker_list_containers = simple_file("insights_commands/docker_ps_--all_--no-trunc")
     docker_list_images = simple_file("insights_commands/docker_images_--all_--no-trunc_--digests")
-    dumpdev = simple_file("insights_commands/awk_.ext_234_._print_1_.proc.mounts")
```
bfahr: @xiangce I believe it's OK to remove this spec since it doesn't appear in the uploader JSON.

bfahr:
This depends on [insighs-core-assets PR 227](https://github.com/RedHatInsights/insights-core-assets/pull/227).

JoySnow:
It's good to me.

csams:
@xiangce @bfahr @ppandey-redhat Hey everybody, let's lean away from the "pretend to be a dict" or dynamic attribute setting mixins and just use regular attributes initialized in `__init__`.  See my [comment on the related issue](https://github.com/RedHatInsights/insights-core/issues/2081#issuecomment-513773022).

xiangce:
Hey all, since the update to the `MountOpts` will impact the `FStab` too, I just reverted all the update to the `MountOpts` and `MountEntry` and merge this PR at first. And open another PR to resolve the #2081

xiangce:
The following updates are in #2104

Issue #2071, opened 2019-07-16
RFE: add a parser for FreeIPA's HealthCheck logs


fcami:
The [FreeIPA](https://www.freeipa.org/page/Main_Page) project now ships a tool able to detect and in some cases predict future issues with an IPA server: [freeipa-healthcheck](https://github.com/freeipa/freeipa-healthcheck).

The output of this tool is a JSON-formatted array listing all the checks that were done by its plugins.
The tool is also able to log into /var/log/ipa/healthcheck/healthcheck.log every day.

It would be nice if Insights-Core had a parser for such a tool.

miclark:
Addressed by #2072

bfahr:
Fixed by #2072

PR #2072, opened 2019-07-16
Add freeipa healthcheck log parser (#2071)

fcami:
* add data source for FreeIPA HealthCheck logs
* add parser for /var/log/ipa/healthcheck/healthcheck.log
    
Requires the fix for issue https://github.com/RedHatInsights/insights-core/issues/2056 e.g. PR https://github.com/RedHatInsights/insights-core/pull/2057

### DIFF
--- a/insights/parsers/freeipa_healthcheck_log.py
+++ b/insights/parsers/freeipa_healthcheck_log.py
@@ -1,0 +1,108 @@
+"""
+freeipa_healthcheck_log - File ``/var/log/ipa/healthcheck/healthcheck.log``
+===========================================================================
+
+This module provides plugins access to file ``/var/log/ipa/healthcheck/healthcheck.log``
+
+This file contains a list of items, each item representing a check made on the local IPA server.
+The file can be either single-line or multi-line indented.
+
+Typical content of file ``/var/log/ipa/healthcheck/healthcheck.log`` is::
+
+    [
+      {
+        "source": "ipahealthcheck.dogtag.ca",
+        "check": "DogtagCertsConfigCheck",
+        "severity": 0,
+        "uuid": "57f7d2c9-f71f-4c2f-b831-9e30250b18b2",
+        "when": "20190726213428Z",
+        "duration": "0.150333",
+        "kw": {
+          "key": "subsystemCert cert-pki-ca",
+          "configfile": "/var/lib/pki/pki-tomcat/conf/ca/CS.cfg"
+        }
+      },
+      {
+        "source": "ipahealthcheck.ipa.certs",
+        "check": "IPACertmongerExpirationCheck",
+        "severity": 0,
+        "uuid": "afd047c2-d78b-4c5c-b3a4-f68578ee1595",
+        "when": "20190726213429Z",
+        "duration": "0.007105",
+        "kw": {
+          "key": "20190620165230"
+        }
+      },
+      {
+        "source": "ipahealthcheck.ds.replication",
+        "check": "ReplicationConflictCheck",
+        "severity": 0,
+        "uuid": "99c0a513-447c-46f1-95ea-058fc0db6075",
+        "when": "20190726213429Z",
+        "duration": "0.001409",
+        "kw": {}
+      },
+      {
+        "source": "ipahealthcheck.ipa.files",
+        "check": "IPAFileNSSDBCheck",
+        "severity": 0,
+        "uuid": "7e8a7739-c4d6-4b97-b602-9f9e981f793c",
+        "when": "20190726213432Z",
+        "duration": "0.000765",
+        "kw": {
+          "key": "_etc_pki_pki-tomcat_alias_cert9.db_owner",
+          "type": "owner",
+          "path": "/etc/pki/pki-tomcat/alias/cert9.db"
+        }
+      },
+      {
+        "source": "ipahealthcheck.ipa.topology",
+        "check": "IPATopologyDomainCheck",
+        "severity": 0,
+        "uuid": "c5ee8ee1-98d1-4696-bbf1-8243677b8918",
+        "when": "20190726213432Z",
+        "duration": "0.019070",
+        "kw": {
+          "suffix": "ca"
+        }
+      },
+      {
+        "source": "ipahealthcheck.system.filesystemspace",
+        "check": "FileSystemSpaceCheck",
+        "severity": 2,
+        "uuid": "4d1c71c5-cc37-4ebf-b53c-34f4e4534437",
+        "when": "20190726213432Z",
+        "duration": null,
+        "kw": {
+          "msg": "/var/lib/dirsrv/: free space percentage under threshold: 1% < 20%",
+          "store": "/var/lib/dirsrv/",
+          "percent_free": 1,
+          "threshold": 20
+        }
+      }
+    ]
+
+The list of errors can be accessed via the errors property.
+
+Examples:
+
+    >>> len(healthcheck.errors)
+    1
+    >>> healthcheck.errors[0]['check'] == 'FileSystemSpaceCheck'
+    True
+
+"""
+
+from insights import parser
+from insights import JSONParser
+from insights.specs import Specs
+
+
+@parser(Specs.freeipa_healthcheck_log)
+class FreeIPAHealthCheckLog(JSONParser):
+    """Parses the content of file ``/var/log/ipa/healthcheck/healthcheck.log``."""
+
+    @property
+    def errors(self):
+        """ list: errors in healthcheck log."""
+        return [entry for entry in self.data if entry['severity'] > 0]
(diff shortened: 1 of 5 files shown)


fcami:
I will rebase on top of https://github.com/RedHatInsights/insights-core/pull/2057 when it is merged as otherwise this is not functional at all.

insights/parsers/freeipa_healthcheck_log.py
```diff
@@ -0,0 +1,23 @@
...
+from insights.specs import Specs
+from insights import JSONParser
+
+
+@parser(Specs.freeipa_healthcheck_log)
+class FreeIPAHealthCheckLog(JSONParser):
+    """Class to parse /var/log/ipa/healthcheck/healthcheck.log"""
+
+    @property
+    def get_errors(self):
```
bfahr: @fcami normally a property (or attribute) would be named as a noun like `errors` instead of a verb.

insights/parsers/freeipa_healthcheck_log.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+
+from .. import parser
+from insights.specs import Specs
+from insights import JSONParser
+
+
+@parser(Specs.freeipa_healthcheck_log)
+class FreeIPAHealthCheckLog(JSONParser):
+    """Class to parse /var/log/ipa/healthcheck/healthcheck.log"""
```
bfahr: @fcami please follow the [documentation guidelines for the project](https://insights-core.readthedocs.io/en/latest/docs_guidelines.html) to ensure that your parser documentation is accurate and renders correctly.

Please include testing for your examples you add to the documentation.

insights/parsers/freeipa_healthcheck_log.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @property
+    def get_errors(self):
+        """ dict: Returns a list of errors in healthcheck logs."""
+        errors = []
+        for entry in self.data:
+            if entry['severity'] > 0:
+                errors.append(entry)
+        return errors
```
bfahr: Your docstring indicates the return type is a `dict` but it is actually a `list`.

Also please consider simplifying this as a list comprehension like:
```
return [entry for entry in self.data if entry['severity'] > 0]
```

insights/parsers/tests/test_freeipa_healthcheck_log.py
```diff
@@ -0,0 +1,18 @@
+from insights.parsers.freeipa_healthcheck_log import FreeIPAHealthCheckLog
+from insights.tests import context_wrap
+
+FREEIPA_HEALTHCHECK_LOG_OK = """[{"source": "ipahealthcheck.meta.services", "check": "certmonger", "severity": 0, "uuid": "693c526c-2261-44fe-98e0-94c6b76bf79b", "when": "20190711141703Z", "duration": "0.012848", "kw": {"status": true}}]"""
```
bfahr: When you use long strings, an alternative way to define them is something like the following:
```
LONG_STR = """
a very long string
 with lots of words
 that can be hard to read
""".strip()

long_str = "".join(LONG_STR.splitlines())
```
Just make sure you have space at the end or beginning of the line so that separate words are not mashed together.

insights/specs/__init__.py
```diff
@@ -151,6 +151,7 @@ class Specs(SpecSet):
     foreman_ssl_access_ssl_log = RegistryPoint(filterable=True)
     foreman_rake_db_migrate_status = RegistryPoint()
     foreman_tasks_config = RegistryPoint(filterable=True)
+    freeipa_healthcheck_log = RegistryPoint(filterable=True)
```
xiangce: From the definition of the parser `FreeIPAHealthCheckLog `, the base parser `JSONParser` requires the input file being in good JSON format. 

However, here, the spec is marked as `filterable=True`, which means the collected `/var/log/ipa/healthcheck/healthcheck.log` file must be filtered by some specified string (a whitelist). And its JSON format must be broken and results in Parser failure at last.

That is, once decided to use `JSONParser`, the content should NOT be filtered (the `filterable=Ture` should be removed). 
But if it must be filtered (e.g. to avoid collecting huge size file, or sensitive info), better change to use `LogFileOutput` or others as the base class.

fcami: Thanks for the explanation, I've removed filterable=True.

insights/parsers/freeipa_healthcheck_log.py
fcami: Will fix, thanks.

insights/parsers/freeipa_healthcheck_log.py
fcami: Updated, thank you.

insights/parsers/tests/test_freeipa_healthcheck_log.py
fcami: Updated, thank you.

bfahr:
@fcami #2057 has been merged.

insights/parsers/freeipa_healthcheck_log.py
fcami: @bfahr I've updated the PR, however as you know healthcheck outputs a very long list of long dictionaries that will make either reading or testing the actual docstring near to impossible.
I've included a complete example in the parser though and I'm testing the string in the test.
Let me know if you think any changes should be made.

fcami:
> @fcami #2057 has been merged.

Thanks for the heads-up @bfahr !

insights/parsers/freeipa_healthcheck_log.py
bfahr: @fcami please refer to the documentation guidelines above about including a `docs/shared_parsers/freeipa_healthcheck_log.rst` file so that your parser will be included in the documentation build.  Please make sure to check that it builds and looks correct.

bfahr:
@fcami you are still missing the `.rst` file to include your code in the rendered documentation.

insights/parsers/freeipa_healthcheck_log.py
fcami: hi @bfahr thanks for the heads up, I've added the file and it renders ok.

fcami:
hi @bfahr I've included the rst and it renders fine.

insights/parsers/freeipa_healthcheck_log.py
```diff
@@ -0,0 +1,69 @@
...
+    ... "severity": 2, "uuid": "ceadd564-f42f-4640-a3ae-b67f1aa29dc4", "when": "20190711141807Z",
+    ... "duration": null, "kw": {"msg": "/var/log/: free space under threshold: 400 MiB < 1024 MiB",
+    ... "store": "/var/log/", "free_space": 400, "threshold": 1024}}]
+    ... '''.strip()
+    >>> FREEIPA_HEALTHCHECK_LOG_FAILURES = "".join(LONG_FREEIPA_HEALTHCHECK_LOG_FAILURES.splitlines())
+    >>>
+    >>> log_obj = FreeIPAHealthCheckLog(context_wrap(FREEIPA_HEALTHCHECK_LOG_FAILURES))
+    >>> for error in log_obj.errors:
+    ...     assert error['check'] == 'FileSystemSpaceCheck'
+    ...     assert error['source'] == 'ipahealthcheck.system.filesystemspace'
```
bfahr: These examples should use the `healthcheck` object that you create in your tests on line 66 instead of creating a new object here.  `healthcheck` will be set to whatever test value you use on line 66, and can just be used directly in the examples here.  Generally it is good to use test data that is the same as the "typical contents of the file" shown above the Examples section so that a user can follow the examples.

insights/parsers/freeipa_healthcheck_log.py
```diff
@@ -0,0 +1,69 @@
...
+        "when": "20190719155725Z",
+        "duration": "0.013207",
+        "kw": {
+          "status": true
+        }
+      }
+    ]
+
+The list of errors can be accessed via the errors property.
+Examples:
```
bfahr: You need to add a blank line above `Examples:` so that it will render properly.

fcami: Solved, thanks

insights/parsers/freeipa_healthcheck_log.py
fcami: hi @bfahr thanks for the input, I believe I've solved that as well in the latest push.

insights/parsers/freeipa_healthcheck_log.py
```diff
@@ -0,0 +1,108 @@
...
+    ]
+
+The list of errors can be accessed via the errors property.
+
+Examples:
+
+    >>> len(healthcheck.errors)
+    1
+    >>> healthcheck.errors[0]['check']
+    'FileSystemSpaceCheck'
```
bfahr: @fcami you need to change this example to something like this to avoid the differences in strings and unicode across python versions:
```
>>> healthcheck.errors[0]['check'] == 'FileSystemSpaceCheck'
True
```

fcami: hi @bfahr this is done, thanks.

PR #2075, opened 2019-07-17
check RHEL 6 cgroups location. move to post-update

gravitypriest:
RHEL6 cgroups directory is in a different location


gravitypriest:
this was a terrible idea

PR #2094, opened 2019-07-22
Parser cdc wdm

vishwanathjadhav:
Added parser for `/sys/bus/usb/drivers/cdc_wdm/module/refcnt` device usage.

### DIFF
--- a/insights/parsers/sys_bus.py
+++ b/insights/parsers/sys_bus.py
@@ -14,7 +14,7 @@ CdcWDM - file ``/sys/bus/usb/drivers/cdc_wdm/module/refcnt``
 
 
 from insights import parser, Parser
-from insights.parsers import SkipException, ParseException
+from insights.parsers import ParseException
 from insights.specs import Specs
 
 
@@ -46,13 +46,9 @@ class CdcWDM(Parser):
     """
 
     def parse_content(self, content):
-        self._ref_count = ""
-        if not content:
-            raise SkipException("No Content!")
-
         try:
             self._ref_count = int(content[0].strip())
-        except:
+        except Exception:
             raise ParseException("Invalid Content!")
 
     @property
--- a/insights/parsers/sys_bus.py
+++ b/insights/parsers/sys_bus.py
@@ -1,0 +1,73 @@
+"""
+``/sys/bus`` Device Usage Information
+=====================================
+
+A parser to parse the usage information of devices connected
+on sys/bus.
+
+Parsers included in this module are:
+
+CdcWDM - file ``/sys/bus/usb/drivers/cdc_wdm/module/refcnt``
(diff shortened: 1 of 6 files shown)


insights/specs/insights_archive.py
```diff
@@ -15,6 +15,7 @@ class InsightsArchiveSpecs(Specs):
     bios_uuid = simple_file("insights_commands/dmidecode_-s_system-uuid")
     blkid = simple_file("insights_commands/blkid_-c_.dev.null")
     brctl_show = simple_file("insights_commands/brctl_show")
+    cdc_wdm = simple_file("/sys/bus/usb/drivers/cdc_wdm/module/refcnt")
```
bfahr: @vishwanathjadhav this is not necessary for a `simple_file`.  It will use the setting from default.py for archives.

insights/parsers/sys_bus.py
```diff
@@ -0,0 +1,70 @@
...
+        Returns (int): device usage count.
+        """
+        return self._ref_count
+
+    @property
+    def device_in_use(self):
+        """
+        Returns (bool): ``True`` when device in use else ``False``.
+        """
+        return True if self._ref_count else None
```
bfahr: Since this returns a bool it should either return True or False, so shouldn't it return something like `return self._ref_count == 1`?

insights/parsers/sys_bus.py
```diff
@@ -0,0 +1,70 @@
...
+
+@parser(Specs.cdc_wdm)
+class CdcWDM(Parser):
+
+    """
+    This file `/sys/bus/usb/drivers/cdc_wdm/module/refcnt` contains
+    device usage count, i.e if a device is in use then the non-zero
+    value will be present in the file.
+
+    Sample Content:
```
bfahr: Check the rendered docs, I believe you need a `::` here to keep the following 2 lines separate.

vishwanathjadhav: Done

insights/parsers/sys_bus.py
vishwanathjadhav: Done

insights/specs/insights_archive.py
vishwanathjadhav: Done

insights/parsers/sys_bus.py
```diff
@@ -0,0 +1,70 @@
...
+        >>> device_usage.device_in_use
+        True
+
+    Raises:
+        SkipException: When contents are empty
+        ParseException: When contents are invalid
+    """
+
+    def parse_content(self, content):
+        self._ref_count = ""
```
xiangce: Better remove this line

PR #2099, opened 2019-07-22
remove pyopenssl req, add new require

gravitypriest:
Removing unnecessary PyOpenSSL and adding an additional require tag for QE RHEL 6 testing


bfahr:
@csams I tested everything I was able to test and it seems to work correctly.

Issue #2081, opened 2019-07-18
MountEntry in mount.py store 2 copy of data in both obj.attribute and self.data ways, which may lead to data inconsistency once abuse write operations are applied to it. 


JoySnow:
Even though such write operations won't be used in our normal usage of parsers, this is still a shortcoming in some way.

Here is a abuse writing example:

At file `test_mount.py`'s Line-60, I add following test code for example:
```
    import pprint
    print("\n === Take this as an example: ")
    print("sr0, type(sr0): ", sr0, type(sr0))
    print("sr0.__dict__ : ")
    pprint.pprint(sr0.__dict__)
    print("sr0.data['mount_point'], sr0.mount_point: ", sr0.data['mount_point'], sr0.mount_point)
    print(" === Do the change now: ")
    sr0.mount_point = "attribute mp"               # <-- change attribute here
    print("sr0.data['mount_point'], sr0.mount_point: ", sr0.data['mount_point'], sr0.mount_point)
    sr0.data['mount_point'] = "data dict mp"   # <-- change in data here
    print("sr0.data['mount_point'], sr0.mount_point: ", sr0.data['mount_point'], sr0.mount_point)
```
And here is the result in `py.test`:
```
insights/parsers/tests/test_mount.py::test_mount 
 === Take this as an example: 
sr0, type(sr0):  <insights.parsers.mount.MountEntry object at 0x7f9b6c2dda20> <class 'insights.parsers.mount.MountEntry'>
sr0.__dict__ : 
{'data': {'filesystem': '/dev/sr0',
          'mount_clause': '/dev/sr0 on /run/media/root/VMware Tools type '
                          'iso9660 '
                          '(ro,nosuid,nodev,relatime,uid=0,gid=0,iocharset=utf8,mode=0400,dmode=0500,uhelper=udisks2) '
                          '[VMware Tools]',
          'mount_label': 'VMware Tools',
          'mount_options': <insights.parsers.mount.MountOpts object at 0x7f9b6c3124e0>,
          'mount_point': '/run/media/root/VMware Tools',
          'mount_type': 'iso9660'},
 'filesystem': '/dev/sr0',
 'mount_clause': '/dev/sr0 on /run/media/root/VMware Tools type iso9660 '
                 '(ro,nosuid,nodev,relatime,uid=0,gid=0,iocharset=utf8,mode=0400,dmode=0500,uhelper=udisks2) '
                 '[VMware Tools]',
[...]

PR #2104, opened 2019-07-23
Update entry classes in mount and fstab

xiangce:
- fix: #2081 
- revert the AttributrAsDict as the base class of MountEntry, MountOpts, and FSTabEntry
- remove the fixed attributes from MountOpts, change to all dynamic attributes
- change dynamic attributes of MountEntry and FSTabEntry to fixed attributes
- update docstring


xiangce:
Some existing rules need to change as per this PR, due to the update of MountEntry

xiangce:
Update: no rules are broken as per the latest commit.

PR #2110, opened 2019-07-25
New parsers and combiner for "dnf module" command

xiangce:


insights/specs/default.py
```diff
@@ -228,7 +228,7 @@ def dnf_module_names(broker):
             return [m for m in dml]
         raise SkipComponent()
 
-    dnf_module_info = foreach_execute(dnf_module_names, "/usr/bin/dnf -C --noplugins module info %s", deps=IsRhel8)
+    dnf_module_info = foreach_execute(dnf_module_names, "/usr/bin/dnf -C --noplugins module info %s", deps=[IsRhel8])
```
Rezney: @xiangce thanks for working on this parser. Do I understand it correctly (as the function name suggests) that for every module name we execute "dnf module info $module"?

I just tried to run "dnf module info module1 module2 module3" with list of modules and it works just fine and should be faster. What do you think?

xiangce: @Rezney - Yes, it works. Thanks for the suggestion!

insights/specs/__init__.py
```diff
@@ -114,7 +114,7 @@ class Specs(SpecSet):
     dmsetup_info = RegistryPoint()
     dnf_modules = RegistryPoint()
     dnf_module_list = RegistryPoint()
-    dnf_module_info = RegistryPoint(multi_output=True)
+    dnf_module_info = RegistryPoint()
```
xiangce: @csams - Please have a look to this one.

I'm sure it is **NG** to remove the `multi_output=True` from a [`foreach_execute` spec](https://github.com/RedHatInsights/insights-core/pull/2110/commits/ee8ac0049746d480c7fdde36ddf11f6c7cd6b47c#diff-40c874757ab5b5f019039c435a8334daR231) as below:

```python
dnf_module_info = foreach_execute(dnf_module_names, "/usr/bin/dnf -C --noplugins module info %s", deps=[IsRhel8])
```
The spec `dnf_module_info` depends on another datasource `dnf_module_names`.

According to the above [suggestion](https://github.com/RedHatInsights/insights-core/pull/2110#discussion_r307358091) from @Rezney, in order to get a better execution efficiency, I just want the provider datasource `dnf_module_names` to return a single string which contains all the module names, instead of a list of separated module names. However, the `foreach_execute` will lead the parser `DnfModuleInfo` being wrapped into a list, which brings a bit complicated when using the parser. And we have to write a combiner to make its usage simple.

Can we do the following enhancement to the `simple_command` to let it be able to receive a string (but not a list of strings) generated by another datasource? and then the parser result wouldn't be wrapped anymore.

```python
@datasource(DnfModuleList)
def dnf_module_names(broker):
    dml = broker[DnfModuleList]
    if dml:
        return (' ').join([m for m in dml])
    raise SkipComponent()

dnf_module_info = simple_command("/usr/bin/dnf -C --noplugins module info %s", dnf_module_names, deps=[IsRhel8])
```
I have implemented a version of code for this idea, please have a look as well.
- https://github.com/RedHatInsights/insights-core/pull/2112

xiangce:
Hey @csams , @bfahr - How do we define the `simple_file` for the `command_with_args` helper in the `insights_archive.py`? Since we do not know what the exact `args` are.

Can it be like?:
```
dnf_module_info = simple_file("insights_commands/dnf_-C_--noplugins_module_info_*")
```

xiangce:
Test on RHEL 8:

~~~
(.python3) [root@vm37-146 insights-core]# insights cat insights.specs.default.DefaultSpecs.dnf_module_info
CommandOutputProvider("'/usr/bin/dnf -C --noplugins module info 389-ds ant container-tools freeradius gimp go-toolset httpd idm inkscape javapackages-runtime libselinu
x-python llvm-toolset mailman mariadb maven mercurial mod_auth_openidc mysql nginx nodejs parfait perl perl-App-cpanminus perl-DBD-MySQL perl-DBD-Pg perl-DBD-SQLite pe
rl-DBI perl-FCGI perl-YAML php pki-core pki-deps postgresql python27 python36 redis rhn-tools ruby rust-toolset satellite-5-client scala squid subversion swig varnish
virt'")
Name             : ant
Stream           : 1.10 [d][a]
Version          : 820181213135032
Context          : 5ea3b708
Profiles         : common [d]
Default profiles : common
Repo             : rhel-8-for-x86_64-appstream-rpms
Summary          : Java build tool
Description      : Apache Ant is a Java library and command-line tool whose mission is to drive processes described in build files as targets and extension points depe
ndent upon each other. The main known usage of Ant is the build of Java applications. Ant supplies a number of built-in tasks allowing to compile, assemble, test and r
un Java applications. Ant can also be used effectively to build non Java applications, for instance C or C++ applications. More generally, Ant can be used to pilot any
 type of process which can be described in terms of targets and tasks.
Artifacts        : ant-0:1.10.5-1.module+el8+2438+c99a8a1e.noarch
                 : ant-lib-0:1.10.5-1.module+el8+2438+c99a8a1e.noarch

Name             : container-tools
Stream           : 1.0

...

            : xmlstreambuffer-0:1.5.4-8.module+el8+2468+c564cec5.noarch
            : xmlstreambuffer-javadoc-0:1.5.4-8.module+el8+2468+c564cec5.noarch
            : xsom-0:0-19.20110809svn.module+el8+2468+c564cec5.noarch
            : xsom-javadoc-0:0-19.20110809svn.module+el8+2468+c564cec5.noarch

Hint: [d]efault, [e]nabled, [x]disabled, [i]nstalled, [a]ctive]
~~~

insights/specs/default.py
xiangce: Hey @Rezney, the dependency PR #2112 was merged. Then I have updated the `dnf module info` parser as per your suggestion. Please have a look at the update.

Rezney: @xiangce looks good to me from "dnf" perspective...thanks

PR #2112, opened 2019-07-26
Enhance simple_command to get string parameter from datasource

xiangce:
- See: https://github.com/RedHatInsights/insights-core/pull/2110#discussion_r307619511

- fix: #2113


csams:
Hey @xiangce , I like it. Can we break it out into a separate helper like `command_with_args` instead of extending `simple_command`?  And allow a tuple substitution in addition to a single string?

xiangce:
Hey @csams - a new separate helper is a good idea, I will modify it as per your suggestion in the next commit together with tuple substitution options. Thanks a lot.

insights/core/spec_factory.py
```diff
@@ -736,6 +736,61 @@ def __call__(self, broker):
...
+            CalledProcessError is raised. If None, timeout is infinite.
+        inherit_env (list): The list of environment variables to inherit from the
+            calling process when the command is invoked.
+
+    Returns:
+        function: A datasource that returns the output of a command that takes
+            specified arguments passed by the provider.
+    """
+
+    def __init__(self, cmd, provider, context=HostContext, deps=[], split=True, keep_rc=False, timeout=None, inherit_env=[], **kwargs):
```
bfahr: @xiangce I had a discussion about this with @csams the other day.  For args where the default is a mutable object like `[]` or `{}` there will only be one instance of it for all objects so its better to use the following pattern:
```
def __init__(self, some_arg=None):
    some_arg = some_arg if some_arg is not None else []
```

See [this article](https://blog.florimond.dev/python-mutable-defaults-are-the-source-of-all-evil) for additional info.

xiangce: Ack, however, do we need to modify them in other helpers?

xiangce:
Hey @csams , @bfahr - How do we define the simple_file for the command_with_args helper in the insights_archive.py? Since we do not know what the exact args are.

Can it be like?:
```
dnf_module_info = simple_file("insights_commands/dnf_-C_--noplugins_module_info_*")
```
Does the wildcard `*` work to the `simple_file`?

bfahr:
@xiangce you would need to use a `glob_file` instead of `simple_file`.

xiangce:
@bfahr - OK. So, it won't cause any conflicts when combining a spec without `multi_output=True` with the `glob_file`, right? Thanks.

bfahr:
@xiangce good question. It should only return one file, but the glob spec could potentially return more than one file, which would break end users that didn't expect a `multi_output` spec.  @csams can you think of a better solution than `glob` here?

insights/core/spec_factory.py
```diff
@@ -736,6 +736,62 @@ def __call__(self, broker):
...
+
+    def __call__(self, broker):
+        source = broker[self.provider]
+        ctx = broker[self.context]
+        if isinstance(source, ContentProvider):
+            source = source.content
+        if not isinstance(source, (list, set, tuple)):
+            source = [source]
+        try:
+            self.cmd = self.cmd % ' '.join(source)
```
csams: Instead of `' '.join(source)`, let's do something like `self.cmd % tuple(source)`. That way substitution parameters can appear anywhere in `self.cmd`.

We should also only accept `list` and `tuple` instances since `set` would make the string substitution ambiguous.

xiangce: @csams - Please correct me if I misunderstood you. 
By replacing the above  `' '.join(source)`  with `tuple(source)`, it requires the source datasource to pass a list(or tuple) with the specified number of strings exactly match with the request of the `self.cmd`. And that requires each item in the passed list(or tuple) should be a single string (but not a list). Is it right? e.g.

```python
@datasource(...)
def ds_1_src(lst):
     return ','.join(lst)
@datasource(...)
def ds_2_src(lst):
     return [','.join(lst), "test"]
@datasource(...)
def ds_3_src(lst, lst2):
     return [','.join(lst), "test", ' '.join(lst2)]

ds_1_args = command_with_args("/bin/xxx -a %s", ds_1_src)
ds_2_args = command_with_args("/bin/xxx -a %s -b %s", ds_2_src)
ds_3_args = command_with_args("/bin/xxx -a %s -b %s -c %s", ds_3_src)
```

xiangce:
@csams - Updated, please have a look again.

PR #2120, opened 2019-07-30
Propogate SystemExit so insights run -h doesn't print '0'.

csams:
The top level try/except around the InsightsCli caught BaseException.
That included SystemExit, which is raised by argparse parsers at the
end of handling the -h option. This PR allows SystemExit to propogate.


patchkez:
test this please

PR #2121, opened 2019-07-31
Add combiner for rsyslog_conf

wushiqinlou:
Before RHEL8, the format of rsyslog.conf is like:
~~~
$IncludeConfig /etc/rsyslog.d/*.conf
$ModLoad imtcp
~~~

However, it changes like following in RHEL8:
~~~
include(file="/etc/rsyslog.d/*.conf" mode="optional")
module(load="imuxsock" 
       SysSock.Use="off")
~~~

We also need to consider the `RainerScript` when parsing rsyslog.conf, then it becomes too complicated if we want to parse this file in detail. Per our discussion in https://github.com/RedHatInsights/insights-core/pull/1952, we decide to return the content directly with format line by line. As `LogFileOutput` can not handle multi lines, I do not use this class, and parse the content with my own code. Also, if we use filter for multi lines, we will get only one line and it may not be useful for plugin check, so I remove the filter.

### DIFF
--- a/insights/combiners/rsyslog_confs.py
+++ b/insights/combiners/rsyslog_confs.py
@@ -14,7 +14,7 @@ from .. import LegacyItemAccess
 
 
 @combiner(RsyslogConf)
-class RsyslogAllConf(LegacyItemAccess):
+class RsyslogAllConf(dict):
     """
     Combiner for accessing all the krb5 configuration files.
 
@@ -40,6 +40,8 @@ class RsyslogAllConf(LegacyItemAccess):
                     self.data[conf.file_path] = conf.data
                     break
             self.data[conf.file_path] = conf.data
+
+        self.update(self.data)
 
         super(RsyslogAllConf, self).__init__()
 
--- a/insights/combiners/rsyslog_confs.py
+++ b/insights/combiners/rsyslog_confs.py
@@ -10,7 +10,6 @@ files, user needs to check this situation in plugin if it is necessary.
 """
 from insights.core.plugins import combiner
 from insights.parsers.rsyslog_conf import RsyslogConf
-from .. import LegacyItemAccess
 
 
 @combiner(RsyslogConf)
@@ -27,7 +26,7 @@ class RsyslogAllConf(dict):
         '$ModLoad imuxsock'
     """
     def __init__(self, confs):
-        self.data = {}
+        data = {}
 
         for conf in confs:
             if conf.file_path == "/etc/rsyslog.conf":
@@ -36,14 +35,11 @@ class RsyslogAllConf(dict):
                     if "include(" in item or "$IncludeConfig" in item:
                         include = True
                 if not include:
-                    self.data.clear()
-                    self.data[conf.file_path] = conf.data
+                    data.clear()
+                    data[conf.file_path] = conf.data
                     break
-            self.data[conf.file_path] = conf.data
+            data[conf.file_path] = conf.data
 
-        self.update(self.data)
+        self.update(data)
 
         super(RsyslogAllConf, self).__init__()
-
-    def __len__(self):
-        return len(self.data)
--- a/insights/combiners/rsyslog_confs.py
+++ b/insights/combiners/rsyslog_confs.py
@@ -15,7 +15,7 @@ from insights.parsers.rsyslog_conf import RsyslogConf
 @combiner(RsyslogConf)
 class RsyslogAllConf(dict):
     """
-    Combiner for accessing all the krb5 configuration files.
+    Combiner for accessing all the rsyslog configuration files.
 
     Examples:
         >>> type(confs)
@@ -29,13 +29,12 @@ class RsyslogAllConf(dict):
         super(RsyslogAllConf, self).__init__()
         data = {}
 
+        # Combine rsyslog configuration files into one dict. Key is file name, value is content of configuration file.
         for conf in confs:
             if conf.file_path == "/etc/rsyslog.conf":
-                include = False
-                for item in conf:
-                    if "include(" in item or "$IncludeConfig" in item:
-                        include = True
-                if not include:
+                # Check if there is include option, if not, only parse /etc/rsyslog.conf even
+                # /etc/rsyslog.d/*.conf exist.
+                if not any(["include(" in item or "$IncludeConfig" in item for item in conf]):
                     data.clear()
                     data[conf.file_path] = conf
                     break
--- a/insights/combiners/rsyslog_confs.py
+++ b/insights/combiners/rsyslog_confs.py
@@ -1,0 +1,47 @@
+"""
+RsyslogConfAll - files ``/etc/rsyslog.conf`` and ``/etc/rsyslog.d/*.conf``
+==========================================================================
+
+Combiner for accessing all the rsyslog comfiguration files. There may be
+multiple rsyslog configuration, and the main configuration file is
+``/etc/rsyslog.conf``. This combiner will not check same option in multi
+files, user needs to check this situation in plugin if it is necessary.
+
+"""
+from insights.core.plugins import combiner
+from insights.parsers.rsyslog_conf import RsyslogConf
+from .. import LegacyItemAccess
+
+
+@combiner(RsyslogConf)
+class RsyslogAllConf(LegacyItemAccess):
+    """
+    Combiner for accessing all the krb5 configuration files.
+
+    Examples:
+        >>> type(confs)
+        <class 'insights.combiners.rsyslog_confs.RsyslogAllConf'>
+        >>> len(confs)
+        2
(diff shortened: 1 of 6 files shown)


insights/combiners/rsyslog_confs.py
```diff
@@ -0,0 +1,47 @@
...
+files, user needs to check this situation in plugin if it is necessary.
+
+"""
+from insights.core.plugins import combiner
+from insights.parsers.rsyslog_conf import RsyslogConf
+from .. import LegacyItemAccess
+
+
+@combiner(RsyslogConf)
+class RsyslogAllConf(LegacyItemAccess):
```
bfahr: @wushiqinlou since you are essentially creating a dictionary in `self.data` you should just base this class on `dict` which would provide all of the functions in `LegacyItemAccess`.

insights/specs/__init__.py
```diff
@@ -471,7 +471,7 @@ class Specs(SpecSet):
     root_crontab = RegistryPoint()
     route = RegistryPoint()
     rpm_V_packages = RegistryPoint()
-    rsyslog_conf = RegistryPoint(filterable=True)
+    rsyslog_conf = RegistryPoint()
```
bfahr: Since this spec was already filtered, this change will have to go through the approval process.

insights/parsers/rsyslog_conf.py
```diff
@@ -10,70 +10,78 @@
...
         self.config_items = {}
-        # Config items are e.g. "$Word value #optional comment"
-        config_re = re.compile(r'^\s*\$(?P<name>\S+)\s+(?P<value>.*?)(?:\s+#.*)?$')
-        for line in self.data:
-            lstrip = line.strip()
-            match = config_re.match(lstrip)
-            if match:
-                self.config_items[match.group('name')] = match.group('value')
 
-    def config_val(self, item, default=None):
```
bfahr: This would break backward compatibility for this parser so you need to keep it in the parser.  It seems like it might be better to have two parsers here since that are different, and use the `IsRhel[6|7|8]` components to trigger which parser is executed.

wushiqinlou: self.config_items is used to store configuration like "$ModLoad imtcp", however, former code can not handle multi lines. Per our discussion in https://github.com/RedHatInsights/insights-core/pull/1952, I combine the multi lines into one line and not parse it as key-value format, directly return content line by line. Also, this code can handle rsyslog.conf in RHEL6|7|8, no need to use IsRhel[6|7|8] components. I cretaed a issue to remove this function, https://github.com/RedHatInsights/insights-core/issues/2130

insights/combiners/rsyslog_confs.py
wushiqinlou: Updated, please have a check

insights/parsers/rsyslog_conf.py
```diff
@@ -10,70 +10,76 @@
...
-from .. import Parser, parser, get_active_lines
-
-import re
+from .. import Parser, parser, get_active_lines, LegacyItemAccess
 from insights.specs import Specs
 
 
 @parser(Specs.rsyslog_conf)
-class RsyslogConf(Parser):
+class RsyslogConf(Parser, LegacyItemAccess):
```
xiangce: The `self.data` of this Parser is a list, `LegacyItemAccess` is not suitable to a list but a dict.

Would you mind change to mix it with a `list` (and remove the self.data)? 
```python
 class RsyslogConf(Parser, list):
```

insights/parsers/rsyslog_conf.py
```diff
@@ -10,70 +10,76 @@
...
+        parenthesis_flag = False
+        parenthesis_string = ""
+        brace_string = ""
 
-        self.config_items = {}
-        # Config items are e.g. "$Word value #optional comment"
-        config_re = re.compile(r'^\s*\$(?P<name>\S+)\s+(?P<value>.*?)(?:\s+#.*)?$')
-        for line in self.data:
+        for line in get_active_lines(content):
             lstrip = line.strip()
```
xiangce: Since `lstrip` is a reserved keyword of python. Would you mind changing it to sth like `l_strip`?

insights/parsers/rsyslog_conf.py
```diff
@@ -10,70 +10,76 @@
...
+                        if ")" in lstrip:
+                            self.data.append(lstrip)
+                        else:
+                            parenthesis_flag = True
+                            parenthesis_string = lstrip
+                        continue
+                    else:
+                        self.data.append(lstrip)
 
     def __len__(self):
```
xiangce: After mixing it with `list` this can  be removed.

insights/combiners/rsyslog_confs.py
```diff
@@ -0,0 +1,45 @@
...
+                        include = True
+                if not include:
+                    data.clear()
+                    data[conf.file_path] = conf.data
+                    break
+            data[conf.file_path] = conf.data
+
+        self.update(data)
+
+        super(RsyslogAllConf, self).__init__()
```
xiangce: Better call the `super.__init__()` at the very begging of this `__init__`

wushiqinlou:
Spec JIRA link: https://projects.engineering.redhat.com/browse/INSPEC-262

insights/combiners/rsyslog_confs.py
```diff
@@ -0,0 +1,44 @@
...
+        data = {}
+
+        for conf in confs:
+            if conf.file_path == "/etc/rsyslog.conf":
+                include = False
+                for item in conf:
+                    if "include(" in item or "$IncludeConfig" in item:
+                        include = True
+                if not include:
```
bfahr: Could you simplify the above logic by using a comprehension something like this?
```python
if not any(["include(" in item or "$IncludeConfig" in item for item in conf]):
```

insights/specs/__init__.py
```diff
@@ -471,7 +471,7 @@ class Specs(SpecSet):
     root_crontab = RegistryPoint()
     route = RegistryPoint()
     rpm_V_packages = RegistryPoint()
-    rsyslog_conf = RegistryPoint(filterable=True)
```
bfahr: @wushiqinlou you need to keep `filterable` here.  You can add `{}()` to a filter in the parser to ensure you get your opening and closing chars.

wushiqinlou: @bfahr - I think we can return all the content of configuration file, and no need to add filter. If we keep filterable, and add filter for {}() in parser, we may miss necessary content like line 90, 91 in test_rsyslog_confs.py.

insights/combiners/rsyslog_confs.py
```diff
@@ -0,0 +1,44 @@
...
+                    data.clear()
+                    data[conf.file_path] = conf
+                    break
+            data[conf.file_path] = conf
```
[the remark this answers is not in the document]
wushiqinlou: Added the documentation to to explain the loop.

insights/combiners/rsyslog_confs.py
wushiqinlou: Changed as suggestion.

wushiqinlou:
@bfahr - Could you please have a check? Any further comment on discussion of adding filterable.

insights/specs/__init__.py
wushiqinlou: @bfahr - Will add filterable as your suggestion.

insights/combiners/rsyslog_confs.py
```diff
@@ -0,0 +1,44 @@
...
+
+"""
+from insights.core.plugins import combiner
+from insights.parsers.rsyslog_conf import RsyslogConf
+
+
+@combiner(RsyslogConf)
+class RsyslogAllConf(dict):
+    """
+    Combiner for accessing all the krb5 configuration files.
```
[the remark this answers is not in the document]
wushiqinlou: Updated

insights/specs/__init__.py
wushiqinlou: @bfahr - Updated as your suggestion, please have a check.

wushiqinlou:
@insights-qe-ci-jenkins

bfahr:
@wushiqinlou please rebase your PR to clear up merge conflicts.

xiangce:
@bfahr - I helped rebase this MR, since the "Update branch" is not visible to @wushiqinlou. I think something goes incorrect to the permission set for him.

wushiqinlou:
@xiangce - I updated insights-plugins and gss-rules accordingly, please have a check.

insights-plugins: https://gitlab.cee.redhat.com/insights-rules/insights-plugins/-/merge_requests/2956
gss-rules: https://gitlab.cee.redhat.com/support-insights/gss-rules/-/merge_requests/1841

Issue #2124, opened 2019-08-01
Duplicate data in Partitions


xiangce:
The `partitions` attribute is duplicated with the items inner the `dict` parser `Partitions`.

xiangce:
Duplicated issue #2122

Issue #2122, opened 2019-08-01
Duplicate data stored in Parser/Combiners mixed with dict


xiangce:
### insights/combiners/modinfo.py
Combiner Modinfo has mixed with `dict` but still has the `data` as an attribute which is redundant.
https://github.com/RedHatInsights/insights-core/blob/fb75a55885bc20d164219277e0017196be55c5c7/insights/combiners/modinfo.py#L50-L65

The same issue is also found in:
* insights/combiner/nmcli_dev_show.py
* insights/parsers/nmcli.py
* insights/parsers/fcoeadm_i.py
* insights/parsers/partitions.py
* insights/combiners/sap.py

xiangce:
# insights/parsers/partitions.py
The partitions attribute is duplicated with the items inner the `dict` parser `Partitions`.

xiangce:
###  insights/combiners/sap.py
The `self.data` attribute is duplicated with the items inner the `dict` combiner `Sap`.

xiangce:
### insights/combiner/nmcli_dev_show.py
### insights/parsers/nmcli.py
### insights/parsers/fcoeadm_i.py

xiangce:
all related PR were closed.

PR #2125, opened 2019-08-01
Remove the self.partitions from Partitions

xiangce:
- fix: #2124 
- No rules are broken


bfahr:
Related to issue #2122

PR #2128, opened 2019-08-01
Remove the redundant self.fcoe from FcoeadmI

xiangce:
- See #2122


xiangce:
For this PR, the `self.fcoe` property was already added.

PR #2131, opened 2019-08-02
Add parser: pci_rport_target_disk_paths

xhe123:
Add parser: pci_rport_target_disk_paths
Signed-off-by: Xiaonan He <xhe@redhat.com>


xhe123:
@bfahr  Hi Rob, may I have your review on this parser? since Xiangce is on PTO in this week.

Code coverage is 100%

```
$ pytest -svk pci_rport_target_dist_paths --cov=insights/parsers  |grep pci_rport_target_disk_paths.py
insights/parsers/pci_rport_target_disk_paths.py         74      0     28      0   100%
```

insights/specs/insights_archive.py
```diff
@@ -192,6 +192,8 @@ class InsightsArchiveSpecs(Specs):
     ovs_vsctl_show = simple_file("insights_commands/ovs-vsctl_show")
     parted__l = simple_file("insights_commands/parted_-l_-s")
     passenger_status = simple_file("insights_commands/passenger-status")
+    insights/specs/insights_archive.py
```
chenlizhong: Remove line 195

xhe123: fixed, thanks LI zhong!

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -0,0 +1,219 @@
...
+                temp_pci['host'] = host
+                temp_pci['target'] = target
+                temp_pci['devnode'] = devnode
+                temp_pci['host_channel_id_lun'] = host_channel_id_lun
+
+                if temp_pci not in pdata_list:
+                    pdata_list.append(temp_pci)
+            else:
+                raise ParseException(BADWD)
+        self.update(pci)
```
bfahr: @xhe123 by adding all of the properties to this class it does not really look like a `dict` object anymore.  I think it would be better to just base the class on `CommandParser` and keep `pci` as an attribute of the class.  Since `pci` is a dictionary it would provide all of the same functionality that you are creating here.

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -0,0 +1,219 @@
...
+        pci_id (list): The list the all pci_id(s)
+        host_channel_id_lun (list): The list of all host_channel_id_lun(s)
+
+    Raises:
+        ParseException: Input content is not available to parse
+        SkipException: Input content is empty
+
+    """
+
+    def __attrs__(self, item):
```
bfahr: This could just be created as a private attribute with a descriptive name like `__device_attributes` in the `parse_content` method, instead of calling a method every time a property is accessed.

insights/parsers/pci_rport_target_disk_paths.py
xhe123: @bfahr yes, I agree with you. I leave the property of pci, removed others as you suggested.

Moreover, considering much about I can invoke these functions frequently in the coming insights rule, I keep these class functions so that I am able to get these item list directly through obj.pci_id()[0], alternative , users can also use them like this: obj.pci['PCI'][0]['pci_id'] . Does it look ok to you?

insights/parsers/pci_rport_target_disk_paths.py
xhe123: @bfahr good suggestion!
after line 205, I created few private attributes with corresponding descriptive names as you suggested, these attributes are invoked by class functions. Does it look ok to you?

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -0,0 +1,227 @@
...
+        /sys/devices/pci0000:00/0000:00:01.0/0000:04:00.6/host1/rport-1:0-1/target1:0:0/1:0:0:0/block/sdb
+        /sys/devices/pci0000:00/0000:00:01.0/0000:04:00.7/host2/rport-2:0-2/target2:0:0/2:0:0:0/block/sdc
+        /sys/devices/pci0000:00/0000:00:02.2/0000:02:00.0/host0/target0:1:0/0:1:0:0/block/sda/dev
+        /sys/devices/pci0000:00/0000:00:02.2/0000:02:00.0/host0/target0:1:0/0:1:0:0/block/sda/sda1
+        /sys/devices/pci0000:00/0000:00:02.2/0000:02:00.0/host0/target0:1:0/0:1:0:0/block/sda/sda2
+
+    The Original data parsed looks like::
+
+            {
+                'PCI': [
```
xiangce: If there is ONLY one `key` in the `dict`, why not just set it as a `list` instead of `dict`?

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -0,0 +1,227 @@
...
+                    {
+                        'target': 'target1: 0: 0',
```
xiangce: I believe the `space`s should not be there in the value.

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -0,0 +1,227 @@
...
+                        'devnode': 'sdb',
+                        'host_channel_id_lun': '1: 0: 0: 0',
```
xiangce: The same comment as above.

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -0,0 +1,227 @@
...
+        pci (dict): The original data parsed
+
+    Raises:
+        ParseException: Input content is not available to parse
+        SkipException: Input content is empty
+
+    """
+
+    @property
+    def pci(self):
```
xiangce: Is this `property` necessary? since you have already mixed it with `dict`.

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -0,0 +1,227 @@
...
+        EMPTY = "Input content is empty"
+        BADWD = "No useful data parsed in content: '{0}'".format(content)
+
+        if not content:
+            raise SkipException(EMPTY)
+
+        pdata_list = []
+        pci = {'PCI': pdata_list}
+
+        for line in content:
```
xiangce: Let's have a talk about the following logic.

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -152,10 +152,15 @@ def parse_content(self, content):
         self.__host_channel_id_lun_attributes = []
 
         for line in content:
+            if (('host' not in line) or ('pci' not in line) or
+                    ('target' not in line) or ('block' not in line)):
+                raise ParseException(BADWD.format(line))
+
             line_sp = list(filter(None, line.strip().split('/')))
             temp_pci = {}
+            rport = None
```
xiangce: This `rport` should not be initialized out of the following loop, which means the value of it in the former line could be inherited by the following lines in which the `"rport-"` is not found.

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -169,16 +174,13 @@ def parse_content(self, content):
                     temp_pci['host_channel_id_lun'] = line_sp[i - 1]
                 temp_pci['rport'] = rport
 
-                if len(temp_pci) == 5:
+                if len(temp_pci) == 6 and temp_pci not in pci:
```
xiangce: `temp_pci` is a `dict`, it's costly to compare it, the `in` operation will compare it with the items in the `pci` list one by one.

The better way to remove duplicated lines is to do the removal before re-construct it.

PS, I don't think there will be any duplicated lines, so it's enough for the Exception to only check the length of the `temp_pci`.

xhe123:
Code coverage is 100% on python2 and python3 now.

insights/specs/__init__.py
```diff
@@ -416,6 +416,7 @@ class Specs(SpecSet):
     partitions = RegistryPoint()
     passenger_status = RegistryPoint()
     password_auth = RegistryPoint()
+    pci_rport_target_disk_paths = RegistryPoint(filterable=True)
```
xiangce: With the new command, it's not necessary to filter it anymore.

insights/specs/insights_archive.py
```diff
@@ -192,6 +192,7 @@ class InsightsArchiveSpecs(Specs):
     ovs_vsctl_show = simple_file("insights_commands/ovs-vsctl_show")
     parted__l = simple_file("insights_commands/parted_-l_-s")
     passenger_status = simple_file("insights_commands/passenger-status")
+    pci_rport_target_disk_paths = simple_file("insights_commands/find_pci_rport_target_disk_paths")
```
xiangce: The `path` of the simpe_file is incorrect, please use the `mangle` command to generate it based on the "simple_command" in defult.py

insights/parsers/pci_rport_target_disk_paths.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        return sorted(self.__rport_attributes)
+
+    @property
+    def target(self):
+        """
+        The all rport(s) from parsed content.
+
+        Returns:
+            list: rports
```
xiangce: targets?

PR #2132, opened 2019-08-05
Parser ovs flow dumps

vishwanathjadhav:
Parser is added to process the ovs flows

### DIFF
--- a/insights/parsers/ovs_ofctl_dump_flows.py
+++ b/insights/parsers/ovs_ofctl_dump_flows.py
@@ -8,9 +8,8 @@ output of command ``/usr/bin/ovs-ofctl dump-flows <bridge-name>``.
 
 
 from insights import CommandParser, parser
-from insights.parsers import SkipException
+from insights.parsers import SkipException, split_kv_pairs
 from insights.specs import Specs
-from ..parsers import split_kv_pairs
 
 
 @parser(Specs.ovs_ofctl_dump_flows)
@@ -33,15 +32,7 @@ class OVSofctlDumpFlows(CommandParser):
                     { 'cookie': '0x0', 'duration': '4.617s', 'table': '0', 'n_packets': '0', 'n_bytes': '0', 'idle_timeout': '60', 'priority': '65535', 'arp,in_port': 's1-eth1', 'vlan_tci': '0x0000', 'dl_src': 'd6:fc:9c:e7:a2:f9', 'dl_dst': 'a2:72:e7:06:75:2e', 'arp_spa': '10.0.0.1', 'arp_tpa': '10.0.0.3', 'arp_op': '2' 'actions=output':'s1-eth3'}
             ]
 
-        Attributes:
-            _bridges (list): A list where each element contains the list of dictionary elements having flow dumps.
-
-        Raises:
-            SkipException: When the file is empty or data is not present for a bridge.
-
         Examples:
-            >>> len(ovs_obj._bridges)
-            2
             >>> ovs_obj.bridge_name
             'br0'
             >>> len(ovs_obj.flow_dumps)
@@ -62,9 +53,9 @@ class OVSofctlDumpFlows(CommandParser):
 
         for line in content:
             line = line.split(',')
-            flow_dict = split_kv_pairs(line, split_on='=')
-            if flow_dict:
-                self._bridges.append(flow_dict)
+            flow_list = split_kv_pairs(line, split_on='=')
+            if flow_list:
+                self._bridges.append(flow_list)
         if not self._bridges:
             raise SkipException("Invalid Content!")
 
--- a/insights/parsers/ovs_ofctl_dump_flows.py
+++ b/insights/parsers/ovs_ofctl_dump_flows.py
@@ -43,33 +43,36 @@ class OVSofctlDumpFlows(CommandParser):
             SkipException: When the file is empty or data is not present for a bridge.
 
         Examples:
-            >>> len(data["br0"])
+            >>> len(ovs_obj.data["br0"])
+            2
+            >>> ovs_obj.bridge_iface
+            'br0'
+            >>> len(ovs_obj.dump_flows('br0'))
             2
     """
 
     def parse_content(self, content):
         if not content:
-            import pdb; pdb.set_trace()
-            raise SkipException("Empty Content")
+            raise SkipException("Empty Content!")
 
         self.data = {}
 
         # Extract the bridge name
-        self._bridge_name = self.file_path.split("ovs-ofctl_dump-flows_")[1]
-        if self._bridge_name:
-            self.data[self._bridge_name] = []
-        else:
-            raise SkipException("Empty Content")
+        try:
+            self._bridge_name = self.file_path.split("ovs-ofctl_dump-flows_")[1]
+        except:
+            raise SkipException("Invalid Path!")
 
         for line in content:
             line = line.split(',')
             flow_dict = split_kv_pairs(line, split_on='=')
-            if flow_dict:
+            if flow_dict and self._bridge_name not in self.data:
+                self.data[self._bridge_name] = []
+            if flow_dict and self._bridge_name in self.data:
                 self.data[self._bridge_name].append(flow_dict)
             flow_dict = {}
         if not self.data:
             raise SkipException("Invalid Content!")
-
 
     @property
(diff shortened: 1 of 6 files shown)


insights/parsers/ovs_ofctl_dump_flows.py
```diff
@@ -0,0 +1,90 @@
...
+            'br0'
+            >>> len(ovs_obj.dump_flows('br0'))
+            2
+    """
+
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("Empty Content!")
+
+        self.data = {}
```
bfahr: @vishwanathjadhav why not use a more descriptive name for this than the generic `data`.  What about `self.bridges` instead?  It would be clearer to the user what the data actually is.

insights/parsers/ovs_ofctl_dump_flows.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Extract the bridge name
+        try:
+            self._bridge_name = self.file_path.split("ovs-ofctl_dump-flows_")[1]
```
bfahr: If there is only one `_bridge_name` then why not just have a `bridge_name` as an attribute, and an attribute `flow_dumps` or whatever as a list attribute.  It seems like you need a combiner that collects all of these things and creates the dictionary that you have here in the parser, with all of the bridges.

insights/parsers/tests/test_ovs_ofctl_dump_flows.py
```diff
@@ -0,0 +1,85 @@
...
+def test_ovs_appctl_fdb_show_bridge():
+    ovs_dump = OVSofctlDumpFlows(context_wrap(OVS_FLOW_DUMPS, path=PATH_BR0))
+    assert ovs_dump.bridge_iface == 'br0'
+    assert len(ovs_dump.dump_flows('br0')) == 8
+
+    ovs_dump = OVSofctlDumpFlows(context_wrap(OVS_FLOW_DUMPS, path=PATH_BR_INT))
+    assert ovs_dump.bridge_iface == 'br-int'
+    assert len(ovs_dump.dump_flows('br-int')) == 8
+
+    ovs_dump = OVSofctlDumpFlows(context_wrap(OVS_FLOW_DUMPS_1, path=PATH_BR_INT))
```
bfahr: Is this test correct?  It looks like the lines in the data are commented out, would that be valid output?

insights/parsers/ovs_ofctl_dump_flows.py
vishwanathjadhav: Done

insights/parsers/tests/test_ovs_ofctl_dump_flows.py
vishwanathjadhav: Done

insights/parsers/ovs_ofctl_dump_flows.py
vishwanathjadhav: Done

insights/parsers/ovs_ofctl_dump_flows.py
vishwanathjadhav: Done

insights/parsers/ovs_ofctl_dump_flows.py
```diff
@@ -0,0 +1,90 @@
...
+
+            {
+                'br0': [
+                    { 'cookie': '0x0', 'duration': '8.528s', 'table': '0', 'n_packets': '0', 'n_bytes': '0', 'idle_timeout': '60', 'priority': '65535', 'arp,in_port': 's1-eth2', 'vlan_tci': '0x0000', 'dl_src': '62:ee:31:2b:35:7c', 'dl_dst': 'a2:72:e7:06:75:2e', 'arp_spa': '10.0.0.2', 'arp_tpa': '10.0.0.3', 'arp_op': '2' 'actions=output':'s1-eth3'},
+                    { 'cookie': '0x0', 'duration': '4.617s', 'table': '0', 'n_packets': '0', 'n_bytes': '0', 'idle_timeout': '60', 'priority': '65535', 'arp,in_port': 's1-eth1', 'vlan_tci': '0x0000', 'dl_src': 'd6:fc:9c:e7:a2:f9', 'dl_dst': 'a2:72:e7:06:75:2e', 'arp_spa': '10.0.0.1', 'arp_tpa': '10.0.0.3', 'arp_op': '2' 'actions=output':'s1-eth3'}
+                ]
+            }
+
+        Attributes:
+            bridges (dict): A dictionary where each element contains the bridge-name
```
xiangce: Since there is only one "bridge"  for each Parser, it's not necessary to store the result into a dict, right?
It would be a bit simple to just store it into a list with a `property` returns the name of "bridge".

vishwanathjadhav: No, it could have more than one bridge interfaces

xiangce: for each parser, there is only 1 `file_path` which means only 1 `bridge`.

Better add a combiner to return all the bridge, if necessary

vishwanathjadhav: No need to we will access it like `ethtool_-i*` or `ifcfg-*`

insights/parsers/ovs_ofctl_dump_flows.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Sample parsed output::
+
+            [
+                    { 'cookie': '0x0', 'duration': '8.528s', 'table': '0', 'n_packets': '0', 'n_bytes': '0', 'idle_timeout': '60', 'priority': '65535', 'arp,in_port': 's1-eth2', 'vlan_tci': '0x0000', 'dl_src': '62:ee:31:2b:35:7c', 'dl_dst': 'a2:72:e7:06:75:2e', 'arp_spa': '10.0.0.2', 'arp_tpa': '10.0.0.3', 'arp_op': '2' 'actions=output':'s1-eth3'},
+                    { 'cookie': '0x0', 'duration': '4.617s', 'table': '0', 'n_packets': '0', 'n_bytes': '0', 'idle_timeout': '60', 'priority': '65535', 'arp,in_port': 's1-eth1', 'vlan_tci': '0x0000', 'dl_src': 'd6:fc:9c:e7:a2:f9', 'dl_dst': 'a2:72:e7:06:75:2e', 'arp_spa': '10.0.0.1', 'arp_tpa': '10.0.0.3', 'arp_op': '2' 'actions=output':'s1-eth3'}
+            ]
+
+        Attributes:
+            _bridges (list): A list where each element contains the list of dictionary elements having flow dumps.
```
xiangce: Please remove this attribute, prefix "_" means a private member in python.

insights/parsers/ovs_ofctl_dump_flows.py
```diff
@@ -0,0 +1,85 @@
...
+
+This module provides class ``OVSofctlDumpFlows`` to parse the
+output of command ``/usr/bin/ovs-ofctl dump-flows <bridge-name>``.
+"""
+
+
+from insights import CommandParser, parser
+from insights.parsers import SkipException
+from insights.specs import Specs
+from ..parsers import split_kv_pairs
```
xiangce: Please do not mix `full path` importing and `relative` importing in one single file.

L13 can be merged to L11

insights/parsers/ovs_ofctl_dump_flows.py
```diff
@@ -0,0 +1,85 @@
...
+            ]
+
+        Attributes:
+            _bridges (list): A list where each element contains the list of dictionary elements having flow dumps.
+
+        Raises:
+            SkipException: When the file is empty or data is not present for a bridge.
+
+        Examples:
+            >>> len(ovs_obj._bridges)
```
xiangce: This example line should be removed.

vishwanathjadhav: Done

insights/parsers/ovs_ofctl_dump_flows.py
vishwanathjadhav: Done

insights/parsers/ovs_ofctl_dump_flows.py
vishwanathjadhav: Done

PR #2134, opened 2019-08-07
Remove prefix "!" or "*" for yum repolist

huali027:
I found this parser is used in a lot of plugins, and many of them don't consider that maybe
it starts with exclamation mark. I think it is useless for comparation, so we can remove
it in the parser.

### DIFF
--- a/insights/parsers/tests/test_yum_repolist.py
+++ b/insights/parsers/tests/test_yum_repolist.py
@@ -42,6 +42,15 @@ repo id                              repo name                                  
 repolist: 21,581
 """
 
+YUM_REPOLIST_CONTENT_OUT_OF_DATE_AND_NOT_MATCH_METADATA = """
+Loaded plugins: product-id, search-disabled-repos, subscription-manager
+Repodata is over 2 weeks old. Install yum-cron? Or run: yum makecache fast
+repo id                              repo name                                          status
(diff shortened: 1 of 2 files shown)


huali027:
@bfahr, if we remove the mark before repo id in the parser, then we don't need to update lots of plugins, what's your opinion for this PR? Please help to review it when you have time, thanks.

bfahr:
@huali027 would you please open an issue for this enhancement.  Is there any specific meaning associated with the `!` and `*` chars?

bfahr:
Based on [this solution](https://access.redhat.com/solutions/2267871), the `!` indicates "the repository has expired metadata".  And based on [this solution](https://access.redhat.com/solutions/59244), the `*` indicates "the first column will start with a '*' if the repo. has metalink data and the latest metadata is not local".  So based on the fact that those chars provide information that could be useful we should not just throw those away, unless there is another column that provides the same information.

huali027:
@bfahr, yes, from the yum man page, we can see the description too. Yesterday I checked all the user cases, generally, we just check if some repo like "rhel-server-rhscl-7-rpms" is enabled with this parser. So in most cases, only the first part of "!rhel-server-rhscl-7-rpms/7Server/x86_64" without a mark is useful. What about we update the key to the first part and keep all the info in the values part? Let me update my patch, you can help to review it to see if it is okay, thanks.

PR #2135, opened 2019-08-07
Add parser to parse the output of the command lscpu

psachin:


psachin:
The specs are already merged: https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/default.py#L480

Insights-core-asset: https://github.com/RedHatInsights/insights-core-assets/pull/250

insights/parsers/lscpu.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("No data.")
+        self.data = {}
+        _content = split_kv_pairs(content, split_on=":")
+
+        # Cleanup
+        for i in _content:
+            if "(s)" in i:
```
xiangce: This is an intentional modification to the `key` which needs an explicit explanation in the docstring.

insights/parsers/lscpu.py
```diff
@@ -0,0 +1,67 @@
...
+        >>> output.data['Cores per socket']
+        '1'
+        >>> output.data['Sockets']
+        '1'
+    """
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("No data.")
+        self.data = {}
+        _content = split_kv_pairs(content, split_on=":")
```
xiangce: By using this `split_kv_parirs` function, the code looks simple, but the execution cost of the code is double (the `content` was iterated twice). Since the metadata is quite simple, better do all the parsing in one single loop (by merging the following `for` clause  (L63) and this `split_kv_pairs` into one loop)

insights/parsers/lscpu.py
```diff
@@ -0,0 +1,67 @@
+"""
+LsCPU - command ``lscpu``
+=========================
+
+This module provides the information about the CPU architecture using the output of the command ``lscpu``.
+"""
+from insights.parsers import SkipException, split_kv_pairs
+from insights.specs import Specs
+from .. import CommandParser, parser, LegacyItemAccess
```
xiangce: Better uniform the usage of `import` format in one single file, but not mixing the '..' (relative path) and full path.

insights/parsers/lscpu.py
psachin: Done

insights/parsers/lscpu.py
psachin: Done

insights/parsers/lscpu.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@parser(Specs.lscpu)
+class LsCPU(CommandParser, LegacyItemAccess):
```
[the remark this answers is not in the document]
psachin: Done

insights/parsers/lscpu.py
psachin: @xiangce I'm not sure if I understand your suggestion. What do you mean by "merge the `for` and `split_kv_pairs`"?

xiangce: The `split_kv_pairs` iterated the content 1 pass, and then the `for` at L63 iterated the content 1 pass again, and the logic in both loops are not complex. I mean better do them in one single loop. (do not use the `split_kv_pairs`)

xiangce:
@psachin - Please create an INSPEC Jira for adding this new spec at first. Since there is a similar spec `/proc/cpuinfo` in the existing specs, you need to explain why the existing one `cpuinfo` doesn't work to you.

insights/parsers/lscpu.py
psachin: Got it.

psachin:
> @psachin - Please create an INSPEC Jira for adding this new spec at first. Since there is a similar spec `/proc/cpuinfo` in the existing specs, you need to explain why the existing one `cpuinfo` doesn't work to you.

JIRA created

psachin:
@xiangce Please check now.

insights/parsers/lscpu.py
```diff
@@ -0,0 +1,67 @@
...
+        '2'
+        >>> output.data['Cores per socket']
+        '1'
+        >>> output.data['Sockets']
+        '1'
+    """
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("No data.")
+        self.data = {}
```
bfahr: @psachin why not use a more descriptive name that data to help a user understand what information is provided?

psachin: @bfahr @xiangce How about `self.info`?

psachin: @bfahr  Used `self.info` in place of `self.data`. Please review.

insights/parsers/lscpu.py
xiangce: I mean you should explain it in the docstring which will be published in the documentation of the parsers, but not the comment which is only visible in the code. To let the user understand what kind of `keys` are available in this dict.

insights/parsers/tests/test_lscpu.py
```diff
@@ -0,0 +1,101 @@
...
+L2 cache:              4096K
+NUMA node0 CPU(s):     0
+""".strip()
+
+BLANK = """
+""".strip()
+
+BAD_LSCPU = """
+Architecture:          x86_64
+CPU op-mode(s) =        32-bit, 64-bit
```
xiangce: For this kind of abnormal lines, I think it's necessary to raise a `ParserExeption`, but not skip it.

psachin: AFAIK `ParserException` will skip all the data where in this case the valid key-values will be captured. Finally the example used `=` will not ever happen in the output of `lscpu`. The reason I used the example to have full code coverage.

xiangce: Since it will never happen, lease remove this line

insights/parsers/lscpu.py
xiangce: @psachin - please add corresponding description for the keys in docstring.

psachin: Sure.

insights/parsers/tests/test_lscpu.py
psachin: @xiangce It will reduce the coverage from 100% to 96%. Do you really want me to remove it?

xiangce: As per the following code:
```python
            if split_on in line:
                k, v = line.split(split_on, 1)
                self.info[k.replace("(s)", "s").strip()] = v.strip()
``` 

 The [L68](https://github.com/RedHatInsights/insights-core/pull/2135/files#diff-1ab499a055ecdc312d029809c71dbd71R68) expects a `':'` in the line, and if there is not, the line will be skipped. And since there is such a check, it indicates some lines might not contain a `':'`.

Although, as you mentioned, this abnormal line ("xx= yy") will never happen, how about other cases?
If there are no other cases, why do we check it in L68? Since you have already thought there might be some lines missing the `':'`, and it's abnormal, it's reasonable to stop the parsing and raise an Exception to us.

But if you make sure there will be no lines missing the `':'`,  the L68 can be removed directly, right?

The goal of 100% coverage test is to help us to find any potential risk of the code. We'd better do all we can think of.

psachin: @xiangce What is the issue with the current implementation? As per my understanding having an additional check won't hurt at all.

psachin: @xiangce I removed L68(check for ":") as per your suggestion as the code was not good. Please check now.

Issue #2138, opened 2019-08-09
The SELinux combiner doesn't use grub blscfg for RHEL8


jsvob:
* The grub_conf combiner was updated to support blscfg and RHEL8 - https://github.com/RedHatInsights/insights-core/issues/1650

* The SELinux combiner doesn't use the grub_conf combiner, but it uses the grub parsers directly - 
https://github.com/RedHatInsights/insights-core/blob/3.0/insights/combiners/selinux.py#L43. So it doesn't work on RHEL8.

* Expected: The SELinux combiner uses the grub_conf combiner and works on RHEL8.

Thank you.

jsvob:
Hi @xiangce, I've assigned you to this just to give you visibility, since you've worked on https://github.com/RedHatInsights/insights-core/issues/1650. If you delegate it to someone else, that would be none of my business :)

xiangce:
Hi @jsvob, Thanks for that, I will fix it asap.

jsvob:
@xiangce Thank you for the fix!

PR #2140, opened 2019-08-12
Use GrubConf combiner in SELinux

xiangce:
- fix: #2138


insights/combiners/selinux.py
```diff
@@ -39,17 +39,16 @@
 BOOT_NOT_ENFORCING = 'selinux_conf_not_enforcing'
 
 
-@combiner(SEStatus, SelinuxConfig,
-          optional=[Grub1Config, Grub1EFIConfig, Grub2Config, Grub2EFIConfig])
+@combiner(SEStatus, SelinuxConfig, GrubConf)
```
bfahr: @xiangce the original logic had grub parsers as optionals, you've made the combiner mandatory here.  The logic still checks for the existence of a grub combiner so I think it should still be optional.

xiangce: Done. Thanks @bfahr

bfahr:
@jsvob @skontar can you take a look at this PR that was merged.  It appears to have broken prodsec rules.

bfahr:
@jsvob @skontar nevermind, I believe @kgrant-rh  fixed the problem.

PR #2152, opened 2019-08-20
Minor code comment fix: s/make_pass/Hostname

iNecas:
While going though the code, I've noticed some minor inconsistencies, so sending fixes along.

### DIFF
--- a/insights/tools/insights_inspect.py
+++ b/insights/tools/insights_inspect.py
@@ -7,9 +7,9 @@ ipython session where you can inspect the outcome.
 
 >>> insights-inspect insights.parsers.hostname.Hostname
 IPython Console Usage Info:
-Enter 'make_pass.' and tab to get a list of properties
+Enter 'Hostname.' and tab to get a list of properties
 Example:
-In [1]: make_pass.<property_name>
+In [1]: Hostname.<property_name>
 Out[1]: <property value>
 To exit ipython enter 'exit' and hit enter or use 'CTL D'
 Python 3.6.6 (default, Jul 19 2018, 14:25:17)
@@ -25,9 +25,9 @@ Out[1]: 'lhuett.usersys.redhat.com'"
 
 >>> insights-inspect insights.combiners.hostname.hostname
 IPython Console Usage Info:
-Enter 'make_pass.' and tab to get a list of properties
+Enter 'hostname.' and tab to get a list of properties
 Example:
-In [1]: make_pass.<property_name>
+In [1]: hostname.<property_name>
 Out[1]: <property value>
 To exit ipython enter 'exit' and hit enter or use 'CTL D'
 Python 3.6.6 (default, Jul 19 2018, 14:25:17)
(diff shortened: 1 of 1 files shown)


insights/tools/insights_inspect.py
```diff
@@ -25,9 +25,9 @@
 
 >>> insights-inspect insights.combiners.hostname.hostname
 IPython Console Usage Info:
-Enter 'make_pass.' and tab to get a list of properties
+Enter 'Hostname.' and tab to get a list of properties
```
lhuett: Thanks for catching this @iNecas  
This should actually be ...
Enter 'hostname.'

insights/tools/insights_inspect.py
```diff
@@ -25,9 +25,9 @@
...
 Example:
-In [1]: make_pass.<property_name>
+In [1]: Hostname.<property_name>
```
lhuett: This should actually be ...
Enter 'hostname.'
Line 40 below should also be lowercase 'hostname'

insights/tools/insights_inspect.py
```diff
@@ -43,7 +43,7 @@
 
 >>> insights-inspect insights.specs.Specs.hostname
 IPython Console Usage Info:
-Enter 'make_pass.' and tab to get a list of properties
+Enter 'Hostname.' and tab to get a list of properties
```
lhuett: This should actually be ...
Enter 'hostname.'
Line 58 below should be 'hostname.cmd' instead of 'CommandOutputProvider.cmd'

insights/tools/insights_inspect.py
```diff
@@ -61,7 +61,7 @@
 
 >>> insights-inspect examples.rules.bash_version.report
 IPython Console Usage Info:
-Enter 'make_pass.' and tab to get a list of properties
+Enter 'Hostname.' and tab to get a list of properties
```
lhuett: This should actually be ...
Enter 'report.'
Line 66 should be 'report.<property_name>' instead of make_pass.<property_name>
Line 76 below should be 'report.values()' instead of 'make_pass.values()'
Before line 85 there should be a line with 'import insights'

bfahr:
@iNecas please add --sign-off to your commits to enable passing the DCO tests.

insights/tools/insights_inspect.py
iNecas:
```
insights-inspect insights.parsers.hostname.Hostname ~/tmp/ccx/sosreport-luna122-1-2019-08-19-uakvnas.tar.xz

IPython Console Usage Info:

Enter 'Hostname.' and tab to get a list of properties 
Example:
In [1]: Hostname.<property_name>
Out[1]: <property value>

To exit ipython enter 'exit' and hit enter or use 'CTL D'

Starting IPython Interpreter Now 

```

and I actually can use `Hostname`, while `hostname` would not work.

insights/tools/insights_inspect.py
iNecas: ditto

insights/tools/insights_inspect.py
iNecas: Scratch that, I've missed there is different input here. Will update

iNecas: Nice catch, updated

insights/tools/insights_inspect.py
iNecas: Updated

insights/tools/insights_inspect.py
iNecas: Updated

insights/tools/insights_inspect.py
```diff
@@ -82,9 +82,10 @@
 insights.specs.Specs.installed_rpms,
 <class 'insights.parsers.installed_rpms.InstalledRpms'>,
 <function report at 0x7f0303c679d8>])
-In [3]: p = broker[insights.parsers.installed_rpms.InstalledRpms]
-In [4]: a = p.get_max('bash')
-In [5]: a.nevra
+In [3]: import insights
```
iNecas: Needed to add this, otherwise I was getting

```
NameError: name 'insights' is not defined
```

iNecas:
`--sign-off` added

Issue #2139, opened 2019-08-12
Variables are used in /boot/loader/entries/*.conf on RHEL8


xiangce:
After doing the following update operation on RHEL8:
```
[root@vm37-146 ~]# sudo dnf update kernel
...
Installed:
  kernel-4.18.0-80.7.2.el8_0.x86_64                  kernel-modules-4.18.0-80.7.2.el8_0.x86_64
  kernel-core-4.18.0-80.7.2.el8_0.x86_64

Complete!
```

I found:
```
[root@vm37-146 ~]# cat /boot/loader/entries/08c540cbca4d412c83e44a745aac36eb-4.18.0-80.7.2.el8_0.x86_64.conf
title Red Hat Enterprise Linux (4.18.0-80.7.2.el8_0.x86_64) 8.0 (Ootpa)
version 4.18.0-80.7.2.el8_0.x86_64
linux /vmlinuz-4.18.0-80.7.2.el8_0.x86_64
initrd /initramfs-4.18.0-80.7.2.el8_0.x86_64.img $tuned_initrd
options $kernelopts $tuned_params                   <--<<< variables are used here.
id rhel-20190726105706-4.18.0-80.7.2.el8_0.x86_64
grub_users $grub_users
grub_arg --unrestricted
grub_class kernel
```
The `variables` are defined in `/boot/grub2/grub.cfg` and `/boot/grub2/grubenv`:
~~~
[root@vm37-146 ~]# grep tuned_params /boot/* -rn
/boot/grub2/grub.cfg:74:set tuned_params=""
/boot/loader/entries/08c540cbca4d412c83e44a745aac36eb-4.18.0-80.el8.x86_64.conf:5:options root=/dev/mapper/rhel_vm37--146-root ro crashkernel=auto resume=/dev/mapper/r
hel_vm37--146-swap rd.lvm.lv=rhel_vm37-146/root rd.lvm.lv=rhel_vm37-146/swap $tuned_params
/boot/loader/entries/08c540cbca4d412c83e44a745aac36eb-4.18.0-80.1.2.el8_0.x86_64.conf:5:options root=/dev/mapper/rhel_vm37--146-root ro crashkernel=auto resume=/dev/ma
pper/rhel_vm37--146-swap rd.lvm.lv=rhel_vm37-146/root rd.lvm.lv=rhel_vm37-146/swap $tuned_params
[...]

PR #2155, opened 2019-08-20
De-reference the variables in blscfg

xiangce:
- fix: #2139


xiangce:
#3072 collected the "/etc/default/grub", this patch is not required anymore.

PR #2171, opened 2019-08-30
Added db2licm command parser

chawlanikhil24:
Adde SPEC collects `db2licm -l` command stdout:

```sh
Product name:                     "DB2 Enterprise Server Edition" 
License type:                     "CPU Option" 
Expiry date:                      "Permanent" 
Product identifier:               "db2ese" 
Version information:              "9.7" 
Enforcement policy:               "Soft Stop" 
Features: 
DB2 Performance Optimization ESE: "Not licensed" 
DB2 Storage Optimization:         "Not licensed" 
DB2 Advanced Access Control:      "Not licensed" 
IBM Homogeneous Replication ESE:  "Not licensed" 
```

### DIFF
--- a/insights/parsers/db2licm.py
+++ b/insights/parsers/db2licm.py
@@ -41,8 +41,6 @@ class DB2Info(CommandParser, dict):
 
         >>> parser_result.keys()
         dict_keys(['DB2 Enterprise Server Edition', 'DB2 Connect Server'])
-        >>> parser_result['DB2 Enterprise Server Edition']
-        {'License type': 'CPU Option', 'Expiry date': 'Permanent', 'Product identifier': 'db2ese', 'Version information': '9.7', 'Enforcement policy': 'Soft Stop', 'DB2 Performance Optimization ESE': 'Not licensed', 'DB2 Storage Optimization': 'Not licensed', 'DB2 Advanced Access Control': 'Not licensed', 'IBM Homogeneous Replication ESE': 'Not licensed'}
         >>> parser_result['DB2 Enterprise Server Edition']["Version information"]
         '9.7'
 
--- a/insights/parsers/db2licm.py
+++ b/insights/parsers/db2licm.py
@@ -79,7 +79,7 @@ class DB2Info(LegacyItemAccess, CommandParser):
                 if key == "Features":
                     continue
             else:
-                raise ParseException("Unable to parse db2licm info: {}".format(content))
+                raise ParseException("Unable to parse db2licm info: {0}".format(content))
             if key == "Product name":
                 if name is None:
                     name = val
@@ -97,4 +97,4 @@ class DB2Info(LegacyItemAccess, CommandParser):
             # If no data is obtained in the command execution then throw an exception instead of returning an empty
             # object.  Rules depending solely on this parser will not be invoked, so they don't have to
             # explicitly check for invalid data.
-            raise ParseException("Unable to parse db2licm info: {}".format(content))
+            raise ParseException("Unable to parse db2licm info: {0}".format(content))
--- a/insights/parsers/db2licm.py
+++ b/insights/parsers/db2licm.py
@@ -39,8 +39,8 @@ class DB2Info(CommandParser, dict):
 
     Example:
 
-        >>> parser_result.keys()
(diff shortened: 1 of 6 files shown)


insights/parsers/db2licm.py
```diff
@@ -0,0 +1,100 @@
...
+in one ``Record`` object for each line of output.
+"""
+from insights.core.plugins import parser
+from insights.core import CommandParser, LegacyItemAccess
+from insights.parsers import ParseException, get_active_lines
+from insights.specs import Specs
+
+
+@parser(Specs.db2licm_l)
+class DB2Info(LegacyItemAccess, CommandParser):
```
xiangce: The `LegacyItemAccess` will be deprecated, would you please to mix it with `dict` directly, which provides all the same access features as `LegacyItemAccess`.

See: https://github.com/RedHatInsights/insights-core/blob/b6da835477b70f7898985a7fb61a403152784983/insights/parsers/rndc_status.py#L11

insights/parsers/db2licm.py
```diff
@@ -0,0 +1,100 @@
...
+        Product name:                     "DB2 Connect Server"
+        Expiry date:                      "Expired"
+        Product identifier:               "db2consv"
+        Version information:              "9.7"
+        Concurrent connect user policy:   "Disabled"
+        Enforcement policy:               "Soft Stop"
+
+    Example:
+
+        >>> parser_result.data.keys()
```
xiangce: Better do not use `.data` in the example.

insights/parsers/db2licm.py
```diff
@@ -0,0 +1,100 @@
...
+            output.
+
+    Raises:
+        ParseException: raised if data is not parsable.
+    """
+
+    def parse_content(self, content):
+
+        # Stored data in a dictionary data structure
+        self.data = {}
```
xiangce: After mixing the parser with `dict`, this `self.data` can be removed since all the contents are stored in the `dict`.

insights/parsers/db2licm.py
```diff
@@ -0,0 +1,100 @@
...
+
+        name = None
+        body = {}
+
+        # Input data is available in text file. Reading each line in file and parsing it to a dictionary.
+        for line in get_active_lines(content):
+            if ':' in line:
+                key, val = line.strip().split(":", 1)
```
xiangce: can be simplified as `key, val = [i.strip() for i in line.split(":", 1)]`

insights/parsers/db2licm.py
```diff
@@ -0,0 +1,100 @@
...
+                raise ParseException("Unable to parse db2licm info: {0}".format(content))
+            if key == "Product name":
+                if name is None:
+                    name = val
+                else:
+                    self.data[name] = body
+                    name = val
+                    body = {}
+            else:
+                body[key.strip()] = val.lstrip(" ")
```
xiangce: These `strip()`s can be ignored, since it's done in the above lines.

insights/parsers/db2licm.py
```diff
@@ -0,0 +1,100 @@
+"""
+IBM DB2 Sever details
+=====================
+
+Module for the processing of output from the ``db2licm -l`` command.
+Data is avaliable as rows of the output contained
+in one ``Record`` object for each line of output.
```
xiangce: Seems there is no such `Record` in the following code. Would you please to update it correspondingly?

insights/parsers/db2licm.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # Input data is available in text file. Reading each line in file and parsing it to a dictionary.
+        for line in get_active_lines(content):
+            if ':' in line:
+                key, val = [i.strip() for i in line.strip().split(":", 1)]
+                if key == "Features":
+                    continue
+            else:
+                raise ParseException("Unable to parse db2licm info: {0}".format(content))
+            if key == "Product name":
```
xiangce: This code (L68~L79) works well, however, it's a bit complicated, would you please to simplify it as:

```python
            if key == "Product name":
                body = {}
                self[val] = body
            else:
                body[key] = val
```

Issue #2173, opened 2019-08-30
No parser for "hostname -I"


xiangce:
Command "hostname -I" is collected by insights client, but there is no parser for it yet.

xiangce:
`IpAddr`  can also get the output of this command. Close this issue.

PR #2174, opened 2019-08-30
new parser for hostname -I

xiangce:
- fix: #2173 

### DIFF
--- a/insights/parsers/hostname.py
+++ b/insights/parsers/hostname.py
@@ -17,6 +17,9 @@ Examples:
     >>> hostname.domain
     'example.com'
 
+HostnameI - command ``hostname -I``
+-----------------------------------
+
 """
 
 from . import ParseException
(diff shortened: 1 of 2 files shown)


insights/parsers/hostname.py
```diff
@@ -46,3 +49,30 @@ def parse_content(self, content):
...
+        if len(content) != 1:
+            msg = "hostname data contains multiple non-empty lines"
+            raise ParseException(msg)
+        self.ipv4 = self.ipv6 = None
+        self.raw = content[0].strip()
+        for v in self.raw.split(None, 1):
+            if '.' in v:
+                self.ipv4 = v.strip()
+            elif ':' in v:
+                self.ipv6 = v.strip()
```
chenlizhong: I have one server, the output of `hostname -I` as below, it is no ipv6

~~~
[root@runner ~]# hostname -I
10.0.79.245 10.88.0.1 
~~~

insights/parsers/hostname.py
```diff
@@ -46,3 +49,30 @@ def parse_content(self, content):
...
+        ipv6: The IPv6 address of the host
+    """
+    def parse_content(self, content):
+        content = list(filter(None, content))
+        if len(content) != 1:
+            msg = "hostname data contains multiple non-empty lines"
+            raise ParseException(msg)
+        self.ipv4 = self.ipv6 = None
+        self.raw = content[0].strip()
+        for v in self.raw.split(None, 1):
```
chenlizhong: Here you just split one time. It can not handle multiple nics and ips.

~~~
  -I, --all-ip-addresses all addresses for the host
~~~

insights/parsers/hostname.py
xiangce: @chenlizhong - Thanks for the check. Will update it in the next commit.

xiangce:
Since the `ip addr` command can also cover the output of this `hostname -I` command, I will close this PR and change to use the `IpAddr` parser.

PR #2181, opened 2019-09-06
New rule for LsUsrLib64

xiangce:
- and sort the order of all ls* specs


JoySnow:
Look good to me.

PR #2183, opened 2019-09-09
Reorder lspci file path from sos archive

akshay196:
`lspci_-nvv` provides more verbose information than `lspci`. Hence, reordering
list of lspci files.


xiangce:
@akshay196 - is there any existing issue related to this PR? If no, would you please to create an issue for it? Thanks.

PR #2184, opened 2019-09-10
Docker specs for sosreport

pratikgadiya12:
- Required for OSP fingerprinting to check the trusted containers SHA's


pratikgadiya12:
Thanks @xiangce and @vishwanathjadhav for the review!

PR #2190, opened 2019-09-13
Fix lift2 example in Custom Predicates of Parsr Tutorial Notebook

csams:


bfahr:
@csams would you consider upgrading the `print` statements in the notebook to work with Python 3?

PR #2192, opened 2019-09-14
Fix issue when using local spec with bare file

bfahr:
* When using a local spec with bare file it may not
  include all attributes such as multi_output.  This
  change insures that the bare file option will still
  work with locally defined specs.

### DIFF
--- a/insights/__init__.py
+++ b/insights/__init__.py
@@ -332,7 +332,7 @@ def run(component=None, root=None, print_summary=False,
         broker = dr.Broker()
         broker[ExecutionContext] = ctx
         for spec, content in specs.items():
-            broker[spec] = content if hasattr(spec, 'multi_output') and spec.multi_output else content[-1]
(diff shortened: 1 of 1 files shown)


bfahr:
Fixes #2193

insights/__init__.py
```diff
@@ -332,7 +332,7 @@ class Args(object):
         broker = dr.Broker()
         broker[ExecutionContext] = ctx
         for spec, content in specs.items():
-            broker[spec] = content if spec.multi_output else content[-1]
+            broker[spec] = content if hasattr(spec, 'multi_output') and spec.multi_output else content[-1]
```
csams: Instead of `hasattr` we can use `dr.DELEGATES[spec].multi_output`, which will get the decorator instance of the datasource instead of the datasource itself, and the datasource decorator should always have a `multi_output` attribute.

bfahr: That makes sense, changes have been made and tested.

Issue #2196, opened 2019-09-16
Add support to older sosreport of RHEL6.5 and below


kenneth-dsouza:
On older sosreport version dmesg would be captured under sos_commands/general/dmesg.
Inisght rules would fail in such instances. Add the same in  sos_archive.py file to make it work.
Same goes for blkid, It would be under sos_commands/filesys/blkid.
Let me know if the below changes would be acceptable. I can work on a patch.

Example: For dmesg only.

 # git diff
diff --git a/insights/specs/sos_archive.py b/insights/specs/sos_archive.py
index 3ddb614b..a2acd0fe 100644
--- a/insights/specs/sos_archive.py
+++ b/insights/specs/sos_archive.py
@@ -32,7 +32,7 @@ class SosSpecs(Specs):
     docker_list_containers = first_file(["sos_commands/docker/docker_ps_-a", "sos_commands/docker/docker_ps"])
     docker_list_images = simple_file("sos_commands/docker/docker_images")
     docker_image_inspect = glob_file("sos_commands/docker/docker_inspect_*")
-    dmesg = first_file(["sos_commands/kernel/dmesg", "var/log/dmesg"])
+    dmesg = first_file(["sos_commands/kernel/dmesg", "sos_commands/general/dmesg" "var/log/dmesg"])
     dmidecode = simple_file("sos_commands/hardware/dmidecode")
     dmsetup_info = simple_file("sos_commands/devicemapper/dmsetup_info_-c")
     dumpe2fs_h = glob_file("sos_commands/filesys/dumpe2fs_-h_*")

JoySnow:
Hi @kenneth-dsouza , thanks for raising this issue. 
As we talked, you can go ahead to open a PullRequest for your patch in this repo.

JoySnow:
@kenneth-dsouza  Assign to you, as you are already working on this, to avoid duplicate implementation.

xiangce:
resolved by #2200

PR #2200, opened 2019-09-18
Update sos_archive to process old sosreport V2 files.

kenneth-dsouza:
On older sosreport version dmesg would be captured under sos_commands/general/dmesg.
Inisght rules would fail in such instances. Same goes for blkid,lsblk, pvs and vgs.
Add the same in sos_archive.py file to make it work.
This patch does not address all issues. Going forward I will try to update the sos_archive
as I discover more such files.

Signed-off-by: Kenneth D'souza <kennethdsouza94@gmail.com>

### DIFF
--- a/insights/specs/sos_archive.py
+++ b/insights/specs/sos_archive.py
@@ -10,7 +10,7 @@ simple_file = partial(simple_file, context=SosArchiveContext)
 
 class SosSpecs(Specs):
     auditctl_status = simple_file("sos_commands/auditd/auditctl_-s")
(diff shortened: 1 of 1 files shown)


JoySnow:
Patch for issue https://github.com/RedHatInsights/insights-core/issues/2196 .

JoySnow:
Taking a sosreport of rhel6.3 as an example, all these added files can be found, and it's content structure are same as the related files.
Approve this MR here.

JoySnow:
@xiangce  Could you help review this?

xiangce:
fix: #2196

PR #2206, opened 2019-09-19
Allow filtering by full module name and tags. Add rule tags to response json doc.

csams:
All components have a `tags` keyword in their decorators, and the values defined there can be overridden in configuration. This PR provides a `--tags` option to `insights run` that allows users to express which rules they want to run based on how they're tagged.

Tag expressions are written in a simple language that allows exact match, regular expression match, quoted strings or regexes, and grouping along with operators for negation, `and`, and `or`.

For example, `insights run -p telemetry --tags 'network & apache & !mod_cluster'` means select rules that have the `network` and `apache` tags but not a `mod_cluster` tag.

`insights-run` also has a `-k` option that applies the same expression language to the full module names of loaded components. If `-k` and `--tags` are used at the same time, the effect is an intersection of their selected rules. In other words, load components with `-p`, filter them by full module name with `-k`, and select from what's left with `--tags`.

For example, `insights run -p telemetry -k '/networking & !/bond'` means load all of the telemetry rules and run only those that have "networking" in the full module name but that don't have "bond" in the full module name.

The same `-k` and `--tags` options are available for `insights info`.

This PR also adds the tags for a rule to any result it contributes to the json or yaml output.

### DIFF
--- a/docs/api_index.rst
+++ b/docs/api_index.rst
@@ -62,6 +62,14 @@ insights.core.spec_factory
     :show-inheritance:
     :undoc-members:
 
+insights.core.taglang
+--------------------------
+
+.. automodule:: insights.core.taglang
+    :members:
+    :show-inheritance:
+    :undoc-members:
+
 insights.parsers
 ----------------
 
--- a/docs/manpages/insights-info.rst
+++ b/docs/manpages/insights-info.rst
@@ -37,6 +37,9 @@ OPTIONS
     -i COMPONENT --info COMPONENT
         Comma separated list of components to get dependency info about.
 
+    -k --pkg-query
+        Expression to select rules by package.
+
     -p PLUGINS --preload PLUGINS
         Comma separated list of packages or modules to preload.
 
@@ -49,6 +52,9 @@ OPTIONS
     -t TYPES --types TYPES
         Filter results based on component type; e.g. 'rule,parser'. Names without '.'
         are assumed to be in insights.core.plugins.
+
+    \-\-tags EXPRESSION
+        An expression for selecting which loaded rules to run based on their tags.
 
     -v --verbose
         Print component dependencies.
--- a/docs/manpages/insights-run.rst
+++ b/docs/manpages/insights-run.rst
@@ -63,6 +63,9 @@ OPTIONS
         Ansible inventory file for cluster analysis.  See INVENTORY(5) for more information
         about the options for format of the inventory file.
 
+    -k --pkg-query
+        Expression to select rules by package.
+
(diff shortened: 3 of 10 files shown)


xiangce:
@csams - Great, it's a nice feature and especially useful to the `tagging system`. It works well on my side on both `RHEL` and `Mac OS` environment, however, when I test it under Fedora30, the full traceback info is outputted, for non-existing tags or pkg expressions, e.g:

~~~
❱ insights-run -p telemetry -k '/sap_1' /tmp/
Install colorama if console colors are preferred.
  File "/home/liuxc/work/insights/insights-plugins/.python3/bin/insights-run", line 11, in <module>
    load_entry_point('insights-core', 'console_scripts', 'insights-run')()
  File "/home/liuxc/work/insights/insights-core/insights/__init__.py", line 417, in main
    run(print_summary=True)
  File "/home/liuxc/work/insights/insights-core/insights/__init__.py", line 328, in run
    raise Exception(msg)
Exception: No components for pkg-query expression: /sap_1

❱ insights-run -p telemetry --tags 'sap' /tmp/
Install colorama if console colors are preferred.
  File "/home/liuxc/work/insights/insights-plugins/.python3/bin/insights-run", line 11, in <module>
    load_entry_point('insights-core', 'console_scripts', 'insights-run')()
  File "/home/liuxc/work/insights/insights-core/insights/__init__.py", line 417, in main
    run(print_summary=True)
  File "/home/liuxc/work/insights/insights-core/insights/__init__.py", line 335, in run
    raise Exception(msg)
Exception: No components for tag expression: sap
~~~

On RHEL and Mac, it looks good:
~~~
(.python3) [root@vm37-146 insights-plugins]# insights run -p telemetry -k '/abc' /tmp/tmp49ecr64_/
No components for pkg-query expression: /abc
(.python3) [root@vm37-146 insights-plugins]# insights run -p telemetry --tags '/abc' /tmp/tmp49ecr64_/
No components for tag expression: /abc
~~~

This is not a problem, but it would be better if it works the same to RHEL on Fedora.

csams:
@xiangce I'm glad you like it! For the mac/rhel vs fedora, it looks like you ran `insights-run` on Fedora instead of `insights run`? We handle all the top level exceptions in the [insights cli code](https://github.com/RedHatInsights/insights-core/blob/master/insights/command_parser.py#L92).

xiangce:
@csams - My bad, I should use `insights run` (not `insights-run`) on Fedora, and it works well. That is the `insights run` works well on all the 3 OSs.  Thanks.

PR #2221, opened 2019-09-24
Allow some special characters in httpd conf comment

### LINKED ISSUE #2219: There is special characters in comment in httpd conf files
From one of the production insights archives, I found the following line in http conf file:
```
#Alterações realizadas por issue no Insights
```
then the parser raises the following exception
```
  File "/Users/huali/worktree/insights/insights-core/insights/combiners/httpd_conf.py", line 328, in __call__
    return self.Top(content)
  File "/Users/huali/worktree/insights/insights-core/insights/parsr/__init__.py", line 332, in __call__
    raise Exception(err.read())
Exception: At line 41 column 10:
```

huali027:
Fixes #2219

### DIFF
--- a/insights/parsr/__init__.py
+++ b/insights/parsr/__init__.py
@@ -1040,7 +1040,7 @@ class OneLineComment(Parser):
     """
     def __init__(self, s):
         super(OneLineComment, self).__init__()
-        p = Literal(s) >> Opt(String(set(string.printable) - set("\r\n")), "")
+        p = Literal(s) >> Opt(String(set(string.printable + '\udcc3' + '\udca7' + '\udcb5') - set("\r\n")), "")
         self.add_child(p)
 
     def process(self, pos, data, ctx):


huali027:
@csams, in your PR https://github.com/RedHatInsights/insights-core/pull/2218, I see you added some odd unicode characters too, so I submit a PR with adding these odd characters found in httpd conf files. Please help to review.

csams:
Hey @huali027 thanks for the PR! I didn't see it before taking a look and submitting #2223. My fix for #2218 wasn't very good since the real problem was in AnyChar in parsr. #2223 should fix both the comment issues.

huali027:
> Hey @huali027 thanks for the PR! I didn't see it before taking a look and submitting #2223. My fix for #2218 wasn't very good since the real problem was in AnyChar in parsr. #2223 should fix both the comment issues.
@csams, any char is the best, then we don't need to do any change if we see other odd characters, I'll close this pr, thanks.

PR #2240, opened 2019-10-01
LogFileOutput token_scan returns a bool, reverse is meaningless

PaulWay:
1) The documentation for the return value (or rather the value set in
   the attribute being created by the token_scan) is a boolean, not a
   list of dicts.

2) Because we search for the presence of a matching line anywhere in
   the file, the 'reverse' parameter is useless here.

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -995,7 +995,7 @@ class LogFileOutput(six.with_metaclass(ScanMeta, Parser)):
         cls.scanner_keys.add(result_key)
 
     @classmethod
-    def token_scan(cls, result_key, token, check=all, reverse=False):
+    def token_scan(cls, result_key, token, check=all):
         """
         Define a property that is set to true if the given token is found in
         the log file.  Uses the __contains__ method of the log file.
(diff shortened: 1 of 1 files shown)


insights/core/__init__.py
```diff
@@ -1007,12 +1007,12 @@ def token_scan(cls, result_key, token, check=all, reverse=False):
             reverse(bool): scan start from the head when ``False`` by default, otherwise start from the tail
 
         Returns:
-            (list): list of dictionaries corresponding to the parsed lines contain the `token`.
+            (bool): the property will contain True if a line contained (any
```
psachin: @PaulWay I agree  that the `reverse` is not useful in `token_scan()` but it returns the `list` of `dict` as mentioned.

PaulWay: Um, the code is: `return any(...)` - which returns a boolean.  If you're expecting a list of dict, you should be using `keep_scan` or `last_scan`.  The purpose of `token_scan` is to say "did I see this token anywhere in the output", but without returning any more information than that.

PR #2241, opened 2019-10-01
use package_info instead of client method, check systemd env

gravitypriest:
RHEL 8 systemd complains pretty loudly if you try to run `systemd-notify` outside the systemd job itself. So, don't try to run `systemd-notify` unless `NOTIFY_SOCKET` is available.

In addition, use `package_info` directly instead of calling `InsightsClient().version()` to get the egg versions. `InsightsClient()` does things in its init function that we don't want to do superfluously, so let's not instantiate it just to get the version info.

### DIFF
--- a/insights/client/utilities.py
+++ b/insights/client/utilities.py
@@ -228,11 +228,11 @@ def get_version_info():
     '''
     Get the insights client and core versions for archival
     '''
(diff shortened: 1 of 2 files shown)


patchkez:
test this please

PR #2248, opened 2019-10-09
Parser sockstat

vishwanathjadhav:
A parser is added to parse the contents of `/proc/net/sockstat`

### DIFF
--- a/insights/parsers/sockstat.py
+++ b/insights/parsers/sockstat.py
@@ -1,18 +1,18 @@
 """
 SockStats - file ``/proc/net/sockstat``
 =======================================
-The ``TcpIpStats`` class implements the parsing of ``/proc/net/snmp``
+The ``TcpIpStats`` class implements the parsing of ``/proc/net/sockstat``
 file, which contains TCP/IP stats of individual layer.
 """
 
 
-from .. import Parser, parser, LegacyItemAccess
+from insights import Parser, parser
 from insights.parsers import SkipException
 from insights.specs import Specs
 
 
 @parser(Specs.sockstat)
-class SockStats(Parser, LegacyItemAccess):
+class SockStats(Parser, dict):
     """
     Parser for ``/proc/net/sockstat`` file.
 
@@ -37,6 +37,13 @@ class SockStats(Parser, LegacyItemAccess):
         4109
         >>> sock_obj.seg_element_details('frag', 'inuse')
         0
+        >>> sock_obj.get('sockets')
+        {'used': '3037'}
+        >>> sock_obj.get('sockets').get('used')
+        '3037'
+        >>> sock_obj.seg_element_details('tcp', 'abc') is None
+        True
+
 
     Resultant Data::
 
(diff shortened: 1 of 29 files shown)


insights/parsers/sockstat.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+
+
+from .. import Parser, parser, LegacyItemAccess
+from insights.parsers import SkipException
+from insights.specs import Specs
+
+
+@parser(Specs.sockstat)
+class SockStats(Parser, LegacyItemAccess):
```
xiangce: The `LegacyItemAccess` only works with `self.data` which is not in your class. In fact, the `LegacyItemAccess` will be deprecated soon. If you want the Parser can be accessed like a `dict`, you can mix it with `dict` directly.

FYI: https://github.com/RedHatInsights/insights-core/blob/89828b760399231917414c8326515fc97d79118c/insights/parsers/modinfo.py#L145

insights/parsers/sockstat.py
```diff
@@ -0,0 +1,117 @@
+"""
+SockStats - file ``/proc/net/sockstat``
+=======================================
+The ``TcpIpStats`` class implements the parsing of ``/proc/net/snmp``
+file, which contains TCP/IP stats of individual layer.
+"""
+
+
+from .. import Parser, parser, LegacyItemAccess
```
xiangce: Please use a uniform format for the `importing`s in one python module. E.g. all in relative path or all in full path

insights/parsers/sockstat.py
```diff
@@ -0,0 +1,117 @@
...
+                    'memory': '0'
+                  }
+        }
+
+    Raises:
+        SkipException: When contents are empty
+    """
+
+    def parse_content(self, content):
+        self.sock_stats = {}
```
xiangce: By `mixing` this parser with `dict` this attribute can be removed. If you want to keep it as an attribute, you could make it as a member-function decorated with `@property` and `return self` directly. And you should add it to the `Attributes` section of the docsting in that case as well.

insights/parsers/sockstat.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if not content:
+            raise SkipException("No Contents")
+
+        for line in content:
+            line_split = line.split(":")
+            nt_seg = line_split[0].lower()
+            self.sock_stats[nt_seg] = {}
+            val_lst = line_split[1].strip().split(' ')
```
xiangce: do you think the `split()` is  better than`split(' ')`?

vishwanathjadhav: Changed

insights/parsers/sockstat.py
vishwanathjadhav: Done

insights/parsers/sockstat.py
vishwanathjadhav: Done

insights/parsers/sockstat.py
vishwanathjadhav: Done

vishwanathjadhav:
Rule dependency:
https://gitlab.cee.redhat.com/insights-rules/insights-plugins/merge_requests/1977

PR #2261, opened 2019-10-16
Made a fix for the remove.conf bug

aleccohan:
Made a simple fix to the remove.conf bug that adds the [remove] section to the file if the file is empty.   

See:  https://projects.engineering.redhat.com/browse/RHCLOUD-160 for more info.


aleccohan:
I added the proposed fixes that we talked about, let me know if anything should be changed around!

PR #2263, opened 2019-10-17
Add parser cpupower_frequency_info

shzhou12:


insights/parsers/cpupower_frequency_info.py
```diff
@@ -0,0 +1,88 @@
+"""
+CpupowerFrequencyInfo - Commands ``cpupower frequency-info``
```
xiangce: I believe the command should be `cpupower -c all frequency-info` to get the `frequency-info` of all the cpulist, Otherwise it only returns the info of CPU0.

insights/parsers/cpupower_frequency_info.py
```diff
@@ -0,0 +1,88 @@
...
+============================================================
+"""
+from insights import parser, CommandParser, LegacyItemAccess
+from insights.parsers import SkipException, ParseException
+from insights.specs import Specs
+
+
+@parser(Specs.cpupower_frequency_info)
+class CpupowerFrequencyInfo(CommandParser, LegacyItemAccess):
```
xiangce: The `LegacyItemAccess` is not suggested to be used to access the `self.data`, please mix it with `dict` directly,

For your reference: 

https://github.com/RedHatInsights/insights-core/blob/532ad3d8c6cebec9dff0e49a151307d4ab7713a0/insights/parsers/modinfo.py#L145

https://github.com/RedHatInsights/insights-core/blob/532ad3d8c6cebec9dff0e49a151307d4ab7713a0/insights/parsers/modinfo.py#L226

insights/parsers/cpupower_frequency_info.py
```diff
@@ -0,0 +1,88 @@
...
+            raise ParseException("Incorrect content: '{0}'".format(content))
+
+        power_frequency_info = {}
+        current_data = power_frequency_info
+        parents = []
+        prev_shift = 0
+        prev_key = None
+
+        for line in content:
+            shift = (len(line) - len(line.lstrip()))
```
xiangce: I don't think to calculate the `shift` of the line is a good idea to recognize different `keys` of the parse result. can we have a talk?

insights/parsers/cpupower_frequency_info.py
xiangce: it's aligned to sosreport.

PR #2267, opened 2019-10-21
Delete run_client.py

subpop:
The module `insights.client.config` API has changed and this file did not change with it. It no longer properly creates an runs the client. Is it okay to delete?


subpop:
CC @kylape

PR #2273, opened 2019-10-29
WIP: Remove the Parser for '/proc/mounts'

xiangce:
- No rules use it
- Includes a bit update to the docstring.


xiangce:
The parser can be left here to parse the file in sosreport, no need to remove it.

PR #2282, opened 2019-10-31
Add more info to insights-client user agent

subpop:
This PR enhances the data included in the insights-client User-Agent. Rather than a (relatively) constant `<app_name>/<app_version>` string, it now includes version information about the platform (both host and Python platform information).

### DIFF
--- a/insights/client/constants.py
+++ b/insights/client/constants.py
@@ -10,7 +10,7 @@ class InsightsConstants(object):
         os.path.dirname(os.path.abspath(__file__)))
     sleep_time = 180
     command_blacklist = ('rm', 'kill', 'reboot', 'shutdown')
-    default_conf_dir = '/etc/insights-client'
+    default_conf_dir = os.getenv('INSIGHTS_CONF_DIR', default='/etc/insights-client')
     default_conf_file = os.path.join(default_conf_dir, 'insights-client.conf')
     user_agent = os.path.join(app_name, package_info["VERSION"])
     log_dir = os.path.join(os.sep, 'var', 'log', app_name)
--- a/insights/client/constants.py
+++ b/insights/client/constants.py
@@ -1,7 +1,5 @@
 import os
 
-from insights import package_info
-
 
 class InsightsConstants(object):
     app_name = 'insights-client'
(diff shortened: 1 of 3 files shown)


subpop:
@gravitypriest, this is ready for review

insights/client/constants.py
```diff
@@ -10,9 +8,8 @@ class InsightsConstants(object):
         os.path.dirname(os.path.abspath(__file__)))
     sleep_time = 180
     command_blacklist = ('rm', 'kill', 'reboot', 'shutdown')
-    default_conf_dir = '/etc/insights-client'
+    default_conf_dir = os.getenv('INSIGHTS_CONF_DIR', default='/etc/insights-client')
```
gravitypriest: I assume you added this for dev purposes?

subpop: Yes, it makes running modules in-tree easier, and *shouldn't* have any adverse effects otherwise.

Glutexo: What does it mean by running modules in-tree? Do you mean running stuff with `python insights/client/something.py`?

Glutexo: Do you think this variable can be passed in by insights-client when spawning the subprocess? It would allow to have dynamic paths for the Client being installed in other place than root.

subpop: > Do you mean running stuff with python insights/client/something.py?

Yes, that is correct. It also allows you to more easily spawn the client code as non-root under certain circumstances.

> Do you think this variable can be passed in by insights-client when spawning the subprocess? It would allow to have dynamic paths for the Client being installed in other place than root.

We could look into that. Are you thinking of changing the code that spawns `run.py` to inject that variable into the environment?

subpop: Actually, that could be pretty neat. At "compile" time, we generate a file similar to the classic `config.h` (`config.py`?) that contains all the typical path values set by `./configure`. Then the client execution code that runs within `/usr/bin/insights-client` will always run with the compiled values. It takes those values and sets them as environment variables when it creates the `run.py` subprocess, allowing the insights-core code to be entirely dynamic and relocatable at runtime.

PR #2288, opened 2019-11-07
New parser for `/usr/bin/sealert -l "*"`.

jsvob:


insights/parsers/sealert.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("Input content is empty")
+
+        self.raw_lines = content
+        self.reports = []
+        last_item = [None]  # python 2 doesn't have `nonlocal`
+
+        def finish_item():
```
skontar: Should this be a normal method?

jsvob: Most probably not.

jsvob: Fixed after offline discussion. Thanks.

insights/parsers/sealert.py
```diff
@@ -0,0 +1,120 @@
...
+        self.lines = []
+
+    def __str__(self):
+        return "\n".join(self.lines).strip()
+
+    def append_line(self, x):
+        self.lines.append(x)
+
+    @property
+    def lines_stripped(self):
```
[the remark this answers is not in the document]
jsvob: That's a good note!

insights/parsers/sealert.py
```diff
@@ -0,0 +1,120 @@
...
+        for line in content:
+            if line.startswith("SELinux is preventing "):
+                self.reports.append(Report())
+            if self.reports:  # skips the first report if it contains only partial data
+                self.reports[-1].append_line(line)
+
+        if not self.reports:
+            raise SkipException("No sealert reports")
+
+        self.raw = "\n".join(content)
```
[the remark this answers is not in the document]
jsvob: Hmm, that's also a very good note. It's unnecessary. If a rule would really need it, it can do the join itself, but it's much more probable that a rule would use `.raw_lines` or `.reports`.

jsvob:
@skontar please re-review. I believe I've addressed your concerns but github didn't notice.

insights/parsers/sealert.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def lines_stripped(self):
+        """
+        Returns the lines without trailing empty lines
+
+        Returns:
+            list: The lines without empty lines at the end of the list.
+        """
+        ret = []
+        for l in reversed(self.lines):
```
skontar: Reversing list two times is inefficient and unnecessary. 

This implementation is about 10-15x faster:

```python
def lines_stripped_faster(self):
    """
        Returns the lines without trailing empty lines
        Returns:
            list: The lines without empty lines at the end of the list.
    """
    lines = self.lines[:]
    for index in range(len(self.lines) - 1, -1, -1):
        if lines[index] == '':
            del(lines[index])
        else:
            break
    return lines
```
(not thoroughly tested)

jsvob: That's good!

jsvob:
@xiangce fixed, sorry for that!

PR #2289, opened 2019-11-08
Add tags.conf

subpop:
This PR introduces a new config file, `tags.conf`, as well as the mechanisms to include tags in the collected archive.

The file `/etc/insights-client/tags.conf` is a YAML file containing key/value pairs that can be used to tag the node with desired attributes. For example:

```yaml
---
zone: us-east1
owner: zach
exclude: true
```

Upon collection, the file `/etc/insights-client/tags.conf` is parsed and its contents are written to the insights archive for inclusion in the upload.

In addition, if the `--group` argument is passed to `insights-client`, the value of the argument is written to the tags file with the `group` key. For example:

```
# cat /etc/insights-client/tags.conf
---
zone: us-east1
owner: zach
exclude: true
# insights-client --group=app-db-01
...
# cat /etc/insights-client/tags.conf
---
zone: us-east1
owner: zach
exclude: true
group: app-db-01
#
```

If the `--group` tag is then omitted from subsequent calls to `insights-client`, no change is made to the tags file.

Because the tags file is YAML, it is possible to create multi-value tags using a YAML list. Similarly, it is possible to create namespaces or nested tags using a YAML dictionary. When the tags are written to the insights archive, they are wrapped inside a top-level "insights-client" key in order to avoid possible collision with future tagging sources. Tags are written to the archive as JSON.

### DIFF
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -15,7 +15,9 @@ from .auto_config import try_auto_configuration
 from .utilities import (delete_registered_file,
                         delete_unregistered_file,
                         write_to_disk,
-                        generate_machine_id)
+                        generate_machine_id,
+                        get_tags,
+                        write_tags)
 
 logger = logging.getLogger(__name__)
 net_logger = logging.getLogger("network")
@@ -55,6 +57,11 @@ class InsightsClient(object):
         # used for requests
         self.session = None
         self.connection = None
+
+        if self.config.group:
+            tags = get_tags()
+            tags["group"] = self.config.group
+            write_tags(tags)
 
     def _net(func):
         def _init_connection(self, *args, **kwargs):
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -60,6 +60,8 @@ class InsightsClient(object):
 
         if self.config.group:
             tags = get_tags()
+            if tags is None:
+                tags = {}
             tags["group"] = self.config.group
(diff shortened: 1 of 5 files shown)


SteveHNH:
In testing, it appears that once you attempt to add a group as directed in the PR description, it modifies tag.conf into a JSON format. We need to work out why this is happening

subpop:
So it appears that Python 2.7 (the interpreter that is used when invoking `/usr/bin/insights-client`) writes JSON when `yaml.dump` is called.

```
Python 2.7.5 (default, Jun 11 2019, 14:33:56) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> yaml
<module 'yaml' from '/usr/lib64/python2.7/site-packages/yaml/__init__.pyc'>
>>> yaml.dump({'a':1}, Dumper=yaml.Dumper)
'{a: 1}\n'
```

I am not sure why yet.

subpop:
Right. So YAML appears to have this concept called flow styles that embed what looks identical to JSON inside YAML values. But it's not JSON, it's still YAML. Early versions of PyYAML (what's shipped in RHEL7, for example) default to enabling the default flow style. So we have to explicitly disable it with the keyword argument `default_flow_style=False`. :-]

PR #2292, opened 2019-11-11
Add a parser and spec for the new tags.json file

SteveHNH:
A new file is being added to the archive with PR #2289 and this parser
is meant to go along with it.


insights/parsers/tags.py
```diff
@@ -0,0 +1,9 @@
+from insights.core import JSONParser
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+
+@parser(Specs.tags)
+class Tags(JSONParser):
+    """ Class for parsing the content of ``branch_info``."""
```
subpop: Should this be ``tags.json`` rather than ``branch_info``?

PR #2293, opened 2019-11-12
Deprecate Satellite6Version

huali027:
There is a combiner "SatelliteVersion" to cover all versions of
satellite.

- fix: #2295 


xiangce:
fixed #2295

PR #2298, opened 2019-11-12
Add Parser SCTPSnmp

Rohan21Lobo:
Signed-off-by: Rohan Arora <roarora@redhat.com>

### DIFF
--- a/insights/parsers/sctp.py
+++ b/insights/parsers/sctp.py
@@ -9,6 +9,9 @@ SCTPEps - file ``/proc/net/sctp/eps``
 
 SCTPAsc - file ``/proc/net/sctp/assocs``
 ----------------------------------------
+
+SCTPSnmp - file ``/proc/net/sctp/snmp``
+---------------------------------------
 """
 
 from insights import Parser, parser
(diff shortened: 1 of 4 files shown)


vishwanathjadhav:
@xiangce - Please check why default is failing and jenkins continuous integration is in blocked state.

insights/parsers/sctp.py
```diff
@@ -254,3 +257,72 @@ def search(self, **args):
...
+
+            self[line_strip[0]] = int(line_strip[1]) if line_strip[1].isdigit() else None
+
+    @property
+    def stats(self):
+        """
+        dict: On success, it will return detailed sctp snmp stats in
+        a dictionary data format, on failure it will return ``None``
+        """
+        return {**self}
```
xiangce: Since the core is opensourced project and it could be run in all possible python versions (2.6, 2.7 and 3.6) environment. We need to support all of them. This `{**self}` is not supported by old python 2.6 and 2.7. Could you please change it to `self` directly?

insights/parsers/sctp.py
```diff
@@ -254,3 +257,72 @@ def search(self, **args):
...
+    Examples:
+
+        >>> type(sctp_snmp)
+        <class 'insights.parsers.sctp.SCTPSnmp'>
+        >>> sctp_snmp.get('SctpCurrEstab')
+        5380
+        >>> sctp_snmp.get('SctpChecksumErrors') == 0
+        True
+        >>> sctp_snmp.stats
+        {'SctpCurrEstab': 5380, 'SctpActiveEstabs': 12749, 'SctpPassiveEstabs': 55, 'SctpAborteds': 2142, 'SctpShutdowns': 5295, 'SctpOutOfBlues': 36786, 'SctpChecksumErrors': 0, 'SctpOutCtrlChunks': 1051492}
```
xiangce: Due to the different hash algorithms of python 2 and 3, the order of elements of the `dict`/`set` will be different. And here the test will only compare the converted string which is in a fixed order, if it passes in python 3, it will fail in python 2.

Would you please change it to compatible with all python versions? e.g.
```python
>>> len(sctp_snmp.stats)
8
>>> sctp_snmp.stats.get('SctpCurrEstab')
5380
```
Or `sorted` them and compare with `==`.

insights/parsers/sctp.py
```diff
@@ -254,3 +257,72 @@ def search(self, **args):
...
+
+        for line in content:
+            line_strip = line.split()
+            if len(line_strip) != 2:
+                raise ParseException("Contents are not compatible to this parser")
+
+            self[line_strip[0]] = int(line_strip[1]) if line_strip[1].isdigit() else None
+
+    @property
+    def stats(self):
```
xiangce: TBH,  I'm wondering why adding this interface? The Parser class is already a `dict`, right?

insights/parsers/sctp.py
```diff
@@ -254,3 +257,72 @@ def search(self, **args):
...
+    def parse_content(self, content):
+        if (not content) or (not self.file_path):
+            raise SkipException("No Contents")
+
+        for line in content:
+            line_strip = line.split()
+            if len(line_strip) != 2:
+                raise ParseException("Contents are not compatible to this parser")
+
+            self[line_strip[0]] = int(line_strip[1]) if line_strip[1].isdigit() else None
```
xiangce: As you did this check `line_strip[1].isdigit()`, is it necessary to raise another `ParseException` when the value is not a digit?

insights/parsers/sctp.py
```diff
@@ -9,6 +9,9 @@
 
 SCTPAsc - file ``/proc/net/sctp/assocs``
 ----------------------------------------
+
+SCTPSnmp - file ``/proc/net/sctp/snmp``
+----------------------------------------
```
xiangce: nitpick: please remove one `-`, to keep it in the same length with the title.

insights/parsers/sctp.py
Rohan21Lobo: @xiangce  Raised exception here in case there are more than 2 elements after splitting. Just making sure file is in correct format. 
Should I remove this?

insights/parsers/sctp.py
Rohan21Lobo: Removed the interface fucntion.

insights/parsers/sctp.py
Rohan21Lobo: Added new examples : 
https://github.com/RedHatInsights/insights-core/pull/2298/files#diff-1a1ec796e7a602ccbea2a1fda0999715R289-R294

List in output is already sorted

insights/parsers/sctp.py
Rohan21Lobo: Removed the interface fucntion

insights/parsers/sctp.py
Rohan21Lobo: Done

insights/parsers/sctp.py
xiangce: @Rohan21Lobo , it depends on you. From my view, it seems incorrect to set the value to `None` when the second splits is not a digit. But if you have confidence that it (non-digit) will never happen, better remove this check.

vishwanathjadhav: @Rohan21Lobo - You have already checked that there are 2 values after unpack, so no need to confirm it again. Also the second value will not be empty, it will always be digit. So, please remove the check.

Rohan21Lobo: Done 
https://github.com/RedHatInsights/insights-core/blob/c4b9e132e82631c774998115df6cdc1403c962d4/insights/parsers/sctp.py#L319-L324

PR #2306, opened 2019-11-14
Add sos archive path for httpd_M

akshay196:
* This adds datasource path from sos archive for httpd_M specs.


patchkez:
test this please

PR #2307, opened 2019-11-14
client: Maintain /v/l/i/results.json and /v/l/i/inventory.json files

mvollmer:
I propose to make these two files (or something like it) part of the stable API of the insights-client.

Open issues:

- The /v/li/results.json is supposed to contain the analysis results corresponding to the most recent upload.  With the legacy API, the upload POST request does not complete until the results are available.  I have heard that the platform API might complete a upload before results are available.

- The two files need to be documented somewhere as stable API.

- Where would the tests for this go?

### DIFF
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -394,6 +394,10 @@ class InsightsClient(object):
     def get_results(self, upload_response=None):
         client.get_results(self.config, self.connection, upload_response)
 
+    @_net
+    def get_inventory(self):
+        client.get_inventory(self.config, self.connection)
+
     def rotate_eggs(self):
         """
             moves newest.egg to last_stable.egg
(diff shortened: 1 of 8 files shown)


mvollmer:
The idea is that we consume these files in Cockpit.  

We want to give the user a link to the exact page on cloud.redhat.com with the Insights results for their machine. We need the inventory id for this.

We want to show a summary of the Insights results in the "Health" section of the system overview. For this we need the reports. With the legacy API, the response of the upload API call already includes a summary of the reports, and we can get it from /etc/insights-client/.last-upload.results.  However, my understanding is that this going to go away.  (Also, the response was only a summary and we will want the full details later on, I think.)

mvollmer:
We have different plans already, so closing.

Issue #2305, opened 2019-11-14
SatelliteVersion should be enhanced for Satellite Capsule


xiangce:
The capsule version is needed by some plugins.

PR #2311, opened 2019-11-15
Make satellite_version support Capsule

xiangce:
- fix: #2305 

### DIFF
--- a/insights/combiners/satellite_version.py
+++ b/insights/combiners/satellite_version.py
@@ -6,14 +6,13 @@ The following modules are included:
 
 SatelliteVersion - Version of Satellite Server
 ----------------------------------------------
+Combiner to get Satellite Server version information.
 
 CapsuleVersion - Version of Satellite Capsule (>=6.2)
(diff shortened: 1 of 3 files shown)


insights/combiners/satellite_version.py
```diff
@@ -2,20 +2,34 @@
...
+SatelliteVersion - Version of Satellite Server
+----------------------------------------------
+
+CapsuleVersion - Version of Satellite Capsule (>=6.2)
+-----------------------------------------------------
+
+Combiners to get Satellite Server and Capsule Version information. It uses the
+results of the :class:`insights.parsers.satellite_version:Satellite6Version`
+parser and the :class:`insights.parsers.installed_rpms.InstalledRpms` parser to
+determine the version.
```
huali027: here I think we can explain it more clearly. Since only SatelliteVersion uses Satellite6Version.

huali027:
@xiangce, one comment about the doc, the other part looks good to me.

Issue #2318, opened 2019-11-18
SatelliteEnableFeatures parser specs need to be revised


bfahr:
The specs need to be revised for the SatelliteEnabledFeatures parser added in #2301 so that they will work with insights-core client collection.  The archive spec needs to be updated.

PR #2319, opened 2019-11-19
Remove spec fist since it can't be collected by client

huali027:
Fixes: https://github.com/RedHatInsights/insights-core/issues/2318

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -27,7 +27,6 @@ from insights.core.spec_factory import first_file, listdir
 from insights.parsers.mount import Mount, ProcMounts
 from insights.parsers.dnf_module import DnfModuleList
 from insights.combiners.cloud_provider import CloudProvider
-from insights.combiners.satellite_version import SatelliteVersion
 from insights.components.rhel_version import IsRhel8
 from insights.specs import Specs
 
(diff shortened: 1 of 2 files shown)


insights/specs/default.py
```diff
@@ -388,15 +387,6 @@ def dumpdev(broker):
     jbcs_httpd24_httpd_error_log = simple_file("/opt/rh/jbcs-httpd24/root/etc/httpd/logs/error_log")
     httpd_pid = simple_command("/usr/bin/pgrep -o httpd")
     httpd_limits = foreach_collect(httpd_pid, "/proc/%s/limits")
-
```
xiangce: Better keep this spec in default.py

insights/specs/insights_archive.py
```diff
@@ -239,7 +239,6 @@ class InsightsArchiveSpecs(Specs):
     saphostctl_getcimobject_sapinstance = simple_file("insights_commands/usr.sap.hostctrl.exe.saphostctrl_-function_GetCIMObject_-enuminstances_SAPInstance")
     saphostexec_status = simple_file("insights_commands/usr.sap.hostctrl.exe.saphostexec_-status")
     saphostexec_version = simple_file("insights_commands/usr.sap.hostctrl.exe.saphostexec_-version")
-    satellite_enabled_features = simple_file("insights_commands/python_-m_insights.tools.cat_--no-header_satellite_enabled_features")
```
bfahr: @huali027 rather than removing this, it would be better to change it using the path that would be created by the client for storage in the archive:
```
$ mangle "/usr/bin/curl -sk https://localhost:9090/features --connect-timeout 5"
insights_commands/curl_-sk_https_..localhost_9090.features_--connect-timeout_5
```

huali027: @bfahr, but we agreed on that we don't add the spec "/usr/bin/curl -sk https://localhost:9090/features --connect-timeout 5" since we don't want to run it on all the hosts, right? So do you suggest we still update it to that?

bfahr: @huali027 you've correctly implemented this with a dependency on a datasource that will prevent it from running on all systems.  From an implementation standpoint, using the mangle as I suggested is the correct way to implement the archive spec.

PR #2328, opened 2019-11-22
Enhanced to get vxlan and openvswitch details

vishwanathjadhav:
Enhanced to get all the details of vxlan and openvswitch from `ip -s link`.

fix: https://github.com/RedHatInsights/insights-core/issues/2329

### DIFF
--- a/insights/parsers/ip.py
+++ b/insights/parsers/ip.py
@@ -71,9 +71,11 @@ def parse_ip_addr(content):
         elif line.startswith("link"):
             parse_link(line, current)
         elif 'vxlan' in line:
-            parse_vxlan_details(line, current)
+            split_content = line.split()
+            current['vxlan'] = split_content
         elif 'openvswitch' in line:
-            parse_ovs_details(line, current)
+            split_content = line.split()
+            current['openvswitch'] = split_content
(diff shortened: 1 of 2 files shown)


insights/parsers/ip.py
```diff
@@ -144,6 +148,16 @@ def parse_tx_stats(line, d):
     d["tx_collsns"] = int(split_content[5])
 
 
+def parse_vxlan_details(line, d):
```
xiangce: The computing of the function is quite simple, better remove the function and move the line to the calling place.

insights/parsers/ip.py
```diff
@@ -144,6 +148,16 @@ def parse_tx_stats(line, d):
...
+    split_content = line.split()
+    d['vxlan'] = split_content
+
+
+def parse_ovs_details(line, d):
```
xiangce: The same comment as above

insights/parsers/ip.py
vishwanathjadhav: Done

insights/parsers/ip.py
vishwanathjadhav: Done

PR #2336, opened 2019-11-26
Set client to use stdout instead of stderr

SteveHNH:
JIRA https://projects.engineering.redhat.com/browse/RHCLOUD-2196

We used to log to stderr to avoid creating problems using the
--to-stdout option, but that option no longer exists. We can safely use
stdout again


gravitypriest:
This may require some changes on QE side since they might be looking to stderr for certain output.

PR #2341, opened 2019-11-27
Add parser to report local kpatch patches

theyoyojo:
This patch creates a parser to report on the possible module names of kpatch patches that are installed on a machine.

Also add a supporting spec 'kpatch_patch_files'

Signed-off-by: Joel Savitz <jsavitz@redhat.com>

### DIFF
--- a/insights/parsers/kpatch_patches.py
+++ b/insights/parsers/kpatch_patches.py
@@ -1,0 +1,23 @@
+"""
+KpatchPatches - report locally stored kpatch patches
+====================================================
+
+This parser creates a list of the module names of locally
+stored kpatch modules. If no modules are installed, a
+ContentException will be raised.
+
+"""
+
+from .. import parser, CommandParser
+from insights.specs import Specs
+
+
+@parser(Specs.kpatch_patch_files)
+class KpatchPatches(CommandParser):
+    """
+    A parser for getting modules names of locally stored kpatch-patch files.
+    """
+
+    def parse_content(self, content):
+        # convert dashes to underscores, remove file suffixes, remove duplicates
+        self.patches = list(set([p.split('.')[0].replace("-", "_") for p in content]))
--- a/insights/specs/__init__.py
+++ b/insights/specs/__init__.py
@@ -258,6 +258,7 @@ class Specs(SpecSet):
     keystone_crontab = RegistryPoint()
     keystone_crontab_container = RegistryPoint()
     keystone_log = RegistryPoint(filterable=True)
+    kpatch_patch_files = RegistryPoint()
     krb5 = RegistryPoint(multi_output=True)
     ksmstate = RegistryPoint()
     kubepods_cpu_quota = RegistryPoint(multi_output=True)
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -26,6 +26,7 @@ from insights.core.spec_factory import first_of, foreach_collect, foreach_execut
 from insights.core.spec_factory import first_file, listdir
 from insights.parsers.mount import Mount, ProcMounts
 from insights.parsers.dnf_module import DnfModuleList
+from insights.parsers.uname import Uname
 from insights.combiners.cloud_provider import CloudProvider
 from insights.combiners.satellite_version import SatelliteVersion
 from insights.components.rhel_version import IsRhel8
@@ -494,6 +495,13 @@ class DefaultSpecs(Specs):
     keystone_crontab = simple_command("/usr/bin/crontab -l -u keystone")
     keystone_crontab_container = simple_command("docker exec keystone_cron /usr/bin/crontab -l -u keystone")
     keystone_log = first_file(["/var/log/containers/keystone/keystone.log", "/var/log/keystone/keystone.log"])
+
+    @datasource(Uname, context=HostContext)
+    def kpatch_patches_running_kernel_dir(broker):
+        un = broker[Uname]
+        return r"/var/lib/kpatch/" + un.kernel
+
+    kpatch_patch_files = command_with_args("ls %s", kpatch_patches_running_kernel_dir)
     krb5 = glob_file([r"etc/krb5.conf", r"etc/krb5.conf.d/*"])
     ksmstate = simple_file("/sys/kernel/mm/ksm/run")
     kubepods_cpu_quota = glob_file("/sys/fs/cgroup/cpu/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod[a-f0-9_]*.slice/cpu.cfs_quota_us")
--- a/docs/shared_parsers_catalog/kpatch_patches.rst
+++ b/docs/shared_parsers_catalog/kpatch_patches.rst
@@ -1,0 +1,3 @@
+.. automodule:: insights.parsers.kpatch_patches
+   :members:
+   :show-inheritance:
--- a/insights/parsers/tests/test_kpatch_patches.py
+++ b/insights/parsers/tests/test_kpatch_patches.py
@@ -1,0 +1,38 @@
+from insights.parsers import kpatch_patches
+from insights.tests import context_wrap
+from insights.core.plugins import ContentException
+import pytest
+
+
+ASSORTED_KPATCHES = """
+asdfasdfasdf_asdfasdfasdf-asdfasdfasdf_asdfasdfasdf.ko
+asdfasdfasdf_asdfasdfasdf-asdfasdfasdf_asdfasdfasdf.ko.xz
+foo-bar.ko
+foo-bar.ko.xz
+foo.ko
+foo.ko.xz
+test_klp_callbacks_demo.ko
+test_klp_callbacks_demo.ko.xz
+""".strip()
+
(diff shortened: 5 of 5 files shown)


insights/specs/__init__.py
```diff
@@ -599,6 +600,7 @@ class Specs(SpecSet):
     udev_persistent_net_rules = RegistryPoint()
     ulimit_hard = RegistryPoint()
     uname = RegistryPoint()
+    uname_r = RegistryPoint()
```
xiangce: The existing `uname` spec covers the `uname_r`.

insights/specs/default.py
```diff
@@ -494,6 +494,15 @@ def semid(broker):
     keystone_crontab = simple_command("/usr/bin/crontab -l -u keystone")
     keystone_crontab_container = simple_command("docker exec keystone_cron /usr/bin/crontab -l -u keystone")
     keystone_log = first_file(["/var/log/containers/keystone/keystone.log", "/var/log/keystone/keystone.log"])
+
+    uname_r = simple_command("/usr/bin/uname -r")
+
+    @datasource(uname_r, context=HostContext)
```
xiangce: The `uname_r` is not collected and it's already covered by the `uname` spec (`uname -a`), please change to use the existing `Uname` parser which is based on `uname` spec here. e.g.

```
from insights.parsers.uname import Uname
@datasource(Uname, context=HostContext)
  un = broker[Uname]
  return "/var/lib/kpatch/" + un.kernel
```

insights/parsers/kpatch_patches.py
```diff
@@ -0,0 +1,40 @@
...
+
+"""
+
+from .. import parser, Parser
+from insights.specs import Specs
+import re
+
+
+@parser(Specs.kpatch_patch_files)
+class KpatchPatches(Parser):
```
xiangce: This parser is for a command, better change to `CommandParser` to get the best usage.
See https://github.com/RedHatInsights/insights-core/blob/34505e28187c71c364a7c463d3f236caba192c7d/insights/core/__init__.py#L511-L523

insights/parsers/kpatch_patches.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    A parser for getting locally stored kpatch-patch modules
+    """
+    __re_cannot_access = re.compile(r"No such file or directory$")
```
xiangce: By leveraging the `CommandParser` this check can be saved.

insights/parsers/kpatch_patches.py
```diff
@@ -0,0 +1,40 @@
...
+    __re_archive = re.compile(r"\.xz$")
```
xiangce: TBH, this `regex` check is very simple, the `str_obj.endswith('.xz')` works in better efficiency than this `re` and the `endswith` looks better too. Would you please change it?

insights/parsers/kpatch_patches.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def get_patches(self):
```
xiangce: Seem this function interface is not necessary, since the `self.patches` attribute works in this way already, right?

insights/parsers/kpatch_patches.py
```diff
@@ -0,0 +1,40 @@
...
+        # If kpatch is not installed, the directory will not exist.
+        # In that case, the list of local kpatch files is empty.
+        if self.__re_cannot_access.search(content[0]):
+            self.patches = []
+            return
+
+        # remove module archives, convert dashes to underscores, remove ".ko" suffixes
+        self.patches =  \
+            list(map(lambda x: x.replace(".ko", ""),
+                list(map(lambda x: x.replace("-", "_"),
```
xiangce: I just wonder why here replace the `-` to `_`? After this, it's difficult to get the original name of the patches. E.g.

In the test case, there is `asdfasdfasdf_asdfasdfasdf-asdfasdfasdf_asdfasdfasdf.ko`, but after this `replace`, we'll get `asdfasdfasdf_asdfasdfasdf_asdfasdfasdf_asdfasdfasdf`. from the result, we don't know what the real original name is.
And the following 3 different names, will result in the same
```
asdfasdfasdf-asdfasdfasdf-asdfasdfasdf-asdfasdfasdf.ko
asdfasdfasdf-asdfasdfasdf_asdfasdfasdf-asdfasdfasdf.ko
asdfasdfasdf-asdfasdfasdf_asdfasdfasdf_asdfasdfasdf.ko
```
Do you think so?

insights/parsers/kpatch_patches.py
```diff
@@ -0,0 +1,40 @@
...
+    def parse_content(self, content):
+        self.patches = content
+        # If kpatch is not installed, the directory will not exist.
+        # In that case, the list of local kpatch files is empty.
+        if self.__re_cannot_access.search(content[0]):
+            self.patches = []
+            return
+
+        # remove module archives, convert dashes to underscores, remove ".ko" suffixes
+        self.patches =  \
```
xiangce: Comparing the multiple `lambda` functions, how do you think split it as below?
```
self.patches = [p.split('.')[0].replace(...) for p in self.patches]
```

insights/parsers/kpatch_patches.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def get_patches(self):
+        return self.patches
+
+    def parse_content(self, content):
+        self.patches = content
+        # If kpatch is not installed, the directory will not exist.
+        # In that case, the list of local kpatch files is empty.
+        if self.__re_cannot_access.search(content[0]):
+            self.patches = []
```
xiangce: Better `raise SkipException` here for no kpatch is installed, rasing the SkipException will avoid generating an instance for the Parser. Then you won't need to do further check when using it (`if obj.patches: ...`.

insights/specs/__init__.py
theyoyojo: Removed in latest patch

insights/specs/default.py
theyoyojo: Thanks for the tip, I do it this way in the latest patch

insights/parsers/kpatch_patches.py
theyoyojo: Changed in latest patch

insights/parsers/kpatch_patches.py
theyoyojo: This check is removed in the latest patch

insights/parsers/kpatch_patches.py
theyoyojo: All of these unnecessary regexes are removed in the latest patch

insights/parsers/kpatch_patches.py
theyoyojo: Removed in latest patch

insights/parsers/kpatch_patches.py
theyoyojo: The idea here is to produce a list of module names corresponding to the files in the /var/lib/kpatch/$(uname -r) directory that one could expect to find in the output of lsmod. While you do raise a valid point, I have spoken with some of the kpatch developers and they told me that the patch files will include kernel version information in the filename, so name collisions such as the above will be avoided. However, the filenames will include a combination of dashes and underscores so I test that these are all converted to underscores.

insights/parsers/kpatch_patches.py
theyoyojo: I simplify the parsing greatly using your suggestion in the latest patch

insights/parsers/kpatch_patches.py
theyoyojo: I wasn't quite sure where the appropriate place to raise the SkipException was, so I added an __init__ function for the parser that does a try/catch on super()__init__ and raises a SkipException when it catches a ContentException. Is this the right way to do it?

In any case, I remove this check from parse_content() in the latest patch.

insights/parsers/kpatch_patches.py
xiangce: Looks good to me. Thanks for the explanation.

insights/parsers/kpatch_patches.py
```diff
@@ -0,0 +1,34 @@
...
+@parser(Specs.kpatch_patch_files)
+class KpatchPatches(CommandParser):
+    """
+    A parser for getting modules names of locally stored kpatch-patch files
+    """
+
+    def __init__(self, context):
+        try:
+            super(type(self), self).__init__(context)
+        except ContentException:
```
xiangce: Thanks for updating this. Since the `__init__` of the superclasses will be called implicitly when it's omitted in the sub-class. It's OK to remove it here.

And `ContentException` looks more suitable here than the `SkipException`, since the `SkipException` will be discarded directly by the core, however, the `ContentException` is track-able. It would be useful for insights to do further check about e.g. if there are CUs using the "kpatch" in an abnormal way? (e.g. we cannot find the regular patches even the kpatch is installed)

How do you think?

theyoyojo: Sure, changed.

PR #2352, opened 2019-12-02
New parser for max_uid

jobselko:


skontar:
@xiangce Actually, we are already collecting this command and we have this parser in our private repository working just fine. @jobselko is moving our private parsers to insights-core where applicable.

Also the command may look complex, but from the shell standpoint it is very simple. Everything what looks complex is actually just a one parameter for awk.

xiangce:
@skontar - Cool, thanks for the explanation.
It looks good to me

PR #2355, opened 2019-12-03
Enhance hammer ping

huali027:
* Sometimes there are only two lines for some service


bfahr:
@huali027 is there an issue that makes this change necessary?

huali027:
> @huali027 is there an issue that makes this change necessary?

Yes, thanks for your reminder. I created an issue in insights core, and submit it again in a new PR: https://github.com/RedHatInsights/insights-core/pull/2360, I'll close this one first.

Issue #2369, opened 2019-12-06
Fix FreeIPAHealthCheckLog so that Healthcheck 0.3+ logs can be parsed


fcami:
FreeIPA Healthcheck JSON schema changed from 0.2 to 0.3: the 'severity' key was replaced by 'result'. 

The parser's "errors" method should also be renamed to avoid confusion with the "ERROR" possible result.

xiangce:
Resolved by #2370

PR #2370, opened 2019-12-06
Adapt FreeIPAHealthCheckLog parser for Healthcheck 0.3 (#2369)

fcami:
FreeIPA Healthcheck JSON schema changed from 0.2 to 0.3:
the 'severity' key was replaced by 'result'.

### DIFF
--- a/insights/parsers/freeipa_healthcheck_log.py
+++ b/insights/parsers/freeipa_healthcheck_log.py
@@ -11,84 +11,52 @@ Typical content of file ``/var/log/ipa/healthcheck/healthcheck.log`` is::
 
     [
       {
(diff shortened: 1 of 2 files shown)


csams:
Hey @fcami will the parser no longer handle the older version after this change? We have lots of rule repositories now, so we can't make breaking changes. Can you adapt the parser to handle both?

fcami:
hi @csams Thanks for the feedback!
As far as I can tell no rule was done for Healthcheck 0.2 which was never shipped in RHEL.
Healthcheck 0.3 is shipping in RHEL 8.1 and this is the first RHEL version to include the package.
If you'd like I can look at adapting the code but given the above I would rather not. It's your call.

PR #2371, opened 2019-12-09
Remove unicode in repos

SteveHNH:
https://projects.engineering.redhat.com/browse/RHCLOUD-3610

When ingesting data into Insights Inventory, the process fails if unicode is found in the repos. We're seeing this hit on a variety of accounts. I went ahead and added a function to remove it in this particular parser, but it's possible it may be needed elsewhere. This is currently the only place I'm aware that we're seeing it.


SteveHNH:
nevermind on this one. Wrong approach

insights/parsers/tests/test_yum_repos_d.py
```diff
@@ -18,6 +19,12 @@
 gpgcheck=1
     0 # This should be ignored
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-beta,file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
+
+[bad-unicode]
+baseurl = https://example.com
+name = insights-test-💩
```
Glutexo: I am afraid this test can’t actually work, because you are reading a Unicode string here, not a set of bytes.

PR #2374, opened 2019-12-12
Keep original content in case we want to output the whole content

huali027:
- fix: #2375 

### DIFF
--- a/insights/parsers/hammer_ping.py
+++ b/insights/parsers/hammer_ping.py
@@ -73,7 +73,7 @@ class HammerPing(CommandParser, dict):
         self.status_of_service = {}
         self.response_of_service = {}
         self.errors = []
-
+        self.raw_content = content
         content = get_active_lines(content, comment_char="COMMAND>")
         service_name = None
         for line in content:
--- a/insights/parsers/hammer_ping.py
+++ b/insights/parsers/hammer_ping.py
@@ -45,6 +45,7 @@ class HammerPing(CommandParser, dict):
 
     Attributes:
         errors (list): Any error messages encountered during parsing
+        raw_content(list): The original output of hammer ping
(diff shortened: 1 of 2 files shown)


insights/parsers/hammer_ping.py
```diff
@@ -73,7 +73,7 @@ def parse_content(self, content):
         self.status_of_service = {}
         self.response_of_service = {}
         self.errors = []
-
+        self.raw_content = content
```
xiangce: Please add the corresponding docstring for the new attribute.

huali027:
> @huali027 - Which issue does this PR fix?

do we need to create an issue in insights-core for every PR?

insights/parsers/hammer_ping.py
huali027: > Please add the corresponding docstring for the new attribute.

sure, I'll add it in next commit

xiangce:
> > @huali027 - Which issue does this PR fix?
> 
> do we need to create an issue in insights-core for every PR?

For bugfix, enhancement, Yes. Not mandatory for new parsers, but it's recommended.

huali027:
> > > @huali027 - Which issue does this PR fix?
> > 
> > 
> > do we need to create an issue in insights-core for every PR?
> 
> For bugfix, enhancement, Yes. Not mandatory for new parsers, but it's recommended.

sure, I'll create one, thanks.

huali027:
> Approved pending an issue for tracking

I have created an issue https://github.com/RedHatInsights/insights-core/issues/2375,
and I added fixes in the second commit

xiangce:
@huali027 - Do not paste the issue number to git commit, since no link-ship will be created in that way. and reviewers cannot get anything.

PR #2382, opened 2019-12-19
Add sos archive path for FreeIPA Healthcheck log

fcami:
sos archives now collect the FreeIPA Healthcheck log.
Add it as a data source.


fcami:
hi @xiangce thanks for the explanation, this is clear. Closing.

PR #2391, opened 2020-01-03
New parser for "/usr/lib/udev/rules.d/59-fc-wwpn-id.rules"

huali027:


huali027:
discussed with @xiangce, I'll move the logic to the rule since we just check if the file syntax is valid.

huali027:
@xiangce, @wushiqinlou, please help to review this PR when u have time, thanks.

PR #2392, opened 2020-01-03
enhacement cpuinfo apicid

ghost:
issue: https://github.com/RedHatInsights/insights-core/issues/2393
Signed-off-by: Rahul Srivastava rasrivas@redhat.com


insights/parsers/cpuinfo.py
```diff
@@ -50,6 +52,8 @@
     >>> cpu_info = shared[CpuInfo]
     >>> cpu_info.cpu_count
     2
+    >>> cpu_info.apicid
+    ["2", "2"]
```
xiangce: The test in the `Examples` should be a reflection of the above `Sample input:`

xiangce (changes requested):
@rasrivas-redhat - Which issue does this PR relate to? Please open an issue, if there is no exiting issue for it, for changelog tracking convenience.

BTW, please fix the test failures in python@2.7&2.6 and DCO failure.

Thanks.

insights/parsers/cpuinfo.py
rasrivas-redhat: @xiangce - I have fixed the issue as per your suggestion, please review

xiangce (approved):

PR #2400, opened 2020-01-08
Cmd kpatch list

shlao:
Added
insights/parsers/kpatch_list.py
insights/parsers/tests/test_kpatch_list.py
docs/shared_parsers_catalog/kpatch_list.rst

Updated
insights/specs/init.py
insights/specs/default.py


insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,71 @@
...
+        for line in content:
+            if 'Loaded patch modules' in line:
+                cur_dict = self._loaded
+                continue
+
+            if 'Installed patch modules' in line:
+                cur_dict = self._installed
+                continue
+
+            if '[' in line or ']' in line:
```
xiangce: Since you have distinguished the `loaded` and `installed` in the above two if-clauses, it's no need to do the following checks.

And the code could be simplified as:

~~~
try:
  kpatch, kernel = [k.strip('()[]') for k in line.split()]
  cur_dict[kpatch] = kernel
catch Exception:
 raise ParseException('xx_msg')
~~~

insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,71 @@
...
+            if len(fields) != 2:
+                # invalid line
+                continue
+
+            cur_dict[fields[0]] = fields[1]
+
+    @property
+    def loaded(self):
+        """Returns the loaded kpatchs"""
+        return self._loaded
```
xiangce: `return next(iter(self._loaded))`

insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,71 @@
...
+            fields = line.split(' ')
+            if len(fields) != 2:
+                # invalid line
+                continue
+
+            cur_dict[fields[0]] = fields[1]
+
+    @property
+    def loaded(self):
+        """Returns the loaded kpatchs"""
```
xiangce: Please present the type (str) of the return value

insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,71 @@
...
+        return self._loaded
+
+    @property
+    def installed(self):
+        """Returns the installed kpatchs"""
```
xiangce: Same comment as above but `dict` here

insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@parser(Specs.kpatch_list)
+class KpatchList(CommandParser):
+    """Class for command: /usr/sbin/kpatch list"""
+    def parse_content(self, content):
+        if content is None or len(content) == 0:
+            raise SkipException("No Data from command: /usr/sbin/kpatch list")
+
+        self._loaded = {}
```
xiangce: The loaded can be a single string.

Is it possible that multiple kpatches are loaded at the same time?

insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,71 @@
...
+            if '(' in line or ')' in line:
+                line = line.replace('(', '').replace(')', '')
+
+            fields = line.split(' ')
+            if len(fields) != 2:
+                # invalid line
+                continue
+
+            cur_dict[fields[0]] = fields[1]
+
```
xiangce: Better `SkipException` it when `installed` is empty.

insights/parsers/tests/test_kpatch_list.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_kpatch_list():
+    kpatchs = kpatch_list.KpatchList(context_wrap(NORMAL_OUTPUT))
+    assert len(kpatchs.loaded) > 0
+    assert len(kpatchs.installed) > 0
+    assert kpatchs.loaded.get('kpatch_3_10_0_1062_1_1_1_4') == "enabled"
+    assert kpatchs.installed.get('kpatch_3_10_0_1062_1_1_1_4') == "3.10.0-1062.1.1.el7.x86_64"
+
+    kpatchs = kpatch_list.KpatchList(context_wrap(EMPTY_OUTPUT))
```
xiangce: How about skipping it in this case?

insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,71 @@
...
+from insights.specs import Specs
+from insights.parsers import SkipException
+from insights import parser, CommandParser
+
+
+@parser(Specs.kpatch_list)
+class KpatchList(CommandParser):
+    """Class for command: /usr/sbin/kpatch list"""
+    def parse_content(self, content):
+        if content is None or len(content) == 0:
```
xiangce: The more pythonic way is `if not content`:

insights/parsers/kpatch_list.py
shlao: In this [patch](https://www.redhat.com/archives/kpatch/2014-July/msg00003.html), it says:
"One or more patch modules may be configured to load at boot time ... ..."

insights/parsers/kpatch_list.py
shlao: In article [`Applying patches with kernel live patching`](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/kernel_administration_guide/applying_patches_with_kernel_live_patching), there is a example:
```# kpatch uninstall kpatch_3_10_0_1062_1_1
uninstalling kpatch_3_10_0_1062_1_1 (3.10.0-1062.el7.x86_64)

# kpatch list
Loaded patch modules:
kpatch_3_10_0_1062_1_1 [enabled]

Installed patch modules:
<NO_RESULT>
```
That means the `Installed` section maybe empty.

xiangce: Then, you should skip the parser when both of them are empty.

insights/parsers/kpatch_list.py
xiangce: You'd better add a test case for more than patches are loaded. To make sure the current logic could cover those cases.

Then as per the actual cases, to decide the type of the `loaded` (list or dict), I mean when all the 2nd columns are the same for `loaded` kpatches, it's not necessary to return them.

insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,67 @@
...
+
+            if 'Loaded patch modules' in line:
+                cur_dict = self._loaded
+                continue
+
+            if 'Installed patch modules' in line:
+                cur_dict = self._installed
+                continue
+
+            kpatch, info = [k.strip('()[]') for k in line.split()]
```
xiangce: I have got the following output of `kpatch list`

~~~
# kpatch list
Loaded patch modules:
kpatch_7_0_1_el7
Installed patch modules:
kpatch-7-0-1-el7.ko (3.10.0-121.el7.x86_64)
kpatch-7-0-1-el7.ko (3.10.0-123.el7.x86_64)
~~~

This line will fail for the loaded part.

insights/parsers/kpatch_list.py
```diff
@@ -0,0 +1,67 @@
...
+            if not info:
+                raise ParseException("Parser Error: Invalid Line in content of 'kpatch list'")
+
+            cur_dict[kpatch] = info
+
+    @property
+    def loaded(self):
+        """``dict`` the loaded kpatchs"""
```
xiangce: The type of return value is not in this format.

Please check the examples in other parsers:

https://github.com/RedHatInsights/insights-core/blob/f58f26eb08bbea38fd4899e6b8a7eae056522c9f/insights/parsers/blkid.py#L76

or take a look at:

https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

insights/parsers/kpatch_list.py
shlao: I patch two kmod and get the following ouput:
```
# kpatch list
Loaded patch modules:
livepatch_other_path [enabled]
livepatch_test [enabled]

Installed patch modules:
```

shlao:
The ci is Pending.

PR #2401, opened 2020-01-08
Add --version flag to the Insights CLI

### LINKED ISSUE #2408: Add version command/flag to the Insights CLI
Add `version` command and `--version` flag/arg to print version info from `insights.get_nvr()`.

vdymna:
* Added `version` command and `--version` argument, `ArgumentParser.add_argument` `version` action was not used because of a bug in earlier versions of python.
* The version info can be retrieved using `insights version`, `insights --version` or `insights <command> --version`.
* Added `insights/tests/test_command_parser.py` with couple basic unit tests.

Resolves #2408

Signed-off-by: Vitaliy Dymna <vdymna@redhat.com>

### DIFF
--- a/insights/command_parser.py
+++ b/insights/command_parser.py
@@ -11,8 +11,6 @@ from __future__ import print_function
 import argparse
 import sys
 
-from insights import get_nvr
-
 USAGE = """insights <command> [<args>]
 Available commands:
   cat         Execute a spec and show the output
@@ -21,6 +19,7 @@ Available commands:
   info        View info and docs for Insights Core components.
   ocpshell         Interactive evaluation of archives, directories, or individual yaml files.
   run         Run insights-core against host or an archive.
+  version     Show Insights Core version information and exit.
 """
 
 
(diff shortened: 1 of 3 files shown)


vdymna:
@csams @lhuett I forgot to add reviewers to this PR (and for some reason I can't add them after the PR was created). But when you get a chance will you please look at this.

patchkez:
test this please

vdymna:
I've tested this change using python 3.7.5, 3.6.9, 2.7.17 and 2.6.9 on Fedora 31. If you think I should test this on different versions of RHEL, I can do that as well?

patchkez:
@vdymna I apologize for confusion, string above triggers our pipeline which starts different insights client tests on RHEL6/7/8.

vdymna:
ok, I see that now :smiley:

vdymna:
test this please

vdymna:
Accidentally deleted my branch, reopening the PR.

insights/command_parser.py
```diff
@@ -46,6 +53,9 @@ def __init__(self):
         # Use dispatch pattern to execute command
         getattr(self, args.command)()
 
+    def parse_version_arg(self):
```
csams: Can you help me understand the manual check in the arg list here? I see you add `--version` to the arg parser at line 40. Can you check `parser.version` after the args have been parsed?

vdymna: Yes. If `parser.parse_args()` is invoked when only `--version` flag is provided (`insights --run`) then ArgumentParser will throw an exception stating that required positional argument `command` is missing. Additionally, for version info display use cases, `ArgumentParser.add_argument` call provides `version` action which will parse everything correctly; however, older version of python have a bug where version info is printed into `stderr`. So for python >= 3.4, the info goes to stdout and for older python it goes to `stderr`. For those 2 reasons I ended up with manual check for `--version`. Does that make sense?

vdymna: And for the most part, the reason I kept `parser.add_argument('--version'...` is because it will generate correct help for the flag.

vdymna:
Added `version` command. Now can use `insights version` to print version info.

PR #2406, opened 2020-01-13
Add specs to sos_archive

wushiqinlou:
fix: #2407


wushiqinlou:
https://github.com/RedHatInsights/insights-core/issues/2407

PR #2414, opened 2020-01-17
Enable client to retrieve advisor report

subpop:
Part of an effort to integrate Insights into Cockpit involves Cockpit knowing more details about the latest Advisor report. To that end, this PR is added a new command line flag `--show` that does not run through the collection and upload process; instead it gets the latest report as compiled by Advisor and returns the JSON. Responses are cached by ETag (however the certificate-based authentication does not appear to respond with `304` status codes).

There are two well-known files that can be monitored for changes. Both files contain unmodified responses from their respective backend API calls. Both files are updated within a few seconds of each other (both files get updated within the same function call, so realistically, they should be updated at "the same time").

#### /var/lib/insights/host-details.json
`GET /inventory/v1/hosts?insights_id=$(cat /etc/insights-client/machine-
id)`

This file contains a host ID at the `jq` path ".results[0].id". This host ID is used in the subsequent API call below, as well as being the ID to build a permalink to the host's record in inventory at cloud.redhat.com.

#### /var/lib/insights/insights-details.json
`GET /insights/v1/systems/$HOST_ID/reports`

The file contents can either be read directly or can be read by calling `insights-client --show` (this command simply dumps the file contents to stdout).

There is a second flag (and related functions within the insights-client package) that triggers an update (`--check-results`). My plan is to make that opaque to external callers (probably by putting it on a systemd timer or unit that is called a few minutes after the collection unit); we will update the files internally. So there should be no need for external programs to call `--check-results` explicitly.

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -1023,11 +1023,19 @@ class InsightsConnection(object):
         if content is None:
             return None
 
+        os.makedirs("/var/lib/insights", mode=0o755, exist_ok=True)
+
+        with open("/var/lib/insights/%s.json" % generate_machine_id(), mode="w") as f:
+            f.write(content)
+
         host_id = json.loads(content)["results"][0]["id"]
         url = self.base_url + "/insights/v1/system/%s/reports/" % host_id
         content = self._get(url)
         if content is None:
             return None
+
+        with open("/var/lib/insights/%s.json" % host_id, mode="w+b") as f:
+            f.write(content)
 
         return json.loads(content)
 
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -1026,7 +1026,7 @@ class InsightsConnection(object):
 
         host_details = json.loads(content)
         if host_details["total"] < 1:
-            raise Exception("Error: no host detected (insights_id = %s): Run insights-client --status to check registration status" % generate_machine_id())
+            raise Exception("Error: failed to find host with matching machine-id. Run insights-client --status to check registration status")
         if host_details["total"] > 1:
             raise Exception("Error: multiple hosts detected (insights_id = %s)" % generate_machine_id())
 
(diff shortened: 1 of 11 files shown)


insights/client/phase/v1.py
```diff
@@ -131,6 +131,23 @@ def post_update(client, config):
...
+        try:
+            client.show()
+            sys.exit(constants.sig_kill_ok)
+        except Exception as e:
+            print(e)
+            sys.exit(constants.sig_kill_bad)
+
+    if config.check_results:
+        try:
+            client.config.base_url += "/platform"
```
gravitypriest: Instead of concatenating `/platform` here, I would suggest using the `show` and `check_results` options to set legacy_upload = False in `InsightsConfig.imply_options`. That way it'll get set in auto_config.try_auto_configuration -- if someone were to set `legacy_upload=False` manually in the config, `/platform` would get added twice.

insights/client/url_cache.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self, path=None):
+        '''
+            Initialize a URLCache, loading entries from @path, if provided.
+        '''
+        self._path = path
+        self._cache = {}
+        if os.path.isfile(self._path):
+            with open(self._path, "r+b") as f:
+                self._cache = pickle.load(f)
+        os.makedirs(os.path.dirname(self._path), exist_ok=True)
```
gravitypriest: It looks like `exist_ok` doesn't work in Python 2, unfortunately

insights/client/connection.py
```diff
@@ -978,3 +979,58 @@ def get_diagnosis(self, remediation_id=None):
...
+    def get_advisor_report(self):
+        '''
+            Retrieve advisor report
+        '''
+        url = self.base_url + "/inventory/v1/hosts?insights_id=%s" % generate_machine_id()
+        content = self._get(url)
+        if content is None:
+            return None
+
+        os.makedirs("/var/lib/insights", mode=0o755, exist_ok=True)
```
gravitypriest: It looks like `exist_ok` doesn't work in Python 2, unfortunately

Issue #2409, opened 2020-01-16
Command spec readlink_e_etc_mtab is missing path for command


bfahr:
The path to the command is missing from the spec `readlink_e_etc_mtab` that was added in PR #2373.  Since the command is run as root when the data is collected, the path needs to be included in the command.

xiangce:
@bfahr  - I believe this issue and #2399 are duplicates. And the #2402 resolved it.

xhe123:
@bfahr I fixed the path missing issue in simple_command() in https://github.com/RedHatInsights/insights-core/pull/2418

bfahr:
Fixed

PR #2418, opened 2020-01-21
add path to simple_command for readlink

xhe123:
Add path of "/usr/bin/" to simple_command for readlink

Signed-off-by: Xiaonan He <xhe@redhat.com>


xhe123:
this MR fixs issue https://github.com/RedHatInsights/insights-core/issues/2409

PR #2419, opened 2020-01-21
Validate machine ID

SteveHNH:
This PR checks to make sure that the client's machine_id is a valid UUID before making any calls to the API.

https://projects.engineering.redhat.com/browse/RHCLOUD-3962


SteveHNH:
@subpop 
I can write a unit test, I'm sure. I'll give it a go. That's a good idea

PR #2420, opened 2020-01-22
Try to load CSafeLoader and gracefully degrade

### LINKED ISSUE #2271: Update YAMLParser to use libyaml's CSafeLoader if it's available
The [YAMLParser](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/__init__.py#L685) currently uses the pure python SafeLoader of pyyaml. It'll load files about 10 times faster if libyaml is installed and we use `yaml.CSafeLoader` instead. See the [ocpshell code](https://github.com/RedHatInsights/insights-core/blob/master/insights/ocpshell.py#L11) for an example.

subpop:
Rather than always relying on the pure Python SafeLoader, try importing
the CSafeLoader locally, handling the ImportError by falling back to the
Pure Python SafeLoader.

Fixes #2271

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -689,10 +689,15 @@ class YAMLParser(Parser, LegacyItemAccess):
     """
     def parse_content(self, content):
         try:
+            try:
+                from yaml import CSafeLoader as SafeLoader
+            except ImportError:
+                from yaml import SafeLoader
+            
             if type(content) is list:
-                self.data = yaml.safe_load('\n'.join(content))
+                self.data = yaml.load('\n'.join(content), Loader=SafeLoader)
             else:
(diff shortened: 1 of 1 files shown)


insights/core/__init__.py
```diff
@@ -689,10 +689,15 @@ class YAMLParser(Parser, LegacyItemAccess):
     """
     def parse_content(self, content):
         try:
+            try:
```
csams: If the import in the try body fails, an `ImportError` will need to raise every time through parse_content to get the python loader out of the module cache in the except block. Let's move the try/except of the import to the top of the module so we only pay the cost once.

bfahr:
please test

patchkez:
test this please

PR #2421, opened 2020-01-22
Fix a placeholder VERSION of 3.0.9-dev

subpop:
In QE, there is a need to differentiate eggs built from branches and development trees from eggs built as released software. Having the version information inside the module is useful, but development versions end up with a fixed version string of "3.0.8". This PR makes identifying eggs and archives built from branches easier to differentiate from eggs built from tagged releases by setting the version to "3.0.999".


bfahr:
@subpop I can understand the need for the change, but I'm not sure this is the best value to use.  @csams @SteveHNH would it make sense to use 3.1.n for master branch, or 3.0.100n to ensure that there is a predictable pattern for changes in the master/development branch versus the 3.0/production branch?

subpop:
I think any predictable, discernible value would be sufficient to meet QE's needs. If you're following semantic versioning, you could use a prerelease identifier after the x.y.z version.

https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions

bfahr:
@subpop that seems like a reasonable approach since the master version does occasionally change.  So maybe `-dev` as a prerelease tag like `3.0.9-dev`?

subpop:
@bfahr I checked with our QE team, and it sounds like having "-dev" as a prerelease string would work. That way you could keep the x.y.z version tracking whatever scheme makes sense for the project.

bfahr:
@subpop apparently the `-dev` is breaking some of the CI tests.  I'll check with the QE team and see if its really a problem, or if the tests can be changed.

patchkez:
test this please

PR #2427, opened 2020-01-23
Correct the check for remove file validation

SteveHNH:
We need to properly check for the two different ways the mode is
represented for the remove.conf file. There is a difference between
python3 and python2, so we'll check for both


SteveHNH:
closing to fix a handful of issues I discoverd

PR #2428, opened 2020-01-24
Parsers for Simultaneous multithreading

jobselko:
These parsers are already in use in our private repository.

### DIFF
--- a/insights/parsers/smt.py
+++ b/insights/parsers/smt.py
@@ -1,0 +1,117 @@
+"""
+Simultaneous Multithreading (SMT) parsers
+=========================================
+
+Parsers included in this module are:
+
+CpuSMTActive - file ``/sys/devices/system/cpu/smt/active``
+----------------------------------------------------------
+CpuCoreOnline - files matching ``/sys/devices/system/cpu/cpu[0-9]*/online``
+---------------------------------------------------------------------------
+CpuSiblings - files matching ``/sys/devices/system/cpu/cpu[0-9]*/topology/thread_siblings_list``
+------------------------------------------------------------------------------------------------
+"""
+
+from insights.core import Parser
+from insights.core.plugins import parser
+from insights.parsers import SkipException
+from insights.specs import Specs
+import re
+
+
+@parser(Specs.cpu_smt_active)
+class CpuSMTActive(Parser):
+    """
+    Class for parsing ``/sys/devices/system/cpu/smt/active`` file.
+    Reports whether SMT is enabled and active.
+
+    Typical output of this command is::
+
+        1
+
+    Raises:
+        SkipException: When content is empty or cannot be parsed.
+
+    Examples:
+        >>> cpu_smt.on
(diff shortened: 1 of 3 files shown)


insights/parsers/smt.py
```diff
@@ -0,0 +1,117 @@
...
+
+@parser(Specs.cpu_cores)
+class CpuCoreOnline(Parser):
+    """
+    Class for parsing ``/sys/devices/system/cpu/cpu[0-9]*/online`` matching files.
+    Reports whether a CPU core is online.
+
+    Typical output of this command is::
+
+        1
```
skontar: I think that odd number of of cores is very unusual. Also should probably say something about "file content" rather than "output of command."

insights/parsers/smt.py
```diff
@@ -0,0 +1,117 @@
...
+        [0, 2]
+    """
+    cpu_siblings_path = r'/sys/devices/system/cpu/cpu(\d+)/topology/thread_siblings_list'
+
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("No content.")
+
+        # The separator in the sibling list may be either in the format 0-1 or 0,2 depending on the CPU model
+        if "-" in content[0]:
```
skontar: Does this work correctly? Does not 0-2 actually mean 0, 1, and 2?

jobselko: I'm not quite sure but I think it works that way. See https://access.redhat.com/solutions/rhel-smt.

insights/parsers/smt.py
jobselko: The number of lines on output depends on the number of "online" files. I understand that cpu0 is always online, so it does not have the "online" file.
But I agree that the docstring can be misleading.

skontar: I see. May be worth adding a note.

insights/parsers/smt.py
skontar: I think that we may be lucky that it works because `-` separator was always used in the article or the tests like `0-1` or `2-3`, not for example `0-3`. @kgrant-rh any knowledge here?

skontar: Based on the following I think that `-` is really supposed to be a range.
```
$ cat /sys/devices/system/cpu/cpu*/topology/core_siblings_list 
0-3
0-3
0-3
0-3
$ cat /sys/devices/system/cpu/cpu*/topology/thread_siblings_list 
0,2
1,3
0,2
1,3
```

kgrant-rh: Yes, I believe you're correct, `-` is inclusive range, `,` lists individual items. I can't recall where in the source that behavior is specified, but I'll see if I can find it.

kgrant-rh: Just for the record, this behavior is handled the same way as other bitmaps, as specified in /linux/lib/bitmap.c

bfahr:
please test

PR #2430, opened 2020-01-24
Extend insights.core.filters.add_filter() functionality

### LINKED ISSUE #2429: Extend insights.core.filters.add_filter() functionality
Extend the functionality of `insights.core.filters.add_filter()` to allow developers to filters underlying datasources by adding filter(s) to a component like parser or combiner. When a filter is added to a non-datasource component, the function logic needs to locate filterable datasources in the dependency graph and correctly filter those.

vdymna:
* Modified `add_filter` so non-datasouce component can be passed into the function.
* Added relevant unit tests to `test_filters.py`.
* Modified `_intercept_add_filter` function to correctly test for missing filters
  during integration tests, based on changed to `add_filter`.
* Added `test_integration_support.py` with unit tests for integation testing
  support functions.
* Added `plugins/ps_rule_fakes.py` with dummy rules that areused in `test_integration_support.py`
  unit tests.
* Modified `Insights API` docs page and `Filtering of Data in Insights Parsers and Rules`
  notebook (added new paragraph under **Adding Filters to Rules** section) to include information about the non-datasource component level filtering.

Resolves #2429

Signed-off-by: Vitaliy Dymna <vdymna@redhat.com>

### DIFF
--- a/insights/core/filters.py
+++ b/insights/core/filters.py
@@ -1,9 +1,14 @@
 """
-The filters module allows developers to apply filters to datasources. A filter
-is a simple string, and it matches if it is contained anywhere within a line.
+The filters module allows developers to apply filters to datasources,
+by adding them directly or through dependent components like parsers
+and combiners. A filter is a simple string, and it matches if it is contained
+anywhere within a line.
 
 If a datasource has filters defined, it will return only lines matching at
 least one of them. If a datasource has no filters, it will return all lines.
+
+Filters can be added to components like parsers and combiners, to apply consistent
+filtering to multiple underlying datasources that are configured as filterable.
 
 Filters aren't applicable to "raw" datasources, which are created with
 ``kind=RawFileProvider`` and have ``RegistryPoint`` instances with ``raw=True``.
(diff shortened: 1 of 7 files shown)


insights/tests/__init__.py
```diff
@@ -32,10 +32,10 @@
 
 def _intercept_add_filter(func):
     @wraps(func)
-    def inner(ds, pattern):
-        ret = add_filter(ds, pattern)
+    def inner(component, pattern):
```
vdymna: I've changed `ds` to `component` here, if you think that might be an issue I can change that? But after looking around some, I didn't see any explicit references to `ds`.

insights/core/filters.py
```diff
@@ -46,37 +51,49 @@
 ENABLED = parse_bool(os.environ.get("INSIGHTS_FILTERS_ENABLED"), default=True)
 
 
-def add_filter(ds, patterns):
+def add_filter(component, patterns):
```
vdymna: I've changed `ds` to `component` here, if you think that might be an issue I can change that? But after looking around some, I didn't see any explicit references to `ds`.

insights/tests/__init__.py
```diff
@@ -56,14 +56,23 @@ def inner(ds, pattern):
 spec_factory.find = _intercept_find(spec_factory.find)
 
 
-def _get_registry_points(component):
+def _get_registry_points(component, filterable_only=False):
```
csams: Let's have this function just deal with returning registry points and filter down to just the filterable ones at the point where that's needed.

insights/tests/__init__.py
```diff
@@ -106,8 +115,7 @@ def run_test(component, input_data, expected=None):
     if filters.ENABLED:
         mod = component.__module__
         sup_mod = '.'.join(mod.split('.')[:-1])
-        rps = _get_registry_points(component)
-        filterable = set(d for d in rps if dr.get_delegate(d).filterable)
+        filterable = _get_registry_points(component, True)
```
csams: Let's revert this so that filtering to just the  filterable registry points happens here. It lets `_get_registry_points` have one job and makes it more explicit at this site that we only want the filterable ones. `get_registry_points(thing, True)` doesn't tell us much, and once we understand it, it raises the question of why there isn't an option for `raw` as well.

vdymna: Yes, this makes sense. I will change it back.

PR #2437, opened 2020-01-29
copy to destination dir/file after writing to tmp

gravitypriest:


gravitypriest:
TODO: handle shutil.copy* exception when the destination already exists

cmoore-redhat:
@subpop

insights/client/config.py
```diff
@@ -668,6 +670,8 @@ def _validate_options(self):
             parent_dir = os.path.dirname(self.output_file.rstrip('/'))
             if not os.path.exists(parent_dir):
                 raise ValueError('Cannot write to %s. Parent directory %s does not exist.' % (self.output_file, parent_dir))
+            if not os.path.isdir(parent_dir):
+                raise ValueError('Cannot write to %s. %s is not a directory.' % (self.output_dir, parent_dir))
```
fstavela:
```suggestion
                raise ValueError('Cannot write to %s. %s is not a directory.' % (self.output_file, parent_dir))
```

PR #2442, opened 2020-02-03
Fix issue with curl command stats in aws data

### LINKED ISSUE #2441: Curl command is returning stats in data, command should use --silent switch
The curl commands for specs `aws_instance_id_doc`, `aws_instance_id_pkcs7` and `aws_instance_type` are returning the curl stats in the data.  The commands should use the `-s` switch to prevent this data from being output.

```
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   483  100   483    0     0   607k      0 --:--:-- --:--:-- --:--:--  471k
{
    "devpayProductCodes" : null,
    ...
}
```

bfahr:
* Fix #2441 by adding -s switch to aws command specs
* Fix parsers to check for stats in output
* Update tests

### DIFF
--- a/insights/parsers/aws_instance_id.py
+++ b/insights/parsers/aws_instance_id.py
@@ -5,8 +5,8 @@ AWSInstanceID
 These parsers read the output of commands to collect identify information
 from AWS instances.
 
-* ``curl http://169.254.169.254/latest/meta-data/instance-identity/document`` and
-* ``curl http://169.254.169.254/latest/meta-data/instance-identity/pkcs7``
+* ``curl -s http://169.254.169.254/latest/meta-data/instance-identity/document`` and
+* ``curl -s http://169.254.169.254/latest/meta-data/instance-identity/pkcs7``
 """
 from __future__ import print_function
 import json
@@ -21,7 +21,7 @@ class AWSInstanceIdDoc(CommandParser, dict):
     """
     Class for parsing the AWS Instance Identity Document returned by the command::
 
-        curl http://169.254.169.254/latest/meta-data/instance-identity/document
+        curl -s http://169.254.169.254/latest/meta-data/instance-identity/document
 
     Typical output of this command is::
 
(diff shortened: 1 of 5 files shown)


insights/parsers/aws_instance_id.py
```diff
@@ -5,8 +5,8 @@
 These parsers read the output of commands to collect identify information
 from AWS instances.
 
-* ``curl http://169.254.169.254/latest/meta-data/instance-identity/document`` and
-* ``curl http://169.254.169.254/latest/meta-data/instance-identity/pkcs7``
+* ``curl -s http://169.254.169.254/latest/meta-data/instance-identity/document`` and
+* ``curl -s http://169.254.169.254/latest/meta-data/instance-identity/pkcs7``
```
vdymna: I think the URL in this examples and the one below (line 24) should be `http://169.254.169.254/latest/dynamic/instance-identity/*`

insights/parsers/aws_instance_type.py
```diff
@@ -39,16 +39,20 @@ class AWSInstanceType(CommandParser):
     """
```
vdymna: Also, I guess the example on the line 20 should include the `-s` option.

insights/parsers/aws_instance_id.py
bfahr: The specs are retrieving the specific items, `document` and `pkcs7`, so that each one can be parsed separately.

insights/parsers/aws_instance_type.py
bfahr: Yeah, good catch, I'll fix that.

insights/parsers/aws_instance_id.py
vdymna: Well, that I mean is that those 2 links are incorrect. It should be: 
* `curl -s http://169.254.169.254/latest/dynamic/instance-identity/document` and
* `curl -s http://169.254.169.254/latest/dynamic/instance-identity/pkcs7`

You can just compare those two to the actual simple_commands in `insights/specs/default.py`
Hopefully, that make sense?

bfahr: Yes, that makes sense, thanks for the feedback.

PR #2445, opened 2020-02-05
Aws feature tweaks

gravitypriest:
Move the cert_verify=True out of connection.py so it's only set on the AWS and Hydra connections.


gravitypriest:
test this please

PR #2446, opened 2020-02-05
Fix profiles API endpoint to get profiles only for this host

akofink:
CC: @dLobatog 


gravitypriest:
@dLobatog yeah feel free to add that in. It's an unfortunate side effect of a customer RFE.

gravitypriest:
For what it's worth, I would just add `foo.redhat.com` just a blanket CYA in case other subdomains are ever needed.

PR #2449, opened 2020-02-11
Add parser for "/etc/firewalld/firewalld.conf"

huali027:


insights/parsers/firewall_config.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@parser(Specs.firewalld_conf)
+class FirewallDConf(Parser, dict):
+    """
+    Class for parsing ``/etc/firewalld/firewalld.conf`` file.
+    """
+
+    def parse_content(self, content):
+        self.update(split_kv_pairs(content, use_partition=False))
```
xiangce: Better add an `Empty check` at the end of this function and raise a `SkipException` for `empty content`, by doing so, there will be no parser result when the content is empty, which is good for component dependency check.

Issue #2454, opened 2020-02-13
"timestamp" returned by Syslog doesn't comply with the raw format in log in some cases


xiangce:
https://github.com/RedHatInsights/insights-core/blob/4025e9ec27539f3cd9f0d2f9a36d1b9909a6ebd2/insights/core/__init__.py#L1324

From guliu@redhat.com
~~~
MSGINFO = """
...
Feb  9 03:20:01 dev7u7 CROND[486]: (root) CMD (/usr/lib64/sa/sa1 1 1)
Feb  9 03:30:01 dev7u7 CROND[1707]: (root) CMD (/usr/lib64/sa/sa1 1 1)
...
"""

    def test_cron_log():
        msg_info = CronLog(context_wrap(MSGINFO))
        crond_list = msg_info.get('CROND')
        assert 6 == len(crond_list)
>       assert crond_list[0].get('timestamp') == "Feb  9 03:20:01"  <-<< 
E       AssertionError
~~~

In the original raw line, there are two spaces between the `Feb` and `9`. There is a reserved space for the timestamp when the day is earlier than 10 in most of "Syslog" like logs. But the parser re-formatted it and returns the timestamp as `Feb 9 03:20:01` (only one space between them).

PR #2455, opened 2020-02-13
Keep the timestamp returned by Syslog as it is

xiangce:
- Fix: #2454 
- Adjust the order of the log lines in test_syslog.py


xiangce:
@bfahr - Could you please help to take a look at this change, it's related to issue #2454. Thanks.

PR #2458, opened 2020-02-14
Add parser "satellite_mongodb_storage_engine"

huali027:


insights/parsers/tests/test_satellite_mongodb_storage_engine.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+import doctest
+from insights.tests import context_wrap
+from insights.parsers import SkipException
+from insights.parsers import satellite_mongodb_storage_engine
+
+
+MONGO_PULP_STORAGE_ENGINE_OUTPUT1 = '''
```
xiangce: I met this failure in my test:
~~~
~]# /usr/bin/mongo pulp_database --eval 'db.serverStatus().storageEngine'
MongoDB shell version v3.4.9
connecting to: mongodb://127.0.0.1:27017/pulp_database
2020-02-13T23:19:57.750-0500 W NETWORK  [thread1] Failed to connect to 127.0.0.1:27017, in(checking socket for error after poll), reason: Connection refused
2020-02-13T23:19:57.751-0500 E QUERY    [thread1] Error: couldn't connect to server 127.0.0.1:27017, connection attempt failed :
connect@src/mongo/shell/mongo.js:237:13
@(connect):1:6
exception: connect failed
~~~

Please confirm if it's necessary for the Parser to raise a ParserException or SkipException for it.

huali027:
> For the name `satellite_mongodb_storage_engine`, do you want to make a general name for all the MongoDB related commands/operations in Satellite? but I think the `__storage_engine` is too specific for the command in this PR. How about renaming it as `satellite_mongodb` or something else?

I'm afraid we'll add other MongoDB queries for satellite later, so I add storage_engine for it here.

insights/parsers/tests/test_satellite_mongodb_storage_engine.py
huali027: > I met this failure in my test:
> 
> ```
> ~]# /usr/bin/mongo pulp_database --eval 'db.serverStatus().storageEngine'
> MongoDB shell version v3.4.9
> connecting to: mongodb://127.0.0.1:27017/pulp_database
> 2020-02-13T23:19:57.750-0500 W NETWORK  [thread1] Failed to connect to 127.0.0.1:27017, in(checking socket for error after poll), reason: Connection refused
> 2020-02-13T23:19:57.751-0500 E QUERY    [thread1] Error: couldn't connect to server 127.0.0.1:27017, connection attempt failed :
> connect@src/mongo/shell/mongo.js:237:13
> @(connect):1:6
> exception: connect failed
> ```
> 
> Please confirm if it's necessary for the Parser to raise a ParserException or SkipException for it.

The satellite on your VM doesn't run, right? Since I just get "name" from the output, so in this case, it will fail to get the name, then raise SkipException. Let me add ParserException too if the name format isn't expected.

xiangce: Good

xiangce:
> > For the name `satellite_mongodb_storage_engine`, do you want to make a general name for all the MongoDB related commands/operations in Satellite? but I think the `__storage_engine` is too specific for the command in this PR. How about renaming it as `satellite_mongodb` or something else?
> 
> I'm afraid we'll add other MongoDB queries for satellite later, so I append storage_engine for it here.

That's ture, just like you did in another parser "satellite_installer_configurations.py ", you can make a generic name for Satellite MongoDB queries.

huali027:
> > > For the name `satellite_mongodb_storage_engine`, do you want to make a general name for all the MongoDB related commands/operations in Satellite? but I think the `__storage_engine` is too specific for the command in this PR. How about renaming it as `satellite_mongodb` or something else?
> > 
> > 
> > I'm afraid we'll add other MongoDB queries for satellite later, so I append storage_engine for it here.
> 
> That's ture, just like you did in another parser "satellite_installer_configurations.py ", you can make a generic name for Satellite MongoDB queries.

Sure, I got your point.

insights/parsers/satellite_mongodb_storage_engine.py
```diff
@@ -44,7 +44,10 @@ def parse_content(self, content):
         self.name = ''
         for line in content:
             if line.strip().startswith('"name"'):
-                self.name = line.split(':')[1].strip(',').strip().strip('"')
-                break
+                try:
```
xiangce: The following logic makes the usage of this parser very limit, since except for the "name", all other information is discarded.  For a parser, we'd better make is as useful as we can, I mean, for this one, we'd better parser all the items inner the "{}".

huali027: > discarded

Yes, I thought about it. But I don't think our rules will use all the other info. So I just parse the name out. What's your opinion? Do you think we'll use them?

xiangce: It depends. If the workload to parse extra info is heavy but it's obvious the extra info is useless, then we'd better skip them. But in this case, I don't see any difficulty to parse the reset lines other than the "name".  And,  it doesn't make sense that the rule you are working on doesn't use other lines indicate no rules will use them, right?

PR #2461, opened 2020-02-17
Parser for Cmd: `/usr/sbin/zdump -v /etc/localtime`

shlao:
Added
```
insights/parsers/zdump_v.py
insights/parsers/tests/test_zdump_v.py
docs/shared_parsers_catalog/zdump_v.rst
```

Updated
```
insights/specs/__init__.py
insights/specs/default.py
```


insights/parsers/zdump_v.py
```diff
@@ -0,0 +1,103 @@
...
+from datetime import datetime
+from insights.specs import Specs
+from insights.parsers import SkipException
+from insights import parser, CommandParser
+
+
+def str2datetime(timestamp):
+    # remove redundant white spaces
+    # fs stands for 'fields'
+    fs = [s for s in timestamp.split(' ') if len(s) > 0]
```
xiangce: This is not needed.

insights/parsers/zdump_v.py
```diff
@@ -0,0 +1,103 @@
...
+    if len(fs) != 6:
+        return
+
+    # Formatting the output
+    # Can't use `" ".join(fs)` for the Fixed width number of timestamp
+    time_string = "%s %s %2s %s %s %s" % (fs[0], fs[1], fs[2], fs[3], fs[4], fs[5])
+
+    # Fixed the problem that the program running this python code doesn't
+    # has the corresponding TimeZone where strptime will raise ValueError.
+    if 'UTC' in fs:
```
xiangce: `def str2datetime(timestamp, tz=False)`

and call it like `str2datetime(str, True) for `utc_time`

insights/parsers/zdump_v.py
```diff
@@ -0,0 +1,103 @@
...
+def str2datetime(timestamp):
+    # remove redundant white spaces
+    # fs stands for 'fields'
+    fs = [s for s in timestamp.split(' ') if len(s) > 0]
+    if len(fs) != 6:
+        return
+
+    # Formatting the output
+    # Can't use `" ".join(fs)` for the Fixed width number of timestamp
+    time_string = "%s %s %2s %s %s %s" % (fs[0], fs[1], fs[2], fs[3], fs[4], fs[5])
```
xiangce: Define it as `def str2datetime(timestamp, tz=False)`

Simplify the code as:

```python
if tz:
 time_f = "%a %b %d %H:%M:%S %Y"
 time_s = timestamp
else:
 time_f = "%a %b %d %H:%M:%S"
 time_s = timestamp.rsplit(None, 1)[0]

try:
        time = datetime.strptime(time_s, time_f)
except ValueError:
        pass
```

and call it like `str2datetime(str, True) for `utc_time`

insights/parsers/zdump_v.py
```diff
@@ -0,0 +1,103 @@
...
+from insights.parsers import SkipException
+from insights import parser, CommandParser
+
+
+def str2datetime(timestamp):
+    # remove redundant white spaces
+    # fs stands for 'fields'
+    fs = [s for s in timestamp.split(' ') if len(s) > 0]
+    if len(fs) != 6:
+        return
```
xiangce: This should be removed since it will cause an error like:
```
TypeError: cannot unpack non-iterable NoneType object
```

insights/parsers/zdump_v.py
```diff
@@ -0,0 +1,103 @@
...
+            if 'UTC' not in line or 'isdst' not in line:
+                # skip the line that does not include a timestamp
+                continue
+
+            utc_time, remains = line.strip('/etc/localtime').split(' = ')
+            dst['utc_time'], dst['utc_time_s'] = str2datetime(utc_time)
+            if dst['utc_time'] is None:
+                continue
+
+            local_time, _ = remains.split("isdst")
```
xiangce: How about

```
utc_time, remains = line.strip('/etc/localtime').split(' = ')
local_time, remains = [i.strip() for i in remains.split("isdst", 1)]
isdst, remains = [i.strip('=') for i in remains.split(None, 1)
gmtoff = remains.split('=')[1]
# code to combine the namedtuple 
```

insights/parsers/zdump_v.py
```diff
@@ -0,0 +1,103 @@
...
+            dst['isdst'] = len([s for s in remains.split(' ') if 'isdst' in s and '1' in s]) > 0
+
+            gmtoff = [s.split('=')[1] for s in remains.split(' ') if 'gmtoff' in s and '=' in s]
+            if gmtoff:
+                dst['gmtoff'] = int(gmtoff[0])
+
+            self._data.append(dst)
+
+    @property
+    def data(self):
```
xiangce: This is not necessary for new parsers.

insights/specs/default.py
```diff
@@ -1035,6 +1035,7 @@ def xfs_mounts(broker):
     yum_log = simple_file("/var/log/yum.log")
     yum_repolist = simple_command("/usr/bin/yum -C repolist")
     yum_repos_d = glob_file("/etc/yum.repos.d/*")
+    zdump_v = simple_command("/usr/sbin/zdump -v /etc/localtime -c 2019,2039")
```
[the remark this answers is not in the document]
xiangce: @shlao - please create an INSPEC Jira for this new SPEC as well.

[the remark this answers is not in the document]
shlao: I ran the command 'zdump {{-V' on RHEL6、RHEL7、RHEL8}} . Only RHEL8 works with option '-V'
```
RHEL6 ====================================
 # zdump --version
 @(#)zdump.c 8.9

 # zdump -V
 zdump: invalid option – 'V'

RHEL7====================================
 # zdump --version
 zdump (GNU libc) 2.17

 # zdump -V
 zdump: invalid option – 'V'
```
So, I think we can't use '-V' in this.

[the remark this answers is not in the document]
xiangce: Good catch, please keep as it was

insights/parsers/zdump_v.py
shlao: If we write it in the way and command  `zdump` changes the order in the feature :
from:  Sun Mar 10 01:59:59 2019 EST `isdst=0 gmtoff=-18000`
to:      Sun Mar 10 01:59:59 2019 EST `gmtoff=-18000 isdst=0`

We need to modify our code.

insights/parsers/zdump_v.py
shlao: timestamp is string, so the fs is the slice of string. It's iterable object.

insights/parsers/zdump_v.py
shlao: `gmtoff ` isn't documented in the Man Page. It's an option.

insights/parsers/zdump_v.py
xiangce: YOu do not need to split it out and then conjunct them again by using my suggestion.

insights/parsers/zdump_v.py
xiangce: You didn't get the point, In L110, you have assigned the result to two variables. this None will cause error.

insights/parsers/zdump_v.py
xiangce: The original logic is not good from performance and logic perspective, please refer the code above.

insights/parsers/zdump_v.py
xiangce: Then you can add some safety check, what I gave is just a prototype.

insights/parsers/zdump_v.py
shlao: My mistake, thanks

insights/parsers/zdump_v.py
shlao: At first, I wrote the code like this:
```
time_f = "%a %b %d %H:%M:%S %Y"
time_s = timestamp

datetime.strptime(time_s, time_f)
```

But if the output changes in various versions `zdump`, 
From:
```
Sun#Mar#10#06:59:59#2019#UTC
```
To
```
Sun##Mar##10##06:59:59##2019##UTC
```
or
```
Sun#Mar##10#06:59:59###2019#UTC
```

We need to refactor the code.

xiangce: 1. The number of spaces won't affect the parsing processes of `strptime`, just have a try.
2. The format of the output of the `zdump` won't change as you imagine, I believe this Parser is not the only one user depends on it.

insights/parsers/zdump_v.py
```diff
@@ -0,0 +1,122 @@
...
+        # 'UC' is an invalid TimeZone for function `strptime`
+        time_s = time_s + " UTC"
+        time_f = "%a %b %d %H:%M:%S %Y %Z"
+
+    try:
+        time = datetime.strptime(time_s, time_f)
+    except ValueError:
+        pass
+
+    return time, time_string
```
xiangce: Here the `time_string` is not the `RAW` string since you have re-formatted it in L57

PR #2463, opened 2020-02-18
Delete archive dir

SteveHNH:
Delete the leftover empty archive dir after a client run.


insights/client/archive.py
```diff
@@ -192,4 +193,5 @@ def cleanup_tmp(self):
                 return  # return before deleting tmp_dir
         else:
             self.delete_archive_file()
+        self.delete_archive_dir()
```
gravitypriest: We don't actually need this line because this function deletes `archive_dir`, which is inside `tmp_dir`, which is deleted by the function called on the following line. I have an open PR that better explains the variables in this class.

gravitypriest: Apologies for the misleading names... I swear it made sense at one point. :pray:

SteveHNH: @gravitypriest no worries! I removed it. I think I put it in during early testing and just forgot to take it back out

PR #2464, opened 2020-02-19
Enhance satellite_installer_configurations

### LINKED ISSUE #2469: Enhance parser "satellite_installer_configurations"
As a default, the file "/etc/foreman-installer/custom-hiera.yaml" is empty, it means customers haven't done any tuning, it shouldn't be skipped.

huali027:
Fixes #2469 
* As default, the file "/etc/foreman-installer/custom-hiera.yaml" is empty with some comment.
  However this is reasonable, it means customers haven't done tuning, and we should remind
  customers to do some tuning.

### DIFF
--- a/insights/parsers/satellite_installer_configurations.py
+++ b/insights/parsers/satellite_installer_configurations.py
@@ -12,7 +12,7 @@ Parsers the file `/etc/foreman-installer/custom-hiera.yaml`
 
 from insights import parser, YAMLParser
 from insights.specs import Specs
-from insights.parsers import ParseException
+from insights.parsers import SkipException
 
 
 @parser(Specs.satellite_custom_hiera)
@@ -29,5 +29,5 @@ class CustomHiera(YAMLParser):
     def parse_content(self, content):
         try:
             super(CustomHiera, self).parse_content(content)
-        except ParseException:
+        except SkipException:
             pass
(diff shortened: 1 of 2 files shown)


insights/parsers/satellite_installer_configurations.py
```diff
@@ -25,5 +26,8 @@ class CustomHiera(YAMLParser):
         >>> custom_hiera['apache::mod::prefork::serverlimit']
         582
     """
-
-    pass
+    def parse_content(self, content):
+        try:
+            super(CustomHiera, self).parse_content(content)
+        except ParseException:
+            pass
```
bfahr: @huali027 I believe this could results in problems for rules if the Yaml has a bad format.  If the Yaml parser fails with an `ScannerException` instead of parsing a file that simply did not contain anything but comments (which causes a `SkipException`), this code will still pass and the `data` attribute will not be present in the resulting parser object because it was never initialized.  If you add a test with invalid yaml, your test will fail.  I think you'd actually want to fail with the `ParseException` in that case.

huali027: @bfahr yes, @xiangce has pointed it out. So I updated the parent class to raise a SkipException when the file is empty since it is not parsing error. Since the PR for parent class hasn't been merged, so I didn't modify this PR yesterday. I updated it now, please have a check again. Thanks for your review.

patchkez:
test this please

PR #2474, opened 2020-02-19
Add noplugins switch to spec yum repolist

### LINKED ISSUE #2473: Spec `yum -C repolist` should not load plugins
Running the command `yum -C repolist` can cause yum plugins to be loaded.  This command should be run with the `--noplugins` switch to prevent this.

bfahr:
* Add switch to prevent plugins from being loaded running command
* Modify docs in parser and add test for parser
* Modify default spec and add new command to archive spec
* Fix #2473


patchkez:
test this please

PR #2365, opened 2019-12-04
Core collection, final draft (WIP)

gravitypriest:
Took a slightly different approach than the previous PR, trying to focus on a simpler drop-in for the core collection functionality and leave our archive logic as untouched as possible. Unnecessary files have been removed. InsightsSpecs are gone, uploader.json functions, vars, and config opts are gone, all collection logic removed from the client code itself and replaced with a call to insights.collect.collect().

### Summary of Changes
- Use `insights.collect` to interface with core as the data collector
- Remove the classic data collection code
- Remove all uploader.json updating code
- Delete `insights_spec.py`
- Remove references to uploader.json files in /etc/insights-client
- Retool InsightsArchive to initialize with a `None` collected data path, and update it with the path generated by `insights.collect.collect()`
- Remove classic archival functions that add data to the archive
- Create --list option to show all commands/files run by insights-core
- Apply insights-client blacklist to core if insights-client is running the collection
- Disable components based on client blacklist
- Separate regex pattern matching and plain string pattern matching in the core blacklist chain
- Move `.exp.sed` password hiding regex to a new utility function, and pass that into core as part of the blacklist chain
- Migrate commands/files specified in remove.conf to insights-core components
- Include a copy of uploader.json in the package for remove.conf migration
- Disable parallel processing in the default manifest because multithreading is broken on RHEL 8
- Add some additional logging
- Delete unused unit tests

### Potential TODOs
- Move pattern/password redaction to after collection (like a soscleaner step)
- Convert the included uploader.json file to a simpler file to map commands to symbolic names as K/V pairs

PR #2476, opened 2020-02-21
When missing dependencies don't collect it

### LINKED ISSUE #2459: Dependencies 'is_aws'/'is_azure' doesn't work on non-aws/non-azure system
The `insights.specs.default.DefaultSpecs.aws_instance_type` parser got the following result incorrectly from non-aws system:

```
# cat /tmp/insights-xxx-20200216034411/insights_commands/python_-m_insights.tools.cat_--no-header_aws_instance_type
Missing Dependencies:
    At Least One Of:
        insights.specs.default.DefaultSpecs.aws_instance_type
        insights.specs.insights_archive.InsightsArchiveSpecs.aws_instance_type
```
Then the `insights.parsers.aws_instance_type.AWSInstanceType` parser returns `.type` as "INSIGHTS", for all non-aws systems, which result some rules got fired incorrectly.

In fact, this spec `aws_instance_type` shouldn't be collected at all since it depends on `is_aws`:
https://github.com/RedHatInsights/insights-core/blob/0db5e8900507a6a60b58ac458a11644a8871306a/insights/specs/default.py#L100-L109

And the `is_aws` is skipped correctly:
```
# insights-cat insights.specs.default.DefaultSpecs.is_aws /tmp/insights-xxx-20200216034411
[Empty output]
```

And the dependency check works well:
```
# insights-cat insights.specs.default.DefaultSpecs.aws_instance_type /tmp/insights-xxx-20200216034411 
Missing Dependencies:
    Requires:
        insights.core.context.HostContext
        insights.specs.default.DefaultSpecs.is_aws
```
However, the spec is still collected by the client.
```
# cat /tmp/insights-xxx-20200216034411/insights_commands/python_-m_insights.tools.cat_--no-header_aws_instance_type
Missing Dependencies:
    At Least One Of:
        insights.specs.default.DefaultSpecs.aws_instance_type
        insights.specs.insights_archive.InsightsArchiveSpecs.aws_instance_type
```

It shouldn't be collected, right?

BTW, the `azure_*` and the other `aws_*` specs are also affected by this issue.

xiangce:
- fix #2459 

### DIFF
--- a/insights/client/archive.py
+++ b/insights/client/archive.py
@@ -163,13 +163,16 @@ class InsightsArchive(object):
         Add files and commands to archive
         Use InsightsSpec.get_output() to get data
         '''
-        cmd_not_found_regex = "^timeout: failed to run command .+: No such file or directory$"
+        ab_regex = [
+            "^timeout: failed to run command .+: No such file or directory$",
+            "^Missing Dependencies:"
+        ]
         if isinstance(spec, InsightsCommand):
             archive_path = os.path.join(self.cmd_dir, spec.archive_path.lstrip('/'))
         if isinstance(spec, InsightsFile):
             archive_path = self.get_full_archive_path(spec.archive_path.lstrip('/'))
         output = spec.get_output()
-        if output and not re.search(cmd_not_found_regex, output):
+        if output and not any(re.search(rg, output) for rg in ab_regex):
             write_data_to_file(output, archive_path)
 
     def add_metadata_to_archive(self, metadata, meta_path):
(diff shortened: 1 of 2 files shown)


xiangce:
This is related to #2459

gravitypriest:
There's no harm in merging this but it won't really do anything. Core is not currently being used for insights collection, and the work in #2365 removes this function.

xiangce:
@gravitypriest - This patch is trying to fix a false-positive reported by the customer: [bz-1803093](https://bugzilla.redhat.com/show_bug.cgi?id=1803093). And since the new #2365 is in `WIP`, can we merge this PR first to fix the bug from the customer?

gravitypriest:
Yeah, it'll work fine.

PR #2479, opened 2020-02-25
Add parser for 'systemctl show *.target'

psachin:


psachin:
Please don't review.

xiangce:
The spec is very similar to the `systemctl_show_all_services` and there is an opening [bz-1794605](https://bugzilla.redhat.com/show_bug.cgi?id=1794605) requires this Parser, I go ahead and merge it before the spec be approved. Thanks.

PR #2487, opened 2020-03-05
samba: Add parser and spec for 'testparm -s' (/etc/samba/smb.conf)

cryptomilk:
This should replace the smb.conf parser as testparm sanitizes the output.

### DIFF
--- a/insights/parsers/tests/test_samba_smb_conf.py
+++ b/insights/parsers/tests/test_samba_smb_conf.py
@@ -1,0 +1,253 @@
+from insights.parsers import samba_smb_conf
+from insights.tests import context_wrap
+from doctest import testmod
+
+SAMBA_CONFIG_DOCUMENTATION = '''
+# This is the main Samba configuration file. You should read the
+# smb.conf(5) manual page in order to understand the options listed
+#...
+#======================= Global Settings =====================================
+
+[global]
+    workgroup = MYGROUP
+    server string = Samba Server Version %v
+    max log size = 50
+
+[homes]
+    comment = Home Directories
+    browseable = no
+    writable = yes
+;   valid users = %S
+;   valid users = MYDOMAIN\%S
+
+[printers]
+    comment = All Printers
+    path = /var/spool/samba
+    browseable = no
+    guest ok = no
+    writable = no
+    printable = yes
+
+# A publicly accessible directory, but read only, except for people in
+# the "staff" group
+[public]
+   comment = Public Stuff
+   path = /home/samba
+   public = yes
+   read only = no
+   printable = no
+   write list = +staff
+'''
+
+
+SAMBA_CONFIG = """
+# This is the main Samba configuration file. You should read the
+# smb.conf(5) manual page in order to understand the options listed
+#...
+#======================= Global Settings =====================================
+
+# By running `testparm`, it is apparent that options outside the [global]
+# section before the [global] section are treated as if they were in the
+# [global] section.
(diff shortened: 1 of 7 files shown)


insights/specs/default.py
```diff
@@ -837,6 +837,8 @@ def rhev_data_center(broker):
     rpm_V_packages = simple_command("/bin/rpm -V coreutils procps procps-ng shadow-utils passwd sudo", keep_rc=True)
     rsyslog_conf = simple_file("/etc/rsyslog.conf")
     samba = simple_file("/etc/samba/smb.conf")
+    # TODO also parse the stderr output of testparm -s
+    samba_smb_conf = simple_command("/usr/bin/testparm -s /etc/samba/smb.conf 2>/dev/null")
```
bfahr: @cryptomilk this command will not work with the  `2> /dev/null` because all commands are run without the shell.

Also since the client adds this information to the archive as a file you will need to include a `simple_file` spec in `insights/specs/insights_archive.py`.  Use the `mangle` command in insights-core to determine what the file path  should be.  For example:
```
$ mangle "/usr/bin/testparm -s /etc/samba/smb.conf"
```

insights/parsers/tests/test_samba_smb_conf.py
```diff
@@ -0,0 +1,245 @@
...
+    assert not config.has_option('homes', 'valid users')
+
+    assert config.get('printers', 'comment') == 'All Printers'
+    assert config.get('printers', 'path') == '/var/spool/samba'
+    assert config.get('printers', 'browseable') == 'No'
+    assert config.get('printers', 'printable') == 'Yes'
+
+    assert 'netlogin' not in config
+    assert 'Profiles' not in config
+    assert 'public' not in config
```
bfahr: Please add testing for your parser docstring examples.  Here's an [example](https://github.com/RedHatInsights/insights-core/blob/f3b02e09a8784b9e3d0618d2319b9c34f457c16c/insights/parsers/tests/test_rhsm_releasever.py#L45) of how that can be done.

insights/parsers/tests/test_samba_smb_conf.py
```diff
@@ -0,0 +1,245 @@
...
+
+[printers]
+        browseable = No
+        comment = All Printers
+        path = /var/spool/samba
+        printable = Yes
+'''
+
+
+def test_match():
```
bfahr: I see that you have the `allow_no_values` set for the parser, are you testing any keys with no values?

cryptomilk:
Thank you very much for the review. I will improve this but it will take some time.

insights/specs/default.py
cryptomilk: Done. Thanks!

insights/parsers/tests/test_samba_smb_conf.py
cryptomilk: We have a test, I renamed it and move it to the end now.

insights/parsers/tests/test_samba_smb_conf.py
cryptomilk: We don't have it set, but I set it and added a test now.

cryptomilk:
Ping!

cryptomilk:
Rebased on master to fix merge conflicts

cryptomilk:
Rebased on master, this code is still bitrotting here ...

cryptomilk:
Closing in favor of  #2786

PR #2491, opened 2020-03-10
Add RHSM (Red Hat Subscription-Manager) Parser

lveyde:
Signed-off-by: Lev Veyde <lveyde@redhat.com>

### DIFF
--- a/insights/parsers/rhsm.py
+++ b/insights/parsers/rhsm.py
@@ -1,0 +1,157 @@
+"""
+Red Hat Subscription-Manager - Commands
+==============
+Parsers for ``subscription-manager`` commands.
+
+This module contains the classes that parse the output of the commands
+`subscription-manager repos --list-enabled`.
+
+RHSMEnabledRepoList - command ``subscription-manager repos --list-enabled``
+-----------------------------------------
+"""
+
+from insights import parser, CommandParser
+from insights.parsers import SkipException, parse_fixed_table
+from insights.specs import Specs
+
+eus = [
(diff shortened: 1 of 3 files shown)


xiangce:
@lveyde - As you might notice, there is an existing parser [YumRepoList](https://github.com/RedHatInsights/insights-core/blob/900148134a25d4bbaed17455dd63171667dcc64d/insights/parsers/yum.py#L44) which parses the output of `yum -C --noplugins repolist` command and can achieve the same function with this one.

https://github.com/RedHatInsights/insights-core/blob/900148134a25d4bbaed17455dd63171667dcc64d/insights/parsers/yum.py#L44

It's not suggested to add duplicated parsers. may I know the reason to add this spec and parser? Can the `YumRepoList` work for you?

lveyde:
Hey,

The difference is that my parser takes the information directly from
subscription-manager and thus allows to properly check the available RHN
channels versus similarly named Yum repos, that may or may not include the
expected packages.

We basically need to check that customers machine is properly configured as
far as the RHN channels are concerned, and simply checking Yum doesn't
provide that accuracy.

The attempt was made to use the same rule interface as the Yum parser, so
that the rules could be easily modified to switch between parsers, if more
strict or less strict check is required.

Thanks in advance,
Lev Veyde,

bfahr:
@lveyde you will need to get this spec approved before we add the parser.

xiangce:
Closing it for now, please re-open it when necessary

PR #2499, opened 2020-03-16
Parsers for Domain information groper

jobselko:


insights/parsers/dig.py
```diff
@@ -0,0 +1,171 @@
...
+from insights.core.plugins import parser
+from insights.parsers import SkipException
+from insights.specs import Specs
+
+
+HEADER_TEMPLATE = re.compile(r';; ->>HEADER<<-.*status: (\S+),')
+RRSIG_TEMPLATE = re.compile(r'RRSIG')
+
+
+class Dig(Parser):
```
bfahr: @jobselko this should use the `CommandParser` base class instead to check for some of the typical error messages commands return.

PR #2500, opened 2020-03-17
Add new parser for "/proc/sys/kernel/sched_rt_runtime_us"

huali027:

### DIFF
--- a/insights/parsers/sys_kernel.py
+++ b/insights/parsers/sys_kernel.py
@@ -1,11 +1,11 @@
 """
-System kernel files under ``/proc/sys/kernel`` or ``/sys/kernel/``
-==================================================================
+System kernel files under ``/proc/sys/kernel`` or ``/sys/kernel``
+=================================================================
 
 This module contains the following parsers:
 
-SchedRTRuntime - (File ``/proc/sys/kernel/sched_rt_runtime_us``)
-----------------------------------------------------------------
+SchedRTRuntime - file ``/proc/sys/kernel/sched_rt_runtime_us``
+--------------------------------------------------------------
 """
 
 from insights import Parser, parser, get_active_lines
@@ -29,7 +29,10 @@ class SchedRTRuntime(Parser):
         950000
 
     Attributes:
-        runtime_us (int): the value of sched_rt_runtime_us
+        runtime_us (int): The value of sched_rt_runtime_us
+
+    Raises:
+        ParseException: Raised when there is more than one line or the value isn't interger.
     """
 
     def parse_content(self, content):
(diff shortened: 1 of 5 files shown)


insights/parsers/sys_kernel.py
```diff
@@ -0,0 +1,38 @@
+"""
+SysKernel - file ``/proc/sys/kernel`` or ``/sys/kernel/``
```
xiangce: Generally, the first item in the title should be a `ParserClass`, when it contains multiple Parsers, better use as `Descriptive title for the parsers`

See https://github.com/RedHatInsights/insights-core/wiki#checklist-for-reviewers

insights/parsers/sys_kernel.py
```diff
@@ -0,0 +1,38 @@
...
+--------------------------------------------------------------
+"""
+
+from insights import Parser, parser, LegacyItemAccess, get_active_lines
+from insights.parsers import ParseException
+from insights.specs import Specs
+
+
+@parser(Specs.sched_rt_runtime_us)
+class SchedRTRuntime(Parser, LegacyItemAccess):
```
xiangce: `LegacyItemAccess` is only useful for `ParserClass` which has a `data` attribute in `dict` type. Here in this parser it's useless.

BTW, The `LegacyItemAccess` going to be deprecated.

insights/parsers/sys_kernel.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Class for parsing the `/proc/sys/kernel/sched_rt_runtime_us` file.
+
+    Typical content of the file is::
+        950000
+
+    Examples:
+        >>> type(srt)
+        <class 'insights.parsers.sys_kernel.SchedRTRuntime'>
+        >>> srt.runtime_us
```
xiangce: For public `attributes` that need to be exposed to the user, it's necessary to describe them in the `Attributes` docstring.

See https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google-style-python-docstrings

insights/parsers/sys_kernel.py
```diff
@@ -0,0 +1,38 @@
...
+from insights.parsers import ParseException
+from insights.specs import Specs
+
+
+@parser(Specs.sched_rt_runtime_us)
+class SchedRTRuntime(Parser, LegacyItemAccess):
+    """
+    Class for parsing the `/proc/sys/kernel/sched_rt_runtime_us` file.
+
+    Typical content of the file is::
```
xiangce: One more blank line is needed between the "::" and it's content.

Issue #2502, opened 2020-03-19
Block-size of 'df -al' is not always default to 1K. DiskFree_AL and DiskFree_ALP need handle this.


JoySnow:
Currently, we treat the block-size as `1K`, and ignore the real block-size used in headline. 

Refer to man-page of `df`, the block-size could be determined by many env variables:
```
Display values are in units of the first available SIZE from --block-size, and the DF_BLOCK_SIZE, BLOCK_SIZE and BLOCK-
       SIZE environment variables.  Otherwise, units default to 1024 bytes (or 512 if POSIXLY_CORRECT is set).
```

For example, by tuning env variables,  the output of `df -al` could be like 
```
Filesystem     512B-blocks     Used Available Use% Mounted on
rootfs            83861248 14012128  69849120  17% /
```
Or
```
Filesystem     1M-blocks  Used Available Use% Mounted on
rootfs            40948M 6842M    34107M  17% /
```

To fix this bug, we can get the  `block-size` from headline, and recalculate each value by multiplying the  `block-size` .

xiangce:
@JoySnow - Nice catch!!

However, the solution you mentioned maybe not the best way to fix this issue.
> To fix this bug, we can get the block-size from headline, and recalculate each value by multiplying the block-size .

This is a really flexible logic for the parser to get the values, however, it may lose precision in some cases, e.g. when the DF_BLOCK_SIZE is set to 1G, especially for some rule that needs to check a threshold, e.g. 5%~10% free space.

And it will make the logic of the parser very complex since there are so many different cases need to consider.
```python
[root@insights-engine insights-engine]# df -B 1G
Filesystem                   1G-blocks  Used Available Use% Mounted on
devtmpfs                             1     0         1   0% /dev
tmpfs                                1     0         1   0% /dev/shm
tmpfs                                1     1         1   4% /run
tmpfs                                1     0         1   0% /sys/fs/cgroup
/dev/mapper/rhel_unused-root        13     3        11  18% /
/dev/vda1                            1     1         1  23% /boot
tmpfs                                1     0         1   0% /run/user/0
[root@insights-engine insights-engine]# df -B 1MB
Filesystem                   1MB-blocks  Used Available Use% Mounted on
devtmpfs                            952     0       952   0% /dev
tmpfs                               964     0       964   0% /dev/shm
tmpfs                               964    35       930   4% /run
tmpfs                               964     0       964   0% /sys/fs/cgroup
[...]

PR #2505, opened 2020-03-23
Update DiskFree to return block_size (for sosreport)

xiangce:
- Add an interface to return the block size to compatible with sos archives
- Remove the deprecated `shared` in examples
- Add test and refine the Examples
- Remove the dependant on defaultdict
- Fix: change the return value of two properties to list
- fix: #2502 


bfahr:
test this please

insights/parsers/df.py
```diff
@@ -289,5 +313,24 @@ class DiskFree_AL(DiskFree):
                 Available      available
                 Use%           capacity
                 Mounted on     mounted_on
+        block_size (str): The unit of display values, '1K' by default.
```
JoySnow: @xiangce 
As sosreport is using `df_al`, we can check `block_size` in rules to workaround this.       
For this point, do we consider to provide an `int` value of `block_size` at same time?  So we do the B/K/M/G converting here instead in each rule that use this parser.   
Like: 
```suggestion
        block_size_raw (str): The unit of display values, '1K' by default.
        block_size (int): The calculatable value of block_size_raw in byte, 1024 for '1K', 1048576 for '1M'.
```

bfahr:
@xiangce given that we have cancelled the spec change to you want to keep any of these changes, or close this PR?

xiangce:
> @xiangce given that we have cancelled the spec change to you want to keep any of these changes, or close this PR?

@bfahr, I'd like to keep the changes and will update for sosreport, since the command collected in sosreport archives is impacted.

insights/parsers/df.py
xiangce: @JoySnow - Updates applied, please have a look.

PR #2506, opened 2020-03-23
Parser for Named-checkconf

jobselko:


jobselko:
@xiangce I will take a look at it, thank you.

jobselko:
@xiangce I will not be capable of rewriting that with using a "parsr". So if it is all right, I would leave it as it is. Nevertheless, I will appreciate the contribution from anyone else.

skontar:
@xiangce I think it is fine if we leave it as is. This code is tested and used in security rules for a long time. It can be re-written using parsr in the future and not necessarily by security team.

Issue #2403, opened 2020-01-09
KpatchPatches (kpatch_patch_files) can be replaced with KpatchList (kpatch_list)


xiangce:
The two specs are as follows:
```python
    @datasource(Uname, context=HostContext)
    def kpatch_patches_running_kernel_dir(broker):
        un = broker[Uname]
        return r"/var/lib/kpatch/" + un.kernel

    kpatch_list = simple_command("/usr/sbin/kpatch list")
    kpatch_patch_files = command_with_args("ls %s", kpatch_patches_running_kernel_dir)
```

The content of KpatchList (kpatch_list) covers the content of KpatchPatches (kpatch_patch_files)  and kpatch_list can be collected by the current client tool.

We'd better deprecate the KpatchPatches (kpatch_patch_files) and use KpatchList (kpatch_list) instead.

PR #2508, opened 2020-03-24
Deprecate KpatchPatches

xiangce:
- fix: #2403


shlao:
insights-plugins code is updated.

PR #2511, opened 2020-03-24
Adding posibility of specifying virtual memory limit usage for archive extraction

joselsegura:
This commit adds the needed code for specifying a virtual memory limit when extracting an archive.
If not specified, it will work with no memory limitation, as it already does

### DIFF
--- a/insights/core/archives.py
+++ b/insights/core/archives.py
@@ -26,28 +26,37 @@ class InvalidContentType(InvalidArchive):
         self.content_type = content_type
 
 
-class ZipExtractor(object):
-    def __init__(self, timeout=None):
-        self.content_type = "application/zip"
+class Extractor(object):
+    """Base class for different extractor types."""
+    def __init__(self, timeout=None, memory_limit=None):
+        """Creates an extractor object with a given timeout and memory_limit."""
         self.timeout = timeout
+        self.memory_limit = memory_limit
         self.tmp_dir = None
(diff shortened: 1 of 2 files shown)


bfahr:
@JoseLSegura is there an open Bugzilla or Issue related to this change?

JoseLSegura:
@bfahr no, it isn't AFAIK. We fond this problem when using `insights-core-messaging` and using a zipbomb for the `extract` method. Should I open a new issue explaining what happened and why this solution is needed?
About configuration and/or command line args, it could be added, you are right. I didn't use insights from CLI at all, so I totally forgot about it.
I guess that if it is possible to define a timeout from the /run args or by configuration, then the memory limit should be added too, but IDK if it is the case.

quarckster:
cc @bsquizz @skarekrow

bsquizz:
FYI @brantleyr @SteveHNH @kylape in case you weren't aware ...

bfahr:
@JoseLSegura an issue is not necessarily required I just wanted to make sure that we have a description of why the enhancement is needed so that the PR can be evaluated.

@csams what's your opinion on adding configuration and/or command line options?

JoseLSegura:
@bfahr ok, nice. I want to be sure, it is my first contribution to `insights-core` repository :-)

Thank you

xiangce:
Closing it for now, we can reopen it in another new PR when necessary

PR #2512, opened 2020-03-24
Deduplicate http response content when upload fails

SteveHNH:
We are currently logging debug information twice when using --net-debug.
This change makes it so that we only report the failure once in stdout.
The output for the failure message still provides relevant info that's
needed to find out what happened. The --net-debug prints some additional
info rather than printing the same thing twice


patchkez:
test this please

patchkez:
test this please

SteveHNH:
test this please

PR #2518, opened 2020-03-27
Add parser for "/usr/bin/firewall-cmd --list-all-zones"

huali027:


insights/parsers/firewall_cmd.py
```diff
@@ -0,0 +1,114 @@
...
+        zone_name = ''
+        zone_attr_index = -1
+        zone_attr_name = ''
+        for line in content:
+            if not line.strip():
+                zone_line = True
+                continue
+            if zone_line:
+                name_info = line.strip().split(None, 1)
+                if len(name_info) > 1 and 'active' in name_info[1]:
```
xiangce: Per the source code of firewall-cmd, there could be other `attributes` in the `()` after the zone name.

It's not the best to only hard-code the `active` like this. better keep all possible `attributes` of a zone to e.g. a list.

https://github.com/firewalld/firewalld/blob/6bfffe65f55b727afc37a8c1fb4068f6589bb890/src/firewall/command.py#L410-L416

insights/parsers/firewall_cmd.py
```diff
@@ -0,0 +1,114 @@
...
+        active_zones (list): A list of the zone name that's active
+        zones (dict): A dict of zone info
+
+    Raises:
+        SkipException: Raised when firewalld is not running
+        ParseException: Raised when the output is in invalid format
+    """
+
+    def parse_content(self, content):
+        error_line = "firewalld is not running"
```
xiangce: You may want to use the `extra_bad_lines` and re-implement the constructor `__init__` of CommandParser to check this `error_line`.

https://github.com/RedHatInsights/insights-core/blob/bf3527ef533ba14dbdbc60edda77d38232b9692b/insights/core/__init__.py#L578

insights/parsers/firewall_cmd.py
huali027: > Per the source code of firewall-cmd, there could be other `attributes` in the `()` after the zone name.
> 
> It's not the best to only hard-code the `active` like this. better keep all possible `attributes` of a zone to e.g. a list.
> 
> https://github.com/firewalld/firewalld/blob/6bfffe65f55b727afc37a8c1fb4068f6589bb890/src/firewall/command.py#L410-L416

Sure, I'll update the PR, thanks.

insights/parsers/firewall_cmd.py
huali027: > You may want to use the `extra_bad_lines` and re-implement the constructor `__init__` of CommandParser to check this `error_line`.
> 
> https://github.com/RedHatInsights/insights-core/blob/bf3527ef533ba14dbdbc60edda77d38232b9692b/insights/core/__init__.py#L578

sure, thanks for your review, I'll update it.

PR #2519, opened 2020-03-28
Add parser for HammerComputeResourceList

pipopopo:
Attempt to create a parser for HammerComputeResourceList

```
# py.test -svk hammer_compute_resource_list | grep hammer_compute_resource_list

/tmp/git/insights-core/insights/combiners/nfs_exports.py:73: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
  if isinstance(nfsexportsd, collections.Iterable):
insights/parsers/tests/test_hammer_compute_resource_list.py::test_hammer_compute_resource_list PASSED
insights/parsers/tests/test_hammer_compute_resource_list.py::test_no_data PASSED

```

### DIFF
--- a/docs/shared_parsers_catalog/hammer_compute_resource_list.rst
+++ b/docs/shared_parsers_catalog/hammer_compute_resource_list.rst
@@ -1,0 +1,3 @@
+. automodule:: insights.parsers.hammer_compute_resource_list
+   :members:
+   :show-inheritance:
--- a/docs/shared_parsers_catalog/hammer_compute_resource_list.rst
+++ b/docs/shared_parsers_catalog/hammer_compute_resource_list.rst
@@ -1,3 +1,3 @@
-. automodule:: insights.parsers.hammer_compute_resource_list
+.. automodule:: insights.parsers.hammer_compute_resource_list
    :members:
    :show-inheritance:
--- a/insights/parsers/hammer_compute_resource_list.py
+++ b/insights/parsers/hammer_compute_resource_list.py
@@ -32,14 +32,14 @@ Examples:
     <class 'insights.parsers.hammer_compute_resource_list.HammerComputeResourceList'>
 """
 
-from insights import parser, CommandParser
-from insights.parsers import SkipException
-from insights.specs import Specs
 import json
+from insights import JSONParser, parser, CommandParser
+from insights.parsers import ParseException, SkipException
+from insights.specs import Specs
 
 
 @parser(Specs.hammer_compute_resource_list)
-class HammerComputeResourceList(CommandParser):
+class HammerComputeResourceList(CommandParser, JSONParser):
     """
     Parse the JSON output from the ``hammer --interactive 0 --output json compute-resource list`` command.
 
@@ -49,8 +49,10 @@ class HammerComputeResourceList(CommandParser):
     Attributes:
         data(list): A list of the parsed output returned by `hammer compute resource list`
     """
-
     def parse_content(self, content):
         if not content:
-            raise SkipException('No content or hammer auth failed.')
-        self.data = json.loads(''.join(content))
+            raise SkipException("No content or hammer auth failed.")
+        try:
+            self.data = json.loads(''.join(content))
+        except:
+            raise ParseException("Could not parse json.", content)
--- a/insights/parsers/hammer_compute_resource_list.py
+++ b/insights/parsers/hammer_compute_resource_list.py
@@ -1,6 +1,6 @@
 """
 HammerComputeResourceList - command ``hammer --interactive 0 --output json compute-resource list``
-===================================================
+==================================================================================================
 
 This parser reads the compute resource list of a Satellite server using hammer in JSON format.
 It relies on the root user running the command and hammer to authenticate based on the hammer configuration file.
@@ -24,7 +24,7 @@ Sample output from the ``hammer --interactive 0 --output json compute-resource l
     "Name": "vmware67",
     "Provider": "VMware"
   }
-]
+    ]
 
 Examples:
 
--- a/insights/parsers/hammer_compute_resource_list.py
+++ b/insights/parsers/hammer_compute_resource_list.py
@@ -1,0 +1,56 @@
+"""
+HammerComputeResourceList - command ``hammer --interactive 0 --output json compute-resource list``
+===================================================
+
+This parser reads the compute resource list of a Satellite server using hammer in JSON format.
+It relies on the root user running the command and hammer to authenticate based on the hammer configuration file.
+If the command is unable to authenticate then no compute resources will be shown and an erro flag will be recorded in the parser
+
+Sample output from the ``hammer --interactive 0 --output json compute-resource list`` command::
+
+    [
+  {
+    "Id": 1,
+    "Name": "kvm-server",
+    "Provider": "Libvirt"
+  },
+  {
+    "Id": 4,
+    "Name": "VMWARE",
+    "Provider": "VMware"
+  },
+  {
+    "Id": 3,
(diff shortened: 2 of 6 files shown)


insights/parsers/hammer_compute_resource_list.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+from insights import parser, CommandParser
+from insights.parsers import SkipException
+from insights.specs import Specs
+import json
+
+
+@parser(Specs.hammer_compute_resource_list)
+class HammerComputeResourceList(CommandParser):
```
xiangce: There is a `JSONParser` that can be used as one of its base classes. Would you please change to use it?

https://github.com/RedHatInsights/insights-core/blob/bf3527ef533ba14dbdbc60edda77d38232b9692b/insights/core/__init__.py#L744-L756

Here is an example:

https://github.com/RedHatInsights/insights-core/blob/bf3527ef533ba14dbdbc60edda77d38232b9692b/insights/parsers/ceph_cmd_json_parsing.py#L44-L56

insights/specs/default.py
```diff
@@ -351,6 +351,7 @@ def dumpdev(broker):
     grubby_default_kernel = simple_command("/usr/sbin/grubby --default-kernel")  # RHEL6 and updwards
     hammer_ping = simple_command("/usr/bin/hammer ping")
     hammer_task_list = simple_command("/usr/bin/hammer --config /root/.hammer/cli.modules.d/foreman.yml --output csv task list --search 'state=running AND ( label=Actions::Candlepin::ListenOnCandlepinEvents OR label=Actions::Katello::EventQueue::Monitor )'")
+    hammer_compute_resource_list = simple_command("hammer --interactive 0 --output json compute-resource list")
```
xiangce: Since the command will be executed by python but without a shell, please specify it's its full path to avoid cannot find the `hammer` command`. 

`/usr/bin/hammer`

insights/specs/insights_archive.py
```diff
@@ -83,6 +83,7 @@ class InsightsArchiveSpecs(Specs):
     gluster_peer_status = simple_file("insights_commands/gluster_peer_status")
     hammer_ping = simple_file("insights_commands/hammer_ping")
     hammer_task_list = simple_file("insights_commands/hammer_--config_.root..hammer.cli.modules.d.foreman.yml_--output_csv_task_list_--search_state_running_AND_label_Actions_Candlepin_ListenOnCandlepinEvents_OR_label_Actions_Katello_EventQueue_Monitor")
+    hammer_compute_resource_list = simple_file("hammer --interactive 0 --output json compute-resource list")
```
xiangce: The string to the `simple_file` is what it would be in the insights_archives. it can be generated by the `mangle` command of insights-core:

```
# cd insights-core
# source [path_to_venv]/bin/activate
# mangle "/usr/bin/hammer --interactive 0 --output json compute-resource list"
"insights_commands/hammer_--interactive_0_--output_json_compute-resource_list"
```

insights/parsers/hammer_compute_resource_list.py
```diff
@@ -0,0 +1,56 @@
+"""
+HammerComputeResourceList - command ``hammer --interactive 0 --output json compute-resource list``
+===================================================
```
xiangce: Better keep the title and the `==` lines in the same length. Otherwise, errors will report when building the Document for it.

You can also check the doc in your local by:

~~~
# cd insights-core
# (cd docs && make html)
OR
# (cd docs && sphinx-build -b html -d _build/doctrees  -W -j 8 . _build/html)
~~~

docs/shared_parsers_catalog/hammer_compute_resource_list.rst
```diff
@@ -0,0 +1,3 @@
+. automodule:: insights.parsers.hammer_compute_resource_list
```
xiangce: There should be two dots (`.`) before the `automodule::` directive.

insights/parsers/hammer_compute_resource_list.py
```diff
@@ -0,0 +1,56 @@
...
+
+This parser reads the compute resource list of a Satellite server using hammer in JSON format.
+It relies on the root user running the command and hammer to authenticate based on the hammer configuration file.
+If the command is unable to authenticate then no compute resources will be shown and an erro flag will be recorded in the parser
+
+Sample output from the ``hammer --interactive 0 --output json compute-resource list`` command::
+
+    [
+  {
```
xiangce: One more block (4 blank spaces) should be added to the following lines  (to L27) to keep the alignment for the `::` section.

insights/parsers/hammer_compute_resource_list.py
```diff
@@ -0,0 +1,56 @@
...
+    "Id": 3,
+    "Name": "vmware67",
+    "Provider": "VMware"
+  }
+]
+
+Examples:
+
+    >>> type(hammer_compute_resources_list)
+    <class 'insights.parsers.hammer_compute_resource_list.HammerComputeResourceList'>
```
xiangce: Would you please add a bit more to the `Examples`, e.g. how many resources are listed (the length of the list, `len(hammer_compute_resources_list)`, or how to access a specific value of a resource?

bfahr:
@pipopopo assuming that you have setup your environment as recommended in [CONTRIBUTING.md](https://github.com/RedHatInsights/insights-core/blob/master/CONTRIBUTING.md) and the [Wiki](https://github.com/RedHatInsights/insights-core/wiki), you can run both `pytest` and `flake8` in the root of your repo to ensure that you have no errors.  You can also run `make clean html_debug` in the `docs` subdir to verify that your docstrings are correct and you can review your generated docs in `docs/_build/html/index.html`.  If all of these are satisfactory then the CI tests are more likely to pass when you commit your changes.  More info on documentation requirements and testing are available in the [Core docs](https://insights-core.readthedocs.io/en/latest/docs_guidelines.html).

insights/specs/default.py
pipopopo: fixed

insights/parsers/hammer_compute_resource_list.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Parse the JSON output from the ``hammer --interactive 0 --output json compute-resource list`` command.
+
+    Raises:
+        SkipException: When nothing is parsed or hammer auth fails
+
+    Attributes:
+        data(list): A list of the parsed output returned by `hammer compute resource list`
+    """
+    def parse_content(self, content):
```
bfahr, line 52: @pipopopo this method is implemented in the base `JSONParser` class so its not needed here.  You can just use `pass` here instead.

insights/parsers/tests/test_hammer_compute_resource_list.py
```diff
@@ -0,0 +1,18 @@
...
+from insights.parsers import SkipException
+
+hammer_compute_resource_list_json_data = '''
+[{"Id": 1, "Name": "kvm-server", "Provider": "Libvirt"}, {"Id": 4, "Name": "VMWARE", "Provider": "VMware"}, {"Id": 3, "Name": "vmware67", "Provider": "VMware"}]
+'''
+
+
+def test_hammer_compute_resource_list():
+    hammer_compute_resource_list.HammerComputeResourceList(context_wrap(hammer_compute_resource_list_json_data))
+
```
bfahr, line 13: This test should at least check for some values to ensure that if something changes in the base parser, it will be surfaced here.

xiangce:
Closing it as no update and it's not required

PR #2523, opened 2020-03-31
parser: add parser for yum list available

theyoyojo:
Add a parser to provide the output of `yum list available`

Signed-off-by: Joel Savitz <jsavitz@redhat.com>

### DIFF
--- a/insights/parsers/yum_list.py
+++ b/insights/parsers/yum_list.py
@@ -7,6 +7,9 @@ The parsers contains in this module are:
 YumListInstalled - Command ``yum list installed``
 -------------------------------------------------
 
+YumListAvailable - Command ``yum list available``
+-------------------------------------------------
+
 """
 from collections import defaultdict
 
@@ -41,9 +44,12 @@ class YumListBase(CommandParser, RpmList):
         contain a ``.repo`` attribute.
     """
 
-    def __init__(self, context):
+    def __init__(self, context, package_status):
         self.expired_cache = False
         """bool: Indicates if the yum repo cache is expired."""
+
+        self.package_status = package_status
+        """str: Indicates if the list is of installed or available packages."""
 
         super(YumListBase, self).__init__(context)
 
@@ -51,7 +57,7 @@ class YumListBase(CommandParser, RpmList):
         for i, c in enumerate(content):
             if 'Repodata is over 2 weeks old' in c:
                 self.expired_cache = True
-            elif c == "Installed Packages":
+            elif c == self.package_status + " Packages":
                 break
         return i + 1
 
@@ -197,4 +203,18 @@ class YumListInstalled(YumListBase):
         >>> rpm1 < rpm2
         True
     """
-    pass
+    def __init__(self, context):
+        super(YumListInstalled, self).__init__(context, "Installed")
+
+
+@parser(Specs.yum_list_available)
+class YumListAvailable(YumListBase):
+    """
+    The ``YumListAvailable`` class parses the output of the ``yum list available``
+    command. Each line is parsed and stored in a ``YumListRpm`` object.
+
+    Input and usage examples are identical to ``YumListInstalled`` but with
+    "Installed" replaced with "Available" wherever applicable.
+    """
+    def __init__(self, context):
+        super(YumListAvailable, self).__init__(context, "Available")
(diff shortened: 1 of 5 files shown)


insights/parsers/yum_list_available.py
```diff
@@ -0,0 +1,126 @@
+
```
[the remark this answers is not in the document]
theyoyojo: I refactored the entire thing into a nearly trivial subclass of YumListBase and I added a short docstring. Would you prefer additional documentation?

insights/specs/default.py
```diff
@@ -1035,6 +1035,7 @@ def xfs_mounts(broker):
     xfs_info = foreach_execute(xfs_mounts, "/usr/sbin/xfs_info %s")
     xinetd_conf = glob_file(["/etc/xinetd.conf", "/etc/xinetd.d/*"])
     yum_conf = simple_file("/etc/yum.conf")
+    yum_list_available = simple_command("yum --noplugins list available")
```
xiangce, line 1038: May I know why you remove the `-C` here? `-C` is added to avoid updating the cache, since the cache-update needs some network traffics, this issue has been complained by some CUs.

xiangce: Since this spec is a `simple_command`, it will be collected into the insights-archive as a `simple_file`. Another entry for this spec needs to be added into the `insights/spec/insights_archive.py`

See: https://github.com/RedHatInsights/insights-core/blob/495c0985a005edb807c41231c9748898a2bc984f/insights/specs/insights_archive.py#L296

insights/parsers/yum_list.py
```diff
@@ -5,6 +5,7 @@
 The parsers contains in this module are:
 
 YumListInstalled - Command ``yum list installed``
+YumListAvailable - Command ``yum list available``
```
xiangce, line 8: The `----` line is needed for each of the sub-title by the `sphinx-doc`.

See: https://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html#headings

insights/parsers/yum_list.py
```diff
@@ -41,17 +42,20 @@ class YumListBase(CommandParser, RpmList):
         contain a ``.repo`` attribute.
     """
 
-    def __init__(self, context):
+    def __init__(self, context, package_status):
         self.expired_cache = False
         """bool: Indicates if the yum repo cache is expired."""
 
+        self.package_status = package_status
```
xiangce, line 49: Thanks for the enhancement! 👍

insights/specs/default.py
theyoyojo: > May I know why you remove the `-C` here? `-C` is added to avoid updating the cache, since the cache-update needs some network traffics, this issue has been complained by some CUs.

The motivation behind this parser is to enable a rule that will check if there are available kernel live patch modules (kpatches) that the user has not already installed. The mechanism for delivery of these modules is via the "kpatch-patch" rpm. kpatch is generally used to deliver critical security patches to systems that must maintain uptime, so by not updating the cache, we may end informing a user that there are no new critical available security patches available when in fact there are, leaving the user vulnerable.

So to make sure we don't give the user any false negative results, I update the cache.

insights/parsers/yum_list.py
theyoyojo: :)

insights/parsers/yum_list.py
theyoyojo: Added.

insights/specs/insights_archive.py
```diff
@@ -294,5 +294,6 @@ class InsightsArchiveSpecs(Specs):
     virt_what = simple_file("insights_commands/virt-what")
     woopsie = simple_file("insights_commands/find_.var.crash_.var.tmp_-path_.reports-_.whoopsie-report")
     yum_list_installed = simple_file("insights_commands/yum_-C_--noplugins_list_installed")
+    yum_list_available = simple_file("insights_commands/yum_--noplugins_list_available")
```
bfahr, line 297: @theyoyojo this spec will need to change to be consistent with the spec in `default`.

theyoyojo: OK, done.

theyoyojo:
Rebased onto latest master as well

PR #2526, opened 2020-04-01
Parser for dhclient in NetworkManager

jobselko:

### DIFF
--- a/insights/parsers/networkmanager_dhclient.py
+++ b/insights/parsers/networkmanager_dhclient.py
@@ -97,10 +97,6 @@ class NetworkManagerDhclient(Parser):
         >>> dhclient.has_vulnerable_block
         True
     """
-    def __init__(self, context):
-        self.has_vulnerable_block = None
-        super(NetworkManagerDhclient, self).__init__(context)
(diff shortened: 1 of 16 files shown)


insights/parsers/networkmanager_dhclient.py
```diff
@@ -0,0 +1,111 @@
...
+        fi
+
+    Examples:
+        >>> type(dhclient)
+        <class 'insights.parsers.networkmanager_dhclient.NetworkManagerDhclient'>
+        >>> dhclient.has_vulnerable_block
+        True
+    """
+    def __init__(self, context):
+        self.has_vulnerable_block = None
```
xiangce, line 101: If this `__init__` constructor is only aiming at adding this `has_vulnerable_block` attribute, it can be removed if you also agree. Since the `parse_content` will be called by the constructor of the base class `Parser` and `has_vulnerable_block` will be added there.

jobselko: Ok, I will remove it.

bfahr:
test this please

PR #2533, opened 2020-04-06
Tags validation

SteveHNH:
Inform the customer if `tags.conf` is improperly formatted. Also enable the ability to use `--validate` to check both tags.conf and remove.conf


SteveHNH:
closing this to fix some merge issues with other changes that happened

PR #2534, opened 2020-04-07
Parser for 'rpm -V coreutils procps procps-ng shadow-utils passwd sudo'

jobselko:

### DIFF
--- a/insights/parsers/rpm_v_packages.py
+++ b/insights/parsers/rpm_v_packages.py
@@ -1,0 +1,54 @@
+"""
+RpmVPackages - command ``/bin/rpm -V coreutils procps procps-ng shadow-utils passwd sudo``
+==========================================================================================
+"""
+
+from insights.core import CommandParser
+from insights.core.plugins import parser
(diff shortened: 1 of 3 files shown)


insights/parsers/rpm_v_packages.py
```diff
@@ -0,0 +1,54 @@
...
+        [('attributes', None), ('file', None), ('line', 'package procps is not installed'), ('mark', None)]
+        >>> sorted(rpm_v_packages.packages_list[1].items())
+        [('attributes', '..?......'), ('file', '/etc/sudoers'), ('line', '..?......  c /etc/sudoers'), ('mark', 'c')]
+"""
+
+    def parse_content(self, content):
+        self.packages_list = []
+
+        if len(content) == 0:  # rpm does not report any changes
+            return self.packages_list
```
xiangce, line 42: It's OK to remove this `if`, when content is empty, the following `for-loop` will be skipped naturally.
And the caller doesn't handle the return value of the `parse_content`, we do not need to return something in this case.

PR #2536, opened 2020-04-09
Add parser for "/etc/cron.daily/rhsmd"

huali027:

### DIFF
--- a/insights/parsers/cron_daily_rhsmd.py
+++ b/insights/parsers/cron_daily_rhsmd.py
@@ -19,7 +19,7 @@ class CronDailyRhsmd(Scannable):
         rhsmd_timeout=$config
         abc=$config
 
-    Examples::
+    Examples:
 
         >>> from insights.tests import context_wrap
         >>> from insights.parsers.cron_daily_rhsmd import CronDailyRhsmd
--- a/insights/parsers/cron_daily_rhsmd.py
+++ b/insights/parsers/cron_daily_rhsmd.py
@@ -13,13 +13,15 @@ class CronDailyRhsmd(Scannable):
     """
     Parse the ``/etc/cron.daily/rhsmd`` file.
 
-    Please refer to its super-class :class:`insights.core.Scannable` for more details.
-
     Sample input::
 
         config=$(grep -E "^processTimeout" /etc/rhsm/rhsm.conf | grep -Po "[0-9]+")
         rhsmd_timeout=$config
         abc=$config
+
+    .. note::
+        Please refer to its super-class :py:class:`insights.core.Scannable`
+        for full usage.
 
     Examples:
 
--- a/insights/parsers/cron_daily_rhsmd.py
+++ b/insights/parsers/cron_daily_rhsmd.py
@@ -13,6 +13,8 @@ class CronDailyRhsmd(Scannable):
     """
     Parse the ``/etc/cron.daily/rhsmd`` file.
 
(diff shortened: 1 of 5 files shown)


insights/parsers/cron_daily_rhsmd.py
```diff
@@ -0,0 +1,16 @@
...
+===============================================
+"""
+
+from insights.core import LogFileOutput
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+
+@parser(Specs.cron_daily_rhsmd)
+class CronDailyRhsmd(LogFileOutput):
```
bfahr, line 12: @huali027 I think its confusing to use `LogFileOutput` as the base parser here as there are methods/attributes in the base parser that would not work/be set.  I think you should just treat it as a text file and use the base `Parser` class.  Or if you prefer you could create a new base class for plain text parsers and provide search options to locate strings.  What do you think?

huali027: @bfahr, sure, thanks for your review.

insights/parsers/tests/test_cron_daily_rhsmd.py
```diff
@@ -0,0 +1,17 @@
...
+abc=$config
+""".strip()
+
+
+def test_docs():
+    env = {
+        'RHSMD_1': RHSMD_1
+    }
+    failed, total = doctest.testmod(cron_daily_rhsmd, globs=env)
+    assert failed == 0
```
xiangce, line 17: We should not rely the test on the test against doc_test only, even they have the same test content. Please add a normal test in the test module.

insights/parsers/cron_daily_rhsmd.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Parse the ``/etc/cron.daily/rhsmd`` file.
+
+    Sample input::
+
+        config=$(grep -E "^processTimeout" /etc/rhsm/rhsm.conf | grep -Po "[0-9]+")
+        rhsmd_timeout=$config
+        abc=$config
+
+    Examples::
```
xiangce, line 22: This should be `Example:`

insights/parsers/cron_daily_rhsmd.py
```diff
@@ -0,0 +1,34 @@
...
+        rhsmd_timeout=$config
+        abc=$config
+
+    Examples:
+
+        >>> from insights.tests import context_wrap
+        >>> from insights.parsers.cron_daily_rhsmd import CronDailyRhsmd
+        >>> CronDailyRhsmd.collect('config_lines', lambda n: n if "$config" in n else "")
+        >>> CronDailyRhsmd.any('one_config_line', lambda n: n if "$config" in n else "")
+        >>> rhsmd=CronDailyRhsmd(context_wrap(RHSMD_1))
```
bfahr, line 28: @huali027 thanks for making the changes in your parser, use of `Scannable` was a good call.  This example is a little misleading in that someone using this parser would not use it in this way with `context_wrap` since that function is only for testing.  I'd recommend you add lines 24-27 to your doc test to create the `rhsmd` object and then leave lines 26-27 here, but comment them out and add a reference to the `Scannable` class which provides more extensive documentation and examples.

huali027: @bfahr, sure, I'll update the PR, thanks.

PR #2538, opened 2020-04-11
adding a scandir() based option for get_all_files

jhjaggars:
This PR replaces the get_all_files function with an `os.scandir()` based implementation.  `os.scandir()` was added in 3.5 and offers a performance increase over `os.walk` since it reduces the number of system calls to stat files.

Signed-off-by: Jesse Jaggars <jjaggars@redhat.com>

### DIFF
--- a/insights/core/hydration.py
+++ b/insights/core/hydration.py
@@ -20,9 +20,10 @@ if hasattr(os, "scandir"):
 
 else:
     def get_all_files(path):
-        for f in archives.get_all_files(path):
-            if os.path.isfile(f) and not os.path.islink(f):
-                yield f
+        for root, _, files in os.walk(path):
+            for f in files:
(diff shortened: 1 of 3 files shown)


insights/core/hydration.py
```diff
@@ -8,13 +8,21 @@
...
+        with os.scandir(path) as it:
+            for ent in it:
+                if ent.is_dir(follow_symlinks=False):
+                    for pth in get_all_files(ent.path):
+                        yield pth
+                elif ent.is_file(follow_symlinks=False):
+                    yield ent.path
+
+else:
+    def get_all_files(path):
```
csams, line 22: I think we should remove `get_all_files` from insights.core.archives and just have the alternate implementations defined here since it's not used anywhere else.

jhjaggars: Like so?

PR #2541, opened 2020-04-13
Add parser for spec "ndctl list -Ni"

huali027:

### DIFF
--- a/insights/parsers/ndctl_list.py
+++ b/insights/parsers/ndctl_list.py
@@ -61,7 +61,7 @@ class NdctlListNi(JSONParser, CommandParser):
 
     @property
     def blockdev_list(self):
-        """ Return a list of the blockdev attribute for all the devices if it has this attribute"""
+        """ Return a list of the blockdev attribute of all the devices if it has the attribute"""
         return [item['blockdev'] for item in self.data if 'blockdev' in item]
 
     def get_block_dev(self, dev_name):
@@ -72,7 +72,7 @@ class NdctlListNi(JSONParser, CommandParser):
             dev_name (str): the blockdev name
 
         Returns:
-            dict: return a dict with all the info if there is block device else empty dict
+            dict: return a dict with all the info if there is the block device else empty dict
 
         """
         for item in self.data:
--- a/insights/parsers/ndctl_list.py
+++ b/insights/parsers/ndctl_list.py
@@ -59,10 +59,14 @@ class NdctlListNi(JSONParser, CommandParser):
         True
     """
 
+    def parse_content(self, content):
+        super(NdctlListNi, self).parse_content(content)
+        self._blockdevs = [item['blockdev'] for item in self.data if 'blockdev' in item]
+
     @property
     def blockdev_list(self):
         """ Return a list of the blockdev attribute of all the devices if it has the attribute"""
-        return [item['blockdev'] for item in self.data if 'blockdev' in item]
+        return self._blockdevs
 
     def get_blockdev(self, dev_name):
         """
--- a/insights/parsers/ndctl_list.py
+++ b/insights/parsers/ndctl_list.py
@@ -1,15 +1,20 @@
 """
+Commands about "libnvdimm" subsystem devices
+============================================
+
+This module contains the following parsers:
+
 NdctlListNi - command ``/usr/bin/ndctl list -Ni``
 =================================================
 """
 
-from insights.core import JSONParser
+from insights.core import JSONParser, CommandParser
 from insights.core.plugins import parser
 from insights.specs import Specs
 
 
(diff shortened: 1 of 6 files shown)


insights/parsers/ndctl_list.py
```diff
@@ -0,0 +1,64 @@
...
+=================================================
+"""
+
+from insights.core import JSONParser
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+
+@parser(Specs.ndctl_list)
+class NdctlListNi(JSONParser):
```
xiangce, line 12: Since this is a parser for command output, you'd better add the `CommandParser` as the bases.

insights/parsers/ndctl_list.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Sample output::
+
+        [
+            {
+                "dev":"namespace1.0",
+                "mode":"fsdax",
+                "map":"mem",
+                "size":811746721792,
+                "uuid":"6a7d93f5-60c4-461b-8d19-0409bd323a94",
```
xiangce, line 24: Is this a UUID from real data? if yes, better so some mask on it.

insights/parsers/ndctl_list.py
```diff
@@ -0,0 +1,64 @@
...
+        >>> type(ndctl_list)
+        <class 'insights.parsers.ndctl_list.NdctlListNi'>
+        >>> ndctl_list.blockdev_list
+        ['pmem1']
+        >>> ndctl_list.get_attr_by_dev('pmem1', 'mode')
+        'fsdax'
+    """
+
+    @property
+    def blockdev_list(self):
```
xiangce, line 58: For public interfaces and attributes (or properties), a brief docstring is needed.

insights/parsers/ndctl_list.py
```diff
@@ -0,0 +1,64 @@
...
+        return [item['blockdev'] for item in self.data if 'blockdev' in item]
+
+    def get_attr_by_dev(self, dev_name, attr_name):
```
xiangce, line 61: The same comment as above. Add better add the description for the `parameters` and `returns` for this function.

insights/parsers/ndctl_list.py
```diff
@@ -0,0 +1,64 @@
...
+        for item in self.data:
+            if dev_name == item.get('blockdev', ''):
+                return item.get(attr_name, '')
```
xiangce, line 64: Per the logic, when there is no such 'item' in `self.data`, `None` will be returned. but when there is no `attr_name` in `item['blockdev']` an empty string will be returned. Better uniform the return type.

And, AFAIK, for JSONParser, there are such cases, that a specified `key` is in the returned dict, but its value is "None" (null). The `key in dict` check can be used to avoid it, please have a try.

insights/specs/default.py
```diff
@@ -622,6 +622,7 @@ def lsmod_all_names(broker):
     mysqld_limits = foreach_collect(mysqld_pid, "/proc/%s/limits")
     named_checkconf_p = simple_command("/usr/sbin/named-checkconf -p")
     namespace = simple_command("/bin/ls /var/run/netns")
+    ndctl_list = simple_command("/usr/bin/ndctl list -Ni")
```
xiangce, line 625: Per the manual of `ndctl` there are various different `ndctl list` sub-commands, would you please rename this spec more specifically?

huali027:
@xiangce, sure, I'll update based on the comments, thanks.

insights/parsers/ndctl_list.py
```diff
@@ -0,0 +1,81 @@
+"""
+Commands about "libnvdimm" subsystem devices
```
xiangce, line 2: This title is from the manual of `ndctl` command but not `ndctl list` would you please double-check?
- https://www.mankier.com/1/ndctl-list

insights/parsers/ndctl_list.py
```diff
@@ -0,0 +1,81 @@
...
+                "sector_size":512,
+                "state":"disabled"
+            }
+        ]
+
+    Examples:
+
+        >>> type(ndctl_list)
+        <class 'insights.parsers.ndctl_list.NdctlListNi'>
+        >>> ndctl_list.blockdev_list
```
xiangce, line 56: Please fix the failed test on this line for python2

insights/parsers/ndctl_list.py
```diff
@@ -0,0 +1,81 @@
...
+        ['pmem1']
+        >>> ndctl_list.get_blockdev('pmem1').get('mode')
+        'fsdax'
```
xiangce, line 59: The same as above.

huali027:
@xiangce, thanks for pointing it out, I'll update it.

PR #2543, opened 2020-04-13
Validate tags

SteveHNH:
Catch invalid YAML issues related to tags.conf

We don't die if this fails, but we do throw in a warning for it.
Validate will now check that `tags.conf` is v alid as well

### DIFF
--- a/insights/client/collection_rules.py
+++ b/insights/client/collection_rules.py
@@ -13,11 +13,6 @@ import yaml
 import stat
 from six.moves import configparser as ConfigParser
 
-try:
-    from yaml import CLoader as Loader
-except ImportError:
-    from yaml import Loader
-
 from subprocess import Popen, PIPE, STDOUT
 from tempfile import NamedTemporaryFile
 from .constants import InsightsConstants as constants
--- a/insights/client/collection_rules.py
+++ b/insights/client/collection_rules.py
@@ -419,8 +419,9 @@ class InsightsUploadConf(object):
         else:
             try:
                 load_yaml(self.tags_file)
+                logger.info("%s loaded successfully", self.tags_file)
             except RuntimeError:
-                logger.warning("Unable to load %s\nInvalid YAML", self.tags_file)
+                logger.warning("Invalid YAML. Unable to load %s", self.tags_file)
                 return None
 
     def validate(self):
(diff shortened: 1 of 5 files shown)


insights/client/config.py
```diff
@@ -242,6 +242,10 @@
         # non-CLI
         'default': os.path.join(constants.default_conf_dir, 'remove.conf')
     },
+    'tags_file': {
+        # non-CLI
+        'default': os.path.join(constants.default_conf_dir, 'tags.conf')
```
gravitypriest, line 247: should be `tags.yaml`

insights/client/constants.py
```diff
@@ -10,6 +10,7 @@ class InsightsConstants(object):
     command_blacklist = ('rm', 'kill', 'reboot', 'shutdown')
     default_conf_dir = os.getenv('INSIGHTS_CONF_DIR', default='/etc/insights-client')
     default_conf_file = os.path.join(default_conf_dir, 'insights-client.conf')
+    default_tags_file = os.path.join(default_conf_dir, 'tags.conf')
```
gravitypriest, line 13: should be `tags.yaml` :)

insights/client/collection_rules.py
```diff
@@ -408,10 +409,25 @@ def get_rm_conf(self):
...
+        Try to load the tags.conf file
+        '''
+        if not os.path.isfile(self.tags_file):
+            logger.info("%s does not exist", self.tags_file)
+            return None
+        else:
+            try:
+                load_yaml(self.tags_file)
+            except RuntimeError as e:
+                logger.info(e)
```
gravitypriest, line 423: maybe `logger.warning` in this function to match remove.conf validation errors?

insights/client/config.py
SteveHNH: ah! didn't realize we had made that switch. On it

SteveHNH:
test this please

koalakangaroo:
@SteveHNH - May I know what the tags.conf file is used for in the client?

SteveHNH:
@koalakangaroo Tags are used to give machines more searchable fields in inventory, I believe. I think the main use right now is for groups. Multiple uploaders will submitting tags. This is the client implementation of it. Satellite can do it as well.

PR #2544, opened 2020-04-14
Remove -R from ls_etc and only list specified sub-dirs

xiangce:
- Remove -R to decrease the size


xiangce:
> Hey @xiangce, the `simple_command` in `default.py` will create a single parser, but the `glob_file` in `insights_archive.py` will create a list of them. I think we'll need to mangle the path from the command and just use `simple_file` for the archive.

@csams, I thought without the `multi_output=True` when `RegistryPoint()` it would create a single parser. but, seems I'm wrong. Thanks.

To active both solutions mentioned in the Jira, I think we have to make a combiner for LsEtc, I submit a draft version, please have a check. And please let me know if you have any better solutions.

xiangce:
OK, to simplify the process and fix the issue quickly, I remove the support of `ls_etc command_with_args` from this patch. @bfahr  please help review. Thanks.

PR #2546, opened 2020-04-14
Prefer igzip over gzip when handling tarballs

jhjaggars:
This allows callers to specify exactly which decompression binary to
use, skipping the content-type check and allowing for use of novel
programs such as igzip.

Signed-off-by: Jesse Jaggars <jjaggars@redhat.com>

### DIFF
--- a/insights/core/archives.py
+++ b/insights/core/archives.py
@@ -4,7 +4,7 @@ import logging
 import os
 import tempfile
 from contextlib import contextmanager
-from insights.util import fs, subproc
+from insights.util import fs, subproc, which
 from insights.util.content_type import from_file as content_type_from_file
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


csams:
It also users `igzip` when available during tarball decompression instead of `gunzip`. Fixes #2547

insights/core/archives.py
```diff
@@ -62,12 +62,15 @@ def _tar_flag_for_content_type(self, content_type):
             raise InvalidContentType(content_type)
         return flag
 
-    def from_path(self, path, extract_dir=None, content_type=None):
+    def from_path(self, path, extract_dir=None, content_type=None, decompressor=None):
```
csams, line 65: This is exposed through the extract function below, and extract can't give zipped files custom decompressors.

csams: I removed the decompressor option since we can bypass mimetype detection by passing content_type explicitly.

PR #2549, opened 2020-04-16
Add parser for kdump-capture.service of dracut module

huali027:

### DIFF
--- a/docs/shared_parsers_catalog/dracut_modueles.rst
+++ b/docs/shared_parsers_catalog/dracut_modueles.rst
@@ -1,3 +1,0 @@
-.. automodule:: insights.parsers.dracut_modueles
-   :members:
-   :show-inheritance:
--- a/docs/shared_parsers_catalog/dracut_modueles.rst
+++ b/docs/shared_parsers_catalog/dracut_modueles.rst
@@ -1,0 +1,3 @@
+.. automodule:: insights.parsers.dracut_modueles
+   :members:
+   :show-inheritance:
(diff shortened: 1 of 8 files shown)


docs/shared_parsers_catalog/dracut_modueles.rst
```diff
@@ -0,0 +1,3 @@
+.. automodule:: insights.parsers.dracut_modueles
```
xiangce, line 1: I believe the name of the module should be `dracut_modules`.

BTW, would you please consider rename it accordingly?

insights/parsers/dracut_modueles.py
```diff
@@ -0,0 +1,24 @@
+"""
+Dracut modules
```
xiangce, line 2: I don't think this title is enough for this module since it's not the module itself but the configuration of the related services.

Please refer to https://github.com/RedHatInsights/insights-core/wiki#checklist-for-reviewers

insights/parsers/dracut_modueles.py
```diff
@@ -0,0 +1,24 @@
...
+from insights import parser, IniConfigFile
+from insights.specs import Specs
+
+
+@parser(Specs.dracut_kdump_capture_service)
+class DracutModuleKdumpCaptureService(IniConfigFile):
+    """
+    Class for parsing the `/usr/lib/dracut/modules.d/99kdumpbase/kdump-capture.service` file.
+
+    .. note::
```
xiangce, line 20: Please add exact examples for this parser here, instead of putting a reference here.

xiangce:
Would you please modify the Git Message shorter?

FYI: http://chris.beams.io/posts/git-commit/

huali027:
@xiangce, sure, I'll update according to the comments, thanks.

PR #2561, opened 2020-04-24
create thread for sdnotify

gravitypriest:
Summary of changes:

- Remove per-command and per-upload attempt calls to sdnotify. This decouples it from the collection code and will make core collection drop-in easier
- Create background thread on collect-and-upload phase to ping sdnotify every 30s
- Check PID, existence of systemd-notify path, and `NOTIFY_SOCKET` before loop init instead of on each call to sdnotify


insights/client/utilities.py
```diff
@@ -318,6 +311,33 @@ def systemd_notify(pid):
...
+        # running standalone, not via systemd job
+        return
+    if not os.path.exists('/usr/bin/systemd-notify'):
+        # RHEL 6, no systemd
+        return
+
+    def _sdnotify_loop():
+        while True:
+            # run sdnotify every 30 seconds
+            _systemd_notify(pid)
```
subpop, line 333: Would it make sense to hook up an exit condition to this thread function to the return value of `_systemd_notify`? Something like:

```python
def _sdnotify_loop():
    ret = True
    while ret:
        global ret  # I think this is how Python's goofy variable scoping works...
        ret = _systemd_notify(pid)
```

That way if `_systemd_notify` needs to abort for any reason (like a non-zero exit code from `systemd-notify`, or a Popen failure trying to create the child process), it can safely abort the thread.

fstavela:
test this please

fstavela:
test this please

fstavela:
test this please

PR #2563, opened 2020-04-25
Fixes improper calculation of core_total

### LINKED ISSUE #2555: Reporting the wrong number of cores_total
Hi folks...

I think we are noticing a bug that comes from a misunderstanding of cpuinfos processor[N]."cpu_cores" value.

That value is not a count of cores per "cpu". Instead it is a count of the cores on the logical cpus parent package.

i.e. for 2x 6 core CPUs each cpu stanza (12 of them) should show "cpu cores : 6"
But there are not 12x6 cpus. There are 12.

In a big.little arch with 1x 4 core and 1x 2 core you'd see
processor       : 0
cpu cores       : 4
...
processor       : 1
cpu cores       : 4
...
processor       : 2
cpu cores       : 4
...
processor       : 3
cpu cores       : 4
...
processor       : 4
cpu cores       : 2
...
processor       : 5
cpu cores       : 2

iphands:
I think we can reject this and instead remove the core_total and instruct people to use lscpu (per @bfahr s suggestion)... but just to be thorough this should be accurate.

Closes #2555

### DIFF
--- a/insights/parsers/cpuinfo.py
+++ b/insights/parsers/cpuinfo.py
@@ -240,7 +240,12 @@ class CpuInfo(LegacyItemAccess, Parser):
     @defaults()
     def core_total(self):
         """
-        str: Returns the total number of cores for the server if available, else None.
+        int: Returns the total number of cores for the server if available, else None.
+
+        .. warning::
+            This function is deprecated.  Please use the
+            :py:class:`insights.parsers.lscpu.LsCPU` class attribute
+            ``info['Cores per socket']`` and ``info['Sockets']`` values instead.
         """
         if self.data and 'cpu_cores' in self.data:
             # I guess we can't get this fancey on older versions of RHEL
--- a/insights/parsers/cpuinfo.py
+++ b/insights/parsers/cpuinfo.py
@@ -243,7 +243,16 @@ class CpuInfo(LegacyItemAccess, Parser):
         str: Returns the total number of cores for the server if available, else None.
         """
         if self.data and 'cpu_cores' in self.data:
-            return sum([int(c) for c in self.data['cpu_cores']])
+            # I guess we can't get this fancey on older versions of RHEL
+            # return sum({e['sockets']: int(e['cpu_cores']) for e in self}.values())
+            physical_dict = {}
+            for e in self:
+                # we should rename sockets here to physical_ids as cpuinfo
+                # has it there can be many physical_ids per socket
+                # see fgrep 'physical id' /proc/cpuinfo on a single
+                # package system
+                physical_dict[e['sockets']] = int(e['cpu_cores'])
+            return sum(physical_dict.values())
         else:
             return None
 
--- a/insights/parsers/tests/test_cpuinfo.py
+++ b/insights/parsers/tests/test_cpuinfo.py
@@ -3,6 +3,962 @@ from insights.parsers import cpuinfo
 from insights.tests import context_wrap
 import doctest
 
+CPUINFO_HETERO = """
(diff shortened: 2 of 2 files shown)


bfahr:
@iphands I think the better approach is to remove the `core_total` attribute from this parser since it is calculated properly by the `lscpu` command which is implemented via the `lscpu` parser.  The goal is to keep the parser consistent with the actual information provided by the spec.  In this case `cpuinfo` could have very different information depending upon which hardware/platform the data is being collected from.  The `lscpu` command should provide the best information regardless of hardware/platform and therefore should be the source for that information.

bfahr:
@iphands also this is a good use case for a combiner that would include information from the `lscpu`, `cpuinfo` and `dmidecode` parsers to provide a complete picture of the processor configuration information.

iphands:
> @iphands I think the better approach is to remove the core_total attribute from this parser since it is calculated properly by the lscpu command which is implemented via the lscpu parser. The goal is to keep the parser consistent with the actual information provided by the spec. In this case cpuinfo could have very different information depending upon which hardware/platform the data is being collected from. The lscpu command should provide the best information regardless of hardware/platform and therefore should be the source for that information.

@bfahr can yall do it? is there any deprecation process yall follow for down stream contributors?

bfahr:
@iphands yes, generally we do not make any changes to break downstream users.  The exception to that is when the change fixes a bug.  Since this attribute is being calculated incorrectly and is therefore a bug it would qualify as an exception.  Its not clear to me that we can cover the proper calculation of total cores for all architectures so in this case I think it would be better to break things then fix a subset of cases.  Does that sound reasonable to you?

iphands:
> @iphands yes, generally we do not make any changes to break downstream users. The exception to that is when the change fixes a bug. Since this attribute is being calculated incorrectly and is therefore a bug it would qualify as an exception. Its not clear to me that we can cover the proper calculation of total cores for all architectures so in this case I think it would be better to break things than fix a subset of cases. Does that sound reasonable to you?

/shrug I dont really care as long as we take action. I think the proposed patch works for the core_total even in a heterogeneous env. Cores per package would be impossible there but not cores_total.

bfahr:
@iphands you can fix the DCO error by following the instructions in the DCO test details.  Take a look at [this page](https://probot.github.io/apps/dco/) for more info about this test and the Developer Certificate of Origin.

insights/parsers/cpuinfo.py
```diff
@@ -243,7 +243,16 @@ def core_total(self):
         str: Returns the total number of cores for the server if available, else None.
         """
```
bfahr, line 244: @iphands can you please add the deprecation feature to this method?  See this [document](https://insights-core.readthedocs.io/en/latest/api.html?#feature-deprecation).  Also please provide a warning in the docstring to notify any users that `insights.parsers.lscpu.LsCPU.info['Cores per socket']` should be used instead, here is an [example](https://github.com/RedHatInsights/insights-core/blob/827faf50b5b34d78a2fd21167d69b9ea2aff3abc/insights/parsers/nfs_exports.py#L142).

iphands: Bob feel free to add it man.

bfahr: @iphands before I push those changes, can you fix your original commit by adding the `--signoff`?  See the [DCO results](https://github.com/RedHatInsights/insights-core/pull/2563/checks?check_run_id=618266058) for the necessary commands.

iphands: For sure

PR #2575, opened 2020-05-05
This PR adds a new lvm parser that depends on the lvmconfig command.

jhjaggars:
This PR adds two new specs for gathering `lvmconfig` data, as well as a parser for that data.

This PR does not attempt to integrate those into the combiner, but that should happen as a follow up to this change.


jhjaggars:
@kergon Please take a look to see if this is going in the right direction.

kergon:
Here's a test case that separates the old and new code (for the lvmconfig parsing):

On a default install, add the following line to the bottom of /etc/lvm/lvm.conf:
tags { tag1{} }

Create the file /etc/lvm/lvm_tag1.conf as follows:
activation {
thin_pool_autoextend_percent = 25
}

This changes the threshold for thin pool autoextension from 20% to 25%. (It defines a host tag that instructs lvm to read in an additional config file with a setting thats override the original.)

The current version of insights-client incorrectly reports 20. The new version should report 25.

insights/specs/default.py
```diff
@@ -558,6 +558,10 @@ def semid(broker):
     lsscsi = simple_command("/usr/bin/lsscsi")
     lvdisplay = simple_command("/sbin/lvdisplay")
     lvm_conf = simple_file("/etc/lvm/lvm.conf")
+    lvmconfig = first_of([
+        simple_command("/usr/sbin/lvmconfig --type full"),
+        simple_command("/usr/sbin/lvm dumpconfig --type full"),
+    ])
```
bfahr, line 564: @jhjaggars you'll also need to add a file spec in `insights_archive.py`.

kergon:
Capture stdout separately from stderr, so the output doesn't get contaminated by any warning messages the tool issues.

jhjaggars:
> Capture stdout separately from stderr, so the output doesn't get contaminated by any warning messages the tool issues. 

This check is in anticipation of a port of this collection to sosreport.  We won't have the option to capture the two parts separately.

PR #2576, opened 2020-05-05
always create archive_tmp_dir for compatibility

gravitypriest:
We opted not to create `archive_tmp_dir` when `obfuscate=True` because soscleaner does its own tempfile management and archive creation.

Compliance is using the `create_tar_file` function which expects this directory to exist, so we kinda pulled the rug out from under them there if obfuscation happens to be enabled. So let's always create the dir just in case.


patchkez:
test this please

PR #2577, opened 2020-05-08
Compliance - only scan internal policies

dLobatog:
Without this external=false change, we are scanning external policies on
'insights-client --compliance', but those weren't manually assigned and
should not be scanned. Furthermore, you could have 2 policies, internal
and external with the same ref_id scanned.


xiangce:
Close it for now, please open it in another PR if necessary

PR #2584, opened 2020-05-11
Disk usage parser

Rohan21Lobo:
Add DiskUsage parser and combiner for directories.


insights/parsers/du.py
```diff
@@ -0,0 +1,91 @@
...
+
+    In output, the size value can be with or without a size unit suffix. If size present, it should be consistent across all records and same as initialized block_unit. So "du -h" which may output in different units is not supported.
+
+    Examples : 
+    For --block-size=4K , block_unit='k' and multiplier=4
+    For --block-size=1M , block_unit='M' and multiplier=1
+    For --bytes,          block_unit='B' and multiplier=1
+    """
+
+    def __init__(self, context, block_unit='K', multiplier=1):
```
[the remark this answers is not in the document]
Rohan21Lobo: I made the changes and removed all the unit parsing logic. Can you review again?

insights/parsers/du.py
```diff
@@ -0,0 +1,91 @@
...
+
+            line_split = line.split(None, 1)
+            if len(line_split) != 2:
+                raise ParseException("Could not parse line: '{}'".format(line))
+            size, path = line_split
+
+            # Strip '.' if relative paths
+            path = path.lstrip('.').rstrip()
+
+            pattern = re.compile(r'^(\d*)([K,M,G,T,P,E,Z,Y][i]*[B]*)*$')
```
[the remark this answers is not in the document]
Rohan21Lobo: Done

Rohan21Lobo:
> A docentry should be added for this new module to "docs/shared_parsers_catalog/"

I add the child class parser to this PR and added the child parser to docs/shared_parsers_catalog/ . Is base parser class documentation also required in docs/shared_parsers_catalog/ ?

insights/parsers/du.py
```diff
@@ -0,0 +1,91 @@
+import re
+
+from insights import Parser, LegacyItemAccess
+from insights.parsers import ParseException
+
+
+class DiskUsage(Parser, LegacyItemAccess):
```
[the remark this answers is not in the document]
Rohan21Lobo: Done, Please review

insights/parsers/tests/test_du_candlepin_activemq_artemis.py
```diff
@@ -0,0 +1,31 @@
...
+""".strip()
+
+
+def test_du_candlepin_activemq_artemis():
+    du_candlepin_activemq = du_candlepin_activemq_artemis.DuVarLibCandlepinActivemqArtemis(context_wrap(DU_ACTIVEMQ))
+    assert du_candlepin_activemq == {'/var/lib/candlepin/activemq-artemis': 725320}
+    assert '/var/lib/candlepin/activemq-artemis' in du_candlepin_activemq
+    assert du_candlepin_activemq.get('/var/lib/candlepin/activemq-artemis') == 725320
+
+    with pytest.raises(ContentException) as exc:
```
Rohan21Lobo, line 23: @xiangce  Hi, Currently the way parser is written, For singe line error data, tests will have to test ContentException rather than ParserException. Should this be okay?

insights/parsers/du_candlepin_activemq_artemis.py
```diff
@@ -0,0 +1,22 @@
...
+    Parser to get disk usage of "/var/lib/candlepin/activemq-artemis" directory
+
+    Examples:
+        >>> '/var/lib/candlepin/activemq-artemis' in du_candlepin_activemq
+        True
+        >>> du_candlepin_activemq.get('/var/lib/candlepin/activemq-artemis')
+        725320
+        >>> int(du_candlepin_activemq.get('/var/lib/candlepin/activemq-artemis') / 1024) # to MiB
+        708
+    """
```
xiangce, line 22: There should be a `pass`

insights/core/__init__.py
```diff
@@ -1663,6 +1663,63 @@ def path_entry(self, path):
...
+            size, path = line_split
+            # Strip '.' if relative paths and skip if current directory
+            path = path.lstrip('.').rstrip()
+            if path == '':
+                continue
+            try:
+                size = int(size)
+            except ValueError:
+                raise ParseException("Could not parse line: '{0}'".format(line))
+            self[path] = size
```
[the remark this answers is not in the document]
Rohan21Lobo: I implemented the change. However I am still converting to int and storing. 

The intention of using this parser is going to be get the size and do some comparison or calculations in plugins so I think storing as int may be better. What are your views on this?

insights/core/__init__.py
```diff
@@ -1663,6 +1663,63 @@ def path_entry(self, path):
...
+
```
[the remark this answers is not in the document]
Rohan21Lobo: Done, 
https://github.com/RedHatInsights/insights-core/pull/2584/commits/aa07c31b5bac82bd4f2bc75d47315c5839f35bbb#diff-e3ed9166e35444a62975de12bf0d55b7R1721-R1722

insights/core/__init__.py
```diff
@@ -1663,6 +1663,63 @@ def path_entry(self, path):
...
+        for line in content:
+            if du_error in line:
+                continue
+
+            line_split = line.split(None, 1)
+            if len(line_split) != 2:
+                raise ParseException("Could not parse line: '{0}'".format(line))
+            size, path = line_split
+            # Strip '.' if relative paths and skip if current directory
+            path = path.lstrip('.').rstrip()
```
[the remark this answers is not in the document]
Rohan21Lobo: Sounds better. Add the check : 
https://github.com/RedHatInsights/insights-core/commit/aa07c31b5bac82bd4f2bc75d47315c5839f35bbb#diff-e3ed9166e35444a62975de12bf0d55b7R1714-R1715

Rohan21Lobo:
> I just thought about this, the `du` command is open for any different kind of `dirs`, means we cannot predict how many `dirs` will be added to be a new parser. Just like the `FileListing` for these `ls ` commands.
> 
> How about making it an open spec, like the `ls_etc`
> 
> https://github.com/RedHatInsights/insights-core/blob/e990b0d3536dd69a6a575fa47a29998bdc0a29c7/insights/specs/default.py#L525-L528
> 
> We can make a predefined list (e.g. `du_dirs_list` as below) that can be used by the major entry of the `du` spec, every time when new `dir`s are required, we can only add the new `dir` to the predefined list, without adding a new spec and without making a new parser every time as well (but, we need to update the spec (in `uploader.json` and `insights_archive.py` for now, since the current client depends on the `uploader.json` and it doesn't support to read the list dynamically)
> 
> ```
> du_dirs_list = sorted(['/var/lib/candlepin/activemq-artemis'])
> du_dirs = simple_command("du -s -k {0}".format(' '.join(du_dirs_list))) 
> ```
> 
> In fact, we could have merged the `ls` commands with the same parameters into one single spec. I'm not sure why we didn't do that in this way at that time. But for the new `'du'`, we could have a try.
> 
> How do you think? @bfahr any thoughts?



> I just thought about this, the `du` command is open for any different kind of `dirs`, means we cannot predict how many `dirs` will be added to be a new parser. Just like the `FileListing` for these `ls ` commands.
> 
> How about making it an open spec, like the `ls_etc`
> 
> https://github.com/RedHatInsights/insights-core/blob/e990b0d3536dd69a6a575fa47a29998bdc0a29c7/insights/specs/default.py#L525-L528
> 
> We can make a predefined list (e.g. `du_dirs_list` as below) that can be used by the major entry of the `du` spec, every time when new `dir`s are required, we can only add the new `dir` to the predefined list, without adding a new spec and without making a new parser every time as well (but, we need to update the spec (in `uploader.json` and `insights_archive.py` for now, since the current client depends on the `uploader.json` and it doesn't support to read the list dynamically)
> 
> ```
> du_dirs_list = sorted(['/var/lib/candlepin/activemq-artemis'])
> du_dirs = simple_command("du -s -k {0}".format(' '.join(du_dirs_list))) 
> ```
> 
> In fact, we could have merged the `ls` commands with the same parameters into one single spec. I'm not sure why we didn't do that in this way at that time. But for the new `'du'`, we could have a try.
> 
> How do you think? @bfahr any thoughts?

I like the idea and wanted to have this kind of implmentation but if all directories are passed to a single command and the list grows long in future, even a single bad directory among them will cause failure to whole spec. For example if the directory is mounted over nfs, it takes time to calculate size and command will eventually timeout leading to nothing being collected.

What if we maintain a list of directories and then use `foreach_execute`  on them ? 

The combined approach takes out the possibility of getting sizes of files in directory which will use du -a </on/some/dir>

xiangce:
> I like the idea and wanted to have this kind of implmentation but if all directories are passed to a single command and the list grows long in future, even a single bad directory among them will cause failure to whole spec. For example if the directory is mounted over nfs, it takes time to calculate size and command will eventually timeout leading to nothing being collected.
> 
> What if we maintain a list of directories and then use `foreach_execute` on them ?
> 
> The combined approach takes out the possibility of getting sizes of files in directory which will use du -a </on/some/dir>
Makes sense! And the `foreach_execute` sounds a good idea. But better make a combiner to merge all the collected `dirs` and wrap into a dict to provide a uniform access entry for all `dirs`.

insights/core/__init__.py
```diff
@@ -1663,6 +1663,65 @@ def path_entry(self, path):
...
+        4       /var/lib/logrotate
+        5492	/var/lib/mlocate
+        20      /var/lib/NetworkManager
+        186484  /var/lib/pgsql
+        856	    /var/lib/rhsm
+        110712	/var/lib/rpm
+        4	    /var/lib/rsyslog
+        64	    /var/lib/systemd
+        15200	/var/lib/yum
+    """
```
bfahr, line 1693: @Rohan21Lobo since this is a base class that can be used in other parsers please include an `Examples:` section to demonstrate how it is used and please add a `Raises:` section to your documentation here to identify the exceptions raised.  In particular include the `ContentException` that can be raised by the `CommandParser`.

insights/core/__init__.py
```diff
@@ -1663,6 +1663,65 @@ def path_entry(self, path):
...
+    column as space size with integer values only and second as pathname which
+    can be a file or directory. Space size with decimal values or unit suffixes
+    like M, GiB is not supported.
+
+    du command produces output in 1K blocks unless block size is specified in
+    command options or an environment variable. This parser is intended to be
+    used only with default block size of 1K which is also equal to plain "du"
+    or "du -k".
+
+    Sample input data:
```
bfahr, line 1681: Please include `::` at the end of this line to correctly format the sample input below.  You can review the rendered documentation following the instructions in the [documentation guidelines](https://insights-core.readthedocs.io/en/latest/docs_guidelines.html#testing-your-docstring).

insights/parsers/du_candlepin_activemq_artemis.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+from insights import DiskUsage, parser
+from insights.specs import Specs
+
+
+@parser(Specs.du_candlepin_activemq_artemis)
+class DuVarLibCandlepinActivemqArtemis(DiskUsage):
+    """
+    Parser to get disk usage of "/var/lib/candlepin/activemq-artemis" directory
+
```
bfahr, line 14: Please provide sample input data here to make it clear where the data in the examples comes from.

insights/core/__init__.py
```diff
@@ -1663,6 +1663,65 @@ def path_entry(self, path):
...
+        56      /var/lib/alternatives
+        4       /var/lib/logrotate
+        5492	/var/lib/mlocate
+        20      /var/lib/NetworkManager
+        186484  /var/lib/pgsql
+        856	    /var/lib/rhsm
+        110712	/var/lib/rpm
+        4	    /var/lib/rsyslog
+        64	    /var/lib/systemd
+        15200	/var/lib/yum
```
bfahr, line 1692: This looks like it might be a cut/paste formatting error with the alignment of the paths?

bfahr:
> I just thought about this, the `du` command is open for any different kind of `dirs`, means we cannot predict how many `dirs` will be added to be a new parser. Just like the `FileListing` for these `ls ` commands.
> 
> How about making it an open spec, like the `ls_etc`
> https://github.com/RedHatInsights/insights-core/blob/e990b0d3536dd69a6a575fa47a29998bdc0a29c7/insights/specs/default.py#L525-L528
> 
> We can make a predefined list (e.g. `du_dirs_list` as below) that can be used by the major entry of the `du` spec, every time when new `dir`s are required, we can only add the new `dir` to the predefined list, without adding a new spec and without making a new parser every time as well (but, we need to update the spec (in `uploader.json` and `insights_archive.py` for now, since the current client depends on the `uploader.json` and it doesn't support to read the list dynamically)
> 
> ```
> du_dirs_list = sorted(['/var/lib/candlepin/activemq-artemis'])
> du_dirs = simple_command("du -s -k {0}".format(' '.join(du_dirs_list))) 
> ```
> 
> In fact, we could have merged the `ls` commands with the same parameters into one single spec. I'm not sure why we didn't do that in this way at that time. But for the new `'du'`, we could have a try.
> 
> How do you think? @bfahr any thoughts?

@xiangce I think we can wait until we add the second `du` command and then consider that change.

insights/core/__init__.py
Rohan21Lobo: There were mix of spaces and tabs before here. I have corrected them with only tab characters as in original output but it still shows misaligned in github. I do not know how single tab characters are adjusted in terminal to show like fix width column data. If same file is viewed in vim,  the columns are aligned.

insights/parsers/du_candlepin_activemq_artemis.py
Rohan21Lobo: removed the child parser class and implemented the combiner approach

insights/core/__init__.py
Rohan21Lobo: Done.

insights/core/__init__.py
Rohan21Lobo: Done

Rohan21Lobo:
@bfahr @xiangce  I have implemented the combined approach. Please review again and let me know if any changes suggested.

insights/parsers/du.py
```diff
@@ -0,0 +1,95 @@
+"""
+Disk Usage parsers
+==================
+
+Module for the processing of output from the ``du`` command.
+
+Parsers provided by this module include:
+
+DiskUsageDirs - command ``du -s -k {directory}``
```
xiangce, line 9: Typo here: DiskUsageDirs -> DiskUsageDir

Rohan21Lobo: Corrected

insights/combiners/du.py
```diff
@@ -0,0 +1,43 @@
...
+    values.
+
+    Sample input data for du commands as parsed by the parsers::
+
+    # Output of the command:
+    # /bin/du -s -k /var/log
+    553500	/var/log
+
+    # Output of the command:
+    # /bin/du -s -k /var/log
```
bfahr, line 29: Should this be `/var/lib`?  Looks like a cut-and-paste error.  Also the example input data needs to be indented.  Please following the [documentation guidelines](https://insights-core.readthedocs.io/en/latest/docs_guidelines.html#testing-your-docstring) to generate the documentation output and check it for errors.

Rohan21Lobo: Apology, Fixed flake8, I will be careful from next time.

Fixed the typo, /var/log to /var/lib/pgsql

Also did some minor doc changes.

PR #2585, opened 2020-05-11
Parser for ImageMagick policy.xml file

jobselko:

### DIFF
--- a/insights/parsers/imagemagick_policy.py
+++ b/insights/parsers/imagemagick_policy.py
@@ -57,8 +57,8 @@ class ImageMagickPolicy(XMLParser):
         <class 'insights.parsers.imagemagick_policy.ImageMagickPolicy'>
         >>> len(imagemagick_policy.policies)
         10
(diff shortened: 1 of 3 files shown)


insights/parsers/imagemagick_policy.py
```diff
@@ -0,0 +1,70 @@
...
+          <policy domain="coder" rights="none" pattern="TEXT" />
+          <policy domain="coder" rights="none" pattern="LABEL"/>
+          <policy domain="path" rights="none" pattern="@*"/>
+        </policymap>
+
+    Examples:
+        >>> type(imagemagick_policy)
+        <class 'insights.parsers.imagemagick_policy.ImageMagickPolicy'>
+        >>> len(imagemagick_policy.policies)
+        10
```
jsvob, line 59: I think it might be good to have at least one test testing/showing the parsed policies, not just their length. That might help others to quickly gain an accurate expectation of the output when integrating it in rules.

PR #2586, opened 2020-05-12
Add links to render data for content formatters

Rohan21Lobo:
This makes links attribute available for use in jinja CONTENT.

If the rule writer wants to show links in the content explicitly, the links dict will be available to use rather than passing extraneous data to make_ funcs.


Rohan21Lobo:
I had first made changes in text, markdown and template formatters individually but since the render function is commonly used by all of them, implemented the change in render function.

PR #2587, opened 2020-05-12
Use determine_hostname instead of get_canonical_facts()['fqdn']

akofink:
@jcrafts recommended this method instead of canonical facts to determine
the hostname of the client system. This fixes RHCLOUD-6495.


jcrafts:
Please remove @jcrafts. I should not be getting emails regarding this subject " the real jcrafts"

PR #2593, opened 2020-05-14
Add markers to syslog format tests

bfahr:
* Add new marker syslog_format to syslog formatter tests to allow
  turning them off when testing in a container that doesn't allow calls
  to ``getuser``
* Turn off tests with the `-m 'not syslog_format'` switch to pytest


bfahr:
Fix #2594

PR #2598, opened 2020-05-18
Add parser "corosync_cmapctl"

huali027:

### DIFF
--- a/insights/parsers/corosync_cmapctl.py
+++ b/insights/parsers/corosync_cmapctl.py
@@ -5,15 +5,18 @@ CorosyncCmapctl - Command ``corosync-cmapctl [params]``
 This module parses the output of the ``corosync-cmapctl [params]`` command.
 """
 
-from insights import parser, CommandParser, LegacyItemAccess
+from insights import parser, CommandParser
 from insights.parsers import SkipException, ParseException
 from insights.specs import Specs
 
 
 @parser(Specs.corosync_cmapctl)
-class CorosyncCmapctl(CommandParser, LegacyItemAccess):
+class CorosyncCmapctl(CommandParser, dict):
     """
     Class for parsing the `/usr/sbin/corosync-cmapctl [params]` command.
+    All lines are stored in the dictionary with the left part of the equal
+    sign witout parenthese info as the key and the right part of equal sign
+    as the value.
 
     Typical output of the command is::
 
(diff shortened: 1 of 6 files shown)


insights/parsers/corosync_cmapctl.py
```diff
@@ -0,0 +1,65 @@
...
+This module parses the output of the ``corosync-cmapctl [params]`` command.
+"""
+
+from insights import parser, CommandParser, LegacyItemAccess
+from insights.parsers import SkipException, ParseException
+from insights.specs import Specs
+
+
+@parser(Specs.corosync_cmapctl)
+class CorosyncCmapctl(CommandParser, LegacyItemAccess):
```
xiangce, line 14: Please inherit it from `dict` instead of `LegacyItemAccess`

huali027: > Please inherit it from `dict` instead of `LegacyItemAccess`

Sure, I'll update it, thanks.

insights/parsers/corosync_cmapctl.py
```diff
@@ -0,0 +1,64 @@
...
+        True
+        >>> corosync['internal_configuration.service.0.name']
+        'corosync_cmap'
+
+    Raises:
+        SkipException: When there is no content
+        ParseException: When there is no "=" in the content
+
+    Attributes:
+        stats_schedmiss (dict): The lines which start with
```
xiangce, line 44: This is not a good idea to access the `"keys"` of `stats.schedmiss`, better change it to a `get_stats_schedmiss` interface and fetch it from the `dict-parser` instead, to avoid storing duplicate data.

insights/parsers/corosync_cmapctl.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Typical output of the command is::
+
+        config.totemconfig_reload_in_progress (u8) = 0
+        internal_configuration.service.0.name (str) = corosync_cmap
+        internal_configuration.service.0.ver (u32) = 0
+        internal_configuration.service.1.name (str) = corosync_cfg
+        internal_configuration.service.1.ver (u32) = 0
+        internal_configuration.service.2.name (str) = corosync_cpg
+        internal_configuration.service.2.ver (u32) = 0
```
xiangce, line 29: Btw, don't you think the unit (e.g. u32) is necessary?

insights/parsers/corosync_cmapctl.py
huali027: > This is not a good idea to access the `"keys"` of `stats.schedmiss`, better change it to a `get_stats_schedmiss` interface and fetch it from the `dict-parser` instead, to avoid storing duplicate data.

Sure, I'll update it.

insights/parsers/corosync_cmapctl.py
huali027: > Btw, don't you think the unit (e.g. u32) is necessary?

Yes, it is mentioned in the Jira ticket too. The info is just the type of value. so no need to store it.

xiangce:
No need to update the uploader.json, since the foreach_execute is not supported for now, and it's necessary to run the three commands in strict order.

PR #2604, opened 2020-05-29
Deprecate the aws_instance_type

xiangce:


insights/parsers/aws_instance_type.py
```diff
@@ -1,58 +0,0 @@
...
-            if ' ' not in l.strip() and len(l.strip()) > 0:
-                self.raw = l.strip()
-                if '.' in self.raw:
-                    self.type = self.raw.split('.')[0].upper()
-
-        if not self.type:
-            raise ParseException('Unrecognized type: "{0}"'.format(content))
-
-    def __repr__(self):
-        return "<aws_type: {t}, raw: {r}>".format(t=self.type, r=self.raw)
```
bfahr, line 58: @xiangce instead of removing this it should be deprecated.  To allow it to continue to be functional you can change the parser to pull type from `aws_instance_id_doc`.

xiangce: @bfahr - I prefer to deprecate them. Since it's covered by the AwsInstanceIdDoc, it's not necessary to parser the aws_instance_id_doc to only get the `instanceType` one more time.

insights/specs/insights_archive.py
```diff
@@ -14,7 +14,6 @@ class InsightsArchiveSpecs(Specs):
     auditctl_status = simple_file("insights_commands/auditctl_-s")
     aws_instance_id_doc = simple_file("insights_commands/python_-m_insights.tools.cat_--no-header_aws_instance_id_doc")
     aws_instance_id_pkcs7 = simple_file("insights_commands/python_-m_insights.tools.cat_--no-header_aws_instance_id_pkcs7")
-    aws_instance_type = simple_file("insights_commands/python_-m_insights.tools.cat_--no-header_aws_instance_type")
```
bfahr, line 17: Hey @xiangce  this needs to stay since we have existing archives that include this spec.

xiangce: Done, @bfahr , Thanks.

PR #2611, opened 2020-06-03
Obfuscate hostname using sha1(hostname) rather than "host0"

gravitypriest:
For increased usability with Satellite, instead of identifying the host as "host0" as soscleaner does by default, let's use a hash function so the system can be identified even when obfuscated.

This obfuscated name should also be written as part of a JSON block to /etc/rhsm/facts so that Satellite can reference it.


gravitypriest:
test this please

gravitypriest:
test this please

gravitypriest:
test this please

PR #2619, opened 2020-06-10
Update ceph version mapping

xiangce:
- This a synchronization that should be done timely.


chenlizhong:
Looks good to me

PR #2623, opened 2020-06-11
ignore failure to write .registered file if non-root

gravitypriest:


insights/client/connection.py
```diff
@@ -880,7 +880,12 @@ def upload_archive(self, data_collected, content_type, duration):
             # 202 from platform, no json response
             logger.debug(upload.text)
             # upload = registration on platform
-            write_registered_file()
+            try:
+                write_registered_file()
+            except OSError as e:
+                # most likely permissions error from running as non-root
```
subpop, line 886: This makes me a bit nervous. While it's probably true that it "most likely" is a permissions error, it would be safer to check the `errno` value of the `OSError` so that other `OSError`s can be properly re-raised.

```python
if e.errno == e.EPERM:
```

Or whatever the right Python syntax is.

PR #2624, opened 2020-06-13
Add new attributes to dmsetup parser

bfahr:
* Added new attributes with more detail
* Updated tests


kergon:
More example output to add coverage for the new fields.

[dmsetup_examples.txt](https://github.com/RedHatInsights/insights-core/files/4780704/dmsetup_examples.txt)

insights/parsers/tests/test_dmsetup_info.py
```diff
@@ -16,6 +18,11 @@
 VG00-opt           253   7 L--w    1    4      0 LVM-gy9uAwD7LuTIApplr2sogbOx5iS0FTaxIiCYm5hcvgQdXynPGBfHQLrtE3sqUKT2
 """.strip()
 
+DMSETUP_INFO_2 = """
+Name               Maj Min Stat Open Targ Event  UUID
+VG00-tmp           253   8 xIsr    1    1      a LVM-gy9uAwD7LuTIApplr2sogbOx5iS0FTax6lLmBji2ueSbX49gxcV76M29cmukQiw4
```
kergon, line 23: It that output genuine?  How did you get 'a' to appear in the Event column?

kergon: and how do you get 'x' in the first Stat column?

Or is this a test to check how it handles corrupt input? (If so, maybe add a comment)

kergon:
Also add a test that checks the field limits are handled correctly i.e. maximum value in each numeric column and maximum length for each string.

kergon:
From the C definitions:
name is 127 characters + trailing nul (=128)
uuid is 128 characters + trailing nul (=129)
uint32_t target_count;  (segments - in practice it can never reach anywhere near the theoretical limit)
int32_t open_count;  (can never get anywhere near the limit and should be impossible for this to be negative unless there is a kernel bug)
uint32_t event_nr;  (in theory this could get big though unlikely)
major - 12 bits (4095)
minor - 20 bits (1048575)

kergon:
Genuine output showing some of the limits.  (Other limits will need to be faked.)
[dmsetup_examples2.txt](https://github.com/RedHatInsights/insights-core/files/4781054/dmsetup_examples2.txt)

insights/parsers/tests/test_dmsetup_info.py
bfahr: Its just fabricated data for testing @kergon

PR #2631, opened 2020-06-17
catch parse_content errors

dpensi:
the errors thrown by the implementations of `parse_content` method are uncatched.

We should catch them to avoid crashes and output what file generated the error to ease bug replication

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -92,7 +92,10 @@ class Parser(object):
         self._handle_content(context)
 
     def _handle_content(self, context):
-        self.parse_content(context.content)
+        try:
+            self.parse_content(context.content)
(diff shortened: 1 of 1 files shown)


csams:
Hey @dpensi , Thanks for the PR!

Those exceptions are caught and logged in the [top level evaluation loop](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/dr.py#L972) and are necessary to make invalid parsers unavailable to things that depend on them. If you want to see tracebacks, pass `-t` to `insights run`. You can also run the parser inside of [insights shell](https://insights-core.readthedocs.io/en/latest/shell.html) to interact with it and use `models.show_exceptions()` to see problems.

PR #2634, opened 2020-06-17
Introduce egg router

SteveHNH:
Grab the proper url to get the insights client egg based on the account being designated as testing or release. This basically just grabs the proper url segment by reaching out to the module egg router service prior to updating the egg.


insights/client/__init__.py
```diff
@@ -95,6 +95,25 @@ def branch_info(self):
...
+    def get_egg_url(self):
+        """
+        Get the proper url based on the configured egg release branch
+        """
+        protocol = "https://"
+        if self.config.insecure_connection:
+            protocol = "http://"
+            self.cert_verify = False
+
+        url = protocol + constants.base_url + constants.module_router_path
```
gravitypriest, line 108: Also using constants.base_url here will force it to use cloud.redhat which will conflict with what's in the config/won't work in Sat/conflicts with the legacy cert

gravitypriest:
Also using constants.base_url here is bad news bears cause it'll goof up in Satellite, we gotta use whatever is set in the config

SteveHNH:
@gravitypriest thanks for the tips. I found a couple more wonky things as well. working on it

insights/client/__init__.py
```diff
@@ -95,6 +95,23 @@ def branch_info(self):
         """
         return client.get_branch_info(self.config, self.connection)
 
+    @_net
+    def get_egg_url(self):
+        """
+        Get the proper url based on the configured egg release branch
+        """
+        if self.config.legacy_upload:
+            url = self.connection.base_url + '/platform' + constants.module_router_path
```
gravitypriest, line 104: I think would probably be fine

PR #2638, opened 2020-06-23
client: return True if unregister with --force

subpop:
When the `--unregister` option is invoked with `--force`, we circumvent the remote URL checks. This resulted in a `False` being returned by the unregistration operation, but in the case of `--force`, we know the remote URL is unreachable; we want to force the unregistration operation anyway, so we explicitly return `True`.


patchkez:
test this please

PR #2640, opened 2020-06-30
Support cib.xml in sosreport

huali027:
* The localtion of cib.xml in sosreport is at "sos_commands/pacemaker/crm_report/<hostname>/cib.xml"
* since hostname is unknown, so use glob here and then get the first one

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -23,7 +23,7 @@ from insights.core.plugins import datasource
 from insights.core.spec_factory import CommandOutputProvider, ContentException, DatasourceProvider, RawFileProvider
 from insights.core.spec_factory import simple_file, simple_command, glob_file, command_with_args
 from insights.core.spec_factory import first_of, foreach_collect, foreach_execute
-from insights.core.spec_factory import first_file, listdir, head
+from insights.core.spec_factory import first_file, listdir
 from insights.parsers.mount import Mount, ProcMounts
 from insights.parsers.dnf_module import DnfModuleList
 from insights.combiners.sap import Sap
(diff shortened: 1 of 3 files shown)


insights/specs/default.py
```diff
@@ -185,7 +185,14 @@ def is_ceph_monitor(broker):
...
     chronyc_sources = simple_command("/usr/bin/chronyc sources")
-    cib_xml = simple_file("/var/lib/pacemaker/cib/cib.xml")
+    cib_xml = first_of(
+        [
+            simple_file("/var/lib/pacemaker/cib/cib.xml"),
+            head(
+                glob_file("sos_commands/pacemaker/crm_report/*/cib.xml")
+            )
+        ]
+    )
```
xiangce, line 195: @csams, @bfahr - This usage seems a good approach, would you please take a look?

And I believe it works well with the `foreach_*` items. By using the "head/first_of", it seems we can implement the "command_with_args" in the following way (another way to avoid wrapping the Parser result into a list when there is only one item in the `"pre_command"`):
```python
pre_sub_cmds = [a,b,....]

cmd_a = head(foreach_execute(pre_sub_cmds, "/bin/cmd %s"))
```

bfahr:
@huali027 is this change just for sosreport?

huali027:
@bfahr, yes, it is for compatible with sosreport. I added the detail in the commit-msg
* The localtion of cib.xml in sosreport is at "sos_commands/pacemaker/crm_report/<hostname>/cib.xml"
* since hostname is unknown, so use glob here and then get the first one
No, there won't be multiple cib.xml under sos_commands, there is only one cib.xml file under "sos_commands/pacemaker/crm_report/"
Sure, I'll move it to sos_archive.py

huali027:
@bfahr, could you please help to review it again when u have time? Thanks.

PR #2641, opened 2020-06-30
Fix %f regex for get_after

Rohan21Lobo:
strptime supports Microseconds with %f but not Milliseconds. When logs
have milliseconds, %f still works fine to convert to datetime object as zeros
are padded to right.

When LogFileOutput uses %f in time_format but there are 3 digit milliseconds,
get_after function does not work correctly as it looks for a 6 digit
number. This change adjusts the regex to {1,6} digits for get_after to
work.

Source : Comments for %f in following documentation :
https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior

"%f is an extension to the set of format characters in the C standard (but implemented separately in datetime objects, and therefore always available). When used with the strptime() method, the %f directive accepts from one to six digits and zero pads on the right."

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -1186,7 +1186,7 @@ class LogFileOutput(six.with_metaclass(ScanMeta, Parser)):
             'p': r'\w{2}',  # AM / PM
             'M': r'([012345]\d)',  # Minutes
             'S': r'([012345]\d|60)',  # Seconds, including leap second
-            'f': r'\d{6}',  # Microseconds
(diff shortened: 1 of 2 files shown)


Rohan21Lobo:
@csams @gravitypriest @xiangce please review

bfahr:
@Rohan21Lobo would you mind to add a test for this condition?

Rohan21Lobo:
> @Rohan21Lobo would you mind to add a test for this condition?

@bfahr Added a test case.

bfahr (approved):

PR #2647, opened 2020-07-02
load filters if available

gravitypriest:


gravitypriest:
Core Collection Part 1

gravitypriest:
A thought: should we panic if called from the client and filters are missing?

gravitypriest:
In addition, fall back to `hostname` if `hostname -f` fails

insights/collect.py
```diff
@@ -246,7 +246,21 @@ def collect(manifest=default_manifest, tmp_path=None, compress=False):
 
     to_persist = get_to_persist(client.get("persist", set()))
 
-    hostname = call("hostname -f", env=SAFE_ENV).strip()
+    try:
+        filters.load()
+    except IOError as e:
+        if e.errno == 0:
```
bfahr, line 252: I get errno = 2 when no file is available.

PR #2654, opened 2020-07-08
load client denylist in core collector

gravitypriest:
also disable parallel run strategy because RHEL 8 threading is broken

### DIFF
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -66,7 +66,7 @@ client:
           enabled: true
 
     run_strategy:
-        name: parallel
+        name: serial
         args:
             max_workers: null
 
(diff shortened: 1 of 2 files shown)


insights/collect.py
```diff
@@ -214,7 +214,7 @@ def get_pool(parallel, kwargs):
         yield None
 
 
-def collect(manifest=default_manifest, tmp_path=None, compress=False):
+def collect(manifest=default_manifest, tmp_path=None, compress=False, rm_conf={}):
```
bfahr, line 217: @gravitypriest I think it'd be better to avoid mutable default args like`rm_conf`, maybe this or something similar would be better:
```
def colllect (..., rm_conf=None):
    ...
   rm_conf = rm_conf or {}
```

PR #2648, opened 2020-07-02
map files and commands to core components

gravitypriest:
Signed-off-by: Jeremy Crafts <jcrafts@redhat.com>

PR #2656, opened 2020-07-09
Two collectors with mapping

gravitypriest:
This is a combination of the following PRs:

https://github.com/RedHatInsights/insights-core/pull/2635
https://github.com/RedHatInsights/insights-core/pull/2648

With the necessary gaps bridged:
- only map components when `core_collect=True`
- pass the denylist into `collect.collect()`


gravitypriest:
Created for QE testing only. These features have been merged from their respective branches.

PR #2661, opened 2020-07-13
Add spec for sosreport

huali027:


insights/specs/sos_archive.py
```diff
@@ -33,6 +33,7 @@ class SosSpecs(Specs):
             )
         ]
     )
+    corosync_cmapctl = simple_file("sos_commands/corosync/corosync-cmapctl")
```
xiangce, line 36: it should be a `glob_file`.

PR #2663, opened 2020-07-14
Remove specs not used by any rules

bfahr:
* Prepare specs for core collection
* Remove all default specs that there are not being used to reduce archive size
  requirements
* This includes specs that are not currently being
  collected by the current client and specs that would not be evaluated
  by any rules if collected
* Remove assocated archive specs
* Add all file-like specs to sos_archive specs to allow collection


insights/specs/default.py
```diff
@@ -556,32 +587,32 @@ def semid(broker):
     ls_var_tmp = simple_command("/bin/ls -ln /var/tmp")
     ls_var_run = simple_command("/bin/ls -lnL /var/run")
     ls_var_www = simple_command("/bin/ls -la /dev/null /var/www")  # https://github.com/RedHatInsights/insights-core/issues/827
-    ls_tmp = simple_command("/bin/ls -la /tmp")
+    # ls_tmp = simple_command("/bin/ls -la /tmp")
```
xiangce, line 590: @bfahr - This one will be used in a new rule.

insights/specs/default.py
```diff
@@ -544,8 +575,8 @@ def semid(broker):
     ls_run_systemd_generator = simple_command("/bin/ls -lan /run/systemd/generator")
     ls_R_var_lib_nova_instances = simple_command("/bin/ls -laR /var/lib/nova/instances")
     ls_sys_firmware = simple_command("/bin/ls -lanR /sys/firmware")
-    ls_usr_lib64 = simple_command("/bin/ls -lan /usr/lib64")
-    ls_usr_sbin = simple_command("/bin/ls -ln /usr/sbin")
+    # ls_usr_lib64 = simple_command("/bin/ls -lan /usr/lib64")
```
xiangce, line 578: @bfahr - Would you please hold off removing this spec, it might be used in a new rule.

PR #2675, opened 2020-07-23
Client invocation

SteveHNH:
Get the name of the parent pid of insights-client

Will result in a string added to the User-Agent with the name of that process

We can detect a few different scenarios:
* shells (user execution)
* systemd
* crond
* python (which is about as close as I can get to ansible using the method here)


SteveHNH:
After digging further, it looks like the parent process of the ansible executed insights-client is python, and the parent of that is just 'sh'. So there's no indicator that it's ansible. I'll have to see if I can get closer to it

subpop:
Maybe we still keep this generic approach of collecting the PPID and then can start to determine what other characteristics are necessary to identify the "invocation" source.

SteveHNH:
I think we want to go ahead and go this direction and swing back around to the ansible stuff later. I'm working on the tests now

insights/client/connection.py
```diff
@@ -216,9 +220,10 @@ def user_agent(self):
                 logger.warning("Failed to detect OS version: %s", e)
         kernel_version = "%s %s" % (platform.system(), platform.release())
 
-        ua = "{client_version} ({core_version}; {requests_version}) {os_family} {os_release} ({python_version}; {kernel_version})".format(
+        ua = "{client_version} ({core_version}; {requests_version}) {os_family} {os_release} ({python_version}; {kernel_version}); {invocation}".format(
```
subpop, line 223: Can we name this thing more technically accurate and leave the assumption about "invocation source" to the interpretation? Something like "parent_process"? Since we aren't claiming that the parent process definitively specifies where the invocation source is coming from.

SteveHNH: this is a good idea. I'll update it

insights/tests/client/init/test_write_pidfile.py
```diff
@@ -5,14 +5,17 @@
...
     '''
     Test writing of the pidfile when InsightsClient
     is called initially (when setup_logging=False)
     '''
     InsightsClient(setup_logging=False)
     getpid.assert_called_once()
-    write_to_disk.assert_called_with(InsightsConstants.pidfile, content=str(getpid.return_value))
+    calls = [write_to_disk(InsightsConstants.pidfile, content=str(getpid.return_value)),
+             write_to_disk(InsightsConstants.ppidfile, content=get_parent_process.return_value)]
+    write_to_disk.has_calls(calls)
```
Glutexo, line 18: _unittest.mock_ doesn’t work like this. `write_to_disk.has_calls` does not assert anything. Moreover calls need to be results of _unittest.mock.call_, not actual mock object calls. Correct usage is:

```suggestion
    from mock.mock import call
    calls = [call(InsightsConstants.pidfile, content=str(getpid.return_value)),
             call(InsightsConstants.ppidfile, content=get_parent_process.return_value)]
    write_to_disk.assert_has_calls(calls)
```

insights/tests/client/init/test_write_pidfile.py
```diff
@@ -5,14 +5,17 @@
 
 @patch("insights.client.write_to_disk")
 @patch("insights.client.os.getpid")
-def test_write_pidfile(getpid, write_to_disk):
+@patch("insights.client.utilities.get_parent_process")
```
Glutexo, line 8: This patch is not correct. You need the path in the namespace from which _get_parent_process_ is used, not where it is defined.

```suggestion
@patch("insights.client.get_parent_process")
```

Glutexo (commented):
This pull request broke the tests. They now don’t test anything and produce false positivies.  Going to fix in #2795.

PR #2595, opened 2020-05-14
consolidate network calls and logging

gravitypriest:
create functions for http reqs
move retry logic to connection.upload_archive
create special exceptions for connection errors
move some code to __init__.py

Signed-off-by: Jeremy Crafts <jcrafts@redhat.com>

PR #2678, opened 2020-07-24
consolidate network calls and logging, simplify functions

gravitypriest:
redo of #2595 

## Design philosophy
Basically the connection lib has had bolt-on after bolt-on of additional functions, all reusing a lot of the same logging, request, and failure handling logic. I think it can all be consolidated into a simple set of self-contained HTTP functions that handle 99% of the logging and failure cases. 

With this change set:
- No additional logging is required from methods that perform HTTP calls, and logging is consistent for all requests
- Error handling can be reduced to a simple true/false check on the response, and because of how python-requests works, even "false" responses can be introspected upon (e.g. additional logging for failure codes in some cases)
- Connection errors and timeouts are the only "hard" stop as they raise an exception that can bubble up to the top level insights-client execution (phase, etc), or if need be, the outer method making the request can handle these if necessary

### Main HTTP request function - `_http_request`
This function does the following:
- Logs the URL and HTTP method to the NETWORK logger
- Performs the request using the requests Session object
- Logs HTTP response and status code
- Calls `raise_for_status` in order to raise an exception on HTTP failure codes
- Returns the response object
- Catches timeouts and connection errors, and re-raises exceptions to be handled by the appropriate logic
- Catches the exception raised by `raise_for_status` and calls our `handle_fail_rcs` function to do the appropriate logging, and re-raise an exception if necessary (i.e. a 413, upload size exceeded)

### Top level HTTP functions - `get`, `post`, `put`, `patch`, `delete`
These are wrappers to `_http_request` exposing the appropriate HTTP methods for ease of use.

### Custom exceptions
- `UnregisteredException(Exception)` - raised on HTTP 412, i.e. when system is unregistered. This is for CLASSIC API only.
- `PayloadTooLargeException(Exception)` - raised on HTTP 413, i.e. when the upload size is too large.
- `InvalidContentTypeException(Exception)` - raised on HTTP 415, i.e. when the content-type is not recognized by ingress.
- `TimeoutException(RuntimeError)` - raised by `_http_request` after catching and logging the `Timeout` exception raised by python-requests on a timeout. This is inherited from `RuntimeError` for ease of exception handling in the top-level execution logic.
- `HostNotFoundException(Exception)` - raised when the system cannot be found in the inventory when performing an ultralight check-in or running `--check-results`. This may be extraneous due to its narrow use case and is subject to removal.

### Changes to existing functions

- All functions within `InsightsConnection` that do HTTP work have had network logging calls removed, since it's all been displaced to `_http_request` function
- Error handling of exceptions is also modified, and response checking is done (in most cases) by a check on the falsiness of the response. Some functions need more granluar error checking, but this is also possible with the Response object returned by `_http_request`
- Raised exceptions will, in most cases, be handled by the top-level code of the phase in `insights.client.phase.v1`. One notable exception is that the upload function will handle `Timeout` exceptions and attempt to retry
- Calls to `InsightsConnection().session.<request>` in `insights.client.__init__` changed to `InsightsConnection().<request>` in order to use the new instance-level functions 

## Future improvements
- Full separation of legacy functions from current ones
- HTTP logic in a fully separate module
- Removal of the `insights.client.client` module and all surplus logic moved to `insights.client.__init__`


gravitypriest:
test this please

insights/client/client.py
```diff
@@ -352,6 +332,28 @@ def _legacy_upload(config, pconn, tar_file, content_type, collection_duration=No
                 logger.error("All attempts to upload have failed!")
                 logger.error("Please see %s for additional information", config.logging_file)
                 raise RuntimeError('Upload failed.')
+
+    api_response = json.loads(upload.text)
```
diegorafrh, line 336: Assuming `upload` is a `requests.Response` object, you can simply call `upload.json()` instead.

insights/client/__init__.py
```diff
@@ -495,8 +492,7 @@ def get_registration_status(self):
             returns (json):
                 {'messages': [dotfile message, api message],
                  'status': (bool) registered = true; unregistered = false
-                 'unreg_date': Date the machine was unregistered | None,
-                 'unreachable': API could not be reached}
+                 'unreg_date': Date the machine was unregistered | None}
```
Glutexo, line 495: This too is true only if _legacy_upload_ is True. [_insights.client.client.get_registration_status_](https://github.com/RedHatInsights/insights-core/blob/a1ffac37ba3e721dadaab6bb891c9800a8ef5a83/insights/client/client.py#L192) returns the result of [_insights.client.support.registration_check_](https://github.com/RedHatInsights/insights-core/blob/a1ffac37ba3e721dadaab6bb891c9800a8ef5a83/insights/client/support.py#L54), which returns
- either a result of [_insights.client.support.\_legacy\_registration\_check_](https://github.com/RedHatInsights/insights-core/blob/a1ffac37ba3e721dadaab6bb891c9800a8ef5a83/insights/client/support.py#L24), which is this dictionary (not JSON), if _legacy_upload_ is true,
- or result of [_insights.client.connection.InsightsConnection.api_registration_check_](https://github.com/RedHatInsights/insights-core/blob/a1ffac37ba3e721dadaab6bb891c9800a8ef5a83/insights/client/connection.py#L729) that is a boolean.

insights/client/connection.py
```diff
@@ -616,92 +661,87 @@ def do_group(self):
...
+        Reach out to the inventory API to check
+        whether a machine exists.
 
-            Returns
-                True    system exists in inventory
-                False   system does not exist in inventory
-                None    error connection or parsing response
+        Returns
+            True    system exists in inventory
+            False   system does not exist in inventory
```
Glutexo, line 736: This can still return _None_ in case of legacy upload. [__legacy_api_registration_check_](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/connection.py#L661) [does so](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/connection.py#L681) on error 404.

insights/client/connection.py
```diff
@@ -616,92 +661,87 @@ def do_group(self):
     def _legacy_api_registration_check(self):
         '''
         Check registration status through API
+
+        Returns:
+            unreg_date if unregistered
```
Glutexo, line 666: Where did this _unreg_date_ come from? In the response, it’s called _unregistered_at_, in the code, it’s _unreg_status_.

insights/client/connection.py
```diff
@@ -616,92 +661,87 @@ def do_group(self):
...
+            True if registered
+            None is unregistered
```
Glutexo, line 668: If I understand it correctly, _None_ means that the host [cannot be found](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/connection.py#L681) (404). A timestamp is returned if the host [has been found](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/connection.py#L697), but unregistered.

```suggestion
            unreg_date if unregistered
            True if registered
            None if not registered
```

insights/client/connection.py
```diff
@@ -616,92 +661,87 @@ def do_group(self):
...
+
+            TODO: FIX THIS
```
Glutexo, line 670:
```suggestion
```

This does not even say how.

insights/client/support.py
```diff
@@ -37,12 +36,8 @@ def _legacy_registration_check(pconn):
...
-        if api_reg_status:
-            api_record = 'Insights API confirms registration.'
-        else:
-            api_record = 'Insights API could not be reached to confirm registration status.'
-            unreachable = True
+    if api_reg_status is True:
+        api_record = 'Insights API confirms registration.'
     elif api_reg_status is None:
         api_record = 'Insights API says this machine is NOT registered.'
         api_reg_status = False
```
Glutexo, line 43: This overwrite of the value retrieved from _api_registration_check_ is pretty ugly, or at least non-semantic.

insights/client/support.py
```diff
@@ -53,8 +48,7 @@ def _legacy_registration_check(pconn):
 
     return {'messages': [local_record, api_record],
             'status': api_reg_status,
-            'unreg_date': unreg_date,
-            'unreachable': unreachable}
+            'unreg_date': unreg_date}
 
 
 def registration_check(pconn):
```
Glutexo, line 54: Do I understand it correctly that the modern API does not track unregistered systems? It looks like [_insights.client.connection.InsightsConnection.api_registration_check_](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/connection.py#L729) [does not distinguish](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/connection.py#L743) between not found and unregistered if _legacy_upload_ is _False_.

Glutexo (commented):
The return value documentation string for [_insights.client.client.get_registration_status_](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/client.py#L192) says it can return _None_, which is not true. If _legacy_upload_ is _False_, it eventually goes down to [_insights.client.connection.InsightsConnection.api_registration_check_](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/connection.py#L729), which only returns _True_ or _False_. Also there is not mentioned that the function can return a dictionary if _legacy_upload_ is _True_. In that case, the dictionary comes from [_insights.client.support.\_legacy\_registration\_check_](https://github.com/RedHatInsights/insights-core/blob/bb04fb85035fbb0eb8bab72486328bb2ad8cc9d1/insights/client/support.py#L24).

insights/client/client.py
```diff
@@ -311,36 +310,17 @@ def get_connection(config):
...
+        try:
+            upload = pconn.upload_archive(tar_file, '', collection_duration)
+        except TimeoutException:
+            # allow timeouts here because we want to retry the upload
+            upload = None
+        except (PayloadTooLargeException, UnregisteredException):
             raise RuntimeError('Upload failed.')
+
+        if upload:
+            break
```
Glutexo, line 323: Just thinking whether to keep the _else_ branch here. The _break_ interrupts the cycle. But the else on the other hand provides a nice symmetry.

insights/client/connection.py
```diff
@@ -299,167 +376,152 @@ def get_proxies(self):
...
+        Parameters
+            url     - URL to send a test request to
+            method  - HTTP method to use
+
+        Returns
+            True on success
+            False on failure
         """
         # tell the api we're just testing the URL
         test_flag = {'test': 'test'}
```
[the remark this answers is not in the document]
Glutexo: Why “flag”?

insights/client/connection.py
```diff
@@ -299,167 +376,152 @@ def get_proxies(self):
...
-            logger.error(
-                "Could not successfully connect to: %s", url)
-            print(exc)
-        if last_ex:
-            raise last_ex
-
-    def test_connection(self, rc=0):
+        logger.log(NETWORK, 'Testing %s', url)
+        if method is 'POST':
+            test_tar = TemporaryFile(mode='rb', suffix='.tar.gz')
```
Glutexo, line 423: Requests [support](https://2.python-requests.org/en/master/user/quickstart/#post-a-multipart-encoded-file) sending files from strings, that should remove the need for a temporary file. But again, this was not introduced by this pull request.

insights/client/connection.py
```diff
@@ -299,167 +376,152 @@ def get_proxies(self):
...
+            test_files = {
+                'file': ('test.tar.gz', test_tar, 'application/vnd.redhat.advisor.collection+tgz'),
+                'metadata': '{\"test\": \"test\"}'
+            }
+            test_req = self.post(url, files=test_files)
+        elif method is "GET":
+            test_req = self.get(url)
+        if not test_req:
+            logger.error("Could not successfully connect to: %s", url)
```
Glutexo, line 432: Just to be clear. This exact logging call used to be used on a connection error, which is not a case of a non-success status code. Now, this is logged instead of `logger.info("Connection failed")`. It looks like a good change, just saying.

insights/client/connection.py
```diff
@@ -299,167 +376,152 @@ def get_proxies(self):
...
+            return False
+        logger.info("Successfully connected to: %s", url)
+        return True
+
+    def test_connection(self):
```
Glutexo, line 437: 👏🏻 for removing the argument as it’s apparently not being used anywhere.

But! Will this now actually work? I mean, it’s called from [_insights.client.InsightsClient.test_connection_](https://github.com/RedHatInsights/insights-core/blob/b1914255dc91047764f8f10d4f30b2e9efeb9173/insights/client/__init__.py#L86). That function directly passes through the value returned from here, thus now a boolean. But its docstring still claims “returns (int): 0 if success 1 if failure”. _InsightsClient.test_connection_ is called from [_insights.client.phase.v1.pre_update_](https://github.com/RedHatInsights/insights-core/blob/b1914255dc91047764f8f10d4f30b2e9efeb9173/insights/client/phase/v1.py#L93), where it still works with the return value as if it was an integer return code `if rc == 0:`. Because `False == 0`, the condition produces an exactly opposite result than it used to.

Glutexo (changes requested):

insights/client/connection.py
```diff
@@ -175,6 +201,57 @@ def _init_session(self):
...
+            Calls handle_fail_rcs to handle messaging and exceptions for failure codes
+            Raises RuntimeError on timeout or failure to connect
+        '''
+        try:
+            logger.log(NETWORK, "%s %s", method, url)
+            res = self.session.request(url=url, method=method, timeout=self.config.http_timeout, **kwargs)
+            logger.log(NETWORK, "HTTP Status Code: %d", res.status_code)
+            logger.log(NETWORK, "HTTP Status Text: %s", res.reason)
+            logger.log(NETWORK, "HTTP Response Text: %s", res.text)
+            res.raise_for_status()
```
Glutexo, line 224: Do we even need this? Like, wouldn’t _handle_fail_rcs_ handle this already?

insights/client/connection.py
```diff
@@ -175,6 +201,57 @@ def _init_session(self):
...
+            return res
+        except requests.Timeout as e:
+            logger.error(e)
+            raise TimeoutException("Connection timed out.")
+        except requests.ConnectionError as e:
+            logger.error(e)
+            raise RuntimeError("The Insights API could not be reached.")
+        except requests.exceptions.HTTPError as e:
+            logger.error(e)
+            self.handle_fail_rcs(res)
```
Glutexo, line 234: This could be just as well outside the _except_ block along with the _return_. Despite its name, the function more or less handles the response by its status code, which does not necessarily be a failure one. Well, it’s ok like this too.

insights/client/connection.py
```diff
@@ -175,6 +201,57 @@ def _init_session(self):
...
+            # python requests are falsy for failure codes, so this will work
+            #   for both true/false checks and cases when attributes need
+            #   to be accessed
+            return res
```
Glutexo, line 238: Wouldn’t _handle_fail_rcs_ raise an error itself? I mean, it essentially does similar work like _raise_for_status_. So we may omit the raise-except here and just use our custom function. 🤷🏻‍♂️

patchkez:
test this please

patchkez:
test this please

insights/client/connection.py
```diff
@@ -299,167 +376,152 @@ def get_proxies(self):
...
-                last_ex = exc
-                logger.error(
-                    "Could not successfully connect to: %s", test_url + ext)
-                print(exc)
-        if last_ex:
-            raise last_ex
+        test_url = url.scheme + "://" + url.netloc + url.path
+        logger.log(NETWORK, "Testing: %s", test_url)
+        if method is "POST":
+            test_req = self.post(test_url, data=test_flag)
```
Glutexo, line 398: No Content-Type header? Or maybe rather `json=test_flag`?

```suggestion
            test_req = self.post(test_url, json=test_flag)
```

insights/client/connection.py
```diff
@@ -1086,19 +1119,16 @@ def checkin(self):
         url = self.inventory_url + "/hosts/checkin"
         logger.debug("Sending check-in request to %s with %s" % (url, canonical_facts))
         try:
-            response = self.session.post(url, headers={"Content-Type": "application/json"}, data=json.dumps(canonical_facts))
-            # Change to POST when the API is fixed.
-        except REQUEST_FAILED_EXCEPTIONS as exception:
-            _api_request_failed(exception)
+            response = self.post(url, headers={"Content-Type": "application/json"}, data=json.dumps(canonical_facts))
```
Glutexo, line 1122:
```suggestion
            response = self.post(url, json=canonical_facts)
```

insights/client/connection.py
```diff
@@ -569,41 +625,30 @@ def group_systems(self, group_name, systems):
         """
         api_group_id = None
         headers = {'Content-Type': 'application/json'}
```
Glutexo, line 627: Requests support a _json_ keyword argument that does this automatically.

```suggestion
```

insights/client/connection.py
```diff
@@ -569,41 +625,30 @@ def group_systems(self, group_name, systems):
...
-            logger.log(NETWORK, "POST", group_path)
-            post_group = self.session.post(group_path,
-                                           headers=headers,
-                                           data=data)
-            logger.debug("POST group status: %s", post_group.status_code)
-            logger.debug("POST Group: %s", post_group.json())
-            self.handle_fail_rcs(post_group)
+            post_group = self.post(group_path,
+                                   headers=headers,
+                                   data=data)
```
Glutexo, line 640:
```suggestion
            data = {'display_name': group_name}
            post_group = self.post(group_path, json=data)
```

insights/client/connection.py
```diff
@@ -569,41 +625,30 @@ def group_systems(self, group_name, systems):
...
-        put_group = self.session.put(group_path +
-                                     ('/%s/systems' % api_group_id),
-                                     headers=headers,
-                                     data=data)
-        logger.debug("PUT group status: %d", put_group.status_code)
-        logger.debug("PUT Group: %s", put_group.json())
+        self.put(group_path +
+                 ('/%s/systems' % api_group_id),
+                 headers=headers,
+                 data=data)
```
Glutexo, line 649:
```suggestion
        path = group_path + ('/%s/systems' % api_group_id)
        self.put(path, json=data)
```

insights/client/connection.py
```diff
@@ -871,89 +899,94 @@ def upload_archive(self, data_collected, content_type, duration=None):
         }
         logger.debug("Uploading %s to %s", data_collected, upload_url)
 
-        logger.log(NETWORK, "POST %s", upload_url)
-        upload = self.session.post(upload_url, files=files, headers={})
-
-        logger.log(NETWORK, "Upload status: %s %s %s",
-                     upload.status_code, upload.reason, upload.text)
+        upload = self.post(upload_url, files=files, headers={})
```
Glutexo, line 902: What’s the `headers={}` there used for? I see it has not been added in this pull request, but still.

```suggestion
        upload = self.post(upload_url, files=files)
```

insights/client/connection.py
```diff
@@ -550,13 +609,10 @@ def create_system(self, new_machine_id=False):
...
-        logger.debug("POST System: %s", post_system_url)
-        logger.debug(data)
-        logger.log(NETWORK, "POST %s", post_system_url)
-        return self.session.post(post_system_url,
-                                 headers={'Content-Type': 'application/json'},
-                                 data=data)
+        post_system_url = self.base_url + '/v1/systems'
+        return self.post(post_system_url,
+                         headers={'Content-Type': 'application/json'},
+                         data=data)
```
Glutexo, line 615:
```suggestion
        post_system_url = self.base_url + '/v1/systems'
        return self.post(post_system_url, json=data)
```

Glutexo (commented):
Found some more things that would probably not work, although not necessarily introduced by this pull request. Marked those with ⚠️ .

insights/client/connection.py
```diff
@@ -1086,19 +1119,16 @@ def checkin(self):
         url = self.inventory_url + "/hosts/checkin"
         logger.debug("Sending check-in request to %s with %s" % (url, canonical_facts))
         try:
-            response = self.session.post(url, headers={"Content-Type": "application/json"}, data=json.dumps(canonical_facts))
-            # Change to POST when the API is fixed.
-        except REQUEST_FAILED_EXCEPTIONS as exception:
-            _api_request_failed(exception)
+            response = self.post(url, headers={"Content-Type": "application/json"}, data=json.dumps(canonical_facts))
+        except RuntimeError as e:
+            logger.error(e)
```
Glutexo, line 1124: This is unnecessary. The call is [wrapped](https://github.com/RedHatInsights/insights-core/blob/9651f0c7e12e5e88ad29d3db1e74f119ce7fc9ad/insights/client/phase/v1.py#L115) by _except Exception_ that does the same logging.

insights/client/connection.py
```diff
@@ -1086,19 +1119,16 @@ def checkin(self):
...
             return None
```
Glutexo, line 1125: We may consider raising an exception in case of failure instead of returning success. That would save another conditional and still correctly result in setting process status code to _constants.sig_kill_bad_.

Glutexo (changes requested):
My last bulk of comments. Most of them are inconsistencies and suggestions, but there are some warnings about stuff that actually gets broken.

Glutexo:
So long and thanks for the fish. 🐟  Closing this as @gravitypriest is not going to merge this specific pull request, but create new, smaller ones instead. Hope that my reviews will help.

PR #2679, opened 2020-07-26
Add parser to parse output of 'lsvmbus -vv'

psachin:

### DIFF
--- a/insights/parsers/lsvmbus.py
+++ b/insights/parsers/lsvmbus.py
@@ -1,0 +1,86 @@
+"""
+LsvmBus - Command ``lsvmbus -vv``
+=================================
+
+This module parses the output of the command ``lsvmbus -vv``.
+"""
+import re
+
+from insights import parser, CommandParser
+from insights.parsers import SkipException
+from insights.specs import Specs
+
+
+@parser(Specs.lsvmbus)
+class LsvmBus(CommandParser):
+    """Parse the output of ``lsvmbus -vv`` as list.
+
+    Typical output::
+
+        VMBUS ID 18: Class_ID = {44c4f61d-4444-4400-9d52-802e27ede19f} - PCI Express pass-through
+                Device_ID = {47505500-0001-0000-3130-444531303244}
+                Sysfs path: /sys/bus/vmbus/devices/47505500-0001-0000-3130-444531303244
+                Rel_ID=18, target_cpu=0
+        VMBUS ID 26: Class_ID = {44c4f61d-4444-4400-9d52-802e27ede19f} - PCI Express pass-through
+                Device_ID = {47505500-0002-0000-3130-444531303244}
+                Sysfs path: /sys/bus/vmbus/devices/47505500-0002-0000-3130-444531303244
+                Rel_ID=26, target_cpu=0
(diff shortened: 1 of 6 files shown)


insights/parsers/lsvmbus.py
```diff
@@ -0,0 +1,70 @@
...
+"""
+import re
+
+from insights import LogFileOutput, parser, CommandParser
+from insights.parsers import SkipException
+from insights.specs import Specs
+
+
+@parser(Specs.lsvmbus)
+class LsvmBus(CommandParser, LogFileOutput):
```
xiangce, line 15: I'm OK with the logic, but just curious why adding the `LogFileOutput` as a base?

psachin: Hi @xiangce Actually `LogFileOutput` as particular significance here. It was left out from my side unintentionally and I forgot to mark this PR as **WIP** which caused confusion.

Now it is ready for review. Thanks.

insights/parsers/lsvmbus.py
```diff
@@ -0,0 +1,72 @@
...
+
+        >>> assert len(lsvmbus.data) == 4
+        >>> assert lsvmbus.data[0].get('vmbus_id', None) == '18'
+        >>> assert lsvmbus.data[0].get('device_id', None) == '47505500-0001-0000-3130-444531303244'
+        >>> assert lsvmbus.data[0].get('rel_id', None) == '18'
+        >>> assert lsvmbus.data[0].get('sysfs_path', None) == '/sys/bus/vmbus/devices/47505500-0001-0000-3130-444531303244'
+        >>> assert lsvmbus.data[0].get('target_cpu', None) == '0'
+
+    Returns:
+        (list): List of dict for each VMBUS ID.
```
bfahr, line 47: @psachin since the documentation built from the docstring won't have the dict keys, would you please add that information here so that a developer will understand what is included in each list element?  Also this is not a `return` element, it is an Attribute of the class so please change the documentation to reflect that.  You might also consider changing the `data` attribute name to `devices` or something that is more descriptive of what data is stored.  `data` is very generic.

psachin: @bfahr Please check if the latest changes looks good to you.

insights/parsers/lsvmbus.py
```diff
@@ -0,0 +1,86 @@
...
+    Examples:
+
+        >>> assert len(lsvmbus.devices) == 4
+        >>> assert lsvmbus.devices[0].get('vmbus_id', None) == '18'
+        >>> assert lsvmbus.devices[0].get('device_id', None) == '47505500-0001-0000-3130-444531303244'
+        >>> assert lsvmbus.devices[0].get('rel_id', None) == '18'
+        >>> assert lsvmbus.devices[0].get('sysfs_path', None) == '/sys/bus/vmbus/devices/47505500-0001-0000-3130-444531303244'
+        >>> assert lsvmbus.devices[0].get('target_cpu', None) == '0'
+
+    Returns:
```
bfahr, line 46: @psachin `Returns` should be `Attributes`.  Otherwise it looks good.

psachin: @bfahr Corrected. Sorry, I misinterpreted your initial comment.

PR #2681, opened 2020-07-27
Update the docstring of Lvs

shlao:
According to the spec: lvs_noheadings, The docstring for class Lvs needs to be updated

### DIFF
--- a/insights/parsers/lvm.py
+++ b/insights/parsers/lvm.py
@@ -19,8 +19,8 @@ Vgs - command ``/sbin/vgs --nameprefixes --noheadings --separator='|' -a -o vg_a
 VgsHeadings - command ``vgs -v -o +vg_mda_count,vg_mda_free,vg_mda_size,vg_mda_used_count,vg_tags --config="global{locking_type=0}"``
 -------------------------------------------------------------------------------------------------------------------------------------
 
-Lvs - command ``/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_all``
-------------------------------------------------------------------------------------
+Lvs - command ``/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_name,lv_size,lv_attr,mirror_log,vg_name,devices,region_size,data_percent,metadata_percent,segtype,seg_monitor --config="global{locking_type=0}"``
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 
 LvsHeadings - command ``/sbin/lvs -a -o +lv_tags,devices --config="global{locking_type=0}"``
 --------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


insights/parsers/lvm.py
```diff
@@ -19,8 +19,8 @@
 VgsHeadings - command ``vgs -v -o +vg_mda_count,vg_mda_free,vg_mda_size,vg_mda_used_count,vg_tags --config="global{locking_type=0}"``
 -------------------------------------------------------------------------------------------------------------------------------------
 
-Lvs - command ``/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_all``
-------------------------------------------------------------------------------------
+Lvs - command ``/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_name,lv_size,lv_attr,mirror_log,vg_name,devices,region_size,data_percent,metadata_percent,segtype,seg_monitor --config="global{locking_type=0}"``
```
bfahr, line 22: @shlao this line is a subheading and it would be better to use an abbreviated form even thought it may be slightly incorrect to make the sidebar in readthedocs more readable.

insights/parsers/lvm.py
```diff
@@ -443,7 +443,7 @@ def parse_content(self, content):
 @parser(Specs.lvs_noheadings)
 class Lvs(Lvm):
     """
-    Parse the output of the `/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_all` command.
+    Parse the output of the `/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_name,lv_size,lv_attr,mirror_log,vg_name,devices,region_size,data_percent,metadata_percent,segtype,seg_monitor --config="global{locking_type=0}"` command.
```
bfahr, line 446: This is an appropriate place to include the entire command.

insights/parsers/lvm.py
shlao: Is it ok to write that:
`Lvs - command ``/sbin/lvs --nameprefixes --noheadings --separator='|' -a -o lv_name,lv_size ...```

So the user will check the definition of 'lvs_noheadings' to make sure that what they need is in the list of lvs fields.
The first time I used Lvs parser, I saw the option `-o lv_all` in the docstring and thought there is a filed `lv_dm_path`.
But in fact, there is no lv_dm_path, it takes me time to debug the problem for this incorrect docstring.

insights/parsers/lvm.py
shlao: sue, It is an appropriate place to add it

insights/parsers/lvm.py
bfahr: Yes, if it causes confusion then that is a good reason to change it.

PR #2682, opened 2020-07-27
Add EtcSshConfigD Parser

shlao:
Updated
```
insights/specs/__init__.py
insights/specs/default.py
insights/parsers/ssh_client_config.py
```

### DIFF
--- a/insights/parsers/ssh_client_config.py
+++ b/insights/parsers/ssh_client_config.py
@@ -9,6 +9,9 @@ Parsers provided by this module are:
 EtcSshConfig - file ``/etc/ssh/ssh_config``
 -------------------------------------------
 
+EtcSshConfigD - file ``/etc/ssh/ssh_config.d/*``
+------------------------------------------------
+
 ForemanSshConfig - file ``/usr/share/foreman/.ssh/ssh_config``
 --------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


insights/parsers/ssh_client_config.py
```diff
@@ -9,6 +9,9 @@
 EtcSshConfig - file ``/etc/ssh/ssh_config``
 -------------------------------------------
 
+EtcSshConfigD - file ``/etc/ssh/ssh_config.d/*``
```
bfahr, line 12: @shlao I believe `file` should be `files` since this is a glob spec.

insights/parsers/ssh_client_config.py
```diff
@@ -124,6 +127,14 @@ class EtcSshConfig(SshClientConfig):
     pass
 
 
+@parser(Specs.ssh_config_d)
+class EtcSshConfigD(SshClientConfig):
+    """
+    This Parser reads the files ``/etc/ssh/ssh_config.d/``
```
bfahr, line 133: Missing the trailing `*` on the glob.

bfahr: Please add the docstrings for this parser consistent with the [documentation guidelines](https://insights-core.readthedocs.io/en/latest/docs_guidelines.html).  I also noticed that the base class attributes are not documented so please either include those in the base class docstring or in the docstring for your class.  

Please also add testing for this parser.  The base class could change and break this parser.

bfahr:
@shlao this looks good once you fix the docstring error.

PR #2684, opened 2020-07-28
Add filters to neutron conf parser

bfahr:
* Add additional filters that may not be added by rules to capture some
  additional data


bregman-arie:
Looks good +1

PR #2686, opened 2020-07-29
Restore ethtool -c specs for core collection

bfahr:


rasrivas-redhat:
@bfahr - look goods to me.

PR #2687, opened 2020-07-29
Add designate parser

bregman-arie:
Including tests and and docs.

### DIFF
--- a/insights/parsers/designate_conf.py
+++ b/insights/parsers/designate_conf.py
@@ -1,0 +1,47 @@
+"""
+DesignateConf - file ``/etc/designate/designate.conf``
+======================================================
+
+This class provides parsing for the files:
+    ``/var/lib/config-data/puppet-generated/designate/etc/designate/designate.conf``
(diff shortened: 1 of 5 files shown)


insights/parsers/designate_conf.py
```diff
@@ -0,0 +1,49 @@
...
+
+See the ``IniConfigFile`` class for examples.
+"""
+from .. import IniConfigFile, parser, add_filter
+from insights.specs import Specs
+
+add_filter(Specs.designate_conf, ["["])
+
+ADDITIONAL_FILTERS = [
+    "[",
```
bfahr, line 33: No need to add this twice.

insights/parsers/designate_conf.py
```diff
@@ -0,0 +1,49 @@
+"""
+DesignateConf - file ``/etc/designate/designate.conf``
+====================================================
```
bfahr, line 3: Heading underline needs to be same length as the line above.

PR #2688, opened 2020-07-30
Enhance parser yum_log to handle erased package with version

wushiqinlou:
In the former yum_log version, there is no version in erased packages, like:

~~~
May 23 16:09:09 Erased: redhat-access-insights-batch
~~~ 

So the code handle this situation with following code:

~~~
 if state == self.ERASED:
    pkg = InstalledRpm({'name': pkg})
else:
    pkg = InstalledRpm.from_package(pkg)
~~~

However, in new yum.log version, there is version in erased packages, like:

~~~
Jul 28 23:24:36 Erased: systemd-219-67.el7.x86_64
~~~

So it is necessary to handle change like:

~~~
 if state == self.ERASED and "." not in pkg:
    pkg = InstalledRpm({'name': pkg})
else:
    pkg = InstalledRpm.from_package(pkg)
~~~

### DIFF
--- a/insights/parsers/yumlog.py
+++ b/insights/parsers/yumlog.py
@@ -117,7 +117,7 @@ class YumLog(Parser):
                 timestamp = ' '.join([month, day, time])
                 state = state.rstrip(':')
                 pkg = pkg.split(':')[-1].strip()
-                if state == self.ERASED:
+                if state == self.ERASED and len(pkg.split(".")) == 1:
                     pkg = InstalledRpm({'name': pkg})
                 else:
                     pkg = InstalledRpm.from_package(pkg)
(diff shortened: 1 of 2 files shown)


insights/parsers/yumlog.py
```diff
@@ -117,7 +117,7 @@ def parse_content(self, content):
                 timestamp = ' '.join([month, day, time])
                 state = state.rstrip(':')
                 pkg = pkg.split(':')[-1].strip()
-                if state == self.ERASED:
+                if state == self.ERASED and len(pkg.split(".")) == 1:
```
xiangce, line 120: How about
`if state == self.ERASED and "." in pkg`

wushiqinlou: it should be` if state == self.ERASED and "." not in pkg`, updated, please have a check.

PR #2702, opened 2020-08-06
Add parser and spec for /sys/module/{dm_mod,scsi_mod}/parameters/use_blk_mq

JoySnow:


insights/parsers/sys_module.py
```diff
@@ -0,0 +1,81 @@
...
+        if not content or len(content) != 1:
+            raise SkipException()
+        self.val = content[0].strip()
+
+    @property
+    def is_on(self):
+        """
+        Returns (bool or None): True for on, False for off, None for unknow case.
+        """
+        return True if self.val == 'Y' else (False if self.val == 'N' else None)
```
bfahr, line 43: @JoySnow if the file is empty or has more than one character you are already skipping, what does it mean if the value is set to a single character other than `Y` or `N`?  That seems like an error condition that should be raised in `parse_content` so that the value could only be a known value.  Then `is_on` could just return T/F.

insights/parsers/sys_module.py
```diff
@@ -0,0 +1,81 @@
...
+    Sample Content::
+
+        Y - for on
+        N - for off
+    """
+
+    def parse_content(self, content):
+        if not content or len(content) != 1:
+            raise SkipException()
+        self.val = content[0].strip()
```
bfahr, line 36: Should be a private attribute since you have the `is_on` property?

insights/parsers/sys_module.py
JoySnow: Hi @bfahr ,
I mean to let `self.val` to keep the raw data.
So here, `self.val` takes the **first line** of file `/proc/mounts` 's content if the file is not empty.
From the usual case, the `self.val` is likely to be `Y`/`N`.  Using `is_on` property to tell if `self.val` is `Y` or `N` (known cases).
And in case other value exists, parser user can get data from `self.val` directly, and do the check logic in rule.

insights/parsers/sys_module.py
JoySnow: Question is answered in my above reply.

insights/parsers/sys_module.py
JoySnow: Thanks @xiangce for the explanation. 
> And for any interfaces named with is_* should always return T/F.

If this is one of the implicit rules that insights-core are using, sure, I am okay to make this `is_on` follow it.

> the parse_content should throw a ParseExpection, but not keep its raw value in val. 

I don't get it why we can not keep raw value in `val`.
I see that is good approach. The idea is when later a parser user found out there are other content instead of `Y`/`N`, one can get the value from `self.val` and handle it. If using the `ParseExpection` approach,  'enhance this parser' will be the first blocker under such case.  
I am suggesting this "keep-raw-data when we can not predicate all possible values." as a handling way.  

> AFAIK, on archlinux, it could be 1 or 0.

Thanks. I will include `1`/`0`.

insights/parsers/sys_module.py
```diff
@@ -0,0 +1,81 @@
...
+
+class XModUseBlkMq(Parser):
+    """
+    Parse for file `/sys/module/{dm_mod,scsi_mod}/parameters/use_blk_mq`.
+    File content shows if use_blk_mq parameter is on.
+
+    Sample Content::
+
+        Y - for on
+        N - for off
```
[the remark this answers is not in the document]
JoySnow: Oh, ACK. Will change it.

insights/parsers/sys_module.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Raises:
+            ParseException: When tell is_on for unknown cases.
+        """
+        if self.val in ['Y', '1']:
+            return True
+        elif self.val in ['N', '0']:
+            return False
+        else:
+            raise ParseException("Unknown value: {0}".format(self.val))
```
bfahr, line 57: @JoySnow @xiangce my original comment was related to valid values and whether you want to be able to check for invalid values (not 'YN10') in a rule.  If that is important in a rule then this should probably be a `ValueError` or custom exception to indicate valid parsing with unexpected content.  An even better solution would be to use a negative or "[dirty](https://insights-core.readthedocs.io/en/latest/exception_model.html#case-3b-parsing-data-to-find-errors-dirty-parser)" parser which specifically looks for unexpected content and can be used by rules instead of exception handling.

If you are going to raise a `ParseException` then it should be raised in parse_content. See the [insights-core documentation section on exceptions](https://insights-core.readthedocs.io/en/latest/exception_model.html#case-3-unparsable-data-provides-useful-information) for more information.

JoySnow: Hi @bfahr , I had updated the code to use  `ValueError.

JoySnow:
@bfahr  Could you help review my update?

PR #2708, opened 2020-08-10
Add spec libssh_config

shlao:
Added
```
insights/parsers/libssh_config.py
insights/parsers/tests/test_libssh_config.py
docs/shared_parsers_catalog/libssh_config.rst
```

Updated
```
insights/specs/__init__.py
insights/specs/default.py
```

### DIFF
--- a/insights/parsers/libssh_config.py
+++ b/insights/parsers/libssh_config.py
@@ -1,0 +1,123 @@
+"""
+Libssh configuration files
+==========================
+
+This module provides the ssh options in libssh configuration file
+
+Parsers provided by this module are:
+
+LibsshClientConfig - file ``/etc/libssh/libssh_client.config``
+--------------------------------------------------------------
+
+LibsshServerConfig - file ``/etc/libssh/libssh_server.config``
+--------------------------------------------------------------
+"""
+
+from insights.specs import Specs
+from insights.core import CommandParser
+from .. import parser, get_active_lines
+from insights.parsers import SkipException
+
+
+class LibsshConfig(CommandParser, dict):
+    """
+    Base class for libssh configuration file.
+
+    There are many options in the libssh configuration file and
+    all of them are stored in one of the following formations:
+    a)   Key  [space]  Value
+    b)   Key  =  Value
+
+    There are the keywords used in the configuration file:
+        include
+        hostkey
+        listenaddress
+        port
+        loglevel
+        ciphers
+        macs
+        kexalgorithms
+        match
+        pubkeyacceptedkeytypes
+        hostkeyalgorithms
+        all
+        user
+        group
+        host
+        localaddress
+        localport
+        rdomain
+        address
+
+    Sample output::
+
+        # Parse system-wide crypto configuration file
+        Include /etc/crypto-policies/back-ends/libssh.config
+        # Parse OpenSSH configuration file for consistency
+        Include /etc/ssh/sshd_config
+
+    .. note::
+
+        If there are two or more lines that have the same key, then we
+        store the values in a list.
+
+    Examples:
+        >>> 'Include' in config
+        True
+        >>> config['Include']
+        ['/etc/crypto-policies/back-ends/libssh.config', '/etc/ssh/sshd_config']
+
+    Raises:
+        SkipException: When input content is empty or there is a syntax error.
+    """
+
+    def parse_content(self, content):
+        if not content:
+            raise SkipException('No content.')
+
+        for line in get_active_lines(content):
+            delimiter = None
+            if ' ' in line:
+                delimiter = ' '
+
+            if '=' in line:
+                delimiter = '='
+
+            try:
+                k, v = [i.strip() for i in line.split(delimiter)]
+                if k not in self:
+                    self[k] = v
+                else:
+                    _v = self[k]
+                    _v = [_v] if not isinstance(_v, list) else _v
+                    if v not in _v:
+                        _v.append(v)
+                        self[k] = _v
+            except ValueError:
+                raise SkipException('Syntax error')
+
+
+@parser(Specs.libssh_client_config)
+class LibsshClientConfig(LibsshConfig):
+    """
+    Parser for accessing the ``/etc/libssh/libssh_client.config`` file.
+
+    .. note::
+
+        Please refer to the super-class :py:class:`insights.parsers.libssh_config:LibsshConfig`
+        for more usage information.
+    """
+    pass
+
+
+@parser(Specs.libssh_server_config)
+class LibsshServerConfig(LibsshConfig):
+    """
+    Parser for accessing the ``/etc/libssh/libssh_server.config`` file
+
+    .. note::
+
+        Please refer to the super-class :py:class:`insights.parsers.libssh_config:LibsshConfig`
+        for more usage information.
+    """
+    pass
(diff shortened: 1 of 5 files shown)


insights/parsers/libssh_config.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Sample output::
+
+        # Parse system-wide crypto configuration file
+        Include /etc/crypto-policies/back-ends/libssh.config
+        # Parse OpenSSH configuration file for consistency
+        Include /etc/ssh/sshd_config
+
+    Attributes:
+       self (list): List of included files
```
xiangce, line 38: `self` is not the `attribute` of this class. 

If the `include` directives/lines will be accessed often, better add a `obj.includes` 

And, It's not so reasonable to keep only the `include` directive of the `config` and discard all the others. It should be a common parser, and it's necessary to make it widely useful as much as it can, but not only keep the lines required by a single rule. You could refer to the existing parsers against other `config/conf` files and re-implement it.

insights/parsers/libssh_config.py
```diff
@@ -0,0 +1,80 @@
...
+INCLUDE_RE = re.compile(r'^Include\s+\S+')
+
+
+class LibsshConfig(CommandParser, list):
+    """
+    Base class for libssh configuration file.
+
+    Sample output::
+
+        # Parse system-wide crypto configuration file
```
xiangce, line 32: I can't believe these files only contain the `include` lines?

insights/parsers/libssh_config.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+import re
+from insights.specs import Specs
+from insights.core import CommandParser
+from .. import parser, get_active_lines
+from insights.parsers import SkipException
+
+# regex for matching the 'include' directives
+INCLUDE_RE = re.compile(r'^Include\s+\S+')
```
xiangce, line 23: `regex` is not the recommended way to parse the simple content if there are other simple methods, like `split` .
Would you please re-implement this logic by using `startswith` or `in` and `split`?

insights/parsers/libssh_config.py
shlao: I searched doc about on google but found nothing and there is no man page for them.

xiangce: If you are not sure about the content of this file, I can't believe this spec will get approved. You need to get the spec approved at first.

shlao:
I can't find any doc about the config file of /etc/libssh/libssh_server.config or /etc/libssh/libssh_client.config
So, I copy the codes from libssh-0.9.4-2.el8.src.rpm showing as followings to see how to parser the config file in C language:

1. `obj/config.h:
#define GLOBAL_BIND_CONFIG "/etc/libssh/libssh_server.config"`

2. `src/options.c:        rc = ssh_bind_config_parse_file(sshbind, GLOBAL_BIND_CONFIG);`

3. `src/bind_config.c
 55 static struct ssh_bind_config_keyword_table_s
 56 ssh_bind_config_keyword_table[] = {
 57     {             
 58         .name   = "include",
 60     },            
 61     {             
 62         .name   = "hostkey",
 64     },     

166 static enum ssh_bind_config_opcode_e
167 ssh_bind_config_get_opcode(char *keyword, uint32_t *parser_flags)
168 {
171     for (i = 0; ssh_bind_config_keyword_table[i].name != NULL; i++) {
172         if (strcasecmp(keyword, ssh_bind_config_keyword_table[i].name) == 0) {

270 static int
271 ssh_bind_config_parse_line(ssh_bind bind,
272                                               const char *line,
294     x = s = strdup(line);
308     keyword = ssh_config_get_token(&s);
315     opcode = ssh_bind_config_get_opcode(keyword, parser_flags);

330     case BIND_CFG_INCLUDE:
331         p = ssh_config_get_str_tok(&s, NULL);

605 int ssh_bind_config_parse_file(ssh_bind bind, const char *filename)
627     while (fgets(line, sizeof(line), f)) {
628         count++;
629         rv = ssh_bind_config_parse_line(bind, line, count, &parser_flags, seen);
`

`src/config_parser.c
 68 char *ssh_config_get_token(char **str)
 73     c = ssh_config_get_cmd(str);
 74        
 75     for (r = c; *c; c++) {
 76         if (isblank(*c) || *c == '=') {
 77             *c = '\0';
 78             goto out;
 79         }
 80     }

105 const char *ssh_config_get_str_tok(char **str, const char *def)
106 {  
109     p = ssh_config_get_token(str);
`

shlao:
What we can learn from those codes are:
1. there are the keywords used in the libssh rpm:
```
include
hostkey
listenaddress
port
loglevel
ciphers
macs
kexalgorithms
match
pubkeyacceptedkeytypes
hostkeyalgorithms
all
user
group
host
localaddress
localport
rdomain
address
```

shlao:
2. All options are stored with the following formation:
```
 a)   Key  [space]  Value
 b)   Key  =  Value
```

src/config_parser.c
 68 char *ssh_config_get_token(char **str)
 73     c = ssh_config_get_cmd(str);
 74        
 75     for (r = c; *c; c++) {
 **_76         if (isblank(*c) || *c == '=') {_**
 77             *c = '\0';
 78             goto out;
 79         }
 80     }

xiangce:
> 1. All options are stored with the following formations:
> 
> ```
>  a)   Key  [space]  Value
>  b)   Key  =  Value
> ```
> 
> ```
> 
> src/config_parser.c
>  68 char *ssh_config_get_token(char **str)
>  73     c = ssh_config_get_cmd(str);
>  74        
>  75     for (r = c; *c; c++) {
>  76         if (isblank(*c) || *c == '=') {
>  77             *c = '\0';
>  78             goto out;
>  79         }
>  80     }
> ```

Can you update the logic as per what you investigated above?
- There are different supported `keywords`, but the current code only support `Include`.
- There is `=` or ` ` as the separator, but the current code only processes ` `.

insights/parsers/libssh_config.py
```diff
@@ -0,0 +1,128 @@
...
+class LibsshConfig(CommandParser):
+    """
+    Base class for libssh configuration file.
+
+    There are many options in the libssh configuration file and
+    all of them are stored in one of the following formations:
+        a)   Key  [space]  Value
+        b)   Key  =  Value
+
+    There are the keywords used in the configuration file:
```
bfahr, line 31: @shlao you have CI errors that I believe are caused by lines 27 and 31 missing `::` and a blank line after.

insights/parsers/libssh_config.py
```diff
@@ -0,0 +1,128 @@
...
+                else:
+                    _v = kv_pairs[k]
+                    _v = [_v] if not isinstance(_v, list) else _v
+                    if v not in _v:
+                        _v.append(v)
+                        kv_pairs[k] = _v
+            except ValueError:
+                raise SkipException('Syntax error')
+
+        self.data = kv_pairs
```
bfahr, line 104: Since the only attribute you are providing in this class is `data` which is a `dict`, it would be better to implement this class as a `dict`.  An example of how this works is provided in the [DnfModuleList](https://github.com/RedHatInsights/insights-core/blob/7809af44df433a2a70e42af2b2a724de072719cf/insights/parsers/dnf_module.py#L64) parser and several others.

bfahr (changes requested):

xiangce (approved):

xiangce:
@shlao - The last several updates look good. I'll go ahead and merge this PR.
But please do not use "--amend" and "--force" to push the new commits to the PR in the following, which will cover the former commits and make it difficult for reviewers to check the changes. Thanks.

PR #2716, opened 2020-08-17
Update the spec of DmsetupInfo

shlao:
There is a rule that needs the info from DmsetupInfo, so I create a new PR for this.

Updated
```
insights/specs/default.py
```

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -228,6 +228,7 @@ class DefaultSpecs(Specs):
     dmesg = simple_command("/bin/dmesg")
     dmesg_log = simple_file("/var/log/dmesg")
     dmidecode = simple_command("/usr/sbin/dmidecode")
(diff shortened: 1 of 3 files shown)


xiangce (approved):

xiangce:
@bfahr - This spec was removed by #2663, however, there is an Insights rule used it. Would you please double confirm this PR?

Ps, the spec is in the uploader.json. no need to raise MR to the insights-core-assets repo for this change.

bfahr:
@shlao you will also need to remove this spec from the test that checks for removed specs in `/insights/tests/client/collection_rules/test_map_components.py` [here](https://github.com/RedHatInsights/insights-core/blob/a44cfd63edba98d6ef9d3816631a0e42f40f455f/insights/tests/client/collection_rules/test_map_components.py#L55)

PR #2717, opened 2020-08-19
Enhance that there are two or more values of the same key in a dictio…

shlao:
Updated
```
insights/parsers/__init__.py
```

Here are the codes of parser SystemctlShowServiceAll:
```
 19 class SystemctlShowServiceAll(CommandParser, dict):
 79     def parse_content(self, content): 
 90         for s, e in idx_list:
 91             data = split_kv_pairs(content[s:e], use_partition=False)
 92             name = data.get('Names', data.get('Id'))
 93             if not name:
 94                 raise ParseException('"Names" or "Id" not found!')
 95             self[name] = dict((k, v) for k, v in data.items() if v)
```

Here are the testing data (on RHEL8):
```
# systemctl show sshd.service  > /tmp/sshd.txt
# cat /tmp/sshd.txt | grep Env
EnvironmentFiles=/etc/crypto-policies/back-ends/opensshserver.config (ignore_errors=yes)
EnvironmentFiles=/etc/sysconfig/sshd (ignore_errors=yes)
```
the function split_kv_pairs will overwrite the dictionary value of the same key **_EnvironmentFiles_** and only save the last value.

### DIFF
--- a/insights/parsers/__init__.py
+++ b/insights/parsers/__init__.py
@@ -110,6 +110,10 @@ def split_kv_pairs(lines, comment_char="#", filter_string=None, split_on="=", us
     will be ignored. ::func:`get_active_lines` is called to strip comments and
     blank lines from the data.
 
+    .. note::
+        If there are two or more lines that have the same key, then we store the
+        values in a list.
+
     Parameters:
         lines (list of str): List of the strings to be split.
         comment_char (str): Char that when present in the line indicates all
@@ -147,18 +151,19 @@ def split_kv_pairs(lines, comment_char="#", filter_string=None, split_on="=", us
         # Comment line
         # Blank lines will also be removed
         keyword1 = value1   # Inline comments
+        keyword1 = value2   # Inline comments
         keyword2 = value2a=True, value2b=100M
         keyword3     # Key with no separator
         >>> split_kv_pairs(lines)
-        {'keyword2': 'value2a=True, value2b=100M', 'keyword1': 'value1'}
+        {'keyword2': 'value2a=True, value2b=100M', 'keyword1': ['value1', 'value2']}
         >>> split_kv_pairs(lines, comment_char='#')
-        {'keyword2': 'value2a=True, value2b=100M', 'keyword1': 'value1'}
+        {'keyword2': 'value2a=True, value2b=100M', 'keyword1': ['value1', 'value2']}
         >>> split_kv_pairs(lines, filter_string='keyword2')
         {'keyword2': 'value2a=True, value2b=100M'}
         >>> split_kv_pairs(lines, use_partition=True)
-        {'keyword3': '', 'keyword2': 'value2a=True, value2b=100M', 'keyword1': 'value1'}
+        {'keyword3': '', 'keyword2': 'value2a=True, value2b=100M', 'keyword1': ['value1', 'value2']}
         >>> split_kv_pairs(lines, use_partition=True, ordered=True)
-        OrderedDict([('keyword1', 'value1'), ('keyword2', 'value2a=True, value2b=100M'), ('keyword3', '')])
+        OrderedDict([('keyword1', ['value1', 'value2']), ('keyword2', 'value2a=True, value2b=100M'), ('keyword3', '')])
 
     """
     _lines = lines if comment_char is None else get_active_lines(lines, comment_char=comment_char)
@@ -166,13 +171,23 @@ def split_kv_pairs(lines, comment_char="#", filter_string=None, split_on="=", us
     kv_pairs = OrderedDict() if ordered else {}
 
     for line in _lines:
+        k, v = [None, None]
         if not use_partition:
             if split_on in line:
-                k, v = line.split(split_on, 1)
-                kv_pairs[k.strip()] = v.strip()
+                k, v = [i.strip() for i in line.split(split_on, 1)]
         else:
-            k, _, v = line.partition(split_on)
-            kv_pairs[k.strip()] = v.strip()
+            k, _, v = [i.strip() for i in line.partition(split_on)]
+
+        if k is not None:
+            if k not in kv_pairs:
+                kv_pairs[k] = v
+            else:
+                _v = kv_pairs[k]
+                _v = [_v] if not isinstance(_v, list) else _v
+                if v not in _v:
+                    _v.append(v)
+                    kv_pairs[k] = _v
+
     return kv_pairs
 
 
--- a/insights/parsers/__init__.py
+++ b/insights/parsers/__init__.py
@@ -186,10 +186,9 @@ def split_kv_pairs(lines, comment_char="#", filter_string=None, split_on="=",
                 kv_pairs[k] = v
             else:
                 if k not in kv_pairs:
-                    kv_pairs[k] = v
+                    kv_pairs[k] = [v]
                 else:
                     _v = kv_pairs[k]
-                    _v = [_v] if not isinstance(_v, list) else _v
                     if v not in _v:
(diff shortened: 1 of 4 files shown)


shlao:
If the key has existed then it is well to combine two dictionary values.
for example, what we get from expression service_conf['sshd.service']['EnvironmentFiles'] is:
`['/etc/crypto-policies/back-ends/opensshserver.config', '/etc/sysconfig/sshd']`

shlao:
@xiangce @bfahr Will you help me to review this?

insights/parsers/__init__.py
```diff
@@ -99,6 +99,75 @@ def make_kv(opt):
     return dict(make_kv(opt) for opt in optlist.split(opt_sep))
 
 
+def set_dict_value(this_dict, key, value):
```
xiangce, line 102: I think this function makes things more complex.

insights/parsers/__init__.py
```diff
@@ -169,10 +238,10 @@ def split_kv_pairs(lines, comment_char="#", filter_string=None, split_on="=", us
         if not use_partition:
             if split_on in line:
                 k, v = line.split(split_on, 1)
-                kv_pairs[k.strip()] = v.strip()
+                set_dict_value(kv_pairs, k.strip(), v.strip())
         else:
             k, _, v = line.partition(split_on)
-            kv_pairs[k.strip()] = v.strip()
+            set_dict_value(kv_pairs, k.strip(), v.strip())
```
xiangce, line 244: How about changing it like:
```diff
+        k, v = [None, None]
         if not use_partition:
             if split_on in line:
-                k, v = line.split(split_on, 1)
-                set_dict_value(kv_pairs, k.strip(), v.strip())
+                k, v = [i.strip() for i in line.split(split_on, 1)]
         else:
-            k, _, v = line.partition(split_on)
-            set_dict_value(kv_pairs, k.strip(), v.strip())
+            k, _, v = [i.strip() for i in line.partition(split_on)]
+        if k is not None:
+            if k not in kv_pairs:
+                kv_pairs[k] = v
+            else:
+                _v = kv_pairs[k]
+                _v = [_v] if not isinstance(_v, list) else _v
+                _v.append(v)
```

shlao: Let me try it in my host.

xiangce (commented):
@shlao  Would you please update the `Examples` for this change as well?

xiangce (approved):

xiangce:
@bfahr - This change looks good to me, please help double confirm.

shlao:
@bfahr Will you help me to review this PR, thanks!

csams:
The proposed impl would require users to test every value to see whether it's a list or not. I prefer to keep the existing impl and handle the duplicate key use case separately, either in a different function or with a keyword argument to the current one. e.g. if `allow_duplicates` is true, _all_ values should be lists for consistency. Otherwise, go with the current behavior where the last duplicate key wins.

xiangce:
> The proposed impl would require users to test every value to see whether it's a list or not. I prefer to keep the existing impl and handle the duplicate key use case separately, either in a different function or with a keyword argument to the current one. e.g. if `allow_duplicates` is true, _all_ values should be lists for consistency. Otherwise, go with the current behavior where the last duplicate key wins.

I like the `allow_duplicates` idea. Thank you Chris. @shlao how do you think? Would you please update per Chris's proposal?

shlao:
@csams  Good idea. @xiangce  Will add a parameter `allow_duplicates` to make the parser backward-compatible.

shlao:
@csams if all values should be lists for consistency:
```
        >>> split_kv_pairs(lines, allow_duplicates=True)
        {'keyword2': ['value2a=True, value2b=100M'], 'keyword1': ['value1', 'value2']}
```

The the we need to change a lot in parser SystemctlShowServiceAll:
```
 91             data = split_kv_pairs(content[s:e], use_partition=False, allow_duplicates=True)
 92             name = data.get('Names', data.get('Id'))
 95             self[name] = dict((k, v) for k, v in data.items() if v)
```
Line 95 will hit this error
TypeError: unhashable type: 'list'

shlao:
@csams So I change it to store values in a list when there are two or more lines that have the same key
```
                     _v = kv_pairs[k]
                    _v = [_v] if not isinstance(_v, list) else _v
```

xiangce:
> @csams if all values should be lists for consistency:
> 
> ```
>         >>> split_kv_pairs(lines, allow_duplicates=True)
>         {'keyword2': ['value2a=True, value2b=100M'], 'keyword1': ['value1', 'value2']}
> ```
> 
> The the we need to change a lot in parser SystemctlShowServiceAll:
> 
> ```
>  91             data = split_kv_pairs(content[s:e], use_partition=False, allow_duplicates=True)
>  92             name = data.get('Names', data.get('Id'))
>  95             self[name] = dict((k, v) for k, v in data.items() if v)
> ```
> 
> Line 95 will hit this error
> TypeError: unhashable type: 'list'

@shlao - This is because the `name` is also wrapped into a list, it can be resolved by:
```python
            # There must be one and only one "name" as the key
            if not name or len(name) > 1:
                raise ParseException('"Names" or "Id" not found!')
            name = name[0]
            self[name] = dict((k, v) for k, v in data.items() if v)
```

As Chris mentioned, if only store the keys with multiple values into a list, it would require users to test every value to see whether it's a list or not. I don't think this is what we want.

shlao:
@xiangce  @csams  You can see what are changes in latest pr.
Beside that, we need to change the rules using the Parsers which invoke function split_kv_pairs:
```
postgresql_on_external_disk
httpd_fails_to_start_during_boot
sssd_network_user_auth_issue
```

xiangce:
@shlao - Yeap, that's the impact of this update (applying the "allow_duplicates=Ture" to this parser). It's reasonable for me to make this parser more flexible to support these `duplicated keys` (like a bugfix). And we'd better use this parser in this way (I think we should update the existing rules). @bfahr - How do you think?

insights/parsers/systemctl_show.py
```diff
@@ -29,6 +33,8 @@ class SystemctlShowServiceAll(CommandParser, dict):
         LimitNOFILE=65536
         LimitMEMLOCK=
         LimitLOCKS=18446744073709551615
+        DummyKey1 = Value1
+        DummyKey1 = Value2
```
bfahr, line 37: Is this valid for there to be two keys with the same name for a particular services in the output of this command?

insights/parsers/systemctl_show.py
```diff
@@ -45,6 +51,7 @@ class SystemctlShowServiceAll(CommandParser, dict):
                 "LimitNOFILE"      : "65536",
                 "TimeoutStartUSec" : "1min 30s",
                 "LimitLOCKS"       : "18446744073709551615",
+                "DummyKey1"        : ["Value1", "Value2"]
```
bfahr, line 54: Based on my comment above, these examples should all be lists.

insights/parsers/systemctl_show.py
```diff
@@ -164,7 +172,7 @@ class SystemctlShowTarget(SystemctlShowServiceAll):
         >>> 'network.target' in systemctl_show_target
         True
         >>> systemctl_show_target.get('network.target').get('WantedBy', None)
-        'NetworkManager.service'
+        ['NetworkManager.service']
```
bfahr, line 175: Same comment as above, I believe all values will be a `list` type.  Will this command actually return duplicates that require a list?

bfahr (changes requested):

shlao:
I close this one because there is no requirements now.

PR #2719, opened 2020-08-20
Print deep-link to HBI record if one exists

subpop:
After performing a collect and upload, query the API for the host record. If a record is returned, format and print a deep-link URL to the inventory record.


insights/client/__init__.py
```diff
@@ -547,6 +547,24 @@ def show_results(self):
             else:
                 raise e
 
+    def show_inventory_deep_link(self):
+        """
+        Show a deep link to this host inventory record
+        """
+        system = self.connection._fetch_system_by_machine_id()
+        if system is not None and system:
```
SteveHNH, line 555: Could you just do `if system:` here since None and False are both falsey values?

insights/client/__init__.py
```diff
@@ -547,6 +547,24 @@ def show_results(self):
...
+            if len(system) == 1:
+                try:
+                    id = system[0]["id"]
```
SteveHNH, line 558: It looks like fetch_system_by_machine_id returns a dict, not a list.  Could probably skip the length check above too since a dict length is how many keys are in it. I guess if you get a a system maybe a `if system.get("id")` would be a good check.

insights/client/__init__.py
subpop: Ah, yea probably. This is a side-effect of my "writing Python like Go" behavior. I have a habit of first checking for NULL, and then checking for value.

insights/client/__init__.py
subpop: Oh woops. I thought it was a list.

subpop: By docstring, it returns a `dict`, but in reality, it returns a `list`, so I'm going to leave this as is.

PR #2723, opened 2020-08-21
suppress help output for --check-results

subpop:
`--check-results` is not intended for general use, so we can suppress it from the `--help` output.


subpop:
Rebased on top of master given the changes made by #2725 .

PR #2724, opened 2020-08-24
connection: pull insights-client version directly

subpop:
Pull insights-client version directly from the constants package rather than relying on the pkg_resources info detection.


subpop:
test this please

gravitypriest:
test this please

PR #2735, opened 2020-08-28
do not redact client metadata files

gravitypriest:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1872777

SOSCleaner is run after metadata files are written. As such it's possible for them to get caught in SOSCleaner's obfuscation mechanisms. The problem we hit here was that the display name was set to the system's short hostname, which SOSCleaner will obfuscate to a generic identifier.

To fix it, a check was added to SOSCleaner to skip the client metadata files while scrubbing the archive.

In addition, skip everything outside the "data" dir when doing core collection. This includes our built-in redaction.


insights/contrib/soscleaner.py
```diff
@@ -1014,8 +1014,8 @@ def _clean_line(self, line, filename):
                 if false_positive in filename:
                     process_obfuscation = False
             new_line = self._sub_keywords(line)  # Keyword Substitution
-            if self.obfuscate_macs is True:
-                new_line = self._sub_mac(new_line)  # MAC address obfuscation
+            # if self.obfuscate_macs is True:
```
subpop, line 1017: Committing commented out code is pointless. The commit tracks the changes; if the code ever needs to be restored, a revert or partial revert can restore the code.

subpop (approved):
Looks good, aside from my one comment. The logic seems sound.

patchkez:
test this please

PR #2737, opened 2020-09-01
OpenStack logs are either containerized or not

valleedelisle:
We should look in both /var/log/service or /var/log/containers/service

Signed-off-by: David Vallee Delisle <dvd@redhat.com>


bfahr:
@valleedelisle this request needs to go through the spec approval process.

bfahr (approved):

PR #2745, opened 2020-09-07
added spec for ls_edac_mc parser

rasrivas-redhat:


rasrivas-redhat:
there is some dependency issue in my local environment, will fix the local environment and raise a new PR later.

PR #2749, opened 2020-09-07
The http_conf Name definition is too strict.

### LINKED ISSUE #2748: httpd_conf combiner doesn't handle identifier with digits in their names
The [Name definition](https://github.com/RedHatInsights/insights-core/blob/master/insights/combiners/httpd_conf.py#L278) in the http conf parser doesn't allow digits. That causes it to fail when parsing things like `H2Push`.

Change to something like:

```python
First = InSet(string.ascii_letters + "_/")
Rest = String(string.ascii_letters + "_/" + string.digits)
Name = (First + Rest).map("".join)
```

csams:
Allow digits after the first character in directive names. This allows directives like `H2Push`.

Fixes #2748

### DIFF
--- a/insights/combiners/httpd_conf.py
+++ b/insights/combiners/httpd_conf.py
@@ -275,7 +275,10 @@ class DocParser(object):
         Complex = Forward()
         Comment = (WS >> OneLineComment("#")).map(lambda x: None)
 
-        Name = String(string.ascii_letters + "_/")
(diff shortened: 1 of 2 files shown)


xiangce (approved):
@csams. Looks good to me. Thanks.

bfahr (approved):

PR #2750, opened 2020-09-08
Adding a numastat parser

valleedelisle:
The `numastat -m` parser is leveraging the alredy existing `MemInfo`
class for each numa node.

The plain `numastat` parser returns a dict with the various stats
as well as accessible percentage calculations.

Signed-off-by: David Vallee Delisle <dvd@redhat.com>


insights/parsers/numastat.py
```diff
@@ -0,0 +1,335 @@
...
+              {'Total': 0.0, 'node0': 0.0, 'node1': 0.0}
+
+        ``percent_foreign`` (dict): Percentage of foreign of all nodes:
+
+              {'Total': 0.0, 'node0': 0.0, 'node1': 0.0}
+
+        ``percent_load`` (dict): Show the percentage of hits on each node:
+
+              {'Total': 100.0, 'node0': 63.09, 'node1': 36.91}
+
```
bfahr, line 97: @valleedelisle please include an `Examples:` section and testing for the examples.  See the [documentation guidelines](https://insights-core.readthedocs.io/en/latest/docs_guidelines.html) and this [parser](https://github.com/RedHatInsights/insights-core/blob/625c82500cdcacc20bbd2d23e96d44e7fefa27cb/insights/parsers/ipsec_conf.py#L19) and [test](https://github.com/RedHatInsights/insights-core/blob/625c82500cdcacc20bbd2d23e96d44e7fefa27cb/insights/parsers/tests/test_ipsec_conf.py#L50) as examples.

insights/parsers/numastat.py
```diff
@@ -0,0 +1,335 @@
...
+from insights.specs import Specs
+import re
+from os import linesep
+
+
+@parser(Specs.numastat)
+class NUMAStat(CommandParser):
+    """
+    Class to process the output of ``numastat`` command.
+
```
bfahr, line 31: Please provide some sample output of the command.

insights/parsers/numastat.py
```diff
@@ -0,0 +1,335 @@
...
+             'node1': <insights.parsers.numastat.NUMAMemInfo object at 0x7f42d8633f28>},
+            }
+
+    """
+
+    # Not sure how to get this info from here
+    # so I made it a constant. This is pretty
+    # much standard anyways, at least in OSP
+    HUPAGESIZE = 1048576
+
```
bfahr, line 267: This looks OK.  Could you explain the comment?

insights/parsers/numastat.py
```diff
@@ -0,0 +1,335 @@
...
+            }
+
+       ``meminfo`` (dict): Dicitonnary of nodes with MemInfo objects:
+
+            {
+             'Total': <insights.parsers.numastat.NUMAMemInfo object at 0x7f42d8646278>,
+             'node0': <insights.parsers.numastat.NUMAMemInfo object at 0x7f42d8633860>,
+             'node1': <insights.parsers.numastat.NUMAMemInfo object at 0x7f42d8633f28>},
+            }
+
```
bfahr, line 260: Same comments as above regarding the documentation.

insights/parsers/numastat.py
```diff
@@ -0,0 +1,335 @@
...
+    def percent_foreign(self):
+        return {h: round(i['foreign'] / self.total_hits[h] * 100, 2) for h, i in self.data.items()}
+
+    @property
+    def percent_load(self):
+        return {h: round(self.total_hits[h] / self.total_hits['Total'] * 100, 2) for h, i in self.data.items()}
+
+
+@parser(Specs.numastat_memory)
+class NUMAStatMem(CommandParser):
```
bfahr, line 135: Same comment as above regarding basing the parser on `dict` as well.

insights/specs/default.py
```diff
@@ -495,6 +495,9 @@ def httpd_cmd(broker):
     ntpq_leap = simple_command("/usr/sbin/ntpq -c 'rv 0 leap'")
     ntptime = simple_command("/usr/sbin/ntptime")
     numa_cpus = glob_file("/sys/devices/system/node/node[0-9]*/cpulist")
+    numastat = simple_command("/bin/numastat")
+    numastat_memory = simple_command("/bin/numastat -m")
+    numactl_hw = simple_command("/bin/numactl --hardware")
```
bfahr, line 500: These specs will need to be approved before they can be collected.  Also you'll need to add corresponding `simple_file` specs to insights_archive.py for the data stored in an insights archive.  To find the correct path use the command `mangle {command_string}` and place the corresponding output in a spec in insights/specs/insights_archive.py.

insights/parsers/numastat.py
```diff
@@ -0,0 +1,335 @@
...
+
+from .. import parser, CommandParser
+from ..parsers.meminfo import MemInfo, AnonMemInfo, FileMemInfo, SlabMemInfo, HugePageMemInfo
+from insights.specs import Specs
+import re
+from os import linesep
+
+
+@parser(Specs.numastat)
+class NUMAStat(CommandParser):
```
bfahr, line 28: Instead of using a generic `data` attribute here consider deriving the parser from dict `NUMAStat(CommandParser, dict)` , and store the information directly in the parser object.  Here's an [example parser](https://github.com/RedHatInsights/insights-core/blob/625c82500cdcacc20bbd2d23e96d44e7fefa27cb/insights/parsers/ipsec_conf.py#L19) based on dict.  You may still add other attributes like `nodes` and properties.

bfahr (changes requested):
Please run flake8 from the command line to check for PEP8 errors.  Also please following the instructions in the [documentation guidelines](https://insights-core.readthedocs.io/en/latest/docs_guidelines.html#testing-your-docstring) to build and evaluate your rendered documentation.

insights/parsers/numastat.py
valleedelisle: When we look at `/proc/meminfo`, the HP are shown as numbers, but we can determine the amount of memory allocated by multiplying by the `Hugepagesize` attribute
```
[root@ess13latest-scpu-0 ~]# grep -P "Huge[pP]ages(ize:|_(Total|Free))" /proc/meminfo
HugePages_Total:     110
HugePages_Free:      102
Hugepagesize:    1048576 kB
```
When we look at `numastat -m`, we only have the amount of memory allocated to HP. We don't know how many HP or their size. 
```
[root@ess13latest-scpu-0 ~]# numastat -m | grep -P "Huge[pP]ages(ize:|_(Total|Free))"
HugePages_Total         56320.00        56320.00       112640.00
HugePages_Free          52224.00        52224.00       104448.00
```

The `MemInfo` parser is expecting the number of HP and not the amount of memory. So I need to guess the number of HP based on missing variables. Ideally, I'd grab the HP size from `/proc/meminfo`, from inside that parser but I'm not sure how to do this.

I think 95% of the deployments are using HP size of 1048576 but I've seen deployments with smaller "huge" page size like 4096.

valleedelisle:
Addressed pretty much everything, `flake8` passed, docs compiled and looked good.

bfahr:
@valleedelisle please check the CI results.  You'll need to change this dictionary comprehension which is not supported in Python 2.6.
`return {h: i['hit'] + i['miss'] + i['foreign'] for h, i in self.items()}`

valleedelisle:
> @valleedelisle please check the CI results. You'll need to change this dictionary comprehension which is not supported in Python 2.6.
> `return {h: i['hit'] + i['miss'] + i['foreign'] for h, i in self.items()}`

This is fixed

PR #2755, opened 2020-09-09
Print meaningful config error message

vkrizan:
Not just plain: "ERROR: Could not read configuration file, using defaults"

The message format is the same as for `ValueError` used down below (for key validation).

### DIFF
--- a/insights/client/config.py
+++ b/insights/client/config.py
@@ -586,11 +586,11 @@ class InsightsConfig(object):
         parsedconfig = ConfigParser.RawConfigParser()
         try:
             parsedconfig.read(fname or self.conf)
-        except ConfigParser.Error:
+        except ConfigParser.Error as e:
             if self._print_errors:
                 sys.stdout.write(
-                    'ERROR: Could not read configuration file, '
-                    'using defaults\n')
+                    'ERROR: {0}.\nCould not read configuration file, '
+                    'using defaults\n'.format(e))
             return
         try:
             if parsedconfig.has_section(constants.app_name):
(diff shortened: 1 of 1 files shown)


vkrizan:
@gravitypriest Would you mind having a look. Thank you.

insights/client/config.py
```diff
@@ -586,11 +586,11 @@ def _load_config_file(self, fname=None):
         parsedconfig = ConfigParser.RawConfigParser()
         try:
             parsedconfig.read(fname or self.conf)
-        except ConfigParser.Error:
+        except ConfigParser.Error as e:
             if self._print_errors:
                 sys.stdout.write(
-                    'ERROR: Could not read configuration file, '
-                    'using defaults\n')
+                    'ERROR: {0}.\nCould not read configuration file, '
```
psachin, line 592: @vkrizan 

As per the changes the error message will look like below,

Example:
```
ERROR: ValueError. Could not read configuration file, using defaults
```

How about we remove 'ERROR' so that the message will be as below?

```
ValueError. Could not read configuration file, using defaults
```

vkrizan: Sure we can do that, however not every Exception ends with "Error" to realize it is an error message.

vkrizan: Too soon. :slightly_smiling_face:  The base exception this is catching has all child exception ending with Error: https://docs.python.org/3/library/configparser.html#exceptions

vkrizan:
@psachin can you please have a look? Thanks!

vkrizan:
#stale

PR #2758, opened 2020-09-13
Add deny filtering capabilitiy

bfahr:
* Adds new deny filter to allow removal of lines that match the deny
  criteria
* Add testing for new deny filters

### DIFF
--- a/insights/core/filters.py
+++ b/insights/core/filters.py
@@ -48,6 +48,7 @@ from insights.util import parse_bool
 
 _CACHE = {}
 FILTERS = defaultdict(set)
+DENY_FILTERS = defaultdict(set)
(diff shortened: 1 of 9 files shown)


bfahr:
@csams @gravitypriest here a first stab at adding deny filters to core.  It should be transparent to end users with no changes to `add_filter`.  A new `add_deny_filter` will create the new type of filters.  Deny filters are incorporated into the `blacklist`, making them mutually exclusive of allow filters and they take precedence over allow filters.  This may not be the best approach if the goal is to remove sensitive info, but also reduce the amount of data collected.

Doc will be part of this PR, but I don't want to create them until we work out the details of serialization and precedence/mutual exclusivity.

xiangce:
Closing it, we can reopen it in another PR if necessary.

PR #2759, opened 2020-09-14
only call _sub_hostname if option is enabled, fix typo

gravitypriest:
Fixes: RHCLOUD-9134

Summary of changes:
- Persist `obfuscate_hostname` as an option into soscleaner, and only call the function to sub hostnames if this option is enabled.
- Fix typo: `ofuscateddomain` => `obfuscateddomain`

### DIFF
--- a/insights/client/data_collector.py
+++ b/insights/client/data_collector.py
@@ -422,8 +422,8 @@ class CleanOptions(object):
         self.networks = None
         self.users = None
         self.users_file = None
-        self.obfuscate_macs = False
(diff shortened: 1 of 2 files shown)


jhutar:
Hello. We ran the performance test creating a tarball and with EGG from master, it took 42.17 seconds and with EGG from this PR it took 28.34 seconds. Also archive size changed from 726585 (master) to 693321 (this PR). Is it possible this also changes current behaviour and disables obfuscation by default?

PR #2769, opened 2020-09-22
Spec and parser for mokutil command

jobselko:

### DIFF
--- a/insights/parsers/mokutil_sbstate.py
+++ b/insights/parsers/mokutil_sbstate.py
@@ -1,0 +1,39 @@
+"""
+MokutilSbstate - command ``mokutil --sb-state``
+===============================================
+"""
+
+from insights.core import CommandParser
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+
+@parser(Specs.mokutil_sbstate)
+class MokutilSbstate(CommandParser):
+    """
+    Class for parsing the ``mokutil --sb-state`` command.
+
+    Attributes:
+        secureboot_enabled (bool): True if SecureBoot is enabled,
+                                   False if SecureBoot is disabled, otherwise None.
+
+    Sample output of this command is::
+
+        SecureBoot enabled
+
+    Examples:
+
+        >>> type(mokutil)
+        <class 'insights.parsers.mokutil_sbstate.MokutilSbstate'>
+        >>> mokutil.secureboot_enabled
+        True
+    """
+
+    def parse_content(self, content):
+        self.secureboot_enabled = None
+
+        non_empty_lines = [line for line in content if line]  # get rid of blank lines
+        if "SecureBoot enabled" in non_empty_lines[0]:
+            self.secureboot_enabled = True
+        elif "SecureBoot disabled" in non_empty_lines[0]:
+            self.secureboot_enabled = False
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -432,6 +432,7 @@ class DefaultSpecs(Specs):
     modinfo_veth = simple_command("/sbin/modinfo veth")
     modinfo_vmxnet3 = simple_command("/sbin/modinfo vmxnet3")
     modprobe = glob_file(["/etc/modprobe.conf", "/etc/modprobe.d/*.conf"])
+    mokutil_sbstate = simple_command("/bin/mokutil --sb-state")
     mongod_conf = glob_file([
                             "/etc/mongod.conf",
                             "/etc/mongodb.conf",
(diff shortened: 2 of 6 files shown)


insights/parsers/mokutil_sbstate.py
```diff
@@ -0,0 +1,39 @@
...
+    Examples:
+
+        >>> type(mokutil)
+        <class 'insights.parsers.mokutil_sbstate.MokutilSbstate'>
+        >>> mokutil.secureboot_enabled
+        True
+    """
+
+    def parse_content(self, content):
+        self.secureboot_enabled = None
```
xiangce, line 33: Just a query, from the manual of `mokutil`, it seems that the `mokutil --sb-state` will only print the state of  `SecureBoot`. How do you think to raise a ParseExpection when the output is not one of `"SecureBoot enabled"` or `"SecureBoot disable"`?

jsvob: It seems to me that in case `mokutil` returns `EFI variables are not supported on this system`, `self.secureboot_enabled` is left at None, which also agrees with the class docstring. Perhaps the docstring could mention that case more explicitly?

bfahr: @jsvob @jobselko does returning `None` make sense since it just lets you know that the command is installed.  Is it meaningful information for a rule?  Its not a great pattern to make rules check for None then True/False, unless all 3 values are important.
```
$ mokutil --sb-state
EFI variables are not supported on this system
```

jobselko: I think it is a matter of opinion. I can use `False` or an exception if it is a preferred option. Let me know.

jsvob: @bfahr @jobselko 

IMO, Having a tri-state return value results in the least amount of work when implementing rules depending on this parser. If coalescing the 3 states to 2 return values is necessary, either way (None->False, None->exception) is about the same amount of work:

1. Hypothetical use case: a rule needs to determine whether secureboot is enabled or disabled, without care for BIOS/UEFI (UEFI disabled and BIOS being both considered as disabled), and with a need for trust in the data (in case uploader.json is outdated):
    * Parser being left as is (tri-state): Required parser for mokutil. If the value is True, secureboot is enabled with high confidence. If the value is Falsey, secureboot is explicitly (UEFI) or implicitly (BIOS) disabled with high confidence. If the requirement is not met (Exception), uploader.json is outdated or the spec is blacklisted, with high confidence.
    * Parser modified - None->Exception: Optional parser for mokutil. If the parser's value is a skip == None, a separate check for the `mokutil` RPM is necessary to figure out whether it's outdated uploader.json and dependencies are not met, or whether secureboot is implicitly disabled (BIOS).
    * Parser modified - None->False: Required parser for mokutil. Same as `Parser being left as is (tri-state)`.
2. Hypothetical use case: a rule needs to determine whether the machine is BIOS, UEFI with disabled secureboot, or UEFI with enabled secureboot, with a need for trust in the data (in case uploader.json is outdated):
    * Parser being left as is (tri-state): Required parser for mokutil. True == UEFI secureboot, False == UEFI nonsecureboot, None == BIOS, all with high confidence. If the requirement is not met (Exception), uploader.json is outdated or the spec is blacklisted, with high confidence.
    * Parser modified - None->Exception: Optional parser for mokutil. If the parser's value is a skip == None, a separate check for the `mokutil` RPM is necessary to figure out whether it's outdated uploader.json and dependencies are not met, or whether it's a BIOS system (with secureboot implicitly disabled).
    * Parser modified - None->False: Required parser for mokutil. True == UEFI secureboot. False == either UEFI nonsecureboot, or BIOS. InstalledRpms query necessary to figure out which False state it actually is.

bfahr: @jsvob thanks for the explanation.  I'm OK with keeping the tri-state return based on your explanation, I just wanted to make sure that the `None` value was meaningful to the rule.

insights/specs/default.py
```diff
@@ -432,6 +432,7 @@ def httpd_cmd(broker):
     modinfo_veth = simple_command("/sbin/modinfo veth")
     modinfo_vmxnet3 = simple_command("/sbin/modinfo vmxnet3")
     modprobe = glob_file(["/etc/modprobe.conf", "/etc/modprobe.d/*.conf"])
+    mokutil_sbstate = simple_command("mokutil --sb-state")
```
bfahr, line 435: @jobselko would you please add the path for this command, for example `/usr/bin/mokutil --sb-state` or `/bin/modutil --sb-state`?  It need to match the JSON spec.

bfahr (changes requested):
Approved once the requested change has been made.

bfahr (approved):

PR #2771, opened 2020-09-23
generate log and lib dirs [RHCLOUD-8677]

gravitypriest:
Resolves: RHCLOUD-8677

Generate the log and lib dirs from within the egg, creating them if they don't exist. Some bonus non-root stuff in this commit as well.

TODO: Perhaps do the creation of the directories outside the constants file. Right now it gets done when the module is imported, it may be more prudent to create them as a distinct step as part of the InsightsClient init.


gravitypriest:
The one thing left to decide is whether to create these dirs with any particular perms.

subpop:
Hm, so if I were to run a REPL console and import `insights.client.constants`, it would attempt to create those directories. That means that if for some reason the directories could not be created, the only error handling option available to the runtime is to raise an Exception. Does that fail to import the whole module?

insights/client/constants.py
```diff
@@ -1,8 +1,42 @@
...
+    '''
+    Get the insights-client log dir
+
+    Default: /var/log/insights-client
+    Non-root user: $XDG_CACHE_HOME/insights-client || $HOME/.cache/insights-client
+    '''
+    if _uid == 0:
+        insights_log_dir = os.path.join(os.sep, 'var', 'log', _app_name)
+    else:
+        local_log_dir = os.getenv('XDG_CACHE_HOME', default=os.path.join(_user_home, '.cache'))
```
subpop, line 18: Nice use of the XDG base directory specification. I looked at what options there are in the specification. Looks like we have 3 to choose from:

* `XDG_CONFIG_HOME`: user-specific configuration files
* `XDG_CACHE_HOME`: user-specific non-essential (cached) data
* `XDG_DATA_HOME`: user-specific data files

There isn't really a set of XDG directories that directly map to the various high-level directories in the FHS (`/var`, `/run`, `/usr/share`, `/etc`, etc.). I think it makes sense to put the logs under `~/.cache`. Does it make sense to also put the downloaded eggs into `~/.cache`? What would happen to a registered client if the latest.egg suddenly disappeared? Would it simply download a new egg again? If so, I think it could be reasoned to be "non-essential data". This way we aren't deviating from the specification of XDG base directories and creating our own hierarchy.

insights/client/constants.py
```diff
@@ -1,8 +1,42 @@
...
+
+def _lib_dir():
+    '''
+    Get the insights-client egg cache dir
+
+    Default: /var/lib/insights
+    Non-root user: $HOME/.local/lib/insights-client
+    '''
+    if _uid == 0:
+        insights_lib_dir = os.path.join(os.sep, 'var', 'lib', 'insights')
```
subpop, line 31: See previous comment.

subpop (changes requested):

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

subpop (approved):

patchkez:
test this please

subpop:
Is this waiting for anything in particular?

PR #2779, opened 2020-10-05
Fix legacy test-connection status code assertion and api call

diegorafrh:


bfahr:
@diegorafrh you will need to fix the DCO CI test by signing off on the request with the `-s` switch on all commits.  See the "Details" in the DCO CI test above.

xiangce:
Closing it temporary, please open it again in other PR if required

PR #2780, opened 2020-10-07
Remove unneeded parsers libkeyutils and whoopsie

matysek:
Signed-off-by: Martin Zibricky <mzibrick@redhat.com>

For security rules we no longer need these parsers.


matysek:
@jsvob @skontar @kgrant-rh please review

insights/specs/__init__.py
```diff
@@ -278,8 +278,6 @@ class Specs(SpecSet):
     ksmstate = RegistryPoint()
     kubepods_cpu_quota = RegistryPoint(multi_output=True)
     lastupload = RegistryPoint(multi_output=True)
-    libkeyutils_objdumps = RegistryPoint()
-    libkeyutils = RegistryPoint()
```
bfahr, line 282: @matysek don't forget to remove `woopsie` as well from this file.

bfahr (changes requested):

insights/specs/__init__.py
matysek: Yeah, the variable was called 'woopsie' and not 'whoopsie' so forgot to remove it. Thank you!.

bfahr (approved):

PR #2785, opened 2020-10-14
Add du_dirs Spec to contexts

Rohan21Lobo:
The Spec was not used earlier in the rule and was removed in https://github.com/RedHatInsights/insights-core/pull/2663.


Rohan21Lobo:
@bfahr  Hi Can you review this? 
Seems the spec was removed as it was not used. Now a rule for this is merged and it needs this spec.

xiangce (approved):

PR #2786, opened 2020-10-14
Added parser for testparm data for Samba configuration

skontar:
testparm command parsing is the better way for handling Samba configuration. Especially it allows getting server role.

The code in the review is not considered to be final but rather draft. I would like to start a conversation about how to handle this.

### DIFF
--- a/insights/parsers/samba.py
+++ b/insights/parsers/samba.py
@@ -70,6 +70,7 @@ Examples:
 """
 import re
 
+from . import ParseException
 from .. import add_filter, IniConfigFile, parser
 from insights.specs import Specs
 
@@ -120,12 +121,10 @@ class SambaConfigs(SambaConfig):
     """
     This parser reads the Samba configuration from command `testparm -v -c` which is more
     reliable than parsing the config as well as contains server role.
+
+    Attributes:
+    server_role (string): Server role as reported by the command.
     """
-
-    def __init__(self, *args, **kwargs):
-        self.server_role = None
-        super(SambaConfigs, self).__init__(*args, **kwargs)
-
     def parse_content(self, content):
         # Parse server role
         for line in content:
@@ -133,6 +132,8 @@ class SambaConfigs(SambaConfig):
             if r:
                 self.server_role = r.group(1)
                 break
+        else:
+            raise ParseException("Server role not found.")
 
         # Technically, the implicit [global] section handling is not needed as testparm command
         # processes it, but for simplicity the old parser is used so the logic is not duplicated
--- a/insights/parsers/samba.py
+++ b/insights/parsers/samba.py
@@ -2,17 +2,17 @@
 SambaConfig - file ``/etc/samba/smb.conf``
 ==========================================
 
-This parser reads the SaMBa configuration file ``/etc/samba/smb.conf``, which
+This parser reads the Samba configuration file ``/etc/samba/smb.conf``, which
 is in standard .ini format, with a couple of notable features:
 
-* SaMBa ignores spaces at the start of options, which the ConfigParser class
-  normally does not.  This spacing is stripped by this parser.
-* SaMBa likewise ignores spaces in section heading names.
-* SaMBa allows the same section to be defined multiple times, with the
+* Samba ignores spaces at the start of options, which the ConfigParser class
+  normally does not. This spacing is stripped by this parser.
+* Samba likewise ignores spaces in section heading names.
+* Samba allows the same section to be defined multiple times, with the
   options therein being merged as if they were one section.
-* SaMBa allows options to be declared before the first section marker.
+* Samba allows options to be declared before the first section marker.
   This parser puts these options in a `global` section.
-* SaMBa treats ';' as a comment prefix, similar to '#'.
+* Samba treats ';' as a comment prefix, similar to '#'.
 
 Sample configuration file::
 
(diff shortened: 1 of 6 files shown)


skontar:
https://gitlab.cee.redhat.com/insights-release-eng/insights-core-assets/-/merge_requests/90

insights/parsers/samba.py
```diff
@@ -108,3 +109,31 @@ def parse_content(self, content):
...
+
+@parser(Specs.testparm_v_s)
+class SambaConfigs(SambaConfig):
+    """
+    This parser reads the Samba configuration from command `testparm -v -c` which is more
+    reliable than parsing the config as well as contains server role.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.server_role = None
```
xiangce, line 126: Please add the corresponding documentation for this new attribute. e.g. to the "Attributes:" docstring of this class.


BTW, since the `parse_content` will also be called by the `__init__` of the base class, this `self.server_role` can also be declared in the `parse_content` (so the re-implentmentation of this `__init__` can be saved :-)

insights/parsers/samba.py
```diff
@@ -108,3 +109,31 @@ def parse_content(self, content):
...
+        super(SambaConfigs, self).__init__(*args, **kwargs)
+
+    def parse_content(self, content):
+        # Parse server role
+        for line in content:
+            r = re.search(r"Server role:\s+(\S+)", line)
+            if r:
+                self.server_role = r.group(1)
```
xiangce, line 134: From the above docstring, it seems the `server_role` is a must-exist attribute for this parser. Is it OK to leave its value as `None` (the current logic) when cannot found the `Server role:` line from the spec, OR raise a `ParseException/SkipException` for this case?

insights/parsers/samba.py
skontar: You are right. Liner in PyCharm complains about unknown attributes, so that is why I usually add it in `__init__` but I agree it is unnecessary.

insights/parsers/samba.py
skontar: I looked at the `testparm` source and with the current options in the spec it should be always present. I think parse exception makes sense.

xiangce:
@skontar - The MR looks good to me. Would you please fix the following flake8 errors? So it can be merged.
```
+ flake8

./insights/parsers/tests/test_samba.py:206:1: E101 indentation contains mixed spaces and tabs
./insights/parsers/tests/test_samba.py:206:1: W191 indentation contains tabs
./insights/parsers/tests/test_samba.py:206:28: W291 trailing whitespace
./insights/parsers/tests/test_samba.py:207:1: W191 indentation contains tabs
./insights/parsers/tests/test_samba.py:208:1: W191 indentation contains tabs
./insights/parsers/tests/test_samba.py:208:20: W291 trailing whitespace
./insights/parsers/tests/test_samba.py:209:1: W191 indentation contains tabs
./insights/parsers/tests/test_samba.py:210:1: W191 indentation contains tabs
./insights/parsers/tests/test_samba.py:213:1: W191 indentation contains tabs
./insights/parsers/tests/test_samba.py:219:1: E101 indentation contains mixed spaces and tabs
./insights/parsers/tests/test_samba.py:270:1: W391 blank line at end of file
```

skontar:
@xiangce I think that apart from the last one it is a bug in flake8. I believe that flake8 should not be concerned about mixed tabs and spaces inside the multi-line string. The text is direct copy/paste from the command output. I have just added directives for flake8 to ignore it.

cryptomilk:
What about https://github.com/RedHatInsights/insights-core/pull/2487

skontar:
@cryptomilk I talked to @kgrant-rh (or @jsvob ?) and I think it is useful to have both. `testparm -s /etc/samba/smb.conf` parses just the config file and it is always better to avoid us doing that directly. `testparm -v -s` parses all configuration, including "For example, for RHEL IdM we store Samba configuration in internal registry and parsing smb.conf will not help you as it just redirects to look in the registry which is a binary database." 

In the rule it is important to both know the whole configuration and if it comes from the default config file, so we can tell user if they should fix it in the config file or "elsewhere".

cryptomilk:
Changing `testparm -s /etc/samba/smb.conf` to `testparm -s` is not a big deal. You should not do -v as it gives you all options instead of just the options which differ from the default. It isn't really useful ...

cryptomilk:
I've changed https://github.com/RedHatInsights/insights-core/pull/2487

skontar:
@cryptomilk I believe that it is not trivial to find defaults for different Samba versions, so that is the reason why `-v` was added. However, the current spec and parser is done based on Security Analysts recommendation in https://projects.engineering.redhat.com/browse/PSINSIGHTS-197 , I do not have extensive knowledge, I am just following the information they provided. So any contradiction should be brought up with them.

skontar:
After a discussion with @cryptomilk we decided that we actually need three specs/parsers.

1. For `/etc/samba/smb.conf` so we know what specifically is in the file for better resolution recommendations.
2. For `testparm -s` so we know what changes user configured, either in the config file or elsewhere, like in the binary database.
3. For `testparm -s -v` so we know what options are currently in effect, including defaults which can differ in different versions of Samba.

I will be doing changes to this merge request to accommodate all these requirements.

bfahr (approved):

PR #2787, opened 2020-10-14
Fix cmdline utils for core archives

bfahr:
* To properly hydrate core for core collected archives we need to use
  new hydration function to identify context and hydrate broker


bfahr:
Fixes #2788

lhuett (approved):
Looks good

PR #2792, opened 2020-10-20
add --module option to run modules from cli

gravitypriest:


insights/client/config.py
```diff
@@ -232,6 +232,12 @@ def _core_collect_default():
         'action': 'store_true',
         'group': 'debug'
     },
+    'module': {
```
subpop, line 235: Can we make this option hidden from the `--help` output by default, since it's not really meant for public consumption?

subpop (changes requested):

subpop (approved):

patchkez:
test this please

gravitypriest:
Merging because only test failures are test-unregister, unrelated to change set.

PR #2794, opened 2020-10-21
new parser networkmanager_config

rasrivas-redhat:

### DIFF
--- a/insights/parsers/tests/test_networkmanager_config.py
+++ b/insights/parsers/tests/test_networkmanager_config.py
@@ -109,12 +109,12 @@ domains=ALL
 
 def test_networkmanager_config_match():
     result = NetworkManagerConfig(context_wrap(NETWORKMANAGER_CONF))
-    result.get('main', 'dhcp') == 'dhclient'
+    assert result.get('main', 'dhcp') == 'dhclient'
 
 
 def test_networkmanager_config_notmatch():
     result = NetworkManagerConfig(context_wrap(NETWORKMANAGER_CONF_NOTMATCH))
-    result.has_option('main', 'dhcp') is False
+    assert result.has_option('main', 'dhcp') is False
 
 
 def test_networkmanager_config_doc_examples():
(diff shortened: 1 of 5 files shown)


xiangce (approved):

xiangce (approved):

insights/parsers/tests/test_networkmanager_config.py
```diff
@@ -0,0 +1,71 @@
...
+# in man journald.conf).
+#
+#level=TRACE
+#domains=ALL
+"""
+
+
+def test_match():
+    result = NetworkManagerConfig(context_wrap(NETWORKMANAGER_CONF))
+    assert result.data.get('main', 'dhcp')
```
bfahr, line 63: @rasrivas-redhat could you actually check the value here, and not just that its present in the data?  Also it might be good to have a negative test, perhaps that `[main] plugins` is not present?

bfahr (changes requested):

insights/parsers/tests/test_networkmanager_config.py
rasrivas-redhat: @bfahr - I have updated the test case for it and also added a test case to check for a negative test, please review.

insights/parsers/tests/test_networkmanager_config.py
```diff
@@ -57,10 +57,64 @@
...
+#
+#level=TRACE
+domains=ALL
+"""
+
+
+def test_networkmanager_config_match():
     result = NetworkManagerConfig(context_wrap(NETWORKMANAGER_CONF))
-    assert result.data.get('main', 'dhcp')
+    result.get('main', 'dhcp') == 'dhclient'
```
bfahr, line 112: @rasrivas-redhat you need to add `assert` here.

insights/parsers/tests/test_networkmanager_config.py
```diff
@@ -57,10 +57,64 @@
...
+
+
+def test_networkmanager_config_notmatch():
+    result = NetworkManagerConfig(context_wrap(NETWORKMANAGER_CONF_NOTMATCH))
+    result.has_option('main', 'dhcp') is False
```
bfahr, line 117: Same as comment above.

bfahr (changes requested):

bfahr (approved):

PR #2799, opened 2020-10-23
Save the raised exception

huali027:
* Some gss rules have to check the raised exception to check specific
  issue when subscription-manager can't work

### DIFF
--- a/insights/parsers/subscription_manager_list.py
+++ b/insights/parsers/subscription_manager_list.py
@@ -45,6 +45,9 @@ class SubscriptionManagerList(CommandParser):
         # it starts a new record.
 
         for line in content:
+            if 'Traceback' in line:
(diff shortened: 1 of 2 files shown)


insights/parsers/subscription_manager_list.py
```diff
@@ -25,6 +25,10 @@ class SubscriptionManagerList(CommandParser):
     A general object for parsing the output of ``subscription-manager list``.
     This should be subclassed to read the specific output - e.g. ``--consumed``
     or ``--installed``.
+
+    Attributes:
+        records (list): A list of dict with the output info
```
xiangce, line 30: Better add `, empty list when an `error` occurs`

xiangce (approved):

PR #2801, opened 2020-10-26
New parser for GoogleInstanceType

xiangce:


xiangce:
@bfahr - This PR is ready, please help review whenever free. Thanks.

insights/specs/default.py
```diff
@@ -433,6 +433,7 @@ def is_gcp(broker):
             return True
         raise SkipComponent()
 
+    gcp_instance_type = simple_command("/usr/bin/curl -s -H 'Metadata-Flavor: Google' http://metadata.google.internal/computeMetadata/v1/instance/machine-type --connect-timeout 5", deps=[is_gcp])
```
ryan-blakley, line 436: @xiangce Would it be better to add &alt=json to the url and parse the json output instead of the text output? I was thinking it might be easier to parse the json output, and in the future Google may add more key:val information to the endpoint output in a different order. Which could cause a parse error. Where with json if we parse based on the keys it wouldn't be an issue if they added more endpoint output.

xiangce: @ryan-blakley - Good idea, I didn't notice that google provides this feature in the API. However,  since I have not a google instance to test and not find related information from the internet, I have no idea how the output will be when adding the `alt=json` option? I'll update the PR once getting any clue.

xiangce: @ryan-blakley - I'm lucky and get a gcp vm to have a test. However, I got the following output
```
[root@piqin-test ~]# curl -s -H 'Metadata-Flavor: Google' "http://metadata.google.internal/computeMetadata/v1/instance/machine-type?alt=text" -w'\n'
projects/1042363005003/machineTypes/e2-medium
[root@piqin-test ~]# curl -s -H 'Metadata-Flavor: Google' "http://metadata.google.internal/computeMetadata/v1/instance/machine-type?alt=json" -w'\n'
"projects/1042363005003/machineTypes/e2-medium"
[root@piqin-test ~]# curl -s -H 'Metadata-Flavor: Google' "http://metadata.google.internal/computeMetadata/v1/instance/machine-type" -w'\n'
projects/1042363005003/machineTypes/e2-medium
```
It seems the `alt=json` doesn't work for this API. If it's true, I think we don't need to change it, right. How do you think?

ryan-blakley: @xiangce Interesting I guess Google's api docs are incorrect, so I think no change is needed as you mentioned.

lhuett: @xiangce It appears your PR has conflicts in insights/specs/default.py, would you mind correcting those?

xiangce: @ryan-blakley - Thanks for reviewing.
@lhuett - Conflicts are fixed. Thanks!

xiangce: @lhuett - I go ahead and merge this PR since there is an advisor candidate depends on it. Thanks.

PR #2809, opened 2020-10-30
Ensure hostname comparisons in SAP are accurate

SteveHNH:
There is a possibility that the FQDN of a system is used as the
shortname in SAP configurations. We want to verify that the short
hostname is used in these comparisons.


CodeHeeler (approved):

xiangce (approved):
Good to me. Thanks @SteveHNH

PR #2811, opened 2020-11-03
New parser for etc udev rules

akshay196:
New parser similar to the udev_rules but parse files under
`/etc/udev/rules.d/` directory instead.


akshay196:
@xiangce @JoySnow Can you please review this parser?

JoySnow:
This parser looks good to me.

xiangce (approved):

PR #2821, opened 2020-11-09
Enhance parser grub_conf

wushiqinlou:
To check RHEL8 grub config, it is necessary to add "IsRhel8" in "Grub2Config".


wushiqinlou:
@xiangce - Please have a check

PR #2826, opened 2020-11-17
Add components to manifest for core collection

bfahr:
* Adding new combiners and parsers they depend on to manifest so that
  those components will be loaded for core collection
* If these components are not in the manifest they won't run, and the
  associated specs will not be collected due to dependency failures


bfahr:
These changes are required for the `is_aws`, `is_azure` and `pcp_enabled` specs to function properly with core collection.

gravitypriest (approved):

lhuett (approved):
Looks good

PR #2827, opened 2020-11-18
Skip the insights-client self grep process

chenlizhong:


insights/parsers/ps.py
```diff
@@ -75,6 +75,9 @@ def parse_content(self, content):
             # The above list comprehension assures all rows have a command.
             for proc in self.data:
                 cmd = proc[self.command_name]
+                # skip the insights-client self grep process
+                if cmd.startswith('grep') and 'httpd' in cmd and 'nginx' in cmd and 'corosync' in cmd:
```
xiangce, line 79: @chenlizhong, the `httpd`, `nginx`, and `corosync` are white-filters getting from the specific rules, they should not be hard-coded here in the parser. Better remove them and change the condition as:
```
if cmd.startswith('grep '):
```

chenlizhong: But we can not skip all grep process, maybe it is from customer. add httpd, nginx and cororsync to ensure it is just the collector self process

chenlizhong: OK. Update

xiangce (approved):

PR #2830, opened 2020-11-18
Add grep command to blacklist

bfahr:
* The grep command is used to apply filters for specs but it also shows
  up in the results of the ps command for core collection.  This can
  cause false positives for rules having filters that are too broad.
  Removing this command will help reduce the number of false positives
  from rules.

### DIFF
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -54,7 +54,8 @@ client:
     blacklist:
         files: []
         commands: []
-        patterns: []
+        patterns:
+            - "grep -F"
         keywords: []
 
     # Can be a list of dictionaries with name/enabled fields or a list of strings
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -34,6 +34,14 @@ SAFE_ENV = {
     "LC_ALL": "C",
 }
 
+FILTER_COMMAND = "grep -F"
+"""
+Command used to filter results collected from the system
+in insights/client/insights_spec.py. It will be added to
+the blacklist here so that its not included in the collected
+data.
+"""
+
 if "LANG" in os.environ:
     SAFE_ENV["LANG"] = os.environ["LANG"]
 
@@ -55,7 +63,7 @@ client:
         files: []
         commands: []
         patterns:
-            - "grep -F"
+            - "{filter_cmd}"
         keywords: []
 
     # Can be a list of dictionaries with name/enabled fields or a list of strings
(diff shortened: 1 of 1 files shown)


miclark:
Should we remove the change made to the `Ps` parser as part of this change?

bfahr:
@miclark yes, once this PR is approved, PR #2827 should be reverted.

chenlizhong:
This is best way that we remove the useless line by client.  Looks excellent to me.

chenlizhong (approved):
This is best way that we remove the useless line by client. Looks excellent to me.

xiangce:
> @miclark yes, once this PR is approved, PR #2827 should be reverted.

@miclark, @bfahr - This PR looks good to me. However, I don't think it's good to revert the change of PR #2827 even after merging this MR.

The `"blacklist, patterns"` you added is just the `default_mainfest` hardcoded inner the core:
https://github.com/RedHatInsights/insights-core/blob/048068164ba138a8033be949185e212ad43ae8a0/insights/collect.py#L217

Once the users define their own "blacklist patterns" (which does not contain this default one) on their hosts, e,g. in the `file-content-redaction.yaml`, this `default_mainfest` will be over-written. And then the `"grep -F"` line will be collected in this case.
https://github.com/RedHatInsights/insights-core/blob/048068164ba138a8033be949185e212ad43ae8a0/insights/client/client.py#L288
https://github.com/RedHatInsights/insights-core/blob/048068164ba138a8033be949185e212ad43ae8a0/insights/client/client.py#L303

I think we'd better keep both of these changes as double insurance for this issue.

jhjaggars:
I'm not sure I understand the value of filtering the lines out on the client side at collection time.  I'm sure we're not likely to filter out useful data but we might take out a useful line with `grep -F` in it by mistake and have no way to recover it right?

It'd be a lot more work, but we could filter the ps list based on the processes that we spawned as part of collection which would be safer (but probably not worth it).

bfahr:
@jhjaggars your question leads back to the comment that @csams and @miclark made about rules being more particular with their filters and how they evaluate the filtered data.  Since filters can be added by any number of rules, there is always a possibility that a particular rule will get more data than it requested via filters.  Enhancing the rules to be more selective when evaluating filtered data would probably be the better option.

That said, this method is the biggest hammer in the toolbox, and filtering out all of the artifacts introduced into the ps output would be a better solution.

bfahr:
Switching to a draft PR to allow further discussion.

chenlizhong:
@jhjaggars  

For `grep -F` command, it is transient process not long run.  In the future,  grep command has bug, hang process here, it is not proactively, customer will know it when invoke it. So no rule for grep command.  it is safe to filter out it.

Currently, the grep -F process is from insights-client itself.

bfahr:
Closing this PR as its not required at this time.  Maybe in the future.

PR #2841, opened 2020-12-01
Enhance the package_provides_java and package_provides_httpd

xiangce:
- Enhance the logic of the datasource to get the command from ps
- Add new parsers in packge_provides 
- Deprecate the package_provides_java parser and combiner
- Deprecate the package_provides_httpd parser and combiner
- Add the dependent ps to default dependencies


insights/combiners/package_provides_httpd.py
```diff
@@ -33,6 +34,10 @@ class PackageProvidesHttpdAll(LegacyItemAccess):
     """
 
     def __init__(self, package_provides_httpd):
+        deprecated(
```
bfahr, line 37: @xiangce since this method is not included in the documentation, could you please add info in the docstring about deprecation of this combiner, thanks.

insights/combiners/package_provides_java.py
```diff
@@ -40,6 +41,10 @@ class PackageProvidesJavaAll(LegacyItemAccess):
     """
 
     def __init__(self, package_provides_java):
+        deprecated(
```
bfahr, line 44: Please include information about the deprecation in the docstring so that it will be included in the documentation.

insights/specs/default.py
```diff
@@ -44,29 +44,45 @@ def get_owner(filename):
...
-        pkg = ctx.shell_out("/bin/rpm -qf {0}".format(resolved[0])) if resolved else None
-        if cmd and pkg is not None:
-            results.add("{0} {1}".format(cmd, pkg[0]))
-    return results
+        cmd = p['COMMAND_NAME']
+        if cmd != command:
+            continue
+        which = ctx.shell_out("/usr/bin/which {0}".format(cmd))
+        ret.add(which[0]) if which else None
+    return sorted(ret)
```
bfahr, line 67: @xiangce since a `set` doesn't have an order, should this be `sorted(list(set))`?

insights/specs/default.py
```diff
@@ -335,29 +350,16 @@ def du_dirs_list(broker):
...
-                cmd = p_splits[10].split()[0]
-                # Should compatible with RHEL6
-                # e.g. /usr/sbin/httpd, /usr/sbin/httpd.worker and /usr/sbin/httpd.event
-                #      and SCL's httpd24-httpd
-                if os.path.basename(cmd).startswith('httpd'):
-                    ps_httpds.add(cmd)
-        # Running multiple httpd instances on RHEL is supported
-        # https://access.redhat.com/solutions/21680
-        return list(ps_httpds)
+        return _get_running_commands(broker, 'httpd')
```
bfahr, line 362: Another reason to use `list(set())` in `_get_running_command` since this originally returned a `list`.

insights/specs/default.py
```diff
@@ -544,6 +546,17 @@ def md5chk_file_list(broker):
...
+
+    @datasource(PsAuxcww, context=HostContext)
+    def java_cmd_and_pkg(broker):
+        """Command: echo the command and package string"""
+        pkg_cmd = list()
+        for cmd in _get_running_commands(broker, 'java'):
+            pkg_cmd.append("{0} {1}".format(cmd, _get_package(broker, cmd)))
+        return pkg_cmd
+
+    package_provides_java = foreach_execute(java_cmd_and_pkg, "/usr/bin/echo %s")
```
bfahr, line 558: Since you are just echoing the data, would it make more sense to use create a `DatasourceProvider` object here and return that? Or do you actually want a separate datasource for each one?

insights/parsers/package_provides.py
```diff
@@ -0,0 +1,84 @@
...
+        self.command = l[0]
+        self.package = l[1]
+
+
+@parser(Specs.package_provides_java)
+class PackageProvidesJava(PackageProvidesCMD):
+    """
+    Parse the output of command::
+
+        for jp in `/bin/ps auxcww | grep <cmd> | grep -v grep | awk '{print $11}' | sort -u`; do echo "$jp `readlink -e $jp | xargs rpm -qf`"; done
```
bfahr, line 52: Its not really clear here that this command is not the actual command you are running.  I think it would be useful to explain that the is the effective command and it is implemented as multiple datasources and functions.

bfahr (changes requested):

insights/specs/default.py
bfahr: On second thought, I think the way you implemented this spec is OK.  DatasourceProvider in the datasource is not necessary.

insights/specs/default.py
xiangce: @bfahr - here it returns a `sorted(set)` but not the original `set`. I think the `sorted(set)` is equivalent with `sorted(list(set))`, both of them returned a new sorted list:
- https://docs.python.org/3/library/functions.html#sorted

insights/combiners/package_provides_httpd.py
xiangce: I will add the corresponding `warning` to the docstring in the next commit

insights/specs/default.py
xiangce: @bfahr - Ack, Thanks

insights/parsers/package_provides.py
xiangce: Done

insights/combiners/package_provides_java.py
xiangce: Done

xiangce:
@bfahr - I modified the spec from a `foreach_collect` to a `command_with_args` which outputs all the running commands by a single `echo` command. And added the `Ps` to the default dependencies to ensure it could get the command from ps during collecting. Please review again. Thanks.

insights/specs/__init__.py
```diff
@@ -461,8 +461,8 @@ class Specs(SpecSet):
     ovs_vsctl_show = RegistryPoint()
     ovs_vswitchd_limits = RegistryPoint()
     pacemaker_log = RegistryPoint(filterable=True)
-    package_provides_java = RegistryPoint(multi_output=True)
-    package_provides_httpd = RegistryPoint(multi_output=True)
+    package_provides_java = RegistryPoint()
+    package_provides_httpd = RegistryPoint()
```
bfahr, line 465: @xiangce this change will break the existing code, I think you'll need to leave it as `multi_output`.

bfahr (changes requested):
Other than my current comment it looks good.

insights/specs/__init__.py
xiangce: @bfahr - yep, that was true. I reverted them. And added a new spec `package_provides_command` which works for any new commands. Please check it again. hanks.

bfahr (approved):

PR #2182, opened 2019-09-06
[Parser Enhancement] YumLog Parser is now filterable

crackcodecamp:
Github Issue: https://github.com/RedHatInsights/insights-core/issues/2180

Signed-off-by: Akshay Ghodake <aghodake@redhat.com>

Issue #2180, opened 2019-09-06
YumLog Praser is not filterable.


crackcodecamp:
Currently, YumLog parser inherited from Parser class. As it parses `/var/log/yum.log` file it should be inherited from LogFileOutput class. Also, this parser should be filtered `i.e. filterable=True`.

bfahr:
@crackcodecamp regarding filtering this log file, how large could it be?  How important is it to know the sequence of events in this log?

PR #2842, opened 2020-12-01
[YumLog] Add filter

psachin:


xiangce:
Relates to #2182 #2180.

@psachin - Please point out the problem you are trying to resolve/fix.

psachin:
@xiangce The changes are needed for one of the plugin.

xiangce:
> @xiangce The changes are needed for one of the plugin.

Does not the current parser w/o `filterable=Ture` work for the plugin? I believe that there are reasons for this parser is designed filterable originally, although it's hard to trace.
However, I think it's OK to make it filterable for bug-fixing or removing sensitive data, please provide the evidence for that since the change changes the original API.

psachin:
Hey @xiangce  Sorry to miss your reply. With the recent change in plugin logic, this change is not required any more.

> Does not the current parser w/o filterable=Ture work for the plugin? I believe that there are reasons for this parser is designed filterable originally, although it's hard to trace.

Same here. I was not able to find any hints to why this parser is not having `filterable=True`

> However, I think it's OK to make it filterable for bug-fixing or removing sensitive data, please provide the evidence for that since the change changes the original API.

By making it filterable, the plugin can leverage attributes such as `keep_scan`, `token_scan` etc. 

Let us close this for now, we can re-open for review if there's a future requirement.

PR #2843, opened 2020-12-01
Add spec lvm_conf in sos_archive

JoySnow:


xiangce:
@JoySnow - For `simple_file`, it's enough to add it to the `default.py` file.

JoySnow:
@xiangce ACK, thanks. Close the MR here.

PR #2846, opened 2020-12-03
Add parser MDAdmMetadata

shlao:
Updated
```
insights/specs/__init__.py
insights/specs/default.py
```
Added new files
```
docs/shared_parsers_catalog/mdadm.rst
insights/parsers/mdadm.py
insights/parsers/tests/test_mdadm.py
```


insights/specs/default.py
```diff
@@ -436,6 +436,7 @@ def md5chk_file_list(broker):
         return ["/etc/pki/product/69.pem", "/etc/pki/product-default/69.pem", "/usr/lib/libsoftokn3.so", "/usr/lib64/libsoftokn3.so", "/usr/lib/libfreeblpriv3.so", "/usr/lib64/libfreeblpriv3.so"]
     md5chk_files = foreach_execute(md5chk_file_list, "/usr/bin/md5sum %s")
     mdstat = simple_file("/proc/mdstat")
+    mdadm_E = foreach_execute(mdstat, "/usr/sbin/mdadm -E %s")
```
xiangce, line 439: The output of the `cat /proc/mdstat` looks like:
```
Personalities : [raid1] [raid6] [raid5] [raid4]
md1 : active raid1 sdb2[1] sda2[0]
      136448 blocks [2/2] [UU]

md2 : active raid1 sdb3[1] sda3[0]
      129596288 blocks [2/2] [UU]

md3 : active raid5 sdl1[9] sdk1[8] sdj1[7] sdi1[6] sdh1[5] sdg1[4] sdf1[3] sde1[2] sdd1[1] sdc1[0]
      1318680576 blocks level 5, 1024k chunk, algorithm 2 [10/10] [UUUUUUUUUU]

md0 : active raid1 sdb1[1] sda1[0]
      16787776 blocks [2/2] [UU]

unused devices: <none> 
```
I don't think the `foreach_execute` would work for this `mdadm_E` spec.

You need to leverage the result of the [MdStat](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/mdstat.py#L60) parser to implement a suitable datasource to return the required list for the `mdadm` command, but not use the `mdstat` spec directly.

insights/parsers/mdadm.py
```diff
@@ -0,0 +1,74 @@
+"""
+MDAdm - command ``/usr/sbin/mdadm -E {device}``
+=================================================
```
xiangce, line 3: Please modify the length of the `===` to the same length of the title.

insights/parsers/mdadm.py
```diff
@@ -0,0 +1,74 @@
...
+@parser(Specs.mdadm_E)
+class MDAdmMetadata(CommandParser, dict):
+    """
+    Parser for output of ``mdadm -E`` for each MD device in system.
+
+    This stores the information from the output in the following properties:
+    * ``device`` - the name of the device after /dev/ - e.g. loop0
+
+    Sample output::
+    /dev/loop0:
```
xiangce, line 23: 1. One more blank line is required for this `Sample output::` section.
2. One more indent block is required for the below content of the `Sample output::` section.

insights/parsers/mdadm.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+
+    def __init__(self, context):
+        # E.g context.path: insights_commands/mdadm_-E_.dev.loop0
+        mdadm_dev = "insights_commands/mdadm_-E_.dev."
+
+        if mdadm_dev in context.path:
+            self.device = '/dev/' + context.path.split(mdadm_dev)[1].strip()
+        else:
+            raise ParseException('Cannot parse device name from path {p}'.format(p=context.path))
```
xiangce, line 66: Is the `ParseException` suitable here? If the `devices` got from the `/proc/mstat` is not the target for this parser, I think `SkipException` is better.

insights/parsers/mdadm.py
```diff
@@ -0,0 +1,74 @@
...
+        >>> mdadm["Events"]
+        60
+
+    """
+
+    def __init__(self, context):
+        # E.g context.path: insights_commands/mdadm_-E_.dev.loop0
+        mdadm_dev = "insights_commands/mdadm_-E_.dev."
+
+        if mdadm_dev in context.path:
```
xiangce, line 63: This check is not so friendly to other archives, e.g. sos, since it makes this parser only works with `insights-archive`.

And this check can be moved to the `parse_content` as well, since the `self.file_path` is passed there.

xiangce (changes requested):
1. Please also add this spec to the `spec/insights_archive.py` file.
2. AFAIK, there is a similar command in the sos report archive, ```sos_commands/md/mdadm_-D_.dev.md```. To make this parser can be used more widely, better make it compatible with the existing specs in sos archive.

And here is the difference between the `-D` and `-E`:

```
-D, --detail
Print details of one or more md devices.

-E, --examine
Print contents of the metadata stored on the named device(s). Note the contrast between --examine and --detail. --examine applies to devices which are components of an array, while --detail applies to a whole array which is currently active.
```

If the `-D` works for you, please change to use `-D` in the spec.

xiangce (approved):

PR #2855, opened 2020-12-09
Enhance the vendor of InstalledRpm

xiangce:
- The current `InstalledRpm.vendor` treats the `(none)` and `no vendor outputted` as the same
  Updated as:
   - set the value as `'(none)'` string, when the vendor is `(none)`
   - set the value as `None` (NoneType), when there is no `vendor` column/item.


xiangce:
@chenlizhong @wushiqinlou - please help review. Thanks.

chenlizhong (approved):

wushiqinlou:
@xiangce - Good to me.

PR #2865, opened 2020-12-11
fix(ultralight checkins): wrap thrown exception with catch

gravitypriest:


Glutexo:
I randomly came across this. I guess it was a hot fix, but still it apparently went through without a review. Also, as I coded in the check-in feature, I’d like to know what kind of exceptions were occurring, causing the need for a try-except wrap.

insights/client/phase/v1.py
```diff
@@ -112,7 +112,12 @@ def pre_update(client, config):
         sys.exit(constants.sig_kill_ok)
 
     if config.checkin:
-        checkin_success = client.checkin()
+        try:
+            checkin_success = client.checkin()
+        except Exception as e:
```
Glutexo, line 117: Exceptions are not Pokémon, you don’t have to catch ’em all. What happened there in the _checkin_ method that caused the need for try-except? Wouldn’t catching a specific type be better? Why isn’t it enough that the exception is caught by the phase, printing “Fatal error”?

gravitypriest: This PR is a quick fix intended to be consistent with the other case in which this error occurs. This whole logic set needs to be reworked anyway -- I'll attend to it.

PR #2884, opened 2021-01-08
Doc: Fix data type for spec_factory cmds

nrwahl2:
`cmd` members should be strings, not lists of lists.

Signed-off-by: Reid Wahl <nrwahl@protonmail.com>


nrwahl2:
I've skimmed over the rest of the `spec_factory.py` file, and it looks like other `cmd` arguments are also strings. Maybe they get split into a list of lists later, but the passed argument is a string.

bfahr (approved):

PR #2891, opened 2021-01-12
Ensure that args to get_running_commands is a list

bfahr:
* Commands argument could be a type other than a list which caused
  archive collection to fail
* Identified in bugzilla 1915219
* Added tests for exceptions

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -46,7 +46,7 @@ def get_owner(filename):
     return (name, group)
 
 
-def _get_running_commands(broker, commands=None):
+def _get_running_commands(broker, commands):
     """
     Search for command in ``ps auxcww`` output and determine RPM providing binary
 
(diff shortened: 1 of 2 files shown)


insights/specs/default.py
```diff
@@ -56,8 +56,13 @@ def _get_running_commands(broker, commands=None):
 
     Returns:
         list: List of the full command paths of the ``command``.
+
+    Raises:
+        Exception: Raises an exception if commands object is not a list or is empty
     """
-    commands = [] if commands is None else commands
+    if not commands or not isinstance(commands, list):
+        raise Exception('Commands argument must be a list object and contain at least on item')
```
csams, line 64: typo nitpick: at least s/on/one item

csams (approved):
Looks good to me!

PR #2897, opened 2021-01-15
Change etc_udev_40_redhat_rules spec entry

akshay196:
This includes:
1. Change spec name `etc_udev_40_redhat_rules` to `udev_40_redhat_rules`. Also change `udev_40_redhat_rules` to read "40-redhat.rules" from multiple places. (Check "RULES FILES" section in `man udev` output)
2. Deprecate `etc_udev_rules.UdevRules40Redhat` parser.
3. Move `etc_udev_rules.UdevRules40Redhat` parser to new module `udev_rules.UdevRules40Redhat` as it is more appropriate location. All udev rules related parser can be placed in `udev_rules` module only.


insights/specs/__init__.py
```diff
@@ -153,7 +153,7 @@ class Specs(SpecSet):
     etc_journald_conf_d = RegistryPoint(multi_output=True)
     etc_journald_conf = RegistryPoint()
     etc_machine_id = RegistryPoint()
-    etc_udev_40_redhat_rules = RegistryPoint(filterable=True)
+    udev_40_redhat_rules = RegistryPoint(filterable=True)
```
bfahr, line 156: @akshay196 this breaks backward compatibility for any applications/rules that use this spec.  It would be better not to change this name.

insights/specs/default.py
```diff
@@ -298,7 +298,8 @@ def du_dirs_list(broker):
     etc_journald_conf = simple_file(r"etc/systemd/journald.conf")
     etc_journald_conf_d = glob_file(r"etc/systemd/journald.conf.d/*.conf")
     etc_machine_id = simple_file("/etc/machine-id")
-    etc_udev_40_redhat_rules = simple_file("/etc/udev/rules.d/40-redhat.rules")
+    udev_40_redhat_rules = first_file(["/etc/udev/rules.d/40-redhat.rules", "/run/udev/rules.d/40-redhat.rules",
+                                       "/usr/lib/udev/rules.d/40-redhat.rules", "/usr/local/lib/udev/rules.d/40-redhat.rules"])
```
bfahr, line 302: Just update the existing spec name with the `first_file` and new paths.

bfahr (changes requested):

insights/specs/__init__.py
akshay196: @bfahr The spec name `etc_udev_40_redhat_rules` might confuse users if we add new paths.

How about we keep the both the specs and later deprecate `etc_udev_40_redhat_rules` similar to the parser deprecation? Is it possible?

akshay196:
@bfahr Reverted spec name to its original. Please take a look.

PR #2905, opened 2021-01-21
Add parser for spec "ls -lan /var/cache/pulp"

huali027:


xiangce (approved):
Approved pending the approval of spec.

PR #2907, opened 2021-01-25
set_enabled works on components and their fully qualified names.

### LINKED ISSUE #2903: set_enabled should accept components in addition to strings.
The pydoc for [set_enabled](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/dr.py#L89) says it accepts the fqdn of an object or the object itself, but it raises an exception (logged at DEBUG) if you pass the object.

csams:
Only call `get_component` if the argument is a string.

Fixes #2903


patchkez:
test this please

PR #2910, opened 2021-01-27
Add spec to get the block size of gfs2 file system

huali027:

### DIFF
--- a/insights/parsers/tests/test_gfs2_file_system_block_size.py
+++ b/insights/parsers/tests/test_gfs2_file_system_block_size.py
@@ -1,0 +1,26 @@
+from insights.parsers import gfs2_file_system_block_size
+from insights.tests import context_wrap
+from insights.parsers import SkipException
+import pytest
+import doctest
+
+BLOCK_SIZE_OUTPUT = """
+4096
+""".strip()
+
+BLOCK_SIZE_OUTPUT_2 = """
+abc
+""".strip()
+
+
+def test_exp():
+    with pytest.raises(SkipException):
+        gfs2_file_system_block_size.GFS2FileSystemBlockSize(context_wrap(BLOCK_SIZE_OUTPUT_2))
+
+
+def test_doc():
+    env = {
+            'gfs2_mp': gfs2_file_system_block_size.GFS2FileSystemBlockSize(context_wrap(BLOCK_SIZE_OUTPUT))
+          }
+    failed, total = doctest.testmod(gfs2_file_system_block_size, globs=env)
+    assert failed == 0
--- a/insights/parsers/tests/test_gfs2_file_system_block_size.py
+++ b/insights/parsers/tests/test_gfs2_file_system_block_size.py
@@ -12,10 +12,28 @@ BLOCK_SIZE_OUTPUT_2 = """
 abc
 """.strip()
 
+BLOCK_SIZE_OUTPUT_3 = """
+512
+""".strip()
+
+BLOCK_SIZE_OUTPUT_4 = """
+4096
+512
+""".strip()
+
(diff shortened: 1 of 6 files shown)


insights/parsers/tests/test_gfs2_file_system_block_size.py
```diff
@@ -0,0 +1,26 @@
...
+BLOCK_SIZE_OUTPUT_2 = """
+abc
+""".strip()
+
+
+def test_exp():
+    with pytest.raises(SkipException):
+        gfs2_file_system_block_size.GFS2FileSystemBlockSize(context_wrap(BLOCK_SIZE_OUTPUT_2))
+
+
```
bfahr, line 20: @huali027 you should add some additional testing for the parser here, and not rely solely on the documentation test and negative test.

bfahr (changes requested):

insights/parsers/tests/test_gfs2_file_system_block_size.py
huali027: @bfahr, I thought it was enough to test right value and bad value. Since the output is every simple. Let me check to add more test, thanks for the review.

insights/specs/default.py
```diff
@@ -383,6 +385,30 @@ def du_dirs_list(broker):
...
+        systems.
+        And only run the ``stat`` command on RHEL version that's less than
+        8.3. With 8.3 and later, the command ``blkid`` will also output the
+        block size info.
+
+        Returns:
+            list: a list of mount points of which the file system type is gfs2
+        """
+        gfs2_mount_points = []
+        if (broker[RedHatRelease] and
```
xiangce, line 402: The Components IsRhel* is just enhanced and ".minor" is added, would you please change to use the IsRhel* here?

huali027: @xiangce, sure.

insights/specs/default.py
```diff
@@ -382,6 +383,29 @@ def du_dirs_list(broker):
...
+        gfs2_mount_points = []
+        if (broker.get(IsRhel6) or broker.get(IsRhel7) or
+                (broker.get(IsRhel8) and broker[IsRhel8].minor < 3)):
+            for mnt in broker[Mount]:
+                if mnt.mount_type == "gfs2":
+                    gfs2_mount_points.append(mnt.mount_point)
+        if gfs2_mount_points:
+            return gfs2_mount_points
+        raise SkipComponent
+    gfs2_file_system_block_size = foreach_execute(gfs2_mount_points, "/usr/bin/stat -fc %%s %s")
```
bfahr, line 408: @huali027 you also need to add a spec to `insights_archive.py` to enable access to the data in the archive.  I believe this will be the glob  `insights_commands\stat_-fc_s*`.

bfahr (changes requested):

insights/specs/default.py
huali027: @bfahr, since it is run by "foreach_execute", all the results are saved to "gfs2_file_system_block_size" and I tested it too, it works well. Do we still need to use glob to add a spec in insights_archive.py explicitly?

xiangce: @bfahr - this spec depends on a @datasource which is not supported by the old collection, which means it cannot be collected by the old collection. And the new collection will set up the relationship by generating a JSON file to the `meta_data` dir for all specs. Is it necessary to add it to the `insights_archive.py`?

bfahr: @huali027 @xiangce you are correct.  I had tested it incorrectly, and when I reran some additional tests I see that it works correctly as it.

bfahr (approved):

PR #2911, opened 2021-01-27
Add HostContext to all custom datasources

### LINKED ISSUE #2908: Datasources should provide HostContext as an arg, not a kwarg
All current datasources should only collect data in the `HostContext` so that should be added to all datasources to avoid exceptions being raised in other contexts.  Also `HostContext` should be used only as an `arg` and not a `kwarg` in order to function properly.

Do this: `@datasource(Ps, HostContext)`

Do not do this: `@datasource(Ps, context=HostContext)`

bfahr:
* Since these should only run when performing host collection, they need
  to include the proper context
* HostContext should be used as an arg and not a kwarg
* Fixes #2908


lhuett (approved):
This looks good.

PR #2917, opened 2021-01-28
Enhanced parser insights/parsers/ip.py

crackcodecamp:
Parser updated to get the value of `promiscuity` item.

### DIFF
--- a/insights/parsers/ip.py
+++ b/insights/parsers/ip.py
@@ -119,10 +119,9 @@ def parse_link(line, d):
         d["peer"] = split_content[3]
     elif len(split_content) >= 2:
         d["mac"] = split_content[1]
-        for item in split_content:
-            if item == "promiscuity":
-                index = split_content.index(item)
-                d["promiscuity_mode"] = split_content[index + 1]
+        if "promiscuity" in split_content:
+            d["promiscuity"] = split_content[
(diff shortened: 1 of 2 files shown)


crackcodecamp:
@xiangce Please review.

xiangce:
@crackcodecamp, please add a test for this newly added item.

crackcodecamp:
@xiangce Added test code for the newly added item, please check.

insights/parsers/ip.py
```diff
@@ -119,6 +119,10 @@ def parse_link(line, d):
         d["peer"] = split_content[3]
     elif len(split_content) >= 2:
         d["mac"] = split_content[1]
+        for item in split_content:
+            if item == "promiscuity":
+                index = split_content.index(item)
+                d["promiscuity_mode"] = split_content[index + 1]
```
xiangce, line 125: How about 
```suggestion
        if "promiscuity" in split_content":
            d["promiscuity"] = split_content[split_content.index('promiscuity') + 1]
```

I suggest modifying the name of the key to `promiscuity` as it is in the line, in case, this parser would be enhanced by other utility helpers automatically.

crackcodecamp: Updated according to suggestion.

xiangce (approved):

Issue #2812, opened 2020-11-03
Sap HDB version parser cannot handle multiple inputs


bfahr:
The current spec sap_hdb_version is multi_output, but the implementation in the JSON concatenates all output in one file.  The HDBVersion parser expects one set of command output per file so it will not work correctly if there is more than one SAP SID.  The parser does work correctly for core collection which does the right thing regarding command output.

bfahr:
@xiangce @SteveHNH FYI.

xiangce:
@bfahr, Since the "HDB version" command works for its own SAP HANA instance only, it's incorrect to let the HDBVersion parser parse all the output if there are multiple SAP HANA instances (it's rare, but I'm not sure if it's possible) running.

In other words, I don't think it's correct theoretically to use the "insights-cat" to collect this spec, since it impossible for the rule to check which HDBVersion stands for which SAP HANA, which is done via checking the "file_path"

HTH,

PR #2919, opened 2021-02-01
Change to use all_instance for HDB Version

xiangce:
- It's necessary to get all the instances instead of the local instances
  Since It's possible for SAP HANA instances to use alias hostname

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -733,7 +733,7 @@ class DefaultSpecs(Specs):
     @datasource(Sap, HostContext)
     def sap_sid(broker):
         sap = broker[Sap]
-        return [sap.sid(i).lower() for i in sap.local_instances]
+        return [sap.sid(i).lower() for i in sap.all_instances]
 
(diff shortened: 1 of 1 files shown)


bfahr:
@xiangce is this the fix for #2812 ?

insights/specs/default.py
```diff
@@ -726,7 +726,7 @@ def pcp_enabled(broker):
     @datasource(Sap, HostContext)
     def sap_sid(broker):
         sap = broker[Sap]
```
bfahr, line 728: Please add a docstring to this datasource.

bfahr (changes requested):

insights/specs/default.py
xiangce: Done

xiangce:
> @xiangce is this the fix for #2812 ? Is this related to RHCLOUD-10505 issue regarding collection of the instances?

@bfahr - I think the #2922 is for the #2812. This one is just to extend the `HDB version` to all the SAP instances instead of the instances that has the same hostname with the host.

bfahr (approved):

PR #2921, opened 2021-02-01
Check file existence state before opening it  in cloud_cfg datasource

### LINKED ISSUE #2920: Unexpected Exception is thrown while running insights-client
```
# insights-client 
Starting to collect Insights data for slutade-rhel7.syslab.pnq2.redhat.com
  File "/var/lib/insights/last_stable.egg/insights/core/dr.py", line 962, in run
    result = DELEGATES[component].process(broker)
  File "/var/lib/insights/last_stable.egg/insights/core/dr.py", line 681, in process
    return self.invoke(broker)
  File "/var/lib/insights/last_stable.egg/insights/core/plugins.py", line 90, in invoke
    return self.component(broker)
  File "/var/lib/insights/last_stable.egg/insights/specs/default.py", line 282, in cloud_cfg
    with open(relative_path, 'r') as f:
IOError: [Errno 2] No such file or directory: '/etc/cloud/cloud.cfg'

```

The current `cloud_cfg` datasource caught the OSError instead of IOError.
https://github.com/RedHatInsights/insights-core/blob/1b2d8055dae1f6608901a1cc30f8ac978af71063/insights/specs/default.py#L288

xiangce:

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -13,7 +13,7 @@ import re
 import json
 
 from grp import getgrgid
-from os import stat, path, listdir as os_listdir
+from os import stat, listdir as os_listdir
 from pwd import getpwuid
 
 import yaml
@@ -226,8 +226,9 @@ class DefaultSpecs(Specs):
     cinder_api_log = first_file(["/var/log/containers/cinder/cinder-api.log", "/var/log/cinder/cinder-api.log"])
     cinder_conf = first_file(["/var/lib/config-data/puppet-generated/cinder/etc/cinder/cinder.conf", "/etc/cinder/cinder.conf"])
     cinder_volume_log = first_file(["/var/log/containers/cinder/volume.log", "/var/log/containers/cinder/cinder-volume.log", "/var/log/cinder/volume.log"])
+    cloud_cfg_input = simple_file("/etc/cloud/cloud.cfg")
 
-    @datasource(HostContext)
+    @datasource(cloud_cfg_input, HostContext)
     def cloud_cfg(broker):
         """This datasource provides the network configuration collected
         from ``/etc/cloud/cloud.cfg``.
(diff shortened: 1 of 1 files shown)


psachin (approved):
Tested locally by building lasted client egg.

insights/specs/default.py
```diff
@@ -285,7 +285,7 @@ def cloud_cfg(broker):
                 if network_config:
                     return DatasourceProvider(content=json.dumps(network_config), relative_path=relative_path)
 
-        except OSError:
+        except IOError:
```
bfahr, line 288: @xiangce I believe this should be a generic `Exception` instead because we don't want this to raise any exceptions other than `SkipComponent` during collection.  

Also see my concerns about the original parser noted in the Issue #2920.  That should probably be done in another PR since any rules using this parser would need to change, but I'd prefer not to close the issue until that is fixed.

bfahr (changes requested):

insights/specs/default.py
xiangce: @bfahr - I add a safety check before opening the file, please check it again.

And I'm very sorry, I cannot get your concerns, would you please make it a bit clear?

xiangce: I just removed the `fix #2920`, which will keep it open after merging this PR.

xiangce: Just updated the #2920, and It seems that more customers are complaining about such problems.

bfahr: @xiangce not sure which point you are asking for more detail on.  

This should be a bare `Exception` to catch whatever problems may occur with reading the file, not just an `IOError`.

I opened a new Issue #2935 with more detail about the context the datasource is not capturing.

bfahr: @xiangce I discussed with @csams and we came up with the changes in my commit.  Also had to fix some merge conflicts to get this merged this morning.

lhuett (approved):
Looks good.

PR #2922, opened 2021-02-02
Remove the sap_hdb_version from insights_archive.py

### LINKED ISSUE #2812: Sap HDB version parser cannot handle multiple inputs

xiangce:
- fix #2812 


xiangce:
After this patch, the spec of the `HDB version` will be stored individually in the archive, then it will fix the #2812.

bfahr (approved):

PR #2929, opened 2021-02-03
Add new spec to get the satellite admin settings

huali027:


insights/parsers/satellite_settings.py
```diff
@@ -0,0 +1,77 @@
...
+
+from insights import parser, CommandParser
+from insights.specs import Specs
+from insights.parsers import SkipException, ParseException
+
+
+@parser(Specs.satellite_settings)
+class SatelliteSettings(CommandParser, dict):
+    """
+    Read the ``psql -d foreman -c '"select name, value, "defualt" from settings where name in ('destroy_vm_on_host_delete', 'unregister_delete_host')"``
```
bfahr, line 44: @huali027 I believe the word `defualt` is a misspelling.

insights/parsers/satellite_settings.py
```diff
@@ -0,0 +1,77 @@
...
+                            val = val.rstrip(' +') + os.linesep + new_val
+                        if default_val.endswith('+'):
+                            default_val = default_val.rstrip(' +') + os.linesep + new_default_val
+                    if not (val.endswith('+') or default_val.endswith('+')):
+                        real_val = val or default_val
+                        try:
+                            self[name] = yaml.safe_load(real_val)
+                        except Exception:
+                            raise ParseException('Could not parse the value for satellite settings of name %s ' % name)
+                        name = val = new_val = ''
```
bfahr, line 75: @huali027 this seems more complicated that it needs to be.  You know what columns are being selected and the order, and you know the row values you want to collect so why not just look for the rows with the 2 keys `[destroy_vm_on_host_delete', 'unregister_delete_host']` and then check for the 'true` and `false` values.  It just seems like it could be simplified given that you know what data is being collected.  What do you think?

bfahr (changes requested):

insights/parsers/satellite_settings.py
huali027: @bfahr, yes, I'll update it, thanks.

insights/parsers/satellite_settings.py
huali027: @bfahr, let me explain more:
1. first I check the columns are in the same order with what expected. Because if customers didn't change the setting, the default column should be used, if changes are done, then the value column should be used.
2. currently,  I only need these two settings, so I limit the SQL to it. If in the parser code, I hard code these two column, and consider the value is just boolean and later if any new setting is needed, we have to update the parser too. For these two settings, the value is just a boolean value, but for other settings, the value may be string or int or other types. But all of them are in the yaml format. 
For example:
```
               name                | value |                              default                              
-----------------------------------+-------+-------------------------------------------------------------------
 foreman_tasks_proxy_batch_size    |       | --- 100                                                          +
                                   |       | ...
 foreman_tasks_troubleshooting_url |       | --- https://access.redhat.com/solutions/satellite6-tasks#%{label}+
                                   |       | ...
(2 rows)

```
Do u think we should only consider the two settings now without considering more?

insights/parsers/satellite_settings.py
```diff
@@ -0,0 +1,77 @@
...
+import os
+import yaml
+
+from insights import parser, CommandParser
+from insights.specs import Specs
+from insights.parsers import SkipException, ParseException
+
+
+@parser(Specs.satellite_settings)
+class SatelliteSettings(CommandParser, dict):
```
xiangce, line 42: `list` would be better to make the parser more compatible with the SQL passed to the spec.

Just as mentioned this [comment](https://github.com/RedHatInsights/insights-core/pull/2934#issuecomment-773917878). It's possible to add more columns in the future for other requirements.

insights/parsers/satellite_settings.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    Read the ``psql -d foreman -c '"select name, value, "defualt" from settings where name in ('destroy_vm_on_host_delete', 'unregister_delete_host')"``
+    and save the data to itself as a dict.
+    """
+
+    def parse_content(self, content):
+        columns = ['name', 'value', 'default']
```
xiangce, line 49: It's not so good to hard-code the columns here in the parser, it will make the parser very limited.

xiangce (changes requested):
@huali027 - I just made a prototype that converts the rows to a `list` instead of `dict`, please have a look at the following code slice:

```python
          columns = [item.strip() for item in content[0].split('|')]
          if columns[0] != 'name':
              raise ParseException("The first column must be 'name'")
          columns_len = len(columns)

          for row in content[2:-1]:  # skip head, '---+---', and tail
              row_sp = [item.strip() for item in row.split('|')]
              if len(row_sp) != columns_len:
                  raise ParseException("Invalid line: '{0}'".format(row))
              if row_sp[0]:
                  data = {}
                  self.append(data)
                  data['name'] = row_sp[0]
                  for i in range(1, columns_len:
                      if not row_sp[i]:
                          continue
                      if row_sp[i].startswith('---') and row_sp[i].endswith('+'):
                          data[columns[i]] = [row_sp[i].rstrip('+')]
                      else:
                          data[columns[i]] = row_sp[i].strip('- ')
              else:
                  for i in range(1, columns_len):
                      if not row_sp[i]:
                          continue
                      if isinstance(data[columns[i]], list):
                          data[columns[i]].append(row_sp[i])
                          if row_sp[i] == '...':
                              val = linesep.join(data[columns[i]])
                              try:
                                  data[columns[i]] = yaml.safe_load(val)
                              except Exception:
                                  raise ParseException("Bad format value: '{0}'".format(val))
                      elif row_sp[i]:
                          raise ParseException("Unexpected '{0}' for '{1}'".format(row_sp[i], columns[i]))

          if not len(self):
              raise SkipException("Cannot get the satellite settings")

      def get_value(self, key):
          for row in self:
              if row['name'] == key:
                  return row.get('value', row.get('default'))
```

insights/parsers/satellite_settings.py
huali027: @xiangce, sure, thanks. I'll update the parser to make it more extendable based on your suggested code.

insights/parsers/satellite_settings.py
bfahr: There is always a trade-off between doing what you need today versus trying to anticipate future possibilities.  If you think its likely that you'll add more columns in the future, and you can write a parser that implements those future data types, then it makes sense to have more complicated parsing.  If you go with the more complex parser please make sure you include sufficient testing to ensure it fails if something is added you don't anticipate.

huali027: @bfahr, sure, I'll add more tests too.

bfahr (changes requested):
My concern here is that a user, reading the existing docstrings, would not understand how to use this parser, and that the form the data is presented (list) doesn't represent what might be expected by a user.  Let me know if I missed something here or misunderstood what you're trying to accomplish.

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,177 @@
...
+                            data[self._columns[i]].append(row_data[i])
+        if not len(self):
+            raise SkipException("There is no data")
+        if self.count != len(self):
+            raise ParseException('The count of rows is not equal to the number of the last row')
+
+    def __contains__(self, first_column_value):
+        return True if any(
+            row[self._first_columm] == first_column_value for row in self
+        ) else False
```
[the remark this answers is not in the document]
huali027: @bfahr, you are right. I updated it

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def get_column_value(self, first_column_value, column_name):
```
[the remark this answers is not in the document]
huali027: @bfahr, to make it more readable, I redefine it with "keyword_search". In this way, the user can define the search conditions by himself to find one row. Since the most usage is to get the column of one row. So It is limited to one row only.

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,177 @@
...
+    def parse_content(self, content):
+        super(SatelliteAdminSettings, self).parse_content(content)
+        for row in self:
+            if 'value' in row:
+                row['value'] = self._parse_yaml(row['value'])
+
+            if 'default' in row:
+                row['default'] = self._parse_yaml(row['default'])
+
+    def get_settting(self, setting_name):
```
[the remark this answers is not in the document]
huali027: @bfahr, oh, yes, typo error, I updated it.

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Examples::
+
+        >>> type(settings)
+        <class 'insights.parsers.satellite_postgresql_query.SatelliteAdminSettings'>
+        >>> settings.get_settting('unregister_delete_host')
+        True
+        >>> len(settings)
+        2
+        >>> settings.get_settting('destroy_vm_on_host_delete')
```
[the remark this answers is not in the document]
huali027: @bfahr, in the base class, I'm not sure if the value of one column is unique for all the tables, so I save the rows to a list.
  Although in the settings table, the name is actually unique,  in the rule I just want to get the value of some setting, I don't override  the format by looping it again after extending the base class. Do u think we have to override it?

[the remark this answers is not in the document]
bfahr: @huali027 there's 3 issues with `get_settings`.  First are there only 2 settings that a user would pass to this method, those in the documentation? If not then how would a user get all of the setting names to know what to pass to the method?

Second, you looping every time this method is called.  You could create a `dict` with all valid settings names and then just do a lookup in the dict each time the method is called.

Third, you are checking each time to determine if both `valid` and `default` columns are in the row and then raising an `ParseException` if they are not.  This should only be checked in `parse_content`, and not in the `get_settings` method.  So if you created the `dict` in `parse_content` of all settings then you could verify at that time whether or not the data is valid.  You don't want to have rules checking for ParseExceptions.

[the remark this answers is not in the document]
huali027: @bfahr,
For the first issue, yes, since my rule only uses these two settings, I limit the SQL to only select these two settings. There are about 240 records in the table. To save time, I add a limitation. Because maybe the other settings are never used.
The rule user should know which settings he wants to use. So we don't need to provide a method to query all the settings.
Sure, I'll update the code to save settings table into a dict format.

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,174 @@
...
+        for row in self:
+            row['default'] = self._parse_yaml(row['default'])
+            row['value'] = self._parse_yaml(row['value'])
+
+    def get_setting(self, setting_name):
+        """
+        Get the actual value of setting_name.
+        If the value column isn't empty, the value of the setting_name is the
+        value column, or else it's the default column.
+        """
```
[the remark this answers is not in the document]
bfahr: Also rather than implicitly returning `None` here, wouldn't it be better to return an empty `list`?

bfahr (changes requested):

insights/parsers/satellite_postgresql_query.py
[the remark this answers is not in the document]
huali027: > @huali027 please include the `Parameters:` and `Returns:` information in the docstring for users of this methods.

Sure, I'll add it, thanks.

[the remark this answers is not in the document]
huali027: > Also rather than implicitly returning `None` here, wouldn't it be better to return an empty `list`?

The return value will be boolean if the setting exists in the table, so I think it's okay to return None if the setting doesn't exist, does it make sense?

huali027:
@bfahr @xiangce, please help to review when u have time, thanks.

insights/parsers/satellite_postgresql_query.py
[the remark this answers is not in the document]
xiangce: In the case of the various return types, better add another method `has_setting` to check if the specified setting in the table. And raise a "KeyError" when there is no such setting in the table for the `get_setting`. Just like what the `[IniConfigFile](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/__init__.py#L1382)` do. How do you think?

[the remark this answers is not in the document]
huali027: @xiangce, good to me, I'll update it.

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,196 @@
...
+            a list.
+
+        Raises::
+
+            KeyError: when setting_name does not exist in the table
+
+        """
+        rows = self.search(name=setting_name)
+        if rows:
+            return rows[0].get('default') if rows[0].get('value') == '' else rows[0].get('value')
```
xiangce, line 195: Nit: Save one getting...
```
value = rows[0].get('value')
return value if value else rows[0].get('default')
```

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,196 @@
...
+        rows = self.search(name=setting_name)
+        return True if rows else False
+
+    def get_setting(self, setting_name):
+        """
+        Get the actual value of setting_name.
+        If the value column isn't empty, the value of the setting_name is the
+        value column, or else it's the default column.
+
+        Args::
```
xiangce, line 179: The two colons `::` is not necessary for these proper sections.

see: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,196 @@
...
+
+            setting_name (str): the value of name column which is searched in the table.
+
+        Returns::
```
xiangce, line 183: The same as above

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,196 @@
...
+
+            It depends on the setting, maybe boolean, string, int,
+            a list.
+
+        Raises::
```
xiangce, line 188: The same as above

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,196 @@
...
+    def search(self, **kwargs):
+        """
+        Get the rows by searching the table with kwargs.
+        This uses the :py:func:`insights.parsers.keyword_search` function for
+        searching; see its documentation for usage details. If no search
+        parameters are given, no rows are returned.
+
+        It simplify the value of the column according to actual usage.
+
+        Returns::
```
xiangce, line 93: The same as above

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,196 @@
...
+
+            list: A list of dictionaries of rows that match the given
+            search criteria.
+
+        Examples::
```
xiangce, line 98: The same as above

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,196 @@
...
+        1
+        >>> rows[0]['value']
+        'test'
+        >>> columns=query.get_columns()
+        >>> 'url' in columns
+        True
+        >>> 'name' in columns
+        True
+
+    Raises::
```
xiangce, line 64: The same as above

insights/parsers/satellite_postgresql_query.py
```diff
@@ -0,0 +1,196 @@
...
+        ]
+
+    Sample Output::
+
+        name,url,value
+        abc,,test
+        def,http://xx.com,
+
+
+    Examples::
```
xiangce, line 49: The same as above

xiangce (changes requested):

huali027:
@xiangce, please help to review when u have time, thanks.

xiangce (approved):
Good to me. @bfahr - please double check.

huali027:
@bfahr, please help to review this parser when u have time, thanks.

PR #2934, opened 2021-02-04
Add spec to get satellite compute resources

huali027:

### DIFF
--- a/insights/parsers/satellite_postgresql_query.py
+++ b/insights/parsers/satellite_postgresql_query.py
@@ -6,6 +6,8 @@ This module contains the following parsers:
 
 SatelliteAdminSettings - command ``psql -d foreman -c 'select name, value, "default" from settings where name in (\'destroy_vm_on_host_delete\', \'unregister_delete_host\') --csv'``
 -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+SatelliteComputeResources - command ``psql -d foreman -c 'select name, type from compute_resources'``
+-----------------------------------------------------------------------------------------------------
 """
 
 import os
@@ -174,3 +176,30 @@ class SatelliteAdminSettings(SatellitePostgreSQLQuery):
         if rows:
             value = rows[0].get('value')
             return rows[0].get('default') if value == '' else value
+
+
+@parser(Specs.satellite_compute_resources)
+class SatelliteComputeResources(SatellitePostgreSQLQuery):
+    """
+    Parse the output of the command ``psql -d foreman -c 'select name, type from compute_resources' --csv``.
+
+    .. note::
+        Please refer to its super-class :class:`insights.parsers.satellite_postgresql_query.SatellitePostgreSQLQuery` for more
+        details.
+
+    Sample output::
+
+        name,type
+        test_compute_resource1,Foreman::Model::Libvirt
+        test_compute_resource2,Foreman::Model::RHV
+
+    Examples::
+        >>> type(resources_table)
+        <class 'insights.parsers.satellite_postgresql_query.SatelliteComputeResources'>
+        >>> rows=resources_table.search(type='Foreman::Model::Libvirt')
+        >>> len(rows)
+        1
+        >>> rows[0]['name']
+        'test_compute_resource1'
+    """
+    pass
--- a/insights/parsers/satellite_postgresql_query.py
+++ b/insights/parsers/satellite_postgresql_query.py
@@ -193,7 +193,7 @@ class SatelliteComputeResources(SatellitePostgreSQLQuery):
         test_compute_resource1,Foreman::Model::Libvirt
         test_compute_resource2,Foreman::Model::RHV
 
-    Examples::
+    Examples:
         >>> type(resources_table)
         <class 'insights.parsers.satellite_postgresql_query.SatelliteComputeResources'>
         >>> rows=resources_table.search(type='Foreman::Model::Libvirt')
--- a/insights/parsers/tests/test_satellite_postgresql_query.py
+++ b/insights/parsers/tests/test_satellite_postgresql_query.py
@@ -128,6 +128,12 @@ destroy_vm_on_host_delete,,"--- true
 ..."
 '''
 
+SATELLITE_COMPUTE_RESOURCE_1 = '''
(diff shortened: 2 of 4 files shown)


insights/parsers/satellite_compute_resources.py
```diff
@@ -0,0 +1,50 @@
+"""
+SatelliteComputerResources - command ``psql -d foreman -c 'select name, type from compute_resources'``
```
gnurag, line 2: `SatelliteComputerResources` -> `SatelliteComputeResources`

The correct term is [Compute Resource](https://theforeman.org/manuals/2.3/index.html#5.2ComputeResources). ComputerResource is mentioned in several places below, so would be a good idea to update there too.

gnurag (commented):
Minor cosmetic suggestion, does not affect functionality so feel free to ignore.

gnurag:
@apuntamb @kgaikwad - for you perusal.

insights/parsers/satellite_compute_resources.py
huali027: @gnurag thanks for the correction, I'll update it.

kgaikwad:
@gnurag , looks good to me :+1: 

@huali027, one query - will it be possible to change/override SQL query i.e `select name, type from compute_resources;` from anywhere except by making change [here](https://github.com/RedHatInsights/insights-core/pull/2934/files#diff-3d5665cd194cc52869c22cc075e78581f83602790b2fb1a695dea1f55576754dR762)?

huali027:
> @gnurag , looks good to me 
> 
> @huali027, one query - will it be possible to change/override SQL query i.e `select name, type from compute_resources;` from anywhere except by making change [here](https://github.com/RedHatInsights/insights-core/pull/2934/files#diff-3d5665cd194cc52869c22cc075e78581f83602790b2fb1a695dea1f55576754dR762)?

@kgaikwad, thanks for the review. We collect it for satellite rules. Since we expect to run it on satellite server only, we can only change/override the SQL query in the insights-core with the help of @datasource.

kgaikwad:
> > @gnurag , looks good to me
> > @huali027, one query - will it be possible to change/override SQL query i.e `select name, type from compute_resources;` from anywhere except by making change [here](https://github.com/RedHatInsights/insights-core/pull/2934/files#diff-3d5665cd194cc52869c22cc075e78581f83602790b2fb1a695dea1f55576754dR762)?
> 
> @kgaikwad, thanks for the review. We collect it for satellite rules. Since we expect to run it on satellite server only, we can only change/override the SQL query in the insights-core with the help of @Datasource.

okies, thanks @huali027
I thought to suggest `-t` option inside psql command if the columns are fixed. Example -
`/usr/bin/sudo -iu postgres /usr/bin/psql -d foreman -tAc "select name, type from compute_resources;"`
which will turn off printing of column names & result row count. But parser class is defined into separate file. 
So it may create problem when someone makes change to query and not to parser.

huali027:
> > > @gnurag , looks good to me
> > > @huali027, one query - will it be possible to change/override SQL query i.e `select name, type from compute_resources;` from anywhere except by making change [here](https://github.com/RedHatInsights/insights-core/pull/2934/files#diff-3d5665cd194cc52869c22cc075e78581f83602790b2fb1a695dea1f55576754dR762)?
> > 
> > 
> > @kgaikwad, thanks for the review. We collect it for satellite rules. Since we expect to run it on satellite server only, we can only change/override the SQL query in the insights-core with the help of @Datasource.
> 
> okies, thanks @huali027
> I thought to suggest `-t` option inside psql command if the columns are fixed. Example -
> `/usr/bin/sudo -iu postgres /usr/bin/psql -d foreman -tAc "select name, type from compute_resources;"`
> which will turn off printing of column names & result row count. But parser class is defined into separate file.
> So it may create problem when someone makes change to query and not to parser.

@kgaikwad, without the column names in the output, the parser won't know the data mapping. As you said, if later someone changed the SQL but didn't update the parser. For example, change the name, type order with "select type, name from  compute_resources", then the data parsed by the parser will be in wrong mapping.

apuntamb:
LGTM :+1:

insights/parsers/satellite_compute_resources.py
```diff
@@ -0,0 +1,50 @@
...
+            columns_in_data = [item.strip() for item in content[0].split('|')]
+            if columns_in_data == columns:
+                for row in content[2:]:
+                    if '|' in row:
+                        data = row.split('|')
+                        if len(data) != len(columns):
+                            raise SkipException("The output is in bad format")
+                        self[data[0].strip()] = data[1].strip()
+        if not self:
+            raise SkipException('No valid values')
```
bfahr, line 50: @huali027 however you simplify #2929, you might be able to take advantage of the same code to parse this data since they are both Postgresql output of specific items.  It might actually be more logical to have both this and #2929 in the same module since they are both satellite settings.

bfahr (changes requested):

insights/parsers/satellite_compute_resources.py
huali027: @bfahr, sure, I'll combine them to the same module.

huali027:
@xiangce @bfahr, it is updated to use the parent "SatellitePostgreSQLQuery" class. Please help to review, thanks.

insights/parsers/satellite_postgresql_query.py
```diff
@@ -174,3 +176,30 @@ def get_setting(self, setting_name):
...
+        Please refer to its super-class :class:`insights.parsers.satellite_postgresql_query.SatellitePostgreSQLQuery` for more
+        details.
+
+    Sample output::
+
+        name,type
+        test_compute_resource1,Foreman::Model::Libvirt
+        test_compute_resource2,Foreman::Model::RHV
+
+    Examples::
```
xiangce, line 196: Please remove one colon.

xiangce (changes requested):

insights/parsers/satellite_postgresql_query.py
huali027: @xiangce, updated, thanks.

xiangce (approved):

PR #2939, opened 2021-02-07
New ld_library_path_of_user spec and UserLdLibraryPath parser

xiangce:


insights/specs/default.py
```diff
@@ -503,17 +503,33 @@ def httpd_cmd(broker):
...
+        users = broker[DefaultSpecs.regular_users]
+        ctx = broker[HostContext]
         llds = []
-        for p in pids:
-            with open('/proc/{0}/environ'.format(p), 'r') as fp:
-                vars = fp.read()
-                lld = [v.split('=', 1)[-1] for v in vars.split('\x00') if v.startswith('LD_LIBRARY_PATH=')]
-                llds.append("{0} {1}".format(p, lld[0])) if lld else None
+        for u in users:
+            ret, vvs = ctx.shell_out("/bin/su -l {0} -c export".format(u), keep_rc=True)
```
bfahr, line 525: @xiangce should you use the `env` command instead of `export` to ensure that it will work regardless of the shell configuration?

insights/parsers/ld_library_path.py
```diff
@@ -1,48 +1,96 @@
...
+"""
+.. warning::
+
+    This type is deprecated please use the :py:class:`LdLibraryPathU` instead.
+
+namedtuple: Type for storing the LdLibraryPath of PID
+"""
+
+
+class PidLdLibraryPath(Parser, list):
```
bfahr, line 71: @xiangce this parser didn't work and it was here for such a short period of time I would just remove this parser instead of deprecating it.

bfahr (changes requested):

insights/specs/default.py
xiangce: Done

bfahr (approved):
Approved pending approval of the spec.

PR #2945, opened 2021-02-09
RHCLOUD-11737: Implement playbook signature validation logic in insights-core

aleccohan:
This PR is the implementation of the playbook signature validation logic in the insights-core.  Currently the features of this PR are as follows:

The function verify() has two default parameters for testing purposes as well as for the initial implementation.  Should skipVerify be set to true.  This will ignore the verification logic and automatically return the playbook as received.  Additionally, in order to test the egg-versioning logic I have added a parameter called checkVersion which when set to false will skip this check.  

NOTE:  The final question regarding this implementation is around where the public key will be recieved from.  The initial idea was to fetch this key from the RedHatSecurity key server however the idea has been posed to instead wrap this key in the EGG so it can be accessible at any time.  

Please let me know if you have any questions about the implementation.

UPDATE 02/10/2021:
As opposed to calling out to the red hat key server I have implemented a file where the ascii armored key will live inside of the playbook verifier module so that it can be wrapped with the egg.  This will let us avoid connecting over ssl whenever we want to verify a playbook.

JIRA:  https://issues.redhat.com/browse/RHCLOUD-11737


SteveHNH:
I like the idea of keeping the key in the egg. My only question then is, if we update the signature with a new key, but the customer doesn't update the egg, what happens? Will all playbooks fail until they update?

aleccohan:
@SteveHNH That is a great question.  This ties in to the egg version check that has been implemented in this code.  I believe the idea is that playbook verification will only work if the customer is running the most recent version of the egg.  I _think_ that's how we want to handle it, at least initially.

SteveHNH (approved):

insights/tests/client/apps/test_playbook_verifier.py
```diff
@@ -0,0 +1,74 @@
...
+            'insights_signature': 'TFMwdExTMUNSVWRKVGlCUVIxQWdVMGxIVGtGVVZWSkZMUzB0TFMwS0N==',
+            'insights_signature_exclude': '/vars/insights_signature'
+        }
+    }]
+
+    with raises(PlaybookVerificationError) as error:
+        verify(fake_playbook, checkVersion=False)
+    assert key_error in str(error.value)
+
+@patch("insights.client.apps.ansible.playbook_verifier.executeVerification", return_value=True)
```
dehort, line 68: Similar to my other comment...

If this test is patching executeVerification() and returning True from executeVerification(), then is the test really checking much?  Maybe I'm misunderstanding the patching / test.  Is there something that this test is targeting?

Are you patching executeVerification() so that you avoid hitting the actual gpg code?  In that case, I might patch the gpg.verify_data() method.  Or generate a key pair for the test? (admittedly thats kinda gross...)

insights/tests/client/apps/test_playbook_verifier.py
```diff
@@ -0,0 +1,74 @@
...
+            'insights_signature_exclude': '/vars/insights_signature'
+        }
+    }]
+
+    with raises(PlaybookVerificationError) as error:
+        verify(fake_playbook, checkVersion=False)
+    assert key_error in str(error.value)
+
+@patch('insights.client.apps.ansible.playbook_verifier.verifyPlaybookSnippet', return_value=[])
+def test_playbook_verification_error(call):
```
dehort, line 54: If this test is patching verifyPlaybookSnippet(), then is the test really checking much? Maybe I'm misunderstanding the patching / test. Is there something that this test is targeting?

Are you patching verifyPlaybookSnippet() so that you avoid hitting the actual gpg code? In that case, I might patch the gpg.verify_data() method. Or generate a key pair for the test? (admittedly thats kinda gross...)

insights/client/apps/ansible/playbook_verifier/__init__.py
```diff
@@ -20,18 +36,106 @@ def __str__(self):
...
+
+    # Egg Version Check
+    eggVersioningCheck(checkVersion)
+
+    if not skipVerify:
+        for snippet in playbook:
+            if ('vars' not in snippet.keys()):
+                raise PlaybookVerificationError(message='VARS FIELD NOT FOUND: Verification failed')
+            elif (SIGKEY not in snippet['vars']):
+                raise PlaybookVerificationError(message='SIGNATURE NOT FOUND: Verification failed')
```
dehort, line 133: I'm possibly being nit-picky here...forgive me.  :)

It feels like this logic belongs in the verifyPlaybookSnippet() function.  The verify() method appears to be "borrowing" this logic.

dehort (approved):
Looks good to me!  I left a couple of comments though.

insights/tests/client/apps/test_playbook_verifier.py
aleccohan: ah!  yes, thank you for catching that.  I meant to remove that patch method once the public key file was created.  I'll remove that now.

insights/tests/client/apps/test_playbook_verifier.py
aleccohan: will do!  I'll go ahead and change this to mock gpg.verify_data()

insights/client/apps/ansible/playbook_verifier/__init__.py
aleccohan: done!

insights/client/apps/ansible/playbook_verifier/__init__.py
```diff
@@ -1,37 +1,139 @@
...
+import tempfile
+from logging import getLogger
+from requests.exceptions import HTTPError
+from insights.client.utilities import get_version_info
 
-__all__ = ("verify", "PlaybookValidationError")
+__all__ = ("verify", "PlaybookVerificationError")
 
+SIGKEY = 'insights_signature'
+PUBLIC_KEY_FOLDER = './insights/client/apps/ansible/playbook_verifier/public.gpg'    # Update this when we have the key generated
```
gravitypriest, line 15: You may want to use `pkgutil` to load static files here, I believe this won't work once it's a zipped egg. See `insights/client/map_components.py` for an example.

gravitypriest (changes requested):

insights/client/apps/ansible/playbook_verifier/__init__.py
aleccohan: @gravitypriest Now using pkgutil to load static files.

gravitypriest (approved):

PR #2931, opened 2021-02-03
Spec to collect pcp data via pmlogsummary

patilsuraj767:
This spec will collect the pcp data needed for ROS service.

PR #2950, opened 2021-02-11
Spec to collect pcp data via pmlogsummary

bfahr:
In collaboration with @patilsuraj767 this PR enhances and replaces #2931 to add a new spec and parser for the pmlog summary information


insights/specs/default.py
```diff
@@ -710,6 +712,34 @@ def pcp_enabled(broker):
...
+
+        Raises:
+            SkipComponent: raises this exception when the command is not present or fails
+        """
+        ps = broker[Ps]
+        hostname = broker[Hostname].fqdn
+        if ps.search(COMMAND__contains='pmlogger'):
+            pcp_log_date = (datetime.date.today() - datetime.timedelta(days=1)).strftime("%Y%m%d")
+            try:
+                cmd = "/usr/bin/pmlogsummary /var/log/pcp/pmlogger/%s/%s.index mem.util.used mem.physmem kernel.all.cpu.user kernel.all.cpu.sys kernel.all.cpu.nice kernel.all.cpu.steal kernel.all.cpu.idle disk.all.total" % (hostname, pcp_log_date)
```
csams, line 733: I'd create the datasource as a helper that constructs the file path and checks that it exists. Return it if so but raise the Skip otherwise. Then use [command_with_args](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/spec_factory.py#L751) to interpolate it into the pmlogsummary command. That'll make sure the command gets timeouts applied, exceptions are consistent, etc.

insights/parsers/pmlog_summary.py
```diff
@@ -0,0 +1,46 @@
...
+@parser(Specs.pmlog_summary)
+class PmLogSummary(CommandParser, dict):
+    """
+    Parser to parse the output of the ``pmlogsummary`` command
+
+    Sample output of the command is::
+
+        mem.util.used  3133919.812 Kbyte
+        mem.physmem  3997600.000 Kbyte
+        kernel.all.cpu.user  0.003 none
```
csams, line 20: We should unflatten these into a nest of dictionaries so the structure is consistent with what we're trying to do with models in general. I think it would also be useful to parse the value side into floats and "units."

csams (changes requested):

insights/parsers/pmlog_summary.py
```diff
@@ -0,0 +1,46 @@
...
+    Example:
+        >>> type(pmlog_summary)
+        <class 'insights.parsers.pmlog_summary.PmLogSummary'>
+        >>> 'mem.util.used' in pmlog_summary
+        True
+        >>> pmlog_summary['disk.all.total']
+        '0.252 count / sec'
+    """
+
+    def parse_content(self, content):
```
csams, line 36: Maybe something like this:
```python
#!/usr/bin/env python3


def parse(data):
    result = {}

    def typed(x):
        try:
            return float(x)
        except Exception:
            return x

    def insert(k, v): 
        cur = result
        key_parts = k.split(".")

        # walk down the structure to the correct leaf
        for part in key_parts:
            if part not in cur:
                cur[part] = {}
            cur = cur[part]

        # break the value apart and store it
        l, r = v.split(None, 1)
        cur["val"] = typed(l)
        cur["units"] = r.strip()

    def kvs():
        # deal with whitespace and high level splitting
        for line in data:
            line = line.strip()
            if line:
                yield line.split(None, 1)

    for k, v in kvs():
        insert(k, v)

    return result
```

insights/parsers/pmlog_summary.py
csams: Maybe something like this:
```python
#!/usr/bin/env python3


def parse(data):
    result = {}

    def typed(x):
        try:
            return float(x)
        except Exception:
            return x

    def insert(k, v): 
        cur = result
        key_parts = k.split(".")

        # walk down the structure to the correct leaf
        for part in key_parts:
            if part not in cur:
                cur[part] = {}
            cur = cur[part]

        # break the value apart and store it
        l, r = v.split(None, 1)
        cur["val"] = typed(l)
        cur["units"] = r.strip()

    def kvs():
        # deal with whitespace and high level splitting
        for line in data:
            line = line.strip()
            if line:
                yield line.split(None, 1)

    for k, v in kvs():
        insert(k, v)

    return result
```

insights/specs/default.py
```diff
@@ -711,6 +714,38 @@ def pcp_enabled(broker):
...
+
+        Raises:
+            SkipComponent: raises this exception when the command is not present or
+                the file is not present
+        """
+        ps = broker[Ps]
+        if ps.search(COMMAND__contains='pmlogger'):
+            hostname = broker[Hostname].fqdn
+            pcp_log_date = (datetime.date.today() - datetime.timedelta(days=1)).strftime("%Y%m%d")
+            file = "/var/log/pcp/pmlogger/%s/%s.index" % (hostname, pcp_log_date)
```
patilsuraj767, line 737: @bfahr ROS team has decided to have the custom PCP configuration on the user's system. Can you please change the file path as below.  

```suggestion
            file = "/var/log/pcp/pmlogger/ros/%s.index" % (pcp_log_date)
```

insights/specs/default.py
```diff
@@ -711,6 +714,38 @@ def pcp_enabled(broker):
...
+            try:
+                if os.path.exists(file) and os.path.isfile(file):
+                    return file
+            except Exception as e:
+                SkipComponent("Failed to check for pmlogger file existance: {0}".format(str(e)))
+
+        raise SkipComponent
+
+    pmlog_summary = command_with_args(
+        "/usr/bin/pmlogsummary %s mem.util.used mem.physmem kernel.all.cpu.user kernel.all.cpu.sys kernel.all.cpu.nice kernel.all.cpu.steal kernel.all.cpu.idle disk.all.total",
```
patilsuraj767, line 747: Please add three more metrics to the `pmlogsummary` command. 

```suggestion
        "/usr/bin/pmlogsummary %s mem.util.used mem.physmem kernel.all.cpu.user kernel.all.cpu.sys kernel.all.cpu.nice kernel.all.cpu.steal kernel.all.cpu.idle disk.all.total mem.util.cached mem.util.bufmem mem.util.free",
```

patilsuraj767 (changes requested):

insights/specs/default.py
bfahr: Done

insights/specs/default.py
bfahr: Done

patilsuraj767 (approved):
Tested the PR against the actual/real system. Spec is collected to the insights archive as expected.

insights/specs/default.py
bfahr: Done

patilsuraj767:
Hello @CodeHeeler @beav I would like to invite you on this PR as the pervious [PR](https://github.com/RedHatInsights/insights-core/pull/2931) is closed.  
Regarding the execution of command in the spec, we have introduce a custom way to collect the PCP logs. Check the below custom PCP configurations. 

1) `vi /var/lib/pcp/config/pmlogger/config.ros`

Paste the below configuration. 
```
log mandatory on default {
        mem.util.used
        mem.physmem
        kernel.all.cpu.user
        kernel.all.cpu.sys
        kernel.all.cpu.nice
        kernel.all.cpu.steal
        kernel.all.cpu.idle
        disk.all.total
        mem.util.cached
        mem.util.bufmem
        mem.util.free
}
[access]
disallow .* : all;
disallow :* : all;
allow local:* : enquire;
```

2. `echo "LOCALHOSTNAME	n   y	PCP_LOG_DIR/pmlogger/ros	-r -T24h10m -c config.ros -v 100Mb" >> /etc/pcp/pmlogger/control.d/local`
3. `systemctl restart pmcd pmlogger`
4. Verify - `ps -aux | grep ros `

Above should create custom PCP logs at location - /var/log/pcp/pmlogger/ros/ I hope with this we get log files of lesser size and it should reduce the time require to run pmlogsummary command.

@bfahr has requested for some measurements related to the execution timing with the custom PCP configuration. I will request you both if you can use the above pcp configs and share your inputs.

lhuett (approved):
Looks good to me.

PR #2955, opened 2021-02-17
Fix problem with corosync_cmapctl_cmd_list datasource

### LINKED ISSUE #2954: Corosync-cmapctl datasource should handle errors gracefully
The [datasource corosync_cmapctl_cmd_list datasource](https://github.com/RedHatInsights/insights-core/blob/6fd58aeab11232a62ea97ba576220abca808c660/insights/specs/default.py#L300) is causing the `corosync_cmapctl` spec to throw exceptions when the `corosync-cmapctl` command does not exist.  Also the commands are not consistent in the list created by the datasource.

bfahr:
* Check for presence of command on system before returning results to be
  executed
* Make list of commands consistent for all executables
* Fix #2954

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -9,6 +9,7 @@ data sources that standard Insights `Parsers` resolve against.
 """
 
 import logging
+import os
 import re
 import json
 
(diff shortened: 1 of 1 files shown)


vdymna (approved):
This looks good to me.

lhuett (approved):
Looks good.

PR #2956, opened 2021-02-17
Divert collect output to a separate debug log in archive

### LINKED ISSUE #2943: Capture core3 collection logs in the archive instead of mixing them with the client log
Users are confused and annoyed by insights core's logs during core3 collection. The messages are often verbose and contain tracebacks that aren't harmful but are distracting when trying to diagnose legitimate issues.

Put those logs into their own file in the archive instead of mixing them with the client log.

bfahr:
* During collection in dr.run_all, exceptions may be added to the log
  when in DEBUG mode, so capture all output from dr.run_all into a
  separate log and save it in the archive for analysis.
* These exceptions are expected, and do not normally impact the
  execution of the client, but may be confusing in the client log.
* Fix #2943

### DIFF
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -361,7 +361,7 @@ def collect(manifest=default_manifest, tmp_path=None, compress=False, rm_conf=No
 
     # Capture all logging for dr.run_all in a separate archive log
     LOG_FORMAT = ("%(asctime)s %(levelname)8s %(name)s %(message)s")
-    log_path = os.path.join(output_path, 'metadata', 'collection.log')
+    log_path = os.path.join(output_path, 'meta_data')
     fs.ensure_path(log_path)
     log_file = os.path.join(log_path, 'collection.log')
     file_handler = logging.FileHandler(log_file)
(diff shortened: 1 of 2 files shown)


insights/collect.py
```diff
@@ -359,6 +359,18 @@ def collect(manifest=default_manifest, tmp_path=None, compress=False, rm_conf=No
     fs.ensure_path(output_path)
     fs.touch(os.path.join(output_path, "insights_archive.txt"))
 
+    # Capture all logging for dr.run_all in a separate archive log
+    LOG_FORMAT = ("%(asctime)s %(levelname)8s %(name)s %(message)s")
+    log_path = os.path.join(output_path, 'metadata', 'collection.log')
```
bfahr, line 364: Path should be `meta_data` instead of `metadata`.

csams (commented):
Some of the exception messages are generated by the parser decorator and get logged in insights.core.plugins (https://github.com/RedHatInsights/insights-core/blob/master/insights/core/plugins.py#L165). I don't think the handler only on `dr` will catch those.

xiangce:
Closing it, we can reopen it in another new PR if necessary

PR #2957, opened 2021-02-19
Use yaml.dump() to convert the python dict into a string

dehort:
The change mirros the change in the signing tool:
https://github.com/RedHatInsights/playbook-signature-tools/pull/6


aleccohan (approved):
Looks great to me!  Just walked through this PR with the changes made to the signer and it works like a charm.

PR #2963, opened 2021-02-24
Use specific exception for missing archive spec

### LINKED ISSUE #2962: When deserializing core collection archives don't throw generic exceptions
There's a condition where during core collection a file may be collected but it is not saved in the archive because it is empty or not text or json.  In a core-collection archive there may be metadata that indicates a file that was removed should be present in the archive. When insights-core tries to deserialize files from this type of archive a generic exception will be thrown which is not representative of the condition.

Here's an example of the log messages:

```
WARNING:insights.core.serde:Unrecognized type: insights.core.spec_factory.TextFileProvider
WARNING:insights.core.serde:Unrecognized type: insights.core.spec_factory.CommandOutputProvider
```

The `deserialize` and `hydrate` functions should distinguish between a `ContentException`, which is this case, or some other type of exception and only log `ContentException` messages at DEBUG level and not WARNING level.

bfahr:
* When deserializing a spec from an archive use ContentException instead
  of Exception when file is missing from the archive
* Fix #2962

### DIFF
--- a/insights/core/serde.py
+++ b/insights/core/serde.py
@@ -87,11 +87,11 @@ def serialize(obj, root=None):
 
 
 def deserialize(data, root=None):
-    try:
-        (_type, from_dict) = DESERIALIZERS.get(data["type"])
-        return from_dict(_type, data["object"], root=root)
-    except Exception:
+    type_data = DESERIALIZERS.get(data["type"])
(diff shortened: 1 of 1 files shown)


lhuett (approved):
Looks good to me.

PR #2965, opened 2021-03-01
Update ld_library_path to check SAP users only

xiangce:


insights/specs/default.py
```diff
@@ -824,6 +797,27 @@ def sap_sid(broker):
         sap = broker[Sap]
         return list(set(sap.sid(i).lower() for i in sap.all_instances))
 
+    @datasource(sap_sid, HostContext)
+    def ld_library_path_of_user(broker):
```
bfahr, line 801: @xiangce would it be clearer to name this spec `ld_library_path_of_sap_user` instead?

xiangce: @bfahr - I'd like to make it more compatible as "ld_library_path_of_user", in case we may add users of other products. How do you think?

bfahr: Since the datasource requires `sap.sid` to work and its specific to SAP usernames, I don't think its very generalized as written.  However given the importance of getting this working I think its acceptable to proceed with the current name.

xiangce: @bfahr - Thanks. Maybe we could add the datasource of other products as dependencies when there is such rule.

PR #2968, opened 2021-03-04
Add invalid YAML check to CustomHiera

Rohan21Lobo:
A rule plugin requires to check if the config file has invalid yaml syntax. This change adds an is_invalid_yaml property to the parser.

### DIFF
--- a/insights/parsers/satellite_installer_configurations.py
+++ b/insights/parsers/satellite_installer_configurations.py
@@ -9,10 +9,12 @@ CustomHiera - file ``/etc/foreman-installer/custom-hiera.yaml``
 Parsers the file `/etc/foreman-installer/custom-hiera.yaml`
 
 """
+import sys
+import six
 
 from insights import parser, YAMLParser
 from insights.specs import Specs
-from insights.parsers import SkipException
+from insights.parsers import SkipException, ParseException
 
 
 @parser(Specs.satellite_custom_hiera)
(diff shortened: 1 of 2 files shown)


xiangce (approved):

insights/parsers/satellite_installer_configurations.py
```diff
@@ -27,7 +29,13 @@ class CustomHiera(YAMLParser):
         582
     """
     def parse_content(self, content):
+        self.is_invalid_yaml = False
         try:
             super(CustomHiera, self).parse_content(content)
         except SkipException:
             pass
+        except ParseException as ex:
```
bfahr, line 37: @Rohan21Lobo in the case where a parser is looking for bad or negative results, we recommend writing a second "negative" or "dirty" parser that will detect the bad condition and not do both in the same parser.  See the documentation on [exceptions](https://insights-core.readthedocs.io/en/latest/exception_model.html#case-3b-parsing-data-to-find-errors-dirty-parser).  The purpose of `SkipException` is to remove the parser from the dependency tree, and here its being circumvented.  This means that any rule that is using this parser cannot rely on the dependency checking to control its execution and it will have to check the `is_invalid_yaml` flag.  Having a negative parser would allow rules that are explicitly looking for either the positive or the negative case to simply rely on the dependency checking to execute.

bfahr (changes requested):

Rohan21Lobo:
Closing this MR as suggested by Bob

PR #2969, opened 2021-03-05
Add test to detect lssap issue in BZ1922937

bfahr:
The `lssap` command is producing the following output that causes the parser to thow an exception.  We need to determine how to handle this data in the parser, as valid data, or as data that is not reliable:
```
*** ERROR => CTrcOpen: fopen dev_lssap

---------------------------------------------------
trc file: "dev_lssap", trc level: 1, release: "721"
---------------------------------------------------
 - lssap version 1.0 -
==========================================
  SID   Nr   Instance   SAPLOCALHOST                        Version                 DIR_EXECUTABLE
 D02|  50|       D50|     sapcbapp09|722, patch 201, changelist 1718183|         /usr/sap/D02/D50/exe
==========================================
```

### DIFF
--- a/insights/parsers/tests/test_lssap.py
+++ b/insights/parsers/tests/test_lssap.py
@@ -74,6 +74,19 @@ HA2 | N16 | D16
 HB2 | foo | bar
 """.strip()
 
+Lssap_extra_lines = """
+*** ERROR => CTrcOpen: fopen dev_lssap
+
+---------------------------------------------------
+trc file: "dev_lssap", trc level: 1, release: "721"
+---------------------------------------------------
+ - lssap version 1.0 -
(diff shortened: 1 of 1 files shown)


bfahr:
@xiangce would you mind to take a look at this `lssap` output to determine if its valid and whether its OK to go ahead and parser it?

xiangce:
@bfahr - After checking the related case, I don't think these ERROR messages before the "===" line of the `lssap` output are valid for the Parser, we can just skip them and only parse the lines between the two "===" lines. How do you think?

xiangce:
Just clicked the "close" button by mistake.

xiangce (approved):

PR #2971, opened 2021-03-09
feat: RHICOMPL-1401 Send OS info in profile request

akofink:
Tailoring will now happen against specific rule sets per OS minor version rather than the initial profile of each policy. If a policy has no profile with a matching OS minor version (i.e. the UI tab does not exist), tailoring will not occur; the host will scan against the canonical rule set. If os_minor_version is missing completely from the API response (via old behavior or feature flag), the initial profile on the policy will be used for tailoring, the same behavior we have before COMP-E-133.

Related backend feature flag PR: https://github.com/RedHatInsights/compliance-backend/pull/787

This is part of COMP-E-133.

### DIFF
--- a/insights/client/apps/compliance/__init__.py
+++ b/insights/client/apps/compliance/__init__.py
@@ -27,7 +27,9 @@ class ComplianceClient:
 
     def oscap_scan(self):
         self._assert_oscap_rpms_exist()
-        policies = self.get_policies()
+        initial_profiles = self.get_initial_profiles()
+        matching_os_profiles = self.get_profiles_matching_os()
+        policies = self.profile_union_by_ref_id(matching_os_profiles, initial_profiles)
         if not policies:
             logger.error("System is not associated with any profiles. Assign profiles using the Compliance web UI.\n")
             exit(constants.sig_kill_bad)
@@ -42,7 +44,8 @@ class ComplianceClient:
         return self.archive.create_tar_file(), COMPLIANCE_CONTENT_TYPE
 
     def download_tailoring_file(self, policy):
-        if 'tailored' not in policy['attributes'] or policy['attributes']['tailored'] is False:
+        if ('tailored' not in policy['attributes'] or policy['attributes']['tailored'] is False or
+                ('os_minor_version' in policy['attributes'] and policy['attributes']['os_minor_version'] != self.os_minor_version())):
             return None
 
         # Download tailoring file to pass as argument to run_scan
@@ -66,11 +69,9 @@ class ComplianceClient:
 
         return tailoring_file_path
 
-    # TODO: Not a typo! This endpoint gives OSCAP policies, not profiles
-    # We need to update compliance-backend to fix this
-    def get_policies(self):
+    def get_profiles(self, search):
         response = self.conn.session.get("https://{0}/compliance/profiles".format(self.config.base_url),
-                                         params={'search': 'system_names={0} external=false canonical=false'.format(self.hostname)})
+                                         params={'search': search})
         logger.debug("Content of the response: {0} - {1}".format(response,
                                                                  response.json()))
         if response.status_code == 200:
@@ -78,12 +79,30 @@ class ComplianceClient:
         else:
             return []
 
+    def get_initial_profiles(self):
+        return self.get_profiles('system_names={0} canonical=false external=false'.format(self.hostname))
+
+    def get_profiles_matching_os(self):
+        return self.get_profiles('system_names={0} canonical=false os_minor_version={1}'.format(self.hostname, self.os_minor_version()))
+
+    def profile_union_by_ref_id(self, prioritized_profiles, merged_profiles):
+        profiles = dict((p['attributes']['ref_id'], p) for p in merged_profiles)
+        profiles.update(dict((p['attributes']['ref_id'], p) for p in prioritized_profiles))
+
+        return list(profiles.values())
+
     def os_release(self):
         _, version, _ = linux_distribution()
-        return findall("^[6-8]", version)[0]
+        return version
+
+    def os_major_version(self):
+        return findall("^[6-8]", self.os_release())[0]
+
+    def os_minor_version(self):
+        return findall("\d+$", self.os_release())[0]
 
     def profile_files(self):
-        return glob("{0}*rhel{1}*.xml".format(POLICY_FILE_LOCATION, self.os_release()))
+        return glob("{0}*rhel{1}*.xml".format(POLICY_FILE_LOCATION, self.os_major_version()))
 
     def find_scap_policy(self, profile_ref_id):
         grepcmd = 'grep ' + profile_ref_id + ' ' + ' '.join(self.profile_files())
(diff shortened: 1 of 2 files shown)


insights/client/apps/compliance/__init__.py
```diff
@@ -69,8 +69,9 @@ def download_tailoring_file(self, policy):
     # TODO: Not a typo! This endpoint gives OSCAP policies, not profiles
     # We need to update compliance-backend to fix this
     def get_policies(self):
+        search = 'system_names={0} canonical=false os_minor_version={1}'.format(self.hostname, self.os_minor_version())
```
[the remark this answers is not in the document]
akofink: > If get_policies is to gather policies that the system is assigned to, then it should not filter them by an OS version.

Before, we were looking for the initial profile (`external = false`). Swapping that for os_minor_version still gives us unique 'policies' (used very loosely, any one profile in each policy). This query gives exact tab matches. It's true that this PR does not yet consider when the tab does not exist.

> external=false should be kept to retrieve policy assignment (w/o any OS filtering). Otherwise there can be a confusion when the OS is updated (and/or not configured on a policy). In that case user will receive info on that the system isn't assigned, which in reality it can be. This feature should be only around tailoring and not policy assignment, unless I've missed something.

100% agree with this. Sorry, I didn't make it clear that this PR is only half-baked. Here's what I was thinking:

1. Find policy profiles that match exactly by tab (i.e. os_minor_version)
2. Find initial policy profiles (the old API request with external=false)
3. Scan based on (ref_ids+tailorings from 1) + (ref_ids from 2 without tailorings - ref_ids from 1)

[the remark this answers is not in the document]
vkrizan: I'd reorder the steps to have the "Find initial policy profiles" first, unless you wanted to have it as a fallback. What you are suggesting makes sense, as it starts using the correct objects. It kinda feels that it moves the selection logic to the client, and also considering that it does so by using filters.

My proposal is/was to keep the semantics with just the initial policy profile, by forcing `tailored=true` and passing OS info on the tailoring endpoint. The idea is that client would request assigned policies, and for each policy it would request tailoring. Tailoring selection would be done on backend with the information provided by client. (policy and initial profile is interchangeable)

insights/client/apps/compliance/__init__.py
```diff
@@ -27,63 +27,82 @@ def __init__(self, config):
...
+                tailoring_file_path=self.download_tailoring_file(profile)
             )
 
         return self.archive.create_tar_file(), COMPLIANCE_CONTENT_TYPE
 
-    def download_tailoring_file(self, policy):
-        if 'tailored' not in policy['attributes'] or policy['attributes']['tailored'] is False:
+    def download_tailoring_file(self, profile):
+        if ('tailored' not in profile['attributes'] or profile['attributes']['tailored'] is False or
+                'os_minor_version' not in profile['attributes'] or profile['attributes']['os_minor_version'] != self.os_minor_version()):
```
akofink, line 48: @vkrizan I believe removing this condition here would allow the client to work as it does today if there are no os_minor_versions defined in a policy's profiles. I will test it out to be sure, but this would allow us to merge this work today without fear of breaking customers, and we can revisit this condition later to turn off tailoring again when a matching os_minor_version is not found.

akofink:
@gravitypriest would you mind please glancing over this PR for any client related gotchas?

I just rebased against master, and the build started failing. Hopefully it was just a fluke and master isn't broken :)

insights/client/apps/compliance/__init__.py
akofink: I've now commented out this line and adjusted a test for this scenario.

insights/client/apps/compliance/__init__.py
```diff
@@ -27,63 +27,82 @@ def __init__(self, config):
...
+                # or 'os_minor_version' not in profile['attributes'] or profile['attributes']['os_minor_version'] != self.os_minor_version()): # TODO after completion of COMP-E-133
             return None
 
         # Download tailoring file to pass as argument to run_scan
         logger.debug(
-            "Policy {0} is a tailored policy. Starting tailoring file download...".format(policy['attributes']['ref_id'])
+            "Policy {0} is a tailored policy. Starting tailoring file download...".format(profile['attributes']['ref_id'])
         )
-        tailoring_file_path = "/var/tmp/oscap_tailoring_file-{0}.xml".format(policy['attributes']['ref_id'])
+        tailoring_file_path = "/var/tmp/oscap_tailoring_file-{0}.xml".format(profile['attributes']['ref_id'])
```
vkrizan, line 55: Unrelated to this idea, but this feels unsafe to construct path out of what comes from the API without any input validation. Mind to call basename on the formatted filename?

vkrizan (approved):
Changes make sense.
I haven't tested them, yet.

insights/client/apps/compliance/__init__.py
akofink: Maybe we can revisit this in a future PR. I'm not sure how this could really be unsafe, since the file path is used to temporarily store the tailoring file so it can be passed to `oscap`. It doesn't exist before the client run, and trying to inject some bogus path here would just cause `oscap` to fail. Also, ref_id is immutable in our API, so it'd be impressive if a user could somehow hijack that value.

We should revisit the tmp dir that we use, though. There is `self.archive.tmp_dir`, which gets automatically cleaned up after the client run. Also, I think most of these paths should probably use `os.path.join` instead of just a string path. I'm going to hold off on these changes in this PR, since they're not related to the new functionality. If you find a genuine security threat, we should definitely report a jira and prioritize fixing it.

vkrizan: Ack. It should be okay for the time being.

Here is what I had in mind, but it won't go through
```
$ touch /var/tmp/oscap_tailoring_file-/../test.xml
touch: cannot touch '/var/tmp/oscap_tailoring_file-/../test.xml': No such file or directory
```

We can further improve it as part of RHICOMPL-1196.

gravitypriest:
Look A-OK to me :+1:

insights/tests/client/apps/test_compliance.py
```diff
@@ -156,11 +192,32 @@ def test_tailored_file_is_not_downloaded_if_tailored_is_missing(config):
 
 @patch("insights.client.apps.compliance.open", new_callable=mock_open)
 @patch("insights.client.config.InsightsConfig")
-def test_tailored_file_is_downloaded_if_needed(config, call):
+def test_tailored_file_is_downloaded_from_initial_profile_if_os_minor_version_is_missing(config, call):
```
akofink, line 195: Notice here the test doesn't need to change at all if the `os_minor_version` attribute is missing from the API response. I also added a check for `'tailored': False` for good measure.

akofink:
@vkrizan please test with the COMP-E-133 feature flag enabled and disabled, with tailored and non-tailored policies, and with os_minor_version set and unset on the relevant profile. All those cases are covered in the unit tests, but it never hurts to validate. Thanks!

vkrizan:
@akofink Tried with some cases and it seems to work as expected. Great job!

PS: When I was testing it and forgot to turn the backend on there was an awful error repeated (`ValueError: No JSON object could be decoded`) with traceback.

akofink:
Reported the `ValueError` as https://issues.redhat.com/browse/RHICOMPL-1529.

LightOfHeaven1994:
@akofink No issue, tested with backend FF turned on/off. LGTM

akofink:
@SteveHNH please take a look and merge when convenient. Thanks!

PR #2972, opened 2021-03-10
Change to use 'Hostname' instead of 'hostname' combiner

xiangce:
- The hostname combiner is deprecated, change to use the Hostname
   To avoid the "DeprecationWarning"s in test of rules.


bfahr (approved):
@xiangce thanks for taking the time to fix this.

PR #2630, opened 2020-06-16
use SIGTERM for rpm and yum commands

gravitypriest:
Signed-off-by: Jeremy Crafts <jcrafts@redhat.com>

PR #2974, opened 2021-03-10
Use SIGTERM for rpm and yum commands

bfahr:
* Need to apply #2630 for core collection
* Fixes bugzilla 1935846


csams (approved):
Looks good to me

PR #2975, opened 2021-03-10
Remove lssap command from collection

bfahr:
* The saphostctrl command is a better source of the information
* Fixes two open bugzillas with the lssap command 1922937 and 1936951
* Registry spec and parser will remain for use in the future


xiangce (approved):

xiangce:
Removal of the Spec is approved.

PR #2981, opened 2021-03-16
Add parser to parse satellite custom ca cert chain

huali027:

### DIFF
--- a/insights/parsers/certificate_chain_enddates.py
+++ b/insights/parsers/certificate_chain_enddates.py
@@ -10,7 +10,7 @@ SatelliteCustomCaChain - command ``awk 'BEGIN { pipe="openssl x509 -noout -subje
 
 from insights import parser, CommandParser
 from datetime import datetime
-from insights.parsers import ParseException
+from insights.parsers import ParseException, SkipException
 from insights.specs import Specs
 from insights.parsers.certificates_enddate import ExpirationDate
 
@@ -50,18 +50,28 @@ class CertificateChainEnddates(CommandParser, list):
         Parse the content of crt chain file. And it saves the expiration
         info of each crt in a list of dict.
 
+        Attributes:
+
+            earliest_date(ExpirationDate): The earliest datetime of the cert in the chain
+                                            or None when there isn't "notAfter" for all
+                                            the certs in the chain.
+
         Raises:
 
             ParseException: when the output isn't in key=value format or
                             the notAfter or notBefore isn't expected format.
         """
+        if len(content) < 1:
+            raise SkipException("No cert in the output")
+        data = {}
+        self.append(data)
+        self.earliest_date = None
         for index, line in enumerate(content):
-            if index == 0:
-                data = {}
             if not line.strip():
                 # a new cert starts
-                self.append(data)
-                data = {}
+                if data:
+                    data = {}
+                    self.append(data)
                 continue
             if '=' not in line:
                 raise ParseException('The line %s is not in key=value format' % line)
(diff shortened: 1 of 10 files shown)


huali027:
@xiangce, please help to review this parser when u have time, thanks.

insights/parsers/certificate_chain_enddates.py
```diff
@@ -0,0 +1,115 @@
...
+                value = ExpirationDate(value, date_time)
+            data[key] = value
+            if index == len(content) - 1:
+                # append the last cert
+                self.append(data)
+
+    @property
+    def earliest_expiration_date(self):
+        """This will return the earliest expiration date or None if notAfter not found"""
+        earliest_date = None
```
xiangce, line 83: It's not recommended to do the complex computation in a `property` method, since the property method works as an attribute that could be called times.

Better calculated the result in the parse_content and just return the value in the `property` method.

insights/parsers/certificates_enddate.py
```diff
@@ -42,13 +42,14 @@
 from insights.specs import Specs
 
 
+# namedtuple: contains the expiration date in string and datetime format.
+ExpirationDate = namedtuple('ExpirationDate', ['str', 'datetime'])
```
xiangce, line 46: Actually, this change modifies the API of the Parser.  However, it makes sense for the new parser and doesn't break any existing rule. It looks acceptable to me. How about you @bfahr ?

insights/parsers/certificate_chain_enddates.py
```diff
@@ -0,0 +1,115 @@
...
+            ParseException: when the output isn't in key=value format or
+                            the notAfter or notBefore isn't expected format.
+        """
+        for index, line in enumerate(content):
+            if index == 0:
+                data = {}
+            if not line.strip():
+                # a new cert starts
+                self.append(data)
+                data = {}
```
xiangce, line 64: By `appending` the data before creating a new one, the excess `append` operation for the last cert can be saved. Just a suggestion.

xiangce (changes requested):

insights/parsers/certificate_chain_enddates.py
huali027: @xiangce, sure, I'll remove it to the parse_content, thanks.

huali027:
@xiangce, the dependency is removed, please help to review when u have time, thanks.

insights/parsers/certificate_chain_enddates.py
```diff
@@ -0,0 +1,118 @@
...
+                value = ExpirationDate(value, date_time)
+            data[key] = value
+
+        for one_cert in self:
+            expire_date = one_cert.get('notAfter')
+            if expire_date and (self.earliest_date is None or expire_date.datetime < self.earliest_date.datetime):
+                self.earliest_date = expire_date
+
+    @property
+    def earliest_expiration_date(self):
```
xiangce, line 90: `earliest_expiration_date` is the same as `earliest_date`, better just keep one of them.

Make the `self.earliest_date` a private member ` self._earliest_date` or remove the `property` one.

huali027: @xiangce, good catch. I'll update it, thanks.

insights/parsers/certificate_chain_enddates.py
huali027: Updated as suggestion.

insights/parsers/certificate_chain_enddates.py
huali027: Updated as suggestion.

xiangce (approved):

insights/parsers/certificates_enddate.py
bfahr: @xiangce @huali027 this change isn't really necessary and it does impact anyone that is using this parser and the `ExpirationDate` tuple.  I think we should stick to our rule to maintain backward compatibility and not make this change.

bfahr (changes requested):

huali027:
@bfahr, I reverted it and use it directly. Please help to review, thanks.

PR #2991, opened 2021-03-25
Add parser to get the issuer of satellite client cert

huali027:
* Add parser to get issuer of "/etc/rhsm/ca/katello-default-ca.pem".
* Refactor CertificateChain to reuse the CertificateInfo since chain is just
  a list of cert.


huali027:
@xiangce, please help to review when u have time, thanks.

xiangce (commented):
@huali027 - sorry for the incomplete suggestion, making one parser as the element of another looks a bit weird. How about just extracting the parsing of the section and making a shared method?

huali027:
> @huali027 - sorry for the incomplete suggestion, making one parser as the element of another looks a bit weird. How about just extracting the parsing of the section and making a shared method?

@xiangce, updated based on the suggestion, please review again, thanks.

insights/parsers/ssl_certificate.py
```diff
@@ -0,0 +1,188 @@
...
+            context,
+            extra_bad_lines=['error opening certificate', 'unable to load certificate'])
+
+    def parse_content(self, content):
+        """
+        This uses the :py:func:`insights.parsers.ssl_certificate.parse_openssl_output` function.
+        See its documentation for parsing details.
+        """
+
+        self.update(parse_openssl_output(content))
```
xiangce, line 84: If it's possible that the result of `parse_openssl_output` an empty dict, better skip it by rating a SkipException.

insights/parsers/ssl_certificate.py
```diff
@@ -0,0 +1,188 @@
...
+        start_index = 0
+        for index, line in enumerate(content):
+            if not line.strip():
+                # one cert ends
+                if start_index != index:
+                    self.append(parse_openssl_output(content[start_index:index]))
+                start_index = index + 1
+            if index == len(content) - 1:
+                self.append(parse_openssl_output(content=content[start_index:index + 1]))
+        for one_cert in self:
```
xiangce, line 139: The same comment as above, in case the list is empty.

insights/parsers/ssl_certificate.py
huali027: @xiangce, updated, thanks.

xiangce (approved):
Looks good to me. @bfahr - Please double confirm this PR.
@huali027 removed the newly added parser `certificate_chain` and move the two base classes `CertificateInfo` and `CertificateChain`  to another new parser named `ssl_cerfificate`.

Although this PR changes an existing parser, the parser was just added several days ago. I think it's OK to make this change. 

Thanks.

bfahr:
> Looks good to me. @bfahr - Please double confirm this PR.
> @huali027 removed the newly added parser `certificate_chain` and move the two base classes `CertificateInfo` and `CertificateChain` to another new parser named `ssl_cerfificate`.
> 
> Although this PR changes an existing parser, the parser was just added several days ago. I think it's OK to make this change.
> 
> Thanks.

@xiangce I think that is OK in this case.

@huali027 do you anticipate adding more parsers that would use the `CertificateChain` and `CertificateInfo` base classes? You are only using them once each right?

huali027:
> > Looks good to me. @bfahr - Please double confirm this PR.
> > @huali027 removed the newly added parser `certificate_chain` and move the two base classes `CertificateInfo` and `CertificateChain` to another new parser named `ssl_cerfificate`.
> > Although this PR changes an existing parser, the parser was just added several days ago. I think it's OK to make this change.
> > Thanks.
> 
> @xiangce I think that is OK in this case.
> 
> @huali027 do you anticipate adding more parsers that would use the `CertificateChain` and `CertificateInfo` base classes? You are only using them once each right?

@bfahr, if later we have new spec to get the certificate info or certificate chain info, I'll use them too. 
If it's okay, I plan to update `PuppetCertExpireDate` to extend `CertificateInfo` to avoid duplicate parsing work. 
I'll mark `self.expire_date` deprecated first to keep compatibility. Currently, there is only one advisor rule with the parser, I'll update the rule too.

PR #2993, opened 2021-03-26
Add custom datasource for collecting yum/dnf updates

semtexzv:
Dynamically imports os-specific library for handling DNF/YUM interaction.
Performs a resolution of updateable packages, and outputs it as JSON.
Expected performance is on the order of seconds for both RHEL7 and RHEL8.

cc: @MichaelMraka

### DIFF
--- a/insights/parsers/yum_updates.py
+++ b/insights/parsers/yum_updates.py
@@ -1,0 +1,54 @@
+"""
+YumUpdates - parser for the `yum_updates` datasource
+====================================================
+Provides a list of available package updates, along with related advisories. This information
+is collected using DNF/YUM python interface.
+"""
+
+from insights import JSONParser, parser
+from insights.specs import Specs
+
+
+@parser(Specs.yum_updateinfo)
+class YumUpdates(JSONParser):
+    """
+    Expected output of the command is::
+
+        {
+          "releasever": "8",
+          "basearch": "x86_64",
+          "update_list": {
+            "NetworkManager-1:1.22.8-4.el8.x86_64": {
+              "available_updates": [
+                {
+                  "package": "NetworkManager-1:1.22.8-5.el8_2.x86_64",
+                  "repository": "rhel-8-for-x86_64-baseos-rpms",
+                  "basearch": "x86_64",
+                  "releasever": "8",
+                  "erratum": "RHSA-2020:3011"
+                }
+              ]
+            }
+          },
+          "metadata_time": "2021-01-01T09:39:45Z"
+        }
+
+    Examples:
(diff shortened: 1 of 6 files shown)


insights/specs/default.py
```diff
@@ -964,7 +968,215 @@ def is_mod_loaded_for_ss(broker):
     yum_log = simple_file("/var/log/yum.log")
     yum_repolist = simple_command("/usr/bin/yum -C --noplugins repolist")
     yum_repos_d = glob_file("/etc/yum.repos.d/*.repo")
-    yum_updateinfo = simple_command("/usr/bin/yum -C updateinfo list")
+    #yum_updateinfo = simple_command("/usr/bin/yum -C updateinfo list")
+
+    @datasource(HostContext)
+    def yum_updates(_broker):
+        DEBUG = True
```
MichaelMraka, line 975: DEBUG = True
should be better switched off for production

MichaelMraka: debug  is still on

MichaelMraka (changes requested):

bfahr (changes requested):

bfahr (changes requested):
@semtexzv it looks like you've made all of the requested changes but still need to fix some of your tests.  If you can get those fixed I think it will be ready to merge.

insights/parsers/yum_updates.py
```diff
@@ -0,0 +1,78 @@
...
+from insights.specs import Specs
+
+
+@parser(Specs.yum_updateinfo)
+class YumUpdates(JSONParser):
+    """
+
+    Expected output of the command is::
+
+    {
```
bfahr, line 18: @semtexzv this is causing the unexpected indentation error, it needs to be indented.

bfahr (changes requested):

insights/specs/default.py
semtexzv: Removed

insights/parsers/yum_updates.py
semtexzv: Fixed

bfahr:
@semtexzv this PR will need to be updated after #3088 has been merged.  You can look at that PR to see how the datasources will be moved out of default.py.

bfahr:

semtexzv:
Thank you, updated. I wasn't sure about testing the datasource, since it invokes local dnf/yum.

bfahr (approved):

bfahr (approved):

bfahr:
Verified rebase with master and tests pass.

PR #2995, opened 2021-03-30
Remove the duplicated lspci_kernel

xiangce:
- `lspci_kenrel` is the same as the `lspci`
   It's OK that just keep the `lspci` which is used by rules.


bfahr (commented):
Approved pending spec approval.

PR #3001, opened 2021-04-01
Add lssap spec to insights archives

bfahr:
* This spec is no longer collect by new clients but may still be
  collected by older clients so adding back to support those


bfahr:
Fix #3002

lhuett (approved):
Looks good.

PR #3009, opened 2021-04-08
Extend CertificateInfo instead of duplicate code

huali027:


huali027:
@xiangce, this is the same structure as other openssl command, so I update it to use the base class. Please help to review, thanks.

xiangce (approved):

PR #3010, opened 2021-04-08
Manually overide skipVerify

aleccohan:
Change Validation function to manually skipVerify until all playbook-snippets are signed.

I also commented out most of the tests for the playbook verifier until such a time when we can remove the manual skip verify.

### DIFF
--- a/insights/client/apps/ansible/playbook_verifier/__init__.py
+++ b/insights/client/apps/ansible/playbook_verifier/__init__.py
@@ -41,7 +41,6 @@ class PlaybookVerificationError(Exception):
 
 def eggVersioningCheck(checkVersion):
     currentVersion = requests.get(VERSIONING_URL)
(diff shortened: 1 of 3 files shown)


insights/client/apps/ansible/playbook_verifier/__init__.py
```diff
@@ -132,6 +132,8 @@ def verify(playbook, checkVersion=True, skipVerify=False):
     # Egg Version Check
     eggVersioningCheck(checkVersion)
```
gravitypriest, line 133: IMO if we're wholesale skipping playbook verification til it's ready, we might as well skip the version check as well so the function never fails.

gravitypriest (changes requested):

gravitypriest (approved):

gravitypriest (approved):

PR #3011, opened 2021-04-09
Update spec "certificates_enddate" to make it always success

huali027:
* Since it is to check expired date of certificates for several directories
  by command find, even the directories don't exist, the command shouldn't fail

### DIFF
--- a/insights/parsers/tests/test_certificates_enddate.py
+++ b/insights/parsers/tests/test_certificates_enddate.py
@@ -63,6 +63,13 @@ FileName= /etc/pki/consumer/cert.pem
 
 PATH1 = "/etc/origin/node/cert.pem"
 
+CRT7 = """
+/usr/bin/find: '/etc/origin/node': No such file or directory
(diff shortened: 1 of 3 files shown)


huali027:
@xiangce, please help to review, thanks.

insights/parsers/tests/test_certificates_enddate.py
```diff
@@ -89,6 +96,9 @@ def test_certificates_enddate():
     assert (set(Cert5.certificates_path) == set([
             '/etc/pki/ca-trust/extracted/pem/email-ca-bundle.pem']))
 
+    Cert6 = CertificatesEnddate(context_wrap(CRT7))
+    assert not Cert6.data
```
xiangce, line 100: In this case, `SkipException` should be thrown, Would you mind improve the logic of this Parser?

huali027: @xiangce, updated, please help to review again, thanks.

xiangce (approved):

PR #3012, opened 2021-04-09
Update spec to make it always success

huali027:
* these two commands should still return even some file doen't exist


huali027:
@xiangce, please help to review. From my findings, only these two need to update.

xiangce:
@huali027 - thanks for your efforts on this issue.

xiangce (approved):

PR #3013, opened 2021-04-09
Add conf file path for rh-mongodb34 to mongod_conf

Rohan21Lobo:
/etc/opt/rh/rh-mongodb34/mongod.conf has same YAML format as of rh-mongodb26

### DIFF
--- a/insights/parsers/mongod_conf.py
+++ b/insights/parsers/mongod_conf.py
@@ -1,13 +1,15 @@
 """
-MongodbConf - files ``/etc/mongod.conf``, ``/etc/mongod.conf`` and ``/etc/opt/rh/rh-mongodb26/mongod.conf``
-===========================================================================================================
+MongodbConf - files - Configuration files for MongoDB
+=====================================================
 
 This module contains the following files:
     ``/etc/mongod.conf``,
     ``/etc/mongodb.conf`` ,
-    ``/etc/opt/rh/rh-mongodb26/mongod.conf``.
+    ``/etc/opt/rh/rh-mongodb26/mongod.conf``
+    ``/etc/opt/rh/rh-mongodb34/mongod.conf``
 
-They are provided by package mongodb-server or rh-mongodb26-mongodb-server.
+They are provided by package mongodb-server, rh-mongodb26-mongodb-server or
+rh-mongodb34-mongodb-server.
 These MongoDB configuration files may use the **YAML** format
 or the standard **key-value pair** format.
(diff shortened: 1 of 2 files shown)


insights/parsers/mongod_conf.py
```diff
@@ -1,13 +1,15 @@
 """
-MongodbConf - files ``/etc/mongod.conf``, ``/etc/mongod.conf`` and ``/etc/opt/rh/rh-mongodb26/mongod.conf``
-===========================================================================================================
+MongodbConf - files ``/etc/mongod.conf``, ``/etc/mongod.conf``, ``/etc/opt/rh/rh-mongodb26/mongod.conf`` and ``/etc/opt/rh/rh-mongodb34/mongod.conf``
+=====================================================================================================================================================
```
xiangce, line 3: Better make the title a brief introduction of the parser, but not to reflect each of the files, the following doc will do that in detail.

e.g.
```
MongodbConf - files - Configuration files for MongDB
```

insights/specs/default.py
```diff
@@ -595,7 +595,8 @@ def md_device_list(broker):
     mongod_conf = glob_file([
                             "/etc/mongod.conf",
                             "/etc/mongodb.conf",
-                            "/etc/opt/rh/rh-mongodb26/mongod.conf"
+                            "/etc/opt/rh/rh-mongodb26/mongod.conf",
```
huali027, line 598: @Rohan21Lobo, from https://access.redhat.com/support/policy/updates/rhscl-rhel7,
 I found rh-mongodb26 is already retired, I think we can delete it. How do u think?

Rohan21Lobo: I think RHUI 3 still uses mongodb26. Don't have an instance to check as of now.

insights/parsers/mongod_conf.py
Rohan21Lobo: Done, Please review.

xiangce: NIT: Sorry for the original typo, but it should be `MongoDB`

insights/specs/default.py
xiangce: Yeap, it doesn't hurt to keep it here even `26` is retired.

insights/parsers/mongod_conf.py
Rohan21Lobo: Fixed . ;) I copied blindly

PR #3015, opened 2021-04-12
Update spec "postgresql_conf" and "postgresql_log"

huali027:
* Remove unsupported poggresql conf and log path
* remove the path which is only supported in sosreport to sos_archive


huali027:
@xiangce, please help to review, thanks.

xiangce (approved):

PR #3017, opened 2021-04-12
Azure marketplace parser

SteveHNH:
We'll need to identify azure marketplace systems inside puptoo. Since we
rely on core for this, we need to have this parser in place.

RHCLOUD-13520

### DIFF
--- a/insights/parsers/azure_instance_plan.py
+++ b/insights/parsers/azure_instance_plan.py
@@ -11,7 +11,7 @@ For more details, See: https://docs.microsoft.com/en-us/rest/api/compute/virtual
 """
 import json
 
-from insights.parsers import SkipException
+from insights.parsers import SkipException, ParseException
 from insights import parser, CommandParser
 from insights.specs import Specs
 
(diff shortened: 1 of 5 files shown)


insights/parsers/azure_instance_plan.py
```diff
@@ -0,0 +1,60 @@
...
+        'planProduct'
+        >>> azure_plan.publisher
+        'planPublisher'
+    """
+
+    def parse_content(self, content):
+        if not content or "curl: " in content[0]:
+            raise SkipException()
+
+        plan = json.loads(content[0])
```
bfahr, line 56: @SteveHNH please add an exception handler around this so that we don't throw unexpected exceptions.  You can raise a ParseException if there is a problem with the `json.loads`.

insights/parsers/azure_instance_plan.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Attributes:
+        name (str): The name of the plan for the VM Instance in Azure, e.g: rhel7
+        size (str): The product of the plan for the VM Instance in Azure, e.g: RHEL
+        publisher (str): The publisher of the plan for the VM Instance in Azure, e.g: Red hat
+        raw (str): The full JSON of the plan returned by the ``curl`` command
+
+    Examples:
+        >>> azure_plan.name
+        'planName'
```
bfahr, line 45: You have some test failures for these doctests in RHEL7.  An easy way to fix them is change the example to something like:
```
>>> asure_plan.name == 'planName'
True
```

bfahr (changes requested):

SteveHNH:
@bfahr fixed the issues you pointed out. Thanks!

xiangce (approved):

PR #3018, opened 2021-04-13
Remove the spec because it is not used by any rule

huali027:
* The ssl certificate changes from satellite 6.6,
  however since no rules use this spec, no need to update it,
  we can remove it directly


xiangce (approved):

xiangce:
@huali027 - you should add the removed spec to the skipped spec-list in the `insights/tests/client/collection_rules/test_map_components.py` to make the pytest pass

PR #3025, opened 2021-04-19
[testing] download idna 2.7 directly

SteveHNH:
This package was removed from pypi. We still need it for python2, so
we'll download it directly via git.


Glutexo (approved):
We don’t know why PIP failed to find the matching versions. My uneducated guess would be that as Python 2.6 is unsupported, older version of _pip_ may no longer be able to work correctly with PyPI. Installing directly from GitHub resolves this for now.

This is definitely not a good long-term solution, but this pipeline is intended to go away soon anyway. (Haha.)

bfahr (approved):

PR #3027, opened 2021-04-19
Collect `rpm-ostree status --json`

csams:

### DIFF
--- a/insights/parsers/rpm_ostree_status.py
+++ b/insights/parsers/rpm_ostree_status.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+"""
+RpmOstreeStatus - Command ``/usr/bin/rpm-ostree status --json``
+===============================================================
+
+The ``RpmOstreeStatus`` class parses the output of the ``/usr/bin/rpm-ostree status --json`` command.
+
+rpm-ostree is a hybrid image and package system; as the name suggests, it
+uses OSTree for the image side, and RPM for the package side. It supports
+composing RPMs server-side into an OSTree commit (like an image), and clients
+can replicate that bit-for-bit, with fast incremental updates. Additionally,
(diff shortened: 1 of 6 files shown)


insights/parsers/tests/test_rpm_ostree_status.py
```diff
@@ -0,0 +1,52 @@
...
+    "cached-update" : null
+    }
+""".strip()
+
+
+def test_good_data():
+    data = context_wrap(GOOD)
+    status = RpmOstreeStatus(data)
+    assert status.data["deployments"][0]["booted"]
+    assert len(status.query.deployments.where("booted", True)) == 1
```
bfahr, line 52: @csams would you mind to add a test for your examples in the docstrings?

insights/parsers/rpm_ostree_status.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+from .. import parser, CommandParser, YAMLParser
+from ..parsr.query import from_dict
+from insights.specs import Specs
+
+
+@parser(Specs.rpm_ostree_status)
+class RpmOstreeStatus(CommandParser, YAMLParser):
+    """
+    Class ``RpmOstreeStatus`` parses the output of the ``rpm-ostree status --json`` command.
```
bfahr, line 74: The command and output appear to be JSON, but you are using the YAMLParser instead of the JSONParser.  Is that intentional?

bfahr (changes requested):

insights/parsers/rpm_ostree_status.py
csams: Yep, I just used yaml since it's a superset of json.

insights/parsers/tests/test_rpm_ostree_status.py
csams: Done

PR #3030, opened 2021-04-20
[parser] GCP parser to gather license codes from instances

SteveHNH:
This is needed to identify marketplace systems in the GCP cloud.

### DIFF
--- a/insights/parsers/gcp_license_codes.py
+++ b/insights/parsers/gcp_license_codes.py
@@ -29,6 +29,7 @@ class GCPLicenseCodes(CommandParser):
 
     Raises:
         SkipException: When content is empty or no parse-able content.
+        ParseException: When the json is unable to be parsed
 
     Attributes:
         ids (list): A list containing the IDs found on the instance
--- a/insights/parsers/gcp_license_codes.py
+++ b/insights/parsers/gcp_license_codes.py
@@ -1,0 +1,60 @@
+"""
+GCPLicenseCodes
+===============
+
+This parser reads the output of a command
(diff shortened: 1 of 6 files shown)


insights/parsers/gcp_license_codes.py
```diff
@@ -0,0 +1,60 @@
+"""
+GCPLicenseCodes
+===============
```
bfahr, line 3: @SteveHNH please include the `.rst` file in the directory `insights/docs/share_parsers_catalog/` so that this new parser will be included in the documentation build.

insights/parsers/gcp_license_codes.py
```diff
@@ -0,0 +1,60 @@
...
+    ``curl -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/licenses/?recursive=True"``,
+
+
+    Typical Output of this command is::
+
+        [{"id": "601259152637613565"}]
+
+    Raises:
+        SkipException: When content is empty or no parse-able content.
+
```
bfahr, line 32: Since you are making changes could you add ParseException to this section of the doc since you raises it as well? Thx.

bfahr (changes requested):

insights/parsers/gcp_license_codes.py
SteveHNH: Ah! i keep forgetting that part. on it

insights/parsers/gcp_license_codes.py
SteveHNH: oh snap. yep

bfahr (approved):
Approved pending approval of the spec.

PR #3032, opened 2021-04-21
Fix ValueError exception in Lvm parser.

ryan-blakley:
* Fix issue were leaked file descriptors messages in the output caused
exceptions.
* Add test for parsing with the leaked file descriptor messages.


bfahr:
This PR will fix #3034

PR #3033, opened 2021-04-22
New parser pmrep command

rasrivas-redhat:

### DIFF
--- a/insights/parsers/pmrep.py
+++ b/insights/parsers/pmrep.py
@@ -13,11 +13,11 @@ Sample ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.colli
 Examples:
     >>> type(pmrep_doc_obj)
     <class 'insights.parsers.pmrep.PMREPMetrics'>
-    >>> pmrep_doc_obj.data[1]
+    >>> pmrep_doc_obj[1]
     {'network.interface.out.packets-lo': '1.000'}
-    >>> pmrep_doc_obj.data[4]
+    >>> pmrep_doc_obj[4]
     {'network.interface.collisions-eth0': '4.000'}
-    >>> pmrep_doc_obj.data[5]
+    >>> pmrep_doc_obj[5]
     {'swap.pagesout': '5.000'}
 """
 
@@ -33,7 +33,6 @@ from insights.parsers import SkipException, ParseException
 class PMREPMetrics(CommandParser, list):
     """Parses output of ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout -o csv`` command."""
     def parse_content(self, content):
-        self.data = []
         temp = {}
         if not content or len(content) == 1:
             raise SkipException("There is no data in the table")
@@ -42,10 +41,9 @@ class PMREPMetrics(CommandParser, list):
         except Exception:
             raise ParseException("The content isn't in csv format")
         for row in reader:
-            self.append(row)
-
-        for k, v in dict(row).items():
-            temp[k] = dict(row)[k]
-            self.data.append(temp)
-            temp = {}
-        return self.data
+            # check the "swap.pagesout" metric value is empty, if the value is empty for this key, will be empty for the remaining keys except the "Time" key
+            if row["swap.pagesout"]:
+                for k, v in dict(row).items():
+                    temp[k] = dict(row)[k]
+                    self.append(temp)
+                    temp = {}
--- a/insights/parsers/pmrep.py
+++ b/insights/parsers/pmrep.py
@@ -10,7 +10,7 @@ Sample ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.colli
              eth0             lo            eth0              lo
           count/s        count/s         count/s         count/s     count/s
               N/A            N/A             N/A             N/A         N/A
-            1.997          0.000           0.000           0.000       0.000
+            1.000          2.000           3.000           4.000       5.000
 
 Examples:
     >>> type(pmrep_doc_obj)
--- a/insights/parsers/pmrep.py
+++ b/insights/parsers/pmrep.py
@@ -38,7 +38,6 @@ class PMREPMetrics(CommandParser, list):
         if not content or len(content) == 1:
             raise SkipException("There is no data in the table")
         try:
-            # keep the line break for yaml parse in some table
             reader = DictReader(os.linesep.join(content).splitlines(True))
         except Exception:
             raise ParseException("The content isn't in csv format")
--- a/insights/parsers/pmrep.py
+++ b/insights/parsers/pmrep.py
@@ -13,11 +13,11 @@ Sample ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.colli
 Examples:
     >>> type(pmrep_doc_obj)
     <class 'insights.parsers.pmrep.PMREPMetrics'>
-    >>> pmrep_doc_obj = sorted(pmrep_doc_obj, key=lambda x: sorted(x.keys())[0])
+    >>> pmrep_doc_obj = sorted(pmrep_doc_obj, key=lambda x: x['name'])
     >>> pmrep_doc_obj[1]
-    {'name': 'network.interface.out.packets-lo', 'value': '1.000'}
-    >>> pmrep_doc_obj[4]
     {'name': 'network.interface.collisions-eth0', 'value': '4.000'}
+    >>> pmrep_doc_obj[4]
+    {'name': 'network.interface.out.packets-lo', 'value': '1.000'}
     >>> pmrep_doc_obj[5]
     {'name': 'swap.pagesout', 'value': '5.000'}
 """
--- a/insights/parsers/pmrep.py
+++ b/insights/parsers/pmrep.py
@@ -1,80 +1,52 @@
 """
-Pmrep - command ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout``
-==============================================================================================================
+Pmrep - command ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout -o csv``
+=====================================================================================================================
 
-Parse the content of the ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout`` command.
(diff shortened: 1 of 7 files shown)


insights/parsers/pmrep.py
```diff
@@ -0,0 +1,80 @@
...
+class PMREPMetrics(Parser):
+    """Parses output of ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout`` command."""
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("Empty content.")
+
+        heading_keys = ["n.i.o.packets", "n.i.collisions", "s.pagesout"]
+        ignore_values = ["count/s", "N/A"]
+        header_value = []
+        self.data = {}
```
xiangce, line 41: Since the `self.data` is not a clear attribute to indicate the data type of the parser, it's not recommended anymore. Please make this parser a `list` of `dicts` by inheriting from `list` directly.

For ex:
- https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/satellite_postgresql_query.py

insights/parsers/pmrep.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@parser(Specs.pmrep_metrics)
+class PMREPMetrics(Parser):
+    """Parses output of ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout`` command."""
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("Empty content.")
+
+        heading_keys = ["n.i.o.packets", "n.i.collisions", "s.pagesout"]
```
xiangce, line 38: The [parse_fixed_table](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/__init__.py#L282) would be helpful to simplify the logic of this parse_content.

rasrivas-redhat: @xiangce - I think using the "parse_fixed_table" method will not be helpful in this case because this method overrides the column name when we have multiple columns with the same name. Below is the example of it:

```
  n.i.o.packets  n.i.o.packets  n.i.collisions  n.i.collisions  s.pagesout
           eth0             lo            eth0              lo
        count/s        count/s         count/s         count/s     count/s
            N/A            N/A             N/A             N/A         N/A
          1.000          2.000           3.000           4.000       5.000
```

- Using the "parse_fixed_table" method, as we have "n.i.o.packets" and "n.i.o.packets" column name twice, so it overides for the "eth0" row and only uses for "lo".
```
[{'n.i.o.packets': 'lo', 'n.i.collisions': 'lo', 's.pagesout': ''}, {'n.i.o.packets': 'count/s', 'n.i.collisions': 'count/s', 's.pagesout': 'count/s'}, {'n.i.o.packets': 'N/A', 'n.i.collisions': 'N/A', 's.pagesout': 'N/A'}, {'n.i.o.packets': '2.000', 'n.i.collisions': '4.000', 's.pagesout': '5.000'}]
```

insights/parsers/pmrep.py
rasrivas-redhat: Ack, I will make this parser as a list of dicts.

insights/parsers/pmrep.py
```diff
@@ -0,0 +1,80 @@
...
+Sample ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout`` command output::
+
+    n.i.o.packets  n.i.o.packets  n.i.collisions  n.i.collisions  s.pagesout
+             eth0             lo            eth0              lo
+          count/s        count/s         count/s         count/s     count/s
+              N/A            N/A             N/A             N/A         N/A
+            1.997          0.000           0.000           0.000       0.000
+
+Examples:
+    >>> type(pmrep_doc_obj)
```
xiangce, line 16: Please use the test data in the above sample in the exmples.

rasrivas-redhat: Updated the test data.

xiangce (commented):
After going through the manual page of this `pmrep` command, I think it would be better to use the `-o csv` option to output the result in CSV format
```
            csv
              Print metrics in CSV format (subject to formatting
              options).
```

https://man7.org/linux/man-pages/man1/pmrep.1.html

I will output the full name of each column and make the parser simpler.

see https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/satellite_postgresql_query.py#L75

rasrivas-redhat:
@xiangce - Thanks for the suggestions, the output data looks much simpler when using the `-o csv` option, will update the parser accordingly.

rasrivas-redhat:
@xiangce - I have updated the code as per the "-o csv" command output, please review. Thanks.

insights/parsers/pmrep.py
```diff
@@ -0,0 +1,51 @@
...
+from insights import parser, CommandParser
+from insights.specs import Specs
+from insights.parsers import SkipException, ParseException
+
+
+@parser(Specs.pmrep_metrics)
+class PMREPMetrics(CommandParser, list):
+    """Parses output of ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout -o csv`` command."""
+    def parse_content(self, content):
+        self.data = []
```
xiangce, line 36: With inheriting the parser from `list`, the `self.data` can be removed. since the real data is `appended` to the list itself in L45.

insights/parsers/pmrep.py
```diff
@@ -0,0 +1,51 @@
...
+        temp = {}
+        if not content or len(content) == 1:
+            raise SkipException("There is no data in the table")
+        try:
+            reader = DictReader(os.linesep.join(content).splitlines(True))
+        except Exception:
+            raise ParseException("The content isn't in csv format")
+        for row in reader:
+            self.append(row)
```
xiangce, line 45: Better re-construct the data here.

rasrivas-redhat: updated the code

insights/parsers/pmrep.py
rasrivas-redhat: updated the code

insights/specs/insights_archive.py
```diff
@@ -187,6 +187,7 @@ class InsightsArchiveSpecs(Specs):
     pcp_metrics = simple_file("insights_commands/curl_-s_http_..127.0.0.1_44322.metrics_--connect-timeout_5")
     pcs_quorum_status = simple_file("insights_commands/pcs_quorum_status")
     pcs_status = simple_file("insights_commands/pcs_status")
+    pmrep_metrics = simple_file("insights_commands/pmrep_-t_1s_-T_1s_network.interface.out.packets_network.interface.collisions_swap.pagesout -o csv")
```
xiangce, line 190: it should be `insights_commands/pmrep_-t_1s_-T_1s_network.interface.out.packets_network.interface.collisions_swap.pagesout_-o_csv`

you can use mangle command to get the correct one:

```
# mangle "pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout -o csv"
insights_commands/pmrep_-t_1s_-T_1s_network.interface.out.packets_network.interface.collisions_swap.pagesout_-o_csv
```

xiangce (approved):

insights/specs/insights_archive.py
rasrivas-redhat: Thanks for noticing, it was missed while updating it, updated it now.

insights/parsers/pmrep.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@parser(Specs.pmrep_metrics)
+class PMREPMetrics(CommandParser, list):
+    """Parses output of ``pmrep -t 1s -T 1s network.interface.out.packets network.interface.collisions swap.pagesout -o csv`` command."""
+    def parse_content(self, content):
+        if not content or len(content) == 1:
+            raise SkipException("There is no data in the table")
+        try:
+            reader = DictReader(os.linesep.join(content).splitlines(True))
```
bfahr, line 40: @rasrivas-redhat what's the purpose of this line?  The content has already been split into a list of lines in Linux so you will end up with the same results won't you?

bfahr (changes requested):

insights/parsers/pmrep.py
rasrivas-redhat: @bfahr - This line creates a key, value pair, for this case it creates metrics key and its respective values pair which is much easier to handle in the code.

insights/parsers/pmrep.py
```diff
@@ -0,0 +1,63 @@
...
+        This uses the :py:func:`insights.parsers.keyword_search` function for
+        searching; see its documentation for usage details. If no search
+        parameters are given, no rows are returned.
+
+        Returns:
+            list: A list of dictionaries of rows that match the given
+            search criteria.
+
+        Examples:
+            >>> pmrep_doc_obj.search(network_interface_out_packets_lo__default='1.000')
```
xiangce, line 58: This `search` does not make the query simpler instead the user should know the "name" in detail.

Better reconstruct the `row` to the following format:
```json
{
  'name': 'network.interface.out.packets-lo',
  'value': '0.100'
}
```
by using the following logic:
```python
        for k, v in dict(list(reader)[-1]).items():
            self.append(dict(name=k, value=v))
```

Then the `search` could be more reasonable:

```python
pmrep_doc_obj.search(name__endswith('lo')
```

xiangce (changes requested):

insights/parsers/pmrep.py
rasrivas-redhat: @xiangce - I have updated the logic and added the "endswith" matchers logic to the keyword_search method. Please check.

insights/parsers/tests/test_pmrep.py
```diff
@@ -0,0 +1,68 @@
...
+""".strip()
+
+
+PMREPMETRIC_EMPTY_DATA = """
+""".strip()
+
+
+def test_pmrep_info():
+    pmrep_table = PMREPMetrics(context_wrap(PMREPMETRIC_DATA))
+    pmrep_table = sorted(pmrep_table, key=lambda x: sorted(x.keys())[0])
```
xiangce, line 31: The sort function should be changed since the key is changed.

```pmrep_table = sorted(pmrep_table, key=lambda x: x['name'])`

xiangce (approved):

PR #2603, opened 2020-05-22
[WIP] Converts the time to initialize from mins to secs

psachin:
A very crude implementation. Just in case the parser fails to convert timestring
in 'min'.

Signed-off-by: Sachin Patil <psachin@redhat.com>

PR #3041, opened 2021-04-27
Fix ValueError exception in systemd_analyze parser

### LINKED ISSUE #3040: Systemd_analyze parser is raising lots of exceptions in production
The SystemdAnalyzeBlame parser is throwing a large number of the exception ValueError('too many values to unpack (expected 2)',) in production.

ryan-blakley:
* The parser wasn't setup to parse all of the possible column amounts,
it was only setup to expect two columns, so seconds only. But the
command can have up to 8 columns of output, so I updated it to parse
all of the columns properly.
* Added all of the possible columns to the test file.
* Fix #3040

### DIFF
--- a/insights/parsers/systemd_analyze.py
+++ b/insights/parsers/systemd_analyze.py
@@ -55,10 +55,34 @@ class SystemdAnalyzeBlame(CommandParser, dict):
             raise SkipException
 
         for c in content:
-            time, service = c.split()
-            if time.endswith('ms'):
-                _time = round(float(time.strip('ms')) / 1000, 5)
-            else:
-                _time = round(float(time.strip('ms')), 5)
+            cols = c.split()
+            # Check to make sure that the first character of the first
+            # entry is a number. This will hopefully exclude any errors
+            # that are outputted in the file.
+            if cols[0][0].isdigit():
+                # The service should be the last column, so just
+                # remove the last column from the list before looping.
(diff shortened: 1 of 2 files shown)


insights/parsers/systemd_analyze.py
```diff
@@ -54,11 +54,33 @@ def parse_content(self, content):
...
+
         for c in content:
-            time, service = c.split()
-            if time.endswith('ms'):
-                _time = round(float(time.strip('ms')) / 1000, 5)
-            else:
-                _time = round(float(time.strip('ms')), 5)
+            service = None
+            time = 0
+            for x in c.split():
```
bfahr, line 63: Could this code be simplified?  Does it make sense to use float/round on numbers that are ints?

bfahr (changes requested):

insights/parsers/systemd_analyze.py
ryan-blakley: I simplified it down and removed the unneeded rounds, and switched the floats to ints where needed.

psachin:
Hey @ryan-blakley I see that you are enhancing the parser. Can you please verify if we can close https://github.com/RedHatInsights/insights-core/pull/2603 ?

ryan-blakley:
Hey @psachin yeah it looks like we can close #2603.

PR #3043, opened 2021-04-27
Fix ValueError exception in dmsetup parser

### LINKED ISSUE #3042: The dmsetup parser is raising ValueError exceptions in production
The DmsetupStatus parser is throwing a large number of the exception ValueError('not enough values to unpack (expected 2, got 1)',) in production.

ryan-blakley:
* The parser wasn't setup to parse the output if there weren't any
  devices found.
* Added test for "No devices found" output.
* Fix #3042

### DIFF
--- a/insights/parsers/dmsetup.py
+++ b/insights/parsers/dmsetup.py
@@ -189,6 +189,9 @@ class DmsetupStatus(CommandParser, list):
 
     def parse_content(self, content):
         self.unparseable_lines = []
+        if content[0].lower() == "no devices found":
+            return
+
         for line in content:
             _device_name, _device_info_str = line.rsplit(':', 1)
             device_name = _device_name.strip()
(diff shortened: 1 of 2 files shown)


insights/parsers/dmsetup.py
```diff
@@ -190,7 +190,11 @@ class DmsetupStatus(CommandParser, list):
     def parse_content(self, content):
         self.unparseable_lines = []
         for line in content:
```
bfahr, line 192: @ryan-blakley since the content would only have "No devices found" would it make sense to check explicitly for that string and return if found?  That would leave the list empty, and the `unparsable_lines` list empty as well.  Seems like that would reflect the output more accurately.

bfahr (changes requested):

insights/parsers/dmsetup.py
ryan-blakley: @bfahr I updated it to return if no devices found is in the content, and updated the test to check for empty list.

PR #3046, opened 2021-04-28
fix(compliance): RHICOMPL-1111 Use ID instead of hostname

akofink:
This changes all API calls to Compliance to use the host ID, retrieved
from the Inventory API via machine_id, rather than the hostname. This
fixes the use case when a user has updated the display name using the
client's --display-name option.

### DIFF
--- a/insights/client/apps/compliance/__init__.py
+++ b/insights/client/apps/compliance/__init__.py
@@ -2,7 +2,6 @@ from glob import glob
 from insights.client.archive import InsightsArchive
 from insights.client.connection import InsightsConnection
 from insights.client.constants import InsightsConstants as constants
-from insights.client.utilities import determine_hostname
 from logging import getLogger
 from platform import linux_distribution
 from re import findall
@@ -22,10 +21,10 @@ class ComplianceClient:
     def __init__(self, config):
         self.config = config
         self.conn = InsightsConnection(config)
-        self.hostname = determine_hostname()
         self.archive = InsightsArchive(config)
 
     def oscap_scan(self):
+        self.inventory_id = self._get_inventory_id()
         self._assert_oscap_rpms_exist()
         initial_profiles = self.get_initial_profiles()
         matching_os_profiles = self.get_profiles_matching_os()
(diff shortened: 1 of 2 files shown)


insights/client/apps/compliance/__init__.py
```diff
@@ -154,3 +153,6 @@ def _assert_oscap_rpms_exist(self):
             if len(rpm.strip().split('\n')) < len(REQUIRED_PACKAGES):
                 logger.error('Missing required packages for compliance scanning. Please ensure the following packages are installed: {0}\n'.format(', '.join(REQUIRED_PACKAGES)))
                 exit(constants.sig_kill_bad)
+
+    def _get_inventory_id(self):
+        self.conn._fetch_system_by_machine_id()[0]['id']
```
akofink, line 158: WIP for error handling

vkrizan (approved):
Awesome!

insights/client/apps/compliance/__init__.py
akofink: @gravitypriest (or @SteveHNH) `_fetch_system_by_machine_id` was clearly meant to be private (Python's soft-private leading underscore), but I see other places in the code using this method publically. I just wanted to check with you that this is OK to use.

insights/tests/client/apps/test_compliance.py
```diff
@@ -232,3 +235,17 @@ def test_build_oscap_command_append_tailoring_path(config):
...
+    compliance_client.conn._fetch_system_by_machine_id = lambda: []
+    with raises(SystemExit):
+        compliance_client._get_inventory_id()
+
+    compliance_client.conn._fetch_system_by_machine_id = lambda: [{}]
+    with raises(SystemExit):
+        compliance_client._get_inventory_id()
+
+    compliance_client.conn._fetch_system_by_machine_id = lambda: [{'id': '12345'}]
+    assert compliance_client._get_inventory_id() == '12345'
```
akofink, line 251: I decided to test this private method, since I've mocked it out of all other tests.

insights/client/apps/compliance/__init__.py
gravitypriest: Yep, no worries there. It will return an HTTP response body so if it works for you purposes, by all means. I think this could feasibly be a public method anyway.

akofink:
@gravitypriest any idea what's going on with the build?

gravitypriest:
I believe there are some pipeline issues right now w/ openstack

gravitypriest (approved):

gravitypriest:
Merged with failing checks because 2.6 pipelines are failing. Confirmed working on an alternate system.

PR #3047, opened 2021-04-29
Enhance docstring installrpm

wushiqinlou:
Sometimes, developers forgot to check "epoch" while it is not "null", then the rpm comparison would get unexpected result. Add this docstring to emphasize this point


wushiqinlou:
@xiangce - Please have a review.

PR #3048, opened 2021-04-29
add ruamel.yaml to verifier

aleccohan:
add ruamel.yaml to verifier contrib file.


gravitypriest (approved):

gravitypriest:
Merged with failing checks because 2.6 pipelines are failing. Confirmed working on an alternate system.

PR #3051, opened 2021-05-06
Fix issues in rhel6/python26 CI

bfahr:
* This fix uses a new python26 image with required modules included
* This goes with [this image](https://github.com/SteveHNH/jenkins-s2i-example/pull/4)


lhuett (approved):
Approved

PR #3056, opened 2021-05-10
Fix uname parser for when a debug kernel is being used

### LINKED ISSUE #2709: uname parser fails to parse if the kernel version ends with "debug" keyword
When using uname parser, we get a condition error as per RAT report when the kernel version ends with "debug" keyword

Example: When we run uname -a command, the kernel version is `3.10.0-514.el7.x86_64.debug` and its ends with "debug"  keyword

```
Linux system.example.com 3.10.0-514.el7.x86_64.debug #1 SMP Wed Oct 19 11:30:11 EDT 2016 x86_64 x86_64 x86_64 GNU/Linux
```

Condition error as per the RAT report:
```
condition_error 7 '<' not supported between instances of 'str' and 'int' TypeError
```

ryan-blakley:
* Debug kernels append .debug to the end of the kernel name in the uname
  output, so I updated the parser to check and remove the appended
  .debug because it causes exceptions.
* Fixes #2709

### DIFF
--- a/insights/parsers/tests/test_uname.py
+++ b/insights/parsers/tests/test_uname.py
@@ -11,6 +11,7 @@ UNAME1 = "Linux foo.example.com 2.6.32-504.el6.x86_64 #1 SMP Tue Sep 16 01:56:35
 UNAME2 = "Linux rhel7box 3.10.0-229.el7.x86_64 #1 SMP Mon Mar 3 13:32:45 EST 2014 x86_64 x86_64 x86_64 GNU/Linux"
 UNAME3 = "Linux map1a 2.6.18-53.el5PAE #1 SMP Wed Oct 10 16:48:18 EDT 2007 i686 i686 i386 GNU/Linux"
 UNAME4 = "Linux cvlvtsmsrv01 3.10.0-229.el7.x86_64 #1 SMP Thu Jan 29 18:37:38 EST 2015 x86_64 x86_64 x86_64 GNU/Linux"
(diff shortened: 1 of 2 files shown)


insights/parsers/uname.py
```diff
@@ -333,6 +333,9 @@ def parse_nvr(cls, nvr, data=None, arch=True):
             data = dict(cls.defaults)
 
         data['version'], data['release_arch'] = nvr.split('-', 1)
+        # Debug kernels have .debug appended to the end, so remove it before continuing.
+        if data['release_arch'].endswith('.debug'):
+            data['release_arch'] = '.'.join(data['release_arch'].split('.')[:-1])
```
bfahr, line 338: @ryan-blakley I think it makes sense to create a new attribute, maybe `debug` or `debug_kernel` and set it based on whether this is present.  Also I think you could use `.rsplit('.', 1)[0]` instead to avoid having to rejoin the parts back together.

insights/parsers/tests/test_uname.py
```diff
@@ -83,10 +85,13 @@ def test_uname():
     # Test the equality and inequality operators
     assert uname1 != uname2
     assert uname2 == uname4
+    assert uname2 == uname4_debug
     assert uname2 > uname1
     assert uname4 >= uname3
+    assert uname4_debug >= uname3
```
bfahr, line 91: If you add the debug attribute please add pos/neg tests for it as well.

bfahr (changes requested):

insights/parsers/uname.py
ryan-blakley: @bfahr Added the new attribute, and switched to rsplit.

insights/parsers/tests/test_uname.py
ryan-blakley: @bfahr added new tests to test weather the new attribute was true or false.

PR #3059, opened 2021-05-12
feat: tell the user the largest file in the archive if the upload is too big (WIP)

gravitypriest:


insights/client/connection.py
```diff
@@ -802,6 +804,25 @@ def register(self):
...
             return (message, client_hostname, "None", "")
 
+    def _archive_too_big(self, archive_file):
+        '''
+        Some helpful messaging for when the archive is too large for ingress
+        '''
+        archive_filesize = size_in_mb(
+            os.stat(archive_file).st_size)
+        logger.info("Archive is {fsize} MB which is larger than the maximum allowed size of {flimit} MB.".format(
+            fsize=archive_filesize, flimit="100"))
```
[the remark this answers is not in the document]
subpop: Yes, I believe it is just a well-accepted value that is enforced by ingress. There's probably an equivalent hard-coded value somewhere in ingress's code base.

insights/client/connection.py
```diff
@@ -802,6 +804,25 @@ def register(self):
...
+
+        if not self.config.core_collect:
+            logger.error("Cannot estimate the spec with largest filesize because core collection is not enabled. "
+                    "Enable core collection by setting core_collect=True in %s, and attempt the upload again.", self.config.conf)
+            return
+
+        biggest_file = largest_spec_in_archive(archive_file)
+        if biggest_file is not None:
+            logger.info("The largest file in the archive is %s at %s MB.", biggest_file[0], size_in_mb(biggest_file[1]))
+            logger.info("Please add this spec:\n\n\t%s\n\nto your denylist configuration according to the documentation, and try the upload again.", biggest_file[2])
```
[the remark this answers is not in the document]
subpop: Good question. I don't know what documentation it's referring to. Maybe we can point the user to the file path rather than the ambiguous "the documentation".

clarabez:
Hi, @gravitypriest!
We've identified that setup pipeline is failing for RHEL6 on "QE Pull Request Testing", but it is working as expected for RHEL79, RHEL84 and RHEL90.

This issue has been reported and is being analyzed here: [RHCLOUD-17392](https://issues.redhat.com/browse/RHCLOUD-17392).

insights/client/utilities.py
```diff
@@ -445,3 +447,41 @@ def os_release_info():
...
+
+
+def largest_spec_in_archive(archive_file):
+    logger.info("Checking for large files...")
+    tar_file = tarfile.open(archive_file, 'r')
+    largest_fsize = 0
+    # get the name of the archive
+    for elems in archive_file.split("/"):
+        if "tar" in elems:
+            name = elems.split(".", 1)
```
strider, line 459: Is `archive_file` an absolute path to the archive?

If yes, it would better to use `os.path` instead:

`name = os.path.basename(archive_file).split(".")[0]`

insights/client/utilities.py
```diff
@@ -445,3 +447,41 @@ def os_release_info():
...
+    for file in tar_file.getmembers():
+        if metadata_top in file.name:
+            file_extract = tar_file.extractfile(file.name)
+            specs_metadata = json.load(file_extract)
+            results = specs_metadata.get("results", [])
+            if not results:
+                continue
+            if type(results) != list:
+                results = [results]
+            for result in results:
```
strider, line 474: It would be better to use `isinstance` builtin instead of type here:

`if isinstance(results, list):`

strider:
I think those three following commits could be squashed in one:
- https://github.com/RedHatInsights/insights-core/pull/3059/commits/486468352e6f9fe9f31eae94ee2cce2791e9d894
- https://github.com/RedHatInsights/insights-core/pull/3059/commits/9e38c7cdff93abdd2eba6fa39c82ef34857503d1
- https://github.com/RedHatInsights/insights-core/pull/3059/commits/d1082b214a1a8c4bfa17e9612bb48cce413b72a4

They are touching the same part of the code, adding tests and fixed things like typos etc ...

insights/client/utilities.py
```diff
@@ -445,3 +447,41 @@ def os_release_info():
...
+
+def largest_spec_in_archive(archive_file):
+    logger.info("Checking for large files...")
+    tar_file = tarfile.open(archive_file, 'r')
+    largest_fsize = 0
+    # get the name of the archive
+    for elems in archive_file.split("/"):
+        if "tar" in elems:
+            name = elems.split(".", 1)
+    name[0]
```
strider, line 460: I think this is a leftover or something else. But I guess you can get rid of this line.

strider (changes requested):
Looks good to me in the overall. Some improvements in my comments might be interesting to address though.

insights/client/utilities.py
ahitacat: Yes, it is

insights/client/utilities.py
ahitacat: Yes, it is a leftover

insights/client/utilities.py
strider: Sorry, @ahitacat. I've misled you about this. Actually the archive name is always have this format `insights-$FQDN-$TIMESTAMP.tar.gz` and the FQDN contains dots :/ 

It should be:

`name = os.path.basename(archive_file).split('.tar.gz')[0]`

strider (approved):
Looks good to me. We've peer-reviewed it together and tests are successful.

Good to go.

strider:
@subpop, @Glutexo Could you please review this one? ;-)

PR #3061, opened 2021-05-13
Update spec pmrep metrics

wushiqinlou:


insights/parsers/pmrep.py
```diff
@@ -54,7 +54,7 @@ def search(self, **kwargs):
 
         Examples:
             >>> pmrep_doc_obj.search(name__endswith='lo')
-            [{'name': 'network.interface.out.packets-lo', 'value': '1.000'}, {'name': 'network.interface.collisions-lo', 'value': '3.000'}]
+            [{'name': 'network.interface.collisions-lo', 'value': '3.000'}, {'name': 'network.interface.out.packets-lo', 'value': '1.000'}]
```
xiangce, line 57: Since the py2 and py3 have different sort policies, you'd better re-sort it in the test. See L16

wushiqinlou: @xiangce - Updated in line 61, please have a check.

insights/parsers/pmrep.py
```diff
@@ -58,4 +58,4 @@ def search(self, **kwargs):
             >>> pmrep_doc_obj.search(name__endswith='swap.pagesout')
             [{'name': 'swap.pagesout', 'value': '5.000'}]
         """
-        return keyword_search(self, **kwargs)
+        return sorted(keyword_search(self, **kwargs), key=lambda x: x['name'])
```
xiangce, line 61: @wushiqinlou - It's not a good idea to sort the result inner the `search`. Although it is caused by the different sort-policies in py2 and py3, it should be fixed in the test too.

Please re-sort the result in the test of the example of the search again.

xiangce (approved):

PR #3063, opened 2021-05-14
Add parser umask

wushiqinlou:


wushiqinlou:
insights-core-assets MR: https://gitlab.cee.redhat.com/insights-release-eng/insights-core-assets/-/merge_requests/230

insights/parsers/umask.py
```diff
@@ -0,0 +1,49 @@
...
+    Sample content::
+
+        0022
+
+    Examples:
+
+        >>> type(umask_obj)
+        <class 'insights.parsers.umask.Umask'>
+        >>> umask_obj['value']
+        '0022'
```
bfahr, line 29: A dictionary really doesn't make sense here, at least the way you have used it.  Why not use `umask -S` and create keys for `user`, `group` and `other`, and then have a tuple or list of the settings for that key?  The way it is currently implemented it simply a copy of the spec.

insights/parsers/umask.py
```diff
@@ -0,0 +1,49 @@
...
+
+        "0022"
+
+    Raises:
+        SkipException: When contents are empty or invalid
+
+    """
+
+    def parse_content(self, content):
+
```
bfahr, line 42: There should be a docstring here to replace the default docstring.  A one-liner is sufficient.

bfahr (changes requested):

insights/parsers/umask.py
wushiqinlou: @bfahr - Updated

insights/specs/default.py
```diff
@@ -992,6 +992,7 @@ def is_mod_loaded_for_ss(broker):
     tomcat_vdc_fallback = simple_command("/usr/bin/find /usr/share -maxdepth 1 -name 'tomcat*' -exec /bin/grep -R -s 'VirtualDirContext' --include '*.xml' '{}' +")
     tuned_adm = simple_command("/usr/sbin/tuned-adm list")
     udev_fc_wwpn_id_rules = simple_file("/usr/lib/udev/rules.d/59-fc-wwpn-id.rules")
+    umask = simple_command("/usr/bin/umask")
```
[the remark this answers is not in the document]
wushiqinlou: @bfahr - Updated, please have a check.

insights/parsers/umask.py
```diff
@@ -0,0 +1,49 @@
...
+from insights.parsers import SkipException
+from insights.specs import Specs
+
+
+@parser(Specs.umask)
+class Umask(Parser, dict):
+    """
+    Parser for ``/usr/bin/umask`` file.
+
+    Sample input is provided in the *Examples*.
```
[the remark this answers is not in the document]
wushiqinlou: @bfahr - Updated, please have a check.

insights/parsers/umask.py
wushiqinlou: @bfahr - Updated, please have a check.

xiangce (approved):
Looks good to me

wushiqinlou:
since insights client executes commands as root but without the shell, the umask collected by the client is always "u=rwx,g=rwx,o=rwx" (0000).  As insights-client can not collect accurate "umask" when root is logged in, we have to abandon this candidate. Close this MR

wushiqinlou:
Close this PR as the result of "su -l root" is unexpected and dangerous.

Issue #3065, opened 2021-05-14
The insights archive spec in #3061 also needs to be updated for the change to pmrep_metrics spec


bfahr:
@wushiqinlou in your PR #3061 that was merged you also need to update the [insights archive spec](https://github.com/RedHjatInsights/insights-core/blob/57ab772ea8eb9f3061d6a356ea78bcef07ecdc4f/insights/specs/insights_archive.py#L190) to either a glob spec, or a first_file spec to collect all of the data.

wushiqinlou:
Fixed MR: https://github.com/RedHatInsights/insights-core/pull/3068

xiangce:
@bfahr - Thanks for the double check. and Sorry for missing this point.

PR #3068, opened 2021-05-17
Update pmrep_metrics spec in insights_archive.py

wushiqinlou:
This is used to fix issue https://github.com/RedHatInsights/insights-core/issues/3065

### DIFF
--- a/insights/specs/insights_archive.py
+++ b/insights/specs/insights_archive.py
@@ -187,7 +187,7 @@ class InsightsArchiveSpecs(Specs):
     pcp_metrics = simple_file("insights_commands/curl_-s_http_..127.0.0.1_44322.metrics_--connect-timeout_5")
     pcs_quorum_status = simple_file("insights_commands/pcs_quorum_status")
     pcs_status = simple_file("insights_commands/pcs_status")
-    pmrep_metrics = simple_file("insights_commands/pmrep_-t_1s_-T_1s_network.interface.out.packets_network.interface.collisions_swap.pagesout_-o_csv")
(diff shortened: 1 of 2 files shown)


xiangce (approved):
Looks good.

insights/specs/insights_archive.py
```diff
@@ -187,7 +187,7 @@ class InsightsArchiveSpecs(Specs):
     pcp_metrics = simple_file("insights_commands/curl_-s_http_..127.0.0.1_44322.metrics_--connect-timeout_5")
     pcs_quorum_status = simple_file("insights_commands/pcs_quorum_status")
     pcs_status = simple_file("insights_commands/pcs_status")
-    pmrep_metrics = simple_file("insights_commands/pmrep_-t_1s_-T_1s_network.interface.out.packets_network.interface.collisions_swap.pagesout_-o_csv")
+    pmrep_metrics = simple_file("insights_commands/pmrep_-t_1s_-T_1s_network.interface.out.packets_network.interface.collisions_swap.pagesout_mssql.memory_manager.stolen_server_memory_mssql.memory_manager.total_server_memory_-o_csv")
```
bfahr, line 190: @wushiqinlou in order to support any existing clients that haven't updated you need to make this backward compatible by using a `first_file` spec.  That will ensure that all of the clients, whether they are updated or now, will work.

bfahr (changes requested):

insights/specs/insights_archive.py
wushiqinlou: @bfahr - Updated, please have a check.

PR #3071, opened 2021-05-17
Make testing work with TOX

bfahr:
Testing with tox.


bfahr:
Since fixing this will require some refactoring of doctests to fix dictionary comparison, I"m going to close this for now and I'll return to it later.

PR #3072, opened 2021-05-18
New parser for /etc/sysconfig/grub

qinpingli:


insights/parsers/sysconfig.py
```diff
@@ -630,3 +633,31 @@ class IfCFGStaticRoute(SysconfigOptions):
     def parse_content(self, content):
         self.static_route_name = self.file_name.split("route-", 1)[1]
         super(IfCFGStaticRoute, self).parse_content(content)
+
+
+@parser(Specs.sysconfig_grub)
+class GrubSysconfig(SysconfigOptions):
+    """
+    Class to parse the ``/etc/sysconfig/grub``
```
xiangce, line 641: Please add a short explanation that this parser is for `/etc/default/grub`, and the `/etc/sysconfig/grub` is a symlink points to  `/etc/default/grub`.

insights/specs/default.py
```diff
@@ -950,6 +950,7 @@ def is_mod_loaded_for_ss(broker):
     subscription_manager_installed_product_ids = simple_command("/usr/bin/find /etc/pki/product-default/ /etc/pki/product/ -name '*pem' -exec rct cat-cert --no-content '{}' \;")
     swift_object_expirer_conf = first_file(["/var/lib/config-data/puppet-generated/swift/etc/swift/object-expirer.conf", "/etc/swift/object-expirer.conf"])
     swift_proxy_server_conf = first_file(["/var/lib/config-data/puppet-generated/swift/etc/swift/proxy-server.conf", "/etc/swift/proxy-server.conf"])
+    sysconfig_grub = simple_file("/etc/sysconfig/grub")
```
xiangce, line 953: better collect the original /etc/default/grub file, since the /etc/sysconfig/grub is a symlink to it, in case the symlink is broken.

insights/specs/sos_archive.py
```diff
@@ -274,6 +274,7 @@ class SosSpecs(Specs):
     swift_log = first_file(["/var/log/containers/swift/swift.log", "/var/log/swift/swift.log"])
     sys_kernel_sched_features = simple_file("/sys/kernel/debug/sched_features")
     sysconfig_chronyd = simple_file("/etc/sysconfig/chronyd")
+    sysconfig_grub = simple_file("/etc/sysconfig/grub")
```
xiangce, line 277: For `simple_file`, if the path is the same as it is on the real system, the definition in `default.py` works for all the archives (sos, insights archive) since the archives keep their original path. This line can be removed.

insights/specs/default.py
xiangce: `first_file` can be used in this case to indicate that they are the same

```
sysconfig_grub = first_file(["/etc/default/grub", "/etc/sysconfig/grub"])
```

xiangce (changes requested):

qinpingli:
@xiangce Thanks for the review. Updated the PR per your comments.

xiangce (approved):
LGTM. @bfahr, please double-check. Thanks.

bfahr (approved):
This looks good to me, @lhuett would you please review?

lhuett (approved):
Looks good.

PR #3077, opened 2021-05-19
Fix ValueError exception in docker_list parser

### LINKED ISSUE #3076: The docker_list parser is raising ValueError exceptions in production.
The DockerList parser is throwing a number of the exception ValueError("Line containing 'REPOSITORY' was not found in table",), and ValueError("Line containing 'CONTAINER' was not found in table",) in production.

ryan-blakley:
* In epel 6 there is a another package named docker, that's a kde system
  tray. So when the docker specs run they just gather the help output.
  So I added a check for the help output so it can be skipped.
* Added a test with the help output.
* Fixes #3076


xiangce (approved):
Looks good to me.

PR #3079, opened 2021-05-20
Updated spec tuned_conf

crackcodecamp:
The spec `tuned_conf` spec.

### DIFF
--- a/insights/specs/sos_archive.py
+++ b/insights/specs/sos_archive.py
@@ -265,7 +265,7 @@ class SosSpecs(Specs):
     testparm_s = simple_file("sos_commands/samba/testparm_s")
     tomcat_web_xml = first_of([glob_file("/etc/tomcat*/web.xml"),
                                   glob_file("/conf/tomcat/tomcat*/web.xml")])
-    tuned_conf = simple_file("/etc/tuned.conf")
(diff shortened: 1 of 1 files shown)


crackcodecamp:
@xiangce Please review.

insights/specs/sos_archive.py
```diff
@@ -265,7 +265,7 @@ class SosSpecs(Specs):
     testparm_s = simple_file("sos_commands/samba/testparm_s")
     tomcat_web_xml = first_of([glob_file("/etc/tomcat*/web.xml"),
                                   glob_file("/conf/tomcat/tomcat*/web.xml")])
-    tuned_conf = simple_file("/etc/tuned.conf")
+    tuned_conf = glob_file(["/etc/tuned.conf", "/usr/lib/tuned/*/tuned.conf"])
```
xiangce, line 268: @crackcodecamp - The `glob_file` requires the spec to be set as `multiple_output=True` which results in wrapping the parser results to a list and breaking all the existing rules depends on it.

You may need to add another spec and parser for the new files, (better in gss repo?)

xiangce (changes requested):

insights/specs/sos_archive.py
crackcodecamp: @xiangce Okay, Closed this PR.

Issue #2997, opened 2021-03-30
The Scheduler Parser is missing documentation


bfahr:
[This parser](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/scheduler.py) is old enough there is no documentation and no tests for the documentation.

bfahr:
Fixed by PR #3083

PR #2996, opened 2021-03-30
Add scheduler spec also for Insights archive

skontar:

PR #3083, opened 2021-05-21
Update parser for device schedulers

skontar:
Signed-off-by: Stanislav Kontar <skontar@redhat.com>


skontar:
Followup for https://github.com/RedHatInsights/insights-core/pull/2996

kgrant-rh (approved):

insights/parsers/scheduler.py
```diff
@@ -1,17 +1,50 @@
...
-                    result[self.file_path.split('/')[3]] = active_scheduler.group()
-        self.data = result
+            r = self.ACTIVE_SCHEDULER_PATTERN.search(line)
+            if r:
+                self.active_scheduler = r.group(1)
+
+            self.schedulers = line.replace('[', '').replace(']', '').split()
+
+        # Legacy values
+        self.data = {self.device: '[' + self.active_scheduler + ']'}
```
jsvob, line 50: If this is for legacy access, I'd expect it to be either of:

- The raw file contents (e.g. `noop deadline [cfq]`)
- The active scheduler (e.g. `cfq`)

What is the reason the brackets are put around `active_scheduler`?

jsvob (commented):
LGTM once the single review comment is explained/resolved.

insights/parsers/scheduler.py
skontar: This is to not break backwards compatibility. See old tests and old code.

jsvob: 🤦 My bad.

jsvob (approved):

insights/parsers/scheduler.py
```diff
@@ -1,17 +1,50 @@
-from .. import parser, get_active_lines, Parser
+"""
+Scheduler - file ``/sys/block/*/queue/scheduler``
+=================================================================
```
jobselko, line 3: The length of `=` characters should be the same as the length of the module name line (for the documentation generation purpose).

insights/parsers/scheduler.py
```diff
@@ -1,17 +1,50 @@
...
+
 import re
+
 from insights.specs import Specs
+from .. import parser, get_active_lines, Parser
 
 
 @parser(Specs.scheduler)
 class Scheduler(Parser):
+    ACTIVE_SCHEDULER_PATTERN = re.compile(r'\[(.*)]')
```
bfahr, line 34: @skontar can you please add a minimal docstring here, since you have full docs in the top of the module.  Also can you please document the attributes you are adding to the class as well as the `data` attribute?

insights/parsers/tests/test_scheduler.py
```diff
@@ -31,3 +31,32 @@ def test_scheduler_deadline():
...
+    assert r.device == 'vda'
+    assert r.schedulers == ['mq-deadline', 'kyber', 'none']
+    assert r.active_scheduler == 'mq-deadline'
+
+    # RHEL 8
+    r = scheduler.Scheduler(context_wrap('[mq-deadline] kyber bfq none',
+                                         '/sys/block/vda/queue/scheduler'))
+    assert r.device == 'vda'
+    assert r.schedulers == ['mq-deadline', 'kyber', 'bfq', 'none']
+    assert r.active_scheduler == 'mq-deadline'
```
bfahr, line 62: Would you mind to add a test for the examples in the docstring?  Here's an [example](https://github.com/RedHatInsights/insights-core/blob/e72f426b75abfbefd020692669006163652563e1/insights/parsers/tests/test_abrt_ccpp.py#L83).

bfahr (changes requested):

bfahr:
Fixes #2997

PR #3084, opened 2021-05-21
Add new switch to run to choose the color output

### LINKED ISSUE #2980: Add an option to force colored output to insights run command
When "pipping" the output of `insights run` to `less`, the colors are missing, even using `less -r`.

Several linux commands has the ability to force a colored output with `--color=always`, for example:
~~~
git diff --color=always | less -r
ls -l --color=always | less -r
grep test . --color=always | less -r 
~~~

Add an option to insights to force the color even when the output is "pipped".

ryan-blakley:
* Add new switch --color[=WHEN] to allow the choice of color output
  encoding. The options for when are always, auto, and never. Always
  sets colorama to not strip the encoding, so the color can be piped.
  Auto sets colorama to the default, and never disables color encoding.
* Fixes #2980

### DIFF
--- a/docs/manpages/insights-run.rst
+++ b/docs/manpages/insights-run.rst
@@ -39,6 +39,12 @@ OPTIONS
     -c CONFIG --config CONFIG
         Configure components.
 
+    \-\-color [=WHEN]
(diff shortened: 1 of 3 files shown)


lhuett (approved):
Looks good to me.

PR #3085, opened 2021-05-27
New parser swap mm process usage

rasrivas-redhat:


insights/specs/default.py
```diff
@@ -948,6 +948,46 @@ def is_mod_loaded_for_ss(broker):
     sssd_config = simple_file("/etc/sssd/sssd.conf")
     subscription_manager_id = simple_command("/usr/sbin/subscription-manager identity")  # use "/usr/sbin" here, BZ#1690529
     subscription_manager_installed_product_ids = simple_command("/usr/bin/find /etc/pki/product-default/ /etc/pki/product/ -name '*pem' -exec rct cat-cert --no-content '{}' \;")
+
+    @datasource(Ps, HostContext)
+    def proc_swap_memory(broker):
+        ps = broker[Ps]
+        pids = [p for p in sorted(os_listdir("/proc/")) if p.isdigit()]
```
xiangce, line 955: Better get the `PID` from the Ps combiner directly, so you do not need to list the whole `"/proc"` dir and iterate the Ps again to get the "COMMAND".

insights/specs/default.py
```diff
@@ -948,6 +948,46 @@ def is_mod_loaded_for_ss(broker):
...
+        data = []
+        for pid_id in pids:
+            with open("/proc/{0}/smaps".format(pid_id), "r") as fi:
+                vars = fi.read()
```
xiangce, line 959: how about `readlines`, so do not need to split it in the following loop.

xiangce (changes requested):
@rasrivas-redhat - As we discussed yesterday, I think we can implement this parser a bit generic, by making it supporting all the Keys contained in the `/proc/{PID}/smaps` files,  in the future, we do not need to add more spec&parsers when want to extract other keys for it..

insights/specs/default.py
rasrivas-redhat: Ack, updated the code

insights/specs/default.py
rasrivas-redhat: Ack, updated the code

insights/specs/default.py
```diff
@@ -948,6 +948,46 @@ def is_mod_loaded_for_ss(broker):
...
+                                swap[k] = swap[k] + int(v.split()[0])
+                            else:
+                                continue
+                    if swap:
+                        data.append(dict(pid=int(pid_id), det=swap))
+            except Exception:
+                continue
+
+        smaps_data = []
+        for proc in ps:
```
xiangce, line 980: Is this `loop` essential? I think it could be done inner the above loop of `L956`

insights/specs/default.py
```diff
@@ -948,6 +948,46 @@ def is_mod_loaded_for_ss(broker):
...
+            pid_id = p_id["PID"]
+            try:
+                with open("/proc/{0}/smaps".format(pid_id), "r") as fp:
+                    swap = {}
+                    for line in fp.readlines():
+                        if not len(line):
+                            continue
+                        else:
+                            k, v = line.split(":", 1)
+                            if "00000000" in k or k == "VmFlags" or line[0].isdigit():
```
xiangce, line 966: The first part of this line looks hex number, it's incorrect to check it with `isdigit()` since the `[a-f].isdigit()` returns `False` but it's a valid hex.

xiangce: If `sum`ed these values up in the parser, some of the information will be lost.

For example, it's impossible for a rule to get the `swap` used in the specific range of the virtual memory section (eg. `7f4656856000-7f4656857000 r--p 00007000 fd:00 268516389                  /usr/lib64/libcrypt-2.17.so`) of a PID.


and the `VmFlags` should not be skiped.

insights/specs/default.py
```diff
@@ -948,6 +948,46 @@ def is_mod_loaded_for_ss(broker):
...
+        data = []
+        for p_id in ps:
+            pid_id = p_id["PID"]
+            try:
+                with open("/proc/{0}/smaps".format(pid_id), "r") as fp:
+                    swap = {}
+                    for line in fp.readlines():
+                        if not len(line):
+                            continue
+                        else:
```
xiangce, line 964: This `else` is not necessary, since you `continue` in the `if`.

insights/specs/default.py
```diff
@@ -948,6 +948,46 @@ def is_mod_loaded_for_ss(broker):
...
+                            k, v = line.split(":", 1)
+                            if "00000000" in k or k == "VmFlags" or line[0].isdigit():
+                                continue
+                            elif k not in swap:
+                                swap[k] = int(v.split()[0])
+                            elif k in swap:
+                                swap[k] = swap[k] + int(v.split()[0])
+                            else:
+                                continue
```
xiangce, line 973: This `else continue` is not necessary too.

insights/specs/default.py
```diff
@@ -948,6 +948,46 @@ def is_mod_loaded_for_ss(broker):
...
+                    if swap:
+                        data.append(dict(pid=int(pid_id), det=swap))
```
xiangce, line 975:
```
                       swap.update(dict(PID=int(pid), COMMAND=proc['COMMAND']))
                       smaps_data.append(swap)
```

xiangce (changes requested):

rasrivas-redhat:
@xiangce - Thanks for the feedback. As per the latest discussion with Jaylin, he is not in favor of developing this spec, he has mentioned his concerns in the comment. Apart from it, I am having a discussion with a related KCS SME about his suggestion for the best possible solution we can provide, I think tell that time we can hold on this MR.

bfahr:
@rasrivas-redhat @xiangce I see this change is on hold, but it also will need to be changed after #3088 is merged.  I will hold off on my review until this PR is no longer on hold, and any necessary changes have been made.

bfahr:

rasrivas-redhat:
@bfahr - Thanks for sharing the information and will follow the new method whenever required. As per the discussion with the team, this spec is no more needed for the rule, so I am closing the MR.

PR #3086, opened 2021-05-27
New spec to get satellite version on satellite client

huali027:

### DIFF
--- a/docs/custom_datasources_index.rst
+++ b/docs/custom_datasources_index.rst
@@ -10,3 +10,11 @@ insights.specs.datasources.cloud_init
     :members: cloud_cfg, LocalSpecs
     :show-inheritance:
     :undoc-members:
+
+insights.specs.datasources.get_satellite_api_base_url
+-----------------------------------------------------
+
+.. automodule:: insights.specs.datasources.get_satellite_api_base_url
+    :members: base_url_of_satellite_api
+    :show-inheritance:
+    :undoc-members:
--- a/docs/shared_parsers_catalog/satellite_version_from_client.rst
+++ b/docs/shared_parsers_catalog/satellite_version_from_client.rst
@@ -1,0 +1,3 @@
+.. automodule:: insights.parsers.satellite_version_from_client
+   :members:
+   :show-inheritance:
--- a/insights/parsers/satellite_version_from_client.py
+++ b/insights/parsers/satellite_version_from_client.py
@@ -1,0 +1,61 @@
+"""
+SatelliteVersionFromClient - command ``curl -k https://satellite.com:port/katello/api/status``
+==============================================================================================
+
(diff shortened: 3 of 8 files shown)


insights/parsers/satellite_version_on_client.py
```diff
@@ -0,0 +1,53 @@
...
+class SatelliteVersionOnClient(JSONParser):
+    """
+    Parse the output of ``curl -k https://satellite.com:port/katello/api/status``.
+
+    Sample output::
+
+        {"version":"3.18.1.22","timeUTC":"2021-05-20 13:59:20 UTC"}
+
+    Attributes:
+        satellite_version (string): The satellite version which the client connects to.
```
xiangce, line 33: Since the name of the parser contains "SatelliteVersion" already, how about renaming this attribute as "version"? and add "major", "minor" just like the SatelliteVersion Combiner?
https://github.com/RedHatInsights/insights-core/blob/62dc4062b8808478aecfe737074dec7c83d50eba/insights/combiners/satellite_version.py#L106-L107

huali027: @xiangce, I updated it, and also combine the parser in the combiner SatelliteVersion. Please help to review it again.

insights/combiners/satellite_version.py
```diff
@@ -54,7 +55,7 @@ def _parse_sat_version(version):
     return [major, minor]
 
 
-@combiner(InstalledRpms, optional=[Sat6Ver])
+@combiner([InstalledRpms, SatelliteVersionFromClient], optional=[Sat6Ver])
```
xiangce, line 58: I believe you should not put the `SatelliteVersionFromClient` here, since the original `SatelliteVersion` combiner is used to indicate the host is a `"Satellite Server"`. After adding this as optional param, the client of Satellite Server will also be treated as a Sat Server, It will break mand existing rules

xiangce (changes requested):

insights/parsers/satellite_version_on_client.py
xiangce: @huali027 - no you should not combine this parser to the SatelliteVersion.

insights/combiners/satellite_version.py
huali027: @xiangce, yes, right, I forgot some rules used it for satellite server only. I'll revert the related code.

bfahr:
@huali027 @xiangce this PR will need to be modified after #3088 is merged.  I will hold off on my review until those changes have been made.

bfahr:

huali027:
@xiangce  please help to review

huali027:
We can use the spec "branch_info" to get the version info, I'll close it.

Issue #3087, opened 2021-06-02
Testing needs to be added for custom datasources


bfahr:
Most custom datasources have been added without proper tests.  Tests need to be added for all custom datasources.

PR #3088, opened 2021-06-02
Refactor custom datasources into their own package

bfahr:
This enhancement PR moves all of the custom datasources and associated code out of `insights/specs/default.py` and moves them in the `insights/specs/datasources` package.  This cleans up and simplifies `default.py` and makes it so that only specs with registry entries (data that will be kept in an archive) are maintained in `default.py`.

The default manifest for core-collection in `insights/collect.py` has also been cleaned up to ensure that no duplicate entires are added in the YAML.

Documentation has been updated to include a new section for Custom Datasources.

I've opened issue #3087  to add testing for all custom datasources.


bfahr:
Due to the complexity of this change I'm going to close this PR and submit separate PRs for each datasource.

PR #3092, opened 2021-06-03
Update parser "httpd_M" and "httpd_V" to return right binary path

huali027:
* Also add args interface for rule testing in InputData and Context

### DIFF
--- a/insights/parsers/httpd_M.py
+++ b/insights/parsers/httpd_M.py
@@ -85,9 +85,7 @@ class HttpdM(LegacyItemAccess, CommandParser):
     @property
     def httpd_command(self):
         """
-        str: The full path of a running httpd. An Empty string when nothing
-        is found.  It's to identify which httpd binaries the instance run with.
+        Return the full binary path of a running httpd or None when nothing
+        is found. It's to identify which httpd binaries the instance run with.
         """
-        # Typical `file_path` of HttpdM looks like: '/usr/sbin/httpd_-M'
-        # Remove the trailing '_-M'
-        return self.file_path[:-3] if self.file_path else ''
+        return self.args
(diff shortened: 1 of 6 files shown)


huali027:
@xiangce, please help to review

xiangce (approved):
It looks good to me.
@bfahr - As mentioned in the sync meeting, this PR also modifies the `context_wrap` to keep the `args` (e.g for `foreach_execute`) to make sure the tests of the `HttpdV/HttpdM` being pass. Please have a review when free. Thanks.

insights/parsers/httpd_M.py
```diff
@@ -85,9 +85,7 @@ def parse_content(self, content):
...
         """
-        str: The full path of a running httpd. An Empty string when nothing
-        is found.  It's to identify which httpd binaries the instance run with.
+        Return the full binary path of a running httpd or None when nothing
+        is found. It's to identify which httpd binaries the instance run with.
         """
-        # Typical `file_path` of HttpdM looks like: '/usr/sbin/httpd_-M'
-        # Remove the trailing '_-M'
-        return self.file_path[:-3] if self.file_path else ''
+        return self.args
```
xiangce, line 91: @bfahr  - Just FYI,

The original method returns the `file_path` of the file in the `insights-archives`, and looks like `insights_commands/httpd_-M` (which removed the ".usr.sbin" by the `mangle`) but the other spec `package_provides_command` keeps the `raw path`  of the httpd binary which looks like `/usr/sbin/httpd`.

To make sure the rule could find the correct package of the running httpd, @huali027 made this change to let this method return the same `raw path` as the `package_provides_command`.

see: 
https://github.com/RedHatInsights/insights-core/blob/bee093a5f5967d00204e002be5c5b71b91a11d04/insights/specs/default.py#L691-L707

https://github.com/RedHatInsights/insights-core/blob/bee093a5f5967d00204e002be5c5b71b91a11d04/insights/specs/default.py#L484-L496

lhuett (approved):
Looks good

PR #3098, opened 2021-06-08
Host and datetime parsing for ForemanSSLAccessLog

Rohan21Lobo:
Signed-off-by: Rohan Arora <roarora@redhat.com>

### All Pull Requests:

### Issue/Bug Fix:
Provide complete details of the issue and how this PR fixes the issue. You can link to one or more existing open, publicly-accessible issue(s) that provide details.

### Enhancement:
Provide complete details of the enhancement.

Log format refferred from documentation https://httpd.apache.org/docs/2.4/logs.html

### DIFF
--- a/insights/parsers/foreman_log.py
+++ b/insights/parsers/foreman_log.py
@@ -25,6 +25,9 @@ ProxyLog - file ``/var/log/foreman-proxy/proxy.log``
 SatelliteLog - file ``/var/log/foreman-installer/satellite.log``
 ----------------------------------------------------------------
 
+ForemanSSLAccessLog - file ``/var/log/httpd/foreman-ssl_access_ssl.log``
+------------------------------------------------------------------------
+
 """
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


xiangce (commented):
@Rohan21Lobo - Your update looks good. But this spec is still not added to the "uploader.json" which means it cannot be collected by the old collection. Rules will not hit, If it's used in Adviosor rules. But it's OK to use it in the GSS rules.

Rohan21Lobo:
> @Rohan21Lobo - Your update looks good. But this spec is still not added to the "uploader.json" which means it cannot be collected by the old collection. Rules will not hit, If it's used in Adviosor rules. But it's OK to use it in the GSS rules.

@xiangce works for me, This change is anyways needed for a gss rule.

xiangce (changes requested):
@Rohan21Lobo - Thanks for the response.

To make sure the core could locate the file successfully from the sos-report archives, please add the spec (simple_file) to the sos_archive.py file.

xiangce (approved):
@Rohan21Lobo - My mistake, it's already added to the sos_archive.py

PR #3099, opened 2021-06-09
Add parser for spec "/etc/candlepin/broker.xml"

huali027:

### All Pull Requests:

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -35,7 +35,7 @@ from insights.parsers.lsmod import LsMod
 from insights.combiners.satellite_version import SatelliteVersion, CapsuleVersion
 from insights.parsers.mount import Mount
 from insights.specs import Specs
-from insights.specs.datasources import cloud_init, ps as ps_datasource
+from insights.specs.datasources import cloud_init, ps as ps_datasource, candlepin_broker
 import datetime
 
 
@@ -232,6 +232,7 @@ class DefaultSpecs(Specs):
     cinder_conf = first_file(["/var/lib/config-data/puppet-generated/cinder/etc/cinder/cinder.conf", "/etc/cinder/cinder.conf"])
     cinder_volume_log = first_file(["/var/log/containers/cinder/volume.log", "/var/log/containers/cinder/cinder-volume.log", "/var/log/cinder/volume.log"])
     cloud_cfg = cloud_init.cloud_cfg
+    candlepin_broker = candlepin_broker.candlepin_broker
     cloud_init_custom_network = simple_file("/etc/cloud/cloud.cfg.d/99-custom-networking.cfg")
     cloud_init_log = simple_file("/var/log/cloud-init.log")
     cluster_conf = simple_file("/etc/cluster/cluster.conf")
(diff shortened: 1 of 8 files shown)


huali027:
@xiangce, I found if we want to provide a common interface to allow adding more filters later, it is also complicated, because we have to add namespace before all tags or attributes. So at last I just remove the namespace and then remove the sensitive data before saving it. In this way, we won't be bothered by the namespace. Please help to review if it's okay to do it in this way. For the spec, the namespace is useless.

bfahr:
@xiangce @huali027 this PR needs to wait until #3088 is merged, and it will then require changes to the datasource.  I have not yet reviewed this PR will do so after those changes have been made.

bfahr:

huali027:
@bfahr, sure, I'll update it.

huali027:
@xiangce please help to review

insights/specs/default.py
```diff
@@ -230,6 +231,7 @@ def is_ceph_monitor(broker):
     cinder_conf = first_file(["/var/lib/config-data/puppet-generated/cinder/etc/cinder/cinder.conf", "/etc/cinder/cinder.conf"])
     cinder_volume_log = first_file(["/var/log/containers/cinder/volume.log", "/var/log/containers/cinder/cinder-volume.log", "/var/log/cinder/volume.log"])
     cloud_cfg = cloud_init.cloud_cfg
+    candlepin_broker = candlepin_broker.candlepin_broker
```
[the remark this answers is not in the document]
huali027: @xiangce, sure, I'll update the MR, thanks.

insights/tests/datasources/test_candlepin_broker.py
```diff
@@ -0,0 +1,170 @@
...
+        <cluster-user />
+    </core>
+</configuration>
+""".strip()
+
+RELATIVE_PATH = '/etc/candlepin/broker.xml'
+
+
+def test_candlepin_broker():
+    simple_file.content = CANDLEPIN_BROKER.splitlines()
```
bfahr, line 144: @huali027 this needs to be changed.  Its also wrong in cloud_init.  This may actually cause some side effects to the base `simple_file` object.  You can either create a new simple_file object and then set the content, or just create a new class that only has a content attribute.  You could also use a mock class.

insights/tests/datasources/test_candlepin_broker.py
```diff
@@ -0,0 +1,170 @@
...
+    result = candlepin_broker(broker)
+    assert result is not None
+    assert isinstance(result, DatasourceProvider)
+    expected = DatasourceProvider(content=CANDLEPIN_BROKER_XML.splitlines(), relative_path=RELATIVE_PATH)
+    assert result.content == expected.content
+    assert result.relative_path == expected.relative_path
+
+
+def test_candlepin_broker_bad():
+    simple_file.content = CANDLEPIN_BROKER_BAD.splitlines()
```
bfahr, line 155: Same comment as above.

insights/tests/datasources/test_candlepin_broker.py
```diff
@@ -0,0 +1,170 @@
...
+    broker = {LocalSpecs.candlepin_broker_input: simple_file}
+    with pytest.raises(SkipComponent) as e:
+        candlepin_broker(broker)
+    assert 'Unexpected exception' in str(e)
+
+
+def test_candlpin_broker_no_sensitive_info():
+    simple_file.content = CANDLEPIN_BROKER_NO_SENSITIVE_INFO.splitlines()
```
bfahr, line 163: Same as above.

bfahr (changes requested):

bfahr (approved):
@huali027 you need to rebase this PR to remove the conflicts, but the PR is otherwise approved.

PR #3102, opened 2021-06-15
Add parser ansible_tower_custom

wushiqinlou:

### All Pull Requests:

### Issue/Bug Fix:
Provide complete details of the issue and how this PR fixes the issue. You can link to one or more existing open, publicly-accessible issue(s) that provide details.

### Enhancement:
Provide complete details of the enhancement.


insights/parsers/ansible_tower_config_custom.py
```diff
@@ -0,0 +1,44 @@
...
+AnsibleTowerConfigCustom - file ``/etc/tower/conf.d/custom.py``
+===============================================================
+The AnsibleTowerConfigCustom class parses the file ``/etc/tower/conf.d/custom.py``.
+"""
+from insights import parser, get_active_lines, Parser, LegacyItemAccess
+from insights.core.filters import add_filter
+from insights.specs import Specs
+from insights.parsers import SkipException
+
+add_filter(Specs.ansible_tower_config_custom, ["AWX_CLEANUP_PATHS", "="])
```
bfahr, line 11: @wushiqinlou filters should only be added to a parser if they are necessary for the parser to work correctly.  For instance if you are parsing a INI-type config file with headings like `[section_name]` then you would need to add a filter to the parser `[` to ensure that the config file parser won't raise an exception.  In this case rules using this parser should be adding any necessary filters.  

Also, `=` is not a good filter since it will collect too many lines from the file.

insights/parsers/ansible_tower_config_custom.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@parser(Specs.ansible_tower_config_custom)
+class AnsibleTowerConfigCustom(Parser, LegacyItemAccess):
```
bfahr, line 15: Since this parser will be storing key/value pairs it really should be derived from `dict` instead of `LegacyItemAccess`.  Here is [an example](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/partitions.py) a parser implementing `dict`.

bfahr (changes requested):
@wushiqinlou also please complete the description for your enhancement to insights core.  The description is necessary for us to fully evaluate your PR.

insights/parsers/ansible_tower_config_custom.py
wushiqinlou: @bfahr - In the former code, I checked all of the content without filter. However, per discussion with @xiangce , the file "/etc/tower/conf.d/custom.py" is python file, it is possible that the content format will change, then we will get unexpected result. As the new rule only needs to check option "AWX_CLEANUP_PATHS" now and sensitive information may be configured in such file, we decide to add filter into this parser.

insights/parsers/ansible_tower_config_custom.py
wushiqinlou: @bfahr - Good to me.

xiangce (approved):
Looks good to me now. Approve pending approval of spec.

insights/parsers/ansible_tower_config_custom.py
bfahr: I agree that the spec needs to be filtered, but my recommendation is that the filter should only be added by rules using the parser and not directly by the parser.  Parsers should be general purpose and by adding that filter you are making it specific to a particular rule.  The only parsers that should be adding filters for a filtered spec are those that require certain information to function properly as I described above in the INI file parser example.

wushiqinlou: @bfahr - ACK, removed the filter from the parser.

insights/specs/__init__.py
```diff
@@ -19,6 +19,7 @@ class Specs(SpecSet):
     abrt_ccpp_conf = RegistryPoint(filterable=True)
     abrt_status_bare = RegistryPoint()
     alternatives_display_python = RegistryPoint()
+    ansible_tower_settings = RegistryPoint(filterable=True)
```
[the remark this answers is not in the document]
bfahr: I'll fix this since it was already merged.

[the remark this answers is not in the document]
wushiqinlou: @bfahr - yes, we need to add `multi_output=True`. Thanks for your update!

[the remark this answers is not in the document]
xiangce: @bfahr - Sorry for bringing this trouble.

PR #3118, opened 2021-06-24
force _boolify to always return bool

gravitypriest:
Signed-off-by: Jeremy <jcrafts@redhat.com>

### All Pull Requests:

Check all that apply:

The function for parsing env vars in `insights.client.config` has an nested function `_boolify`. This function will return the same string it is given, if this string is not recognized as True or False. In our logic, this is fine, since strings will evaluate to True, but this will make it a bit cleaner.


gravitypriest:
Never mind, this needs more time in the kitchen.

Glutexo:
lol

PR #3122, opened 2021-06-28
Fix ethtool parse exceptions due to a spec issue

### LINKED ISSUE #1791: Issue in insights_commands/ethtool_-k_* and other ethtool spec
It seems there is a bug in spec creation for `insights_commands/ethtool_-k_*` and other related specs related to **ethtool** command when a **VLAN connection** configured on the host. The spec which is currently created is `ethtool_-i_bond0.123_bond0` which should be actually `ethtool_-i_bond0.123` and due to which we are getting `Cannot get device feature names: No such device` when we open `insights_commands/ethtool_-k_ethtool_-k_bond0.123_bond0` logically this device does not exist on this host. It looks that the interface name is referred from `/sys/class/net` and then referred to ethtool command, but for VLAN interface seems it fails to refer the correct interface name. 

Below is an example for reference.

```
[root@rhel7-box vagrant]# ls insights-rhel7-box-20190403111343/sys/class/net/
bond0/     bond0.123/ enp0s3/    enp0s8/    enp0s9/    lo/
``` 

```
[root@rhel7-box vagrant]# ls insights-rhel7-box-20190403111343/insights_commands/ethtool_-k_
ethtool_-k_bond0            ethtool_-k_bond0.123_bond0  ethtool_-k_enp0s3           ethtool_-k_enp0s8           ethtool_-k_enp0s9
```

ryan-blakley:
* Add new ethernet interfaces datasource, to properly gather a list of
  valid ethernet interface names. That way ethtool is ran against the
  proper interface names.
* Fixes #1791

### All Pull Requests:

Check all that apply:

Switch from just listing the files under the /sys/class/net to run ethtool against, instead parse the ip -o link command for any interfaces with link/ether in it. Also correct the interface name when it is a vlan interface, as it appends the parent interface on the end after an @ symbol. This plus an update to the uploader.json should fix the majority of the ethtool parse exceptions that occur in production.

### DIFF
--- a/insights/specs/datasources/ethernet.py
+++ b/insights/specs/datasources/ethernet.py
@@ -1,0 +1,57 @@
+"""
+Custom datasource for gathering a list of the ethernet interface names.
+"""
+from insights.core.context import HostContext
+from insights.core.dr import SkipComponent
+from insights.core.plugins import datasource
+from insights.core.spec_factory import simple_command
+from insights.specs import Specs
+
+
+class LocalSpecs(Specs):
+    """ Local specs used only by ethernet_interfaces datasource. """
+    ip_link = simple_command("/sbin/ip -o link")
+
+
+@datasource(LocalSpecs.ip_link, HostContext)
+def interfaces(broker):
+    """
+    This datasource provides a list of the ethernet interfaces available.
+
+    Typical content of the spec is::
+
+        1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000\\    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+        2: enp1s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000\\    link/ether 52:54:00:13:14:b5 brd ff:ff:ff:ff:ff:ff
+        3: enp8s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000\\    link/ether 52:54:00:e5:11:d4 brd ff:ff:ff:ff:ff:ff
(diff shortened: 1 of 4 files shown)


insights/tests/datasources/test_ethernet.py
```diff
@@ -0,0 +1,33 @@
...
+5: ib0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 4092 qdisc mq state DOWN group default qlen 256\\    link/infiniband 00:01:02:03:fd:90:0:00:00:00:00:00:ef:0d:8b:02:01:d9:82:fd
+"""
+
+IP_LINK_BAD = ""
+
+EXPECTED = ['enp1s0', 'enp8s0', 'enp1s0.2']
+
+
+def test_ethernet_interfaces():
+    broker = {LocalSpecs.ip_link: context_wrap(IP_LINK.splitlines())}
```
bfahr, line 23: @ryan-blakley instead of using context_wrap here it would be better to use mock or something similar:
```
from mock.mock import Mock
...
ip_link_command = Mock()
ip_link_command.content = "content value"
broker = {LocalSpecs.ip_link: ip_link_command}
```

insights/tests/datasources/test_ethernet.py
```diff
@@ -0,0 +1,33 @@
...
+    result = interfaces(broker)
+    assert result is not None
+    assert isinstance(result, list)
+    assert result == sorted(EXPECTED)
+
+
+def test_ethernet_interfaces_bad():
+    broker = {LocalSpecs.ip_link: context_wrap(IP_LINK_BAD.splitlines())}
```
bfahr, line 31: Same as above.

bfahr (changes requested):

bfahr (approved):

insights/specs/datasources/ethernet.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Raises:
+        SkipComponent: When there is not any content.
+    """
+    content = broker[LocalSpecs.ip_link].content
+    if content:
+        ifaces = []
+        for x in content:
+            # Only process lines that have link/ether, this should exclude non ethernet devices.
+            if re.search("link/ether", x):
```
xiangce, line 51: Nitpick: Since this `search` depends on the built-in library `re` and is a bit heavy, and this check looks quite simple, how about changing it to a simple check like `if "link/ether" in x:`?

ryan-blakley: @xiangce great point, I updated it as you suggested.

xiangce (approved):

PR #3127, opened 2021-06-30
New spec "ipcs -s -i xxx"

huali027:

### All Pull Requests:

Check all that apply:

* [ x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?

Add the spec back since it is used by an existing webserver rule.


xiangce (commented):
Better remove the verb `get_` from the filename and the name of the spec, and use `ipcs` as the filename, `semid` as the name of spec.

insights/specs/datasources/get_semid.py
```diff
@@ -0,0 +1,26 @@
+from insights.core.context import HostContext
```
xiangce, line 1: Please add a title for this module

insights/specs/datasources/get_semid.py
```diff
@@ -0,0 +1,26 @@
...
+from insights.core.plugins import datasource
+from insights.parsers.ipcs import IpcsS
```
xiangce, line 3: If there is possible to get the expected list from an existing spec, do not use the Parser which is heavier.

The existing spec `ipcs_s` is quite simple, and can be used directly for this datasource.

xiangce (changes requested):

insights/collect.py
```diff
@@ -180,6 +180,9 @@
           enabled: true
         - name: insights.components.rhel_version.IsRhel8
           enabled: true
+    # neededfor the 'pre-check' of 'ipcs_s_i' spec
```
xiangce, line 183: By change to base on the `ipcs_s` spec, this pre-check parser is not required.

insights/specs/datasources/ipcs.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    content = broker[LocalSpecs.ipcs_s_cmd].content
+    results = set()
+    for s in content:
+        s_splits = s.split()
+        # key        semid      owner      perms      nsems
+        # 0x00000000 65536      apache     600        1
+        if len(s_splits) == 5 and s_splits[1].isdigit():
+            results.add(s_splits[1])
+    return list(results)
```
xiangce, line 44: How about skipping it when nothing is get from the `ipcs -s`?

xiangce (approved):
Looks good. But it would be better if the datasource `semid` could handle the empty result itself.

Approved pending the approval of spec.

huali027:
> Looks good. But it would be better if the datasource `semid` could handle the empty result itself.
> 
> Approved pending the approval of spec.

Sure, I'll check if it's emtpy, thanks for the review.

xiangce (approved):
Approved pending the approval of Spec

insights/specs/default.py
```diff
@@ -448,7 +448,8 @@ def httpd_cmd(broker):
     ipaupgrade_log = simple_file("/var/log/ipaupgrade.log")
     ipcs_m = simple_command("/usr/bin/ipcs -m")
     ipcs_m_p = simple_command("/usr/bin/ipcs -m -p")
-    ipcs_s = simple_command("/usr/bin/ipcs -s")
+    ipcs_s = ipcs.LocalSpecs.ipcs_s_cmd
```
bfahr, line 451: @huali027 this spec can be left in the original form as a `simple_command`.  There's no need to create the LocalSpec in the datasource.

insights/specs/datasources/ipcs.py
```diff
@@ -0,0 +1,47 @@
...
+from insights.core.spec_factory import simple_command
+from insights.core.plugins import datasource
+from insights.specs import Specs
+from insights.core.dr import SkipComponent
+
+
+class LocalSpecs(Specs):
+    """ Local specs used only by semid datasources """
+
+    ipcs_s_cmd = simple_command("/usr/bin/ipcs -s")
```
bfahr, line 15: There's no need to create this, just reference the spec `insights.specs.Specs.ipcs_s` instead.

bfahr (changes requested):

insights/specs/datasources/ipcs.py
huali027: @bfahr, sure, updated, thanks.

bfahr (approved):

PR #3135, opened 2021-07-02
Add make_none response for rules

### LINKED ISSUE #3026: Return a response on behalf of rules that return None
A rule that returns `None` has `SkipComponent` raised on its behalf, so it is silently ignored. Nothing indicating the skip is logged, and nothing shows up in any of the response documents, which makes debugging and accounting difficult.

The [rule plugin type](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/plugins.py#L294) should return a [Response subclass](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/plugins.py#L372) in these cases so that [execution counts](https://github.com/RedHatInsights/insights-core/blob/master/insights/formats/text.py#L91) can be corrected and [template based formatters](https://github.com/RedHatInsights/insights-core/blob/master/insights/formats/template.py#L56) have something to work with.

bfahr:
### All Pull Requests:

Check all that apply:

* Changes how rules that return none are handled. Instead of raising a
  SkipComponent and ignoring the rule, they will now be counted in the
  results.
* Add a make_none response type
* Update formatters to handle the new type
* Update tests
* Fix #3026

### DIFF
--- a/docs/manpages/insights-run.rst
+++ b/docs/manpages/insights-run.rst
@@ -75,6 +75,9 @@ OPTIONS
     -m --missing
         Show missing requirements.
 
+    -n --none
+        Show rules returning ``None``.
+
     -p PLUGINS --plugins PLUGINS
         Comma-separated list without spaces of package(s) or module(s) containing plugins.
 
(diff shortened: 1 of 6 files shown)


vdymna:
This looks good to me.

PR #3138, opened 2021-07-07
Add new file to haproxy_conf

jobselko:
### All Pull Requests:

Check all that apply:

*Add your description here*


jobselko:
@bfahr Should I also adjust the name of the HaproxyCfg module as three files are collected? In that case, that line would be about 160 characters long. So is there any recommendation on how to proceed? Thank you.

insights/specs/default.py
```diff
@@ -300,7 +300,7 @@ def gfs2_mount_points(broker):
     grubby_default_index = simple_command("/usr/sbin/grubby --default-index")  # only RHEL7 and updwards
     grubby_default_kernel = simple_command("/sbin/grubby --default-kernel")
     hammer_task_list = simple_command("/usr/bin/hammer --config /root/.hammer/cli.modules.d/foreman.yml --output csv task list --search 'state=running AND ( label=Actions::Candlepin::ListenOnCandlepinEvents OR label=Actions::Katello::EventQueue::Monitor )'")
-    haproxy_cfg = first_file(["/var/lib/config-data/puppet-generated/haproxy/etc/haproxy/haproxy.cfg", "/etc/haproxy/haproxy.cfg"])
+    haproxy_cfg = first_file(["/var/lib/config-data/puppet-generated/haproxy/etc/haproxy/haproxy.cfg", "/etc/haproxy/haproxy.cfg", "/etc/opt/rh/rh-haproxy18/haproxy/haproxy.cfg"])
```
ryan-blakley, line 303: @jobselko would the scl haproxy config path need to be in front of the normal config path in case they have both installed?

jobselko: No, the order is not important. I just did not know the list is processed in reverse order.

ryan-blakley: @jobselko Okay cool thank you for confirming that.

jobselko: @ryan-blakley And what about the order? Is it OK?

ryan-blakley: @jobselko The order is fine if the order doesn't matter. My concern was if they have the regular haproxy package installed along with the software collections haproxy package installed. Then it would only parse the regular haproxy package's config file since it's in the list before the software collections haproxy package's config file.

jobselko: @ryan-blakley Now I see a problem. The situation you are describing could occur. But even if we switch the order of that files in the list, the problem is still here. Another solution would be to collect both files (have two different _first_file_ or _use glob_file_).
I will try to find more info tomorrow.

skontar: @jobselko I thought that you should never encounter both, as it is either RHEL-8 where it is available as a normal package, or RHEL-7 where it is SCL. Is there a way to get combination somehow?

jobselko: All config files for both regular and SCL packages should be collected. I adjusted the code.
@ryan-blakley Please review.

jobselko:
@rblakley @bfahr I need to update the HaproxyConf parser to process more files. Moving into WIP.

insights/specs/default.py
ryan-blakley: @skontar on rhel 7 there is a haproxy rpm in the main repo, and there is rh-haproxy18 rpm in the rhscl repo, I just tested to make sure, both can be installed at the same time. 

@jobselko if using glob the parser would need to be updated due to it using multi-collect.

jobselko:
@bfahr @ryan-blakley Please review again.

insights/parsers/haproxy_cfg.py
```diff
@@ -81,6 +90,17 @@ def _parse_content(content):
 
 @parser(Specs.haproxy_cfg)
 class HaproxyCfg(Parser):
-    """Class to parse file ``haproxy.cfg``."""
+    """
```
ryan-blakley, line 93: @jobselko Since both parsers use the _parse_content function. Could you change the _parse_content function to a base class with a parse_content  function that both parsers inherit, that way the parsers can just have pass under them similar to how the ps parser is setup? In addition would it be possible to move the doc string under the new base class?

jobselko: @ryan-blakley Updated. Let me know if I missed anything.

ryan-blakley: @jobselko That looks good, just one thing I see, could you add a pass under the comments on each of the parser classes like the below?

```python
@parser(Specs.haproxy_cfg)
class HaproxyCfg(HaproxyFile):
    """
    Class to parse file ``/var/lib/config-data/puppet-generated/haproxy/etc/haproxy/haproxy.cfg`` or ``haproxy.cfg``.
    """
    pass
```

jobselko: @ryan-blakley According to the pylint, `pass` is not needed, but I will add it.

ryan-blakley: @jobselko Thank you, everything looks good to me.

bfahr (approved):

PR #3139, opened 2021-07-08
Fix SyntaxWarnings

Glutexo:
### All Pull Requests:

Check all that apply:

Comparing string literals by is yields a SyntaxWarning on newer Pythons. This happens in insights.client.connection that uses _is/is not_ to compare with string and number literals. Not only this issues warnings, but is itself incorrect behavior that may lead to bugs. Hence the warnings. Fixed and tested.

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -312,10 +312,10 @@ class InsightsConnection(object):
         for ext in paths:
             try:
                 logger.log(NETWORK, "Testing: %s", test_url + ext)
-                if method is "POST":
+                if method == "POST":
                     test_req = self.session.post(
                         test_url + ext, timeout=self.config.http_timeout, data=test_flag)
-                elif method is "GET":
+                elif method == "GET":
                     test_req = self.session.get(test_url + ext, timeout=self.config.http_timeout)
                 logger.log(NETWORK, "HTTP Status Code: %d", test_req.status_code)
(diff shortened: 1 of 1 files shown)


Glutexo:
There are some failing tests that are unrelated to this change. I was not able to quickly reproduce the failures locally. It may be related to environment variables. Looking into that.

Glutexo:
Oh no! It may be related!!

gravitypriest (approved):

Glutexo:
Removed the tests. Testing this fix is a bigger issue.

First, we don’t want ANY syntax warning, not only those. There is a possibility to configure pytest so that it elevates all warnings to errors. That may be a way to go.

Also, tests for the warnings are neither exactly unit tests, nor integration or functional tests. It’s effectively testing a “compilation”. Mangling with imports in the tests also result in unexpected side effects.

Although I am definitely against merging any untested code, here it may be worth diverging from the rule:

- There are already tests for the affected code. And if not, it is not a matter of this pull request as it only targets the syntax warnings.
- Testing the warnings may require changes to Pytest configuration. That would affect the whole test suite and deserves its own pull request.
- There are other warnings appearing in Python ≥3.8. Any changes that would convert warnings to errors would make this pull request insufficient.

Glutexo:
:shipit: A way to test this beast to follow.

PR #3156, opened 2021-07-21
Add make none ci fix

### LINKED ISSUE #3026: Return a response on behalf of rules that return None

bfahr:
### All Pull Requests:

Check all that apply:

Add make_none response for rules
    
* Changes how rules that return none are handled. Instead of raising a
   SkipComponent and ignoring the rule, they will now be counted in the
    results.
* Add a make_none response type
* Update formatters to handle the new type
* Update tests
* Fix #3026
* Remove none rules from detailed output
* Update to add none option to text format

Fix legacy CI tests that check for None
    
* This change allows the legacy tests that check for None result from a
   rule to work with make_none() without requiring changes to the rules
* The simplified integration testing fixture includes a new flag that
   allows turning switching to the new functionality on a per test basis

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -28,7 +28,7 @@ def run_rule():
     internal support rules that will not be used in the customer facing Insights
     product.
     """
-    def _run_rule(rule, input_data):
+    def _run_rule(rule, input_data, return_make_none=False):
         """
         Fixture for rule integration testing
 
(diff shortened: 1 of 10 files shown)


lhuett (approved):
Looks good to me.

PR #3159, opened 2021-07-22
Fix the issue that cannot find the 'data' of the new archive

### LINKED ISSUE #3158: The validation of ContextProvider for core collection failed
For the `insights-archive` collected by the new Core Collection identified by the `"insights_archive.txt"`, the path of the real collected data is placed in the './data' but not the './' which works for the old archives.

It always reports the following error when `insights-run` against a new archive:

```

  File "/Users/liuxc/work/insights/insights-core/insights/core/plugins.py", line 90, in invoke
    return self.component(broker)
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 569, in __call__
    return self.kind(ctx.locate_path(self.path), root=ctx.root, ds=self, ctx=ctx)
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 175, in __init__
    self.validate()
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 183, in validate
    raise ContentException("%s does not exist." % self.path)
insights.core.plugins.ContentException: /private/tmp/insights-ip-172-31-21-63.us-east-2.compute.internal-20210713080914/etc/redhat-release does not exist.
```

But in fact, the file is placed at 
```
insights-ip-172-31-21-63.us-east-2.compute.internal-20210713080914/data/etc/redhat-release
```

xiangce:

### All Pull Requests:

Check all that apply:

Fix #3158

### DIFF
--- a/insights/core/context.py
+++ b/insights/core/context.py
@@ -153,6 +153,7 @@ class ExecutionContext(six.with_metaclass(ExecutionContextMeta)):
 
     def __init__(self, root="/", timeout=None, all_files=None):
         self.root = root
+        self.data = root
         self.timeout = timeout
         self.all_files = all_files or []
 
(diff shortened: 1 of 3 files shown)


xiangce:
This must not be the best resolution to fix this issue, but I failed to find the best resolution in a short time. Please close it freely, if you can fix it in a better way.

xiangce:
Close this one since the #3183 will fix it in the better way.

PR #3160, opened 2021-07-22
Fix candlepin broker test

Glutexo:
### All Pull Requests:

Check all that apply:

Fixes [ESSNTL-958](https://issues.redhat.com/browse/ESSNTL-958).

_insights/tests/datasources/test_candlepin_broker.py_ contains a circular import causing tests to fail. The problem is that the _insights.specs.candlepin_broker_ module imports _from insights.specs.default_, which itself imports from _insights.specs.candlepin_broker_. Any direct import from _insights.specs.candlepin_broker_ is thus predestined to fail.

(Hot)fixed by importing _candlepin_broker_ from _from insights.specs.default_ instead. Like that, Python handles the circularity well, not failing and making the tests green again. I am not sure whether this is the best solution. The circular reference looks like a bad practice to me.

### DIFF
--- a/insights/tests/datasources/test_candlepin_broker.py
+++ b/insights/tests/datasources/test_candlepin_broker.py
@@ -3,7 +3,7 @@ from mock.mock import Mock
 
 from insights.core.spec_factory import DatasourceProvider
 from insights.core.dr import SkipComponent
-from insights.specs.datasources.candlepin_broker import candlepin_broker, LocalSpecs
+from insights.specs.default import candlepin_broker
 
 
 CANDLEPIN_BROKER = """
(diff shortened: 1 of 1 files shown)


bfahr (approved):

Glutexo:
My guess is it does noc occur if the tests are run in a specific order. That order would cause _insights.specs.default_ to be loaded and cashed by another test, not causing problems later on. But if the Candlepin tests is the first one that tries to load _insights.specs.candlepin_broker_, it fails.

I can’t tell for sure, but we probably don’t have to investigate it precisely. Tests should be independent and _test_candlepin_broker_ failing when run alone definitely is a bug.

Glutexo:
Thanks @bfahr for your review!

PR #3168, opened 2021-07-27
chmod lastupload files to 644

gravitypriest:

### All Pull Requests:

Check all that apply:

After writing the `.lastupload` and `.last-upload-results` file, set the permissions to 644 to disallow global writing.

Using chmod after file write is a little cheap, but the files will already exist in the case of already-registered hosts, and the alternative is that they would need to be deleted and recreated each time with the proper perms, with some umask manipulation. Opted for the simpler move here.

Possible future improvements:
- Expand the `write_to_disk` function to allow for something like a "rewrite" option to fully remove the file before writing, as well as a "permissions" option to allow for calling routines to set write permissions at file creation time
- Remove the `.last-upload-results` file altogether because it no longer provides any useful information. Originally it held the response containing rule hits from the uploads API in classic, but now that ingress handles all uploads, there is nothing written other than the request ID and account number. The `--check-results` service is its spiritual replacement.


subpop (approved):
My first thought was to add this functionality to `write_to_disk` too, but on second thought, I think I prefer the modular (albeit a DRY violation in appearance) use of directly calling `os.chmod`. It's far more obvious what is actually happening here than stuffing it all inside a mysterious function named `write_to_disk` (that actually does some weird file deleting stuff based on function parameters). I'd be in favor or removing `write_to_disk` entirely; the basic file writing API in Python is not so complicated that it's necessary to wrap inside a utility function.

PR #3186, opened 2021-08-13
Update candlepin_broker tests to work with newer python pytest

bfahr:
### All Pull Requests:

Check all that apply:

* [X] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [X] Is this PR to correct an issue?
* [ ] Is this PR an enhancement?

### Complete Description of Additions/Changes:

* This fixes testing issues with python 3.8 and the latest pytest where
  the order of attributes changes in the XML tags.
* The new test doesn't compare everything, but compares most things
  including removal of sensitive tags.

Signed-off-by: Bob Fahr <20520336+bfahr@users.noreply.github.com>

PR #3187, opened 2021-08-14
Update setup.py for multiple versions of python

bfahr:
### All Pull Requests:

Check all that apply:

Update setup.py to support tests for multiple versions of python

* For testing only specify python module version requirements based on the python version

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,10 +77,15 @@ docs = set([
                       # https://github.com/davidhalter/jedi/issues/1714
 ])
 
+# python 2.6 requires setuptools~=36.8.0 to support this syntax
 testing = set([
-    'coverage==4.3.4',
-    'pytest==3.0.6',
-    'pytest-cov==2.4.0',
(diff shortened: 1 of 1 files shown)


bfahr:
This PR requires PR #3186 be merged first to fix issues with the latest pytest on python 3.8.

xiangce (approved):
All tests in py2, py36, and py38 passed on my side. But the test of py39 failed with the following error:
```python
                # remove namespace before save to avoid urgly search
>               for node in root.getiterator():
E               AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getiterator'
```

I think it's another issue that has nothing to do with this PR, See the https://docs.python.org/3.9/whatsnew/3.9.html#removed
we can fix it in another one if py39 is also required.

bfahr:
@xiangce yes that is another problem that will have to be fixed before we can support python 3.9.  There's no replacement for `getiterator` in python 2.6 so it will require some additional code changes to make it work.

PR #3201, opened 2021-08-27
Add parser cups_ppd

wushiqinlou:

### All Pull Requests:

Check all that apply:

This parser is used to parse the content of cups queue file: "/etc/cups/ppd/*"


insights/parsers/cups_ppd.py
```diff
@@ -0,0 +1,52 @@
...
+    Examples:
+        >>> type(cups_ppd)
+        <class 'insights.parsers.cups_ppd.CupsPpd'>
+        >>> cups_ppd["PCFileName"]
+        '"ippeve.ppd"'
+        >>> cups_ppd["cupsFilter2"]
+        ['"application/vnd.cups-pdf application/pdf 10 -"', '"application/vnd.cups-postscript application/postscript 10 -"']
+    """
+    def parse_content(self, content):
+        self.data = {}
```
bfahr, line 38: @wushiqinlou since you are only created a `dict` as the attribute, just go ahead and make the base class `dict` and then you'll get all of the methods for free.  [Here's an example](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/hammer_ping.py) of how you can implement the parser as a `dict`.

bfahr (changes requested):

insights/parsers/cups_ppd.py
wushiqinlou: @bfahr - Updated, please have a check.

insights/specs/__init__.py
```diff
@@ -98,6 +98,7 @@ class Specs(SpecSet):
     crypto_policies_bind = RegistryPoint()
     crt = RegistryPoint()
     current_clocksource = RegistryPoint()
+    cups_ppd = RegistryPoint(multi_output=True)
```
xiangce, line 101: @wushiqinlou - would you please move this spec ahead of the `current_clocksource`, that's to sort them in alphabet order.

insights/parsers/cups_ppd.py
```diff
@@ -0,0 +1,55 @@
...
+        *LanguageVersion: English
+        *LanguageEncoding: ISOLatin1
+        *cupsFilter2: "application/vnd.cups-pdf application/pdf 10 -"
+        *cupsFilter2: "application/vnd.cups-postscript application/postscript 10 -"
+        *cupsFilter2: "application/vnd.cups-banner application/vnd.cups-banner 10 -"
+
+    Examples:
+        >>> type(cups_ppd)
+        <class 'insights.parsers.cups_ppd.CupsPpd'>
+        >>> cups_ppd["PCFileName"]
```
xiangce, line 33: Would you please take the examples in the `Examples` from the above `Sample output`? since I don't see the `PCFileName` from it.

xiangce (changes requested):

insights/specs/__init__.py
wushiqinlou: Updated

insights/parsers/cups_ppd.py
wushiqinlou: Updated

xiangce (approved):
Looks good to me.

PR #3204, opened 2021-08-30
New spec to get if SCA is enabled on satellite

huali027:

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/satellite_postgresql_query.py
+++ b/insights/parsers/satellite_postgresql_query.py
@@ -6,8 +6,10 @@ This module contains the following parsers:
 
 SatelliteAdminSettings - command ``psql -d foreman -c 'select name, value, "default" from settings where name in (\'destroy_vm_on_host_delete\', \'unregister_delete_host\') --csv'``
 -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


insights/parsers/satellite_postgresql_query.py
```diff
@@ -203,3 +205,34 @@ class SatelliteComputeResources(SatellitePostgreSQLQuery):
...
+
+    Examples:
+        >>> type(sat_sca_info)
+        <class 'insights.parsers.satellite_postgresql_query.SatelliteSCAStatus'>
+        >>> sat_sca_info.sca_enabled
+        True
+    """
+
+    @property
+    def sca_enabled(self):
```
xiangce, line 233: Since it's said `If the value of content_access_mode is "org_environment", it means the SCA is enabled for this organization`. 
I think it should be an `attribute` of the `"Organization"` returned by the SQL.

Or, it should be `If one of the organizations is SCA enabled, the Satellite is SCA enabled`, right

xiangce (changes requested):

insights/parsers/satellite_postgresql_query.py
huali027: @xiangce, yes, the latter one is right. Let me update the comment to make it more clear.

xiangce (approved):

Issue #3211, opened 2021-09-10
The httpd conf parser "_HttpdConf" failed to parse some file


huali027:
For one httpd conf file, it contains the following line:
```
RequestHeader set SSL_CLIENT_S_DN ""
RequestHeader set SSL_CLIENT_CERT ""
RequestHeader set SSL_CLIENT_VERIFY ""
```

And the parser failed to parse it with error:
```
    self.parse_content(context.content)
  File "/tmp/insights_core-3.0.8-py3.6.egg/insights/core/__init__.py", line 357, in parse_content
    self.doc = self.parse_doc(content)
  File "/tmp/insights_core-3.0.8-py3.6.egg/insights/combiners/httpd_conf.py", line 355, in parse_doc
    result = self.parse(content)[0]
  File "/tmp/insights_core-3.0.8-py3.6.egg/insights/combiners/httpd_conf.py", line 331, in __call__
    return self.Top(content)
  File "/tmp/insights_core-3.0.8-py3.6.egg/insights/parsr/__init__.py", line 356, in __call__
    raise Exception(err.read())
Exception: At line 31 column 38:
quoted string
    Expected 1 of ['\t', '\n', '\x0b', '\x0c', '\r', ' ', '!', '#', '$', '%', '&', "'", '(', ')', '*', '+', ',', '-', '.', '/', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', ':', ';', '<', '=', '>', '?', '@', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', '[', '\\', ']', '^', '_', '`', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', '{', '|', '}', '~']. Got u'"'.
quoted string
[...]

PR #3218, opened 2021-09-14
Fix bug about some httpd directives may have empty string as attribute

huali027:
Fixes https://github.com/RedHatInsights/insights-core/issues/3211
### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsr/__init__.py
+++ b/insights/parsr/__init__.py
@@ -1240,4 +1240,3 @@ Number = (Lift(_make_number) * Opt(Char("-"), "") * Digits * Opt(Char(".") + Dig
 SingleQuotedString = Char("'") >> String(set(string.printable) - set("'"), "'") << Char("'")
 DoubleQuotedString = Char('"') >> String(set(string.printable) - set('"'), '"') << Char('"')
 QuotedString = Wrapper(DoubleQuotedString | SingleQuotedString) % "quoted string"
-EmptyAttr = String('"\'', min_length=2)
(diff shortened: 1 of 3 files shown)


huali027:
@xiangce, please help to review. The bug is found when I test the httpd ssl cert issue.
Currently, I just check single quote or double quote is used, but not twice of a single quote or double quote. I think if "'\"" is used, it's the wrong format, httpd will fail then. So I didn't check more. Please check if it is okay, or we should do more.

xiangce (approved):

insights/parsr/__init__.py
```diff
@@ -1240,3 +1240,4 @@ def skip_none(x):
 SingleQuotedString = Char("'") >> String(set(string.printable) - set("'"), "'") << Char("'")
 DoubleQuotedString = Char('"') >> String(set(string.printable) - set('"'), '"') << Char('"')
 QuotedString = Wrapper(DoubleQuotedString | SingleQuotedString) % "quoted string"
+EmptyAttr = String('"\'', min_length=2)
```
bfahr, line 1243: @huali027 this seems more specific to the parser you are modifying than something that should be included in `parsr`.  The category of "empty attributes" could cover a lot of different cases.

bfahr (changes requested):

insights/parsr/__init__.py
huali027: @bfahr, besides httpd, nginx is the same. I thought maybe the others have the same configuration to set empty string, so I put it in the `parsr`. Do you suggest to put it to httpd_conf.py first?

bfahr: @huali027 yes, I was suggesting that it be kept in httpd_conf.py instead.  It's still pretty specific even if its both Apache and Nginx config files.  I think that "EmptyAttr" could be interpreted much more broadly and in different ways which would lead to confusion.

huali027: @bfahr, sure, I'll update it to httpd_conf.py.

bfahr (approved):

PR #3231, opened 2021-09-22
preserve alignment in netstat -neopa output in obfuscation

gravitypriest:

### All Pull Requests:

Check all that apply:

SOScleaner is misaligning netstat output with the IP obfuscation, causing false positive rule hits. This tries to address it by padding or removing spaces as needed after the changed IP. Also skip obfuscation of 127.0.0.1 addresses.


skontar:
Off the top of my head I can imagine more reasons why obfuscation of IP addresses may cause issues:

- We may use them in reason / resolution / Ansible playbook
- We use them to detect loopback/private/public sockets
- We use `ip addr` to map wildcard IPs to loopback/private/public

So I believe that any IP address obfuscation has to at least preserve the IP range or meaning (if we ignore the fact that reason / resolution / Ansible playbook may be wrong and potentially wreak havoc on customer system).
For example insights-core/insights/contrib/ipaddress.py should still work without changes.

@kgrant-rh @jsvob , I believe that you have more knowledge to look at this.

skontar:
Also IP address which is obfuscated once, should be obfuscated exactly the same way for all other files/commands, so they can be matched together again.

skontar:
I am thinking, would not it be easier to just obfuscate IP address to a different one, just same length? Then spacing would be preserved and no padding changes would be needed.

kgrant-rh:
I agree, obfuscated IP addresses would ideally still be valid IP addresses after obfuscation. I also agree that some sort of hashing would be ideal so that the same addresses were obfuscated to the same values. To prevent e.g. rainbow table attacks, a different (undisclosed) salt or seed could be used with each iteration. I think the problem of preserving "publicness" through the obfuscation process is intractable. My perspective is that customers choosing to obfuscate data have to realize that analyses depending on that data will be rendered less accurate. My preference would be to avoid false *negatives* in these cases, however, so any sort of transformation should avoid results in standard private use ranges like 10.x.x.x or 192.x.x.x, so that a public address pre-obfuscation doesn't get transformed into a private address.

gravitypriest:
> 
> 
> Also IP address which is obfuscated once, should be obfuscated exactly the same way for all other files/commands, so they can be matched together again.

SOSCleaner currently supports this by keeping a table of matching IPs for the duration of execution.



> 
> 
> I am thinking, would not it be easier to just obfuscate IP address to a different one, just same length (character-wise)? Then spacing would be preserved and no padding changes would be needed.

Would require a significant overhaul of the IP obfuscation algorithm. Current SOScleaner has a "start" address, and just increments by 1 every time a new address is found. There's no control over the length.

Otherwise, I think these are excellent ideas for improvement. Our version of SOScleaner is in need of an update and we're looking at switching to the `sos clean` functionality of the `sos` package, which is more recently maintained and has a much smarter IP obfuscation implementation that I think preserves ranges and subnet relationships.

It's a significant migration effort however, so in the interest of time, would this change set suffice for avoiding the false positive rule hits reported in the issue?

gravitypriest:
test this please

PR #3235, opened 2021-09-24
🔥 Keep lib directory initialization only in client init

Glutexo:
### All Pull Requests:

Check all that apply:

If the lib directory does not exists, it is a sign of a bigger issue, not something we should expect. Kept the directory creation in [client init](https://github.com/RedHatInsights/insights-core/blob/04a9a911b4eda8de3ed753c329f773f5d1ed0300/insights/client/__init__.py#L55) and removed the logic checking for it in other places.


subpop (approved):

gravitypriest:
This was also part of the non-root enhancement (directory creation for those dirs in $HOME). Again, fine to remove, just keep in mind it might come up again one day.

gravitypriest (approved):

PR #3239, opened 2021-09-28
Update the nfnetlink parser

ryan-blakley:
### All Pull Requests:

Check all that apply:

Replace the assert check with if statement and raise a parse exception if met. Updated the parse to inherit from a list, and added the data property for backwards compatibility. Moved the docstring from the top to within the class. Updated the test to look for parse and skip exceptions instead of assertion errors.

### DIFF
--- a/insights/parsers/nfnetlink_queue.py
+++ b/insights/parsers/nfnetlink_queue.py
@@ -1,66 +1,75 @@
 """
 NfnetLinkQueue - file ``/proc/net/netfilter/nfnetlink_queue``
 =============================================================
-
-Reads the ``/proc/net/netfilter/nfnetlink_queue`` file and creates a list of
-dictionaries, one dictionary per row in the file.
-
-The keys of the dictionary are (see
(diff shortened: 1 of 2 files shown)


insights/parsers/nfnetlink_queue.py
```diff
@@ -1,66 +1,83 @@
...
+    Examples:
+
+        >>> # Set up of the environment - ignore this bit:
+        >>> nfnetlink_queue_data = '''
+        ... 0  -4423     0 2 65535     0     0       22  1
+        ... 1  -4424     0 2 65535     0     0       27  1
+        ... '''
+        >>> from insights.tests import context_wrap
+        >>> from insights.parsers.nfnetlink_queue import NfnetLinkQueue
+        >>> nf = NfnetLinkQueue(context_wrap(nfnetlink_queue_data))
```
bfahr, line 48: @ryan-blakley this (L46-48) would normally not be included here.  Just setup the parser object in the doc tests and run examples here.  Since you are updating the docs could you go ahead and remove it?

bfahr (changes requested):

PR #3273, opened 2021-11-02
Strip progress messages from testparm output

kgrant-rh:
On some older releases of Samba, testparm -v -s contains progress
messages such as `Processing section "[homes]"`. Because these
messages contain section names, they aren't removed by filtering.
This change removes them in the testparm parser.

Signed-off-by: Keith Grant <kgrant@redhat.com>

### All Pull Requests:

Check all that apply:

Some older versions of Samba contain messages in the output of `testparm` that cause exceptions in the testparm parsers. This change removes those lines in the parser.

### DIFF
--- a/insights/parsers/samba.py
+++ b/insights/parsers/samba.py
@@ -135,6 +135,10 @@ class SambaConfigs(SambaConfig):
         server_role (string): Server role as reported by the command.
     """
     def parse_content(self, content):
+        # The output of `testparm` sometimes includes progress lines such as
+        # `Processing section "[homes]"`, so those should be removed.
+        content = [line for line in content if not line.startswith("Processing section")]
+
         # Parse server role
         for line in content:
             r = re.search(r"Server role:\s+(\S+)", line)
--- a/insights/parsers/samba.py
+++ b/insights/parsers/samba.py
@@ -135,20 +135,22 @@ class SambaConfigs(SambaConfig):
         server_role (string): Server role as reported by the command.
     """
     def parse_content(self, content):
-        # The output of `testparm` sometimes includes progress lines such as
-        # `Processing section "[homes]"`, so those should be removed.
-        content = [line for line in content if not line.startswith("Processing section")]
-
         # Parse server role
-        for line in content:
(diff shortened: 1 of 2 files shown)


insights/parsers/samba.py
```diff
@@ -135,6 +135,10 @@ class SambaConfigs(SambaConfig):
         server_role (string): Server role as reported by the command.
     """
     def parse_content(self, content):
+        # The output of `testparm` sometimes includes progress lines such as
+        # `Processing section "[homes]"`, so those should be removed.
+        content = [line for line in content if not line.startswith("Processing section")]
```
xiangce, line 140: Nitpick: This will add one more pass of iterating against the whole content. Would it be better to ignore these lines in the following loop?
But please ignore this comment, if the content is always in very few lines.

kgrant-rh: Unfortunately, due to the structure of the samba config parser, there is no loop in which these lines can be ignored. The SambaConfigs parser calls the SambaConfig parser, which calls the IniConfigFile parser, which is where the error occurs. This content should be on average about 300 lines or slightly less; it may be longer on systems where a significant number of samba shares are defined.

skontar: @kgrant-rh I think he meant to combine this with the loop where we are looking for server role on lines 143. Also we can be sure it does not appear after server role. Maybe we could just discard everything before server role? If server role does not exist we raise exception anyway.

skontar: I think we can do something like:
```
 # Parse server role
        for i, line in enumerate(content):
            r = re.search(r"Server role:\s+(\S+)", line)
            if r:
                self.server_role = r.group(1)
                server_role_line = i
                break
        else:
            raise ParseException("Server role not found.")
super(SambaConfigs, self).parse_content(content[server_role_line:])
```

kgrant-rh: @skontar You're right, that's a better solution; I believe you're correct that these messages should always appear before the server role line.

xiangce (approved):
LGTM. Thanks @kgrant-rh.

bfahr (approved):

PR #3281, opened 2021-11-08
Added rhsm dependency for insights-core develop 1596

ahitacat:
### All Pull Requests:

Check all that apply:

Insights-client develop requires of rhsm package. Added this dependency when created the virtualenv.


subpop:
test this please

setup.py
```diff
@@ -62,6 +62,7 @@ def maybe_require(pkg):
 
 develop = set([
     'wheel',
+    'rhsm',
```
ryan-blakley, line 65: @ahitacat Since this is for the client specifically could you please create a client-develop set with only rhsm in it. Then on line 134/135 add the client-develop in the client-develop extra requires. Also could you elaborate on why this is needed, I'm pretty sure rhsm should be pulled in via the subscription-manager-rhsm rpm, and not the module from pypi but I could be wrong.

docs/quickstart_insights_core.rst
```diff
@@ -37,7 +36,8 @@ file associated with the insights-core project.
 .. HINT::
    You might also need to install ``gcc`` to be able to build some python modules,
    ``unzip`` to be able to run `pytest` on the ``insights-core`` repo,
-   and ``pandoc`` to build Insights Core documentation.
+   ``pandoc`` to build Insights Core documentation, and ``platform-python-devel`` 
+   and ``openssl-devel`` to build ``rhsm``.
```
ryan-blakley, line 40: @ahitacat Since this is for the client only could you please remove the updates to the insights core quickstart doc. You may want to try and update the developer setup doc in the insights-client repo. https://github.com/RedHatInsights/insights-client

ryan-blakley (changes requested):

setup.py
ahitacat: Hi!
I found that if you install the client within the virtualenv for development, the rhsm is not pulled and thus you need it in the environment. And sure, I can add it in a client-develop set.

Glutexo: If I’m not mistaken things installed from RPMs are only available for the system-wide Python and not in virtual environments, as @ahitacat says.

Glutexo: @ryan-blakley _rhsm_ is actually imported in [_insights-core_](https://github.com/RedHatInsights/insights-core/blob/e6b64c3c02752a737e61123f9167903cc058c5dd/insights/client/auto_config.py#L102) on auto config. I’d say it’s essential for it to even run and not limited to _insights-client_ development.

docs/quickstart_insights_core.rst
Glutexo: @ryan-blakley This is actually what you need to run code in _insights-core_. _insights-client_ only warns you on make if you are missing the Python _rhsm_ module, but does not use it itself.

Glutexo:
test this please

setup.py
bfahr: @ahitacat @Glutexo you are correct it is in insights-core but it's only imported by the client code in insights-core, and only necessary for client development/testing. I think it makes more sense to include this in the existing `develop-client` set or create a separate set as @ryan-blakley suggested to include that module.  We have several teams that use insights-core but not the client so it seems to make things cleaner and more obvious.

docs/quickstart_insights_core.rst
ryan-blakley: @Glutexo I've never had to build rhsm to run code in insights-core. If it's required for the client code development, I'd recommend creating a section in the document that specifically mentions for client code development you may need to install x additionally. That way it's distinguished from what users viewing this doc need for rule development.

setup.py
Glutexo: @bfahr I’d put that into the _client_ section, because it’s required to even _run_ the client in its own virtual environment. If I am not mistaken, when used from the installed RPM package as an egg, the dependencies are not installed by _setuptools_, but by RPM, so there should be no interference.

ryan-blakley: Adding it to the client set works also if it's required for the client to run. In regards to what I mentioned about the rpm, you can create virtualenv's with the --system-site-packages parameter that gives the venv access to the system's module directory. I mentioned the rpm only because the version of rhsm on pypi is older than what's shipped in rhel, and it doesn't appear to really be maintained as there is only two version, but I could be wrong. So I thought I'd mention the rpm in case there was differences in the code between the versions.

Glutexo: @ryan-blakley Is there a reason for _rhsm_ in PyPI not being updated?

docs/quickstart_insights_core.rst
Glutexo: E.g. the _auto_config_ module here in _insights-core_ uses the _rhsm_ module. https://github.com/RedHatInsights/insights-core/blob/e6b64c3c02752a737e61123f9167903cc058c5dd/insights/client/auto_config.py#L102 It may not be necessary for **rule** development, but it is for development of code in the _insights-core_ repository.

This change is in a page called “Quickstart Insights Development” with “Rule Development Setup” being just a subsection. The green “Hint” box even mentions (for _pytest_ in this case) the “insights-core repo” that includes the client code (and tests of it) as well. Adding a section sounds like a good idea. Thanks! 👍🏻

setup.py
ryan-blakley: @Glutexo I personally don't know I'm not involved with rhsm in anyway I'm on the insights-core team. I just checked pypi to make sure the package existed initially, and saw it only had two version both of which are older than what's in rhel. But taking another look at the pypi project it's maintained by someone outside of Red Hat, so I'm now less inclined to include that requirement. That might be why it hasn't been a requirement to date.

Glutexo: @ryan-blakley Interesting! Maybe grabbing that from PyPI is not the best thing to do and we should rather find a way how to access the RPM library from the virtual environment. I think Chris would be able to help with that. Thank you, Ryan!

docs/quickstart_insights_core.rst
Glutexo: We talked a bit about this with @ahitacat and it may be ok enough to mention in the _Hint_ box something like

> For the Client development, you may need to install `python-devel` and `openssl-devel`. This is not necessary for Rule only development.

However we should make sure first that adding `rhsm` as a PyPI dependency is the right way. If we find a way to directly use `rhsm` installed from an RPM, this is all void.

setup.py
ryan-blakley: @Glutexo No problem, you can access the rpm library by creating a virtualenv with the following command. `$ python3 -m venv --system-site-packages .venv` I tested this and was able to import rhsm from within the freshly created venv.

docs/quickstart_insights_core.rst
Glutexo: After talking with Chris Snyder, we decided we’ll go the PyPI way. It’s not ready yet though and the version of `rhsm` in _PyPI_ is obsolete. We’re not going to install it and thus the `-devel` libraries are not necessary.

Glutexo:
@ahitacat Let’s ditch this pull request and create a new one just updating the documentation. We’ll suggest using the `--system-site-packages` CLI parameter in case of Client development. It’s not the best solution, but now there is no other way to use a virtual environment and still have access to the `rhsm` library. This is going to be replaced by a proper dependency.

I also suggest adding a check to the _setup.py_ that would make it fail if _client_ extra is enabled and `rhsm` not available. That would prevent the library to be installed in an environment where it would not work.

Glutexo (changes requested):
Not doing this until we `rhsm` is in PyPI.

bfahr:
@ahitacat can this be closed?

ahitacat:
> @ahitacat was this replace with #3287? Can this PR be closed?

Yes, at least it is specified in the documentation that the package is missed.

bfahr:
Closing based on previous comment by @ahitacat

PR #3283, opened 2021-11-09
Add parser mssql_tls_file

wushiqinlou:

### All Pull Requests:

Check all that apply:

This parser is used by a insights rule to check mssql tls expiration date.

### DIFF
--- a/insights/parsers/ssl_certificate.py
+++ b/insights/parsers/ssl_certificate.py
@@ -12,6 +12,8 @@ HttpdSSLCertExpireDate - command ``openssl x509 -in httpd_certificate_path -endd
 ============================================================================================
 NginxSSLCertExpireDate - command ``openssl x509 -in nginx_certificate_path -enddate -noout``
 ============================================================================================
+MssqlTLSCertExpireDate - command ``openssl x509 -in mssql_tls_cert_enddate -enddate -noout``
+============================================================================================
 """
 
 from insights import parser, CommandParser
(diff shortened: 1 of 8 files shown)


wushiqinlou:
it is able to collect the spec via test:

~~~
# cat data/insights_commands/openssl_x509_-in_.etc.ssl.certs.mssql.pem_-enddate_-noout
notAfter=Nov 10 06:45:07 2021 GMT
~~~

xiangce (approved):
Looks good to me.

@psachin - please have a look too.

psachin:
Good to me.

insights/parsers/ssl_certificate.py
```diff
@@ -251,3 +253,25 @@ class NginxSSLCertExpireDate(CertificateInfo):
...
+    It parses the output of ``openssl x509 -in mssql_tls_cert_file -enddate -noout``.
+
+    Sample output of ``openssl x509 -in mssql_tls_cert_file -enddate -noout``::
+
+        notAfter=Dec 4 07:04:05 2035 GMT
+
+    Examples:
+        >>> type(mssql_date_info)
+        <class 'insights.parsers.ssl_certificate.MssqlTLSCertExpireDate'>
+        >>> mssql_date_info['notAfter'].datetime
```
psachin, line 274: @wushiqinlou As we are specifically looking for the value of 'notAfter' here can we include that check in the parser logic instead of checking `'notAfter' in date_info` in the rule code ? The basics idea is to return `None` is 'notAfter' does not exist.

xiangce: @psachin - the superclass is a dict, a `KeyError` will be thrown when fetching a non-existed key. I think this is okay for a `dict` parser (but a bit weird to enforce the obj[key] returns None by default), of cause, `.get()`  would return `None` in that case as well.

psachin: @xiangce Yes, I wanted to suggest `.get()` what do you think ?

PR #3288, opened 2021-11-15
Add in IsRhel9 component

ryan-blakley:
Check all that apply:

Updated the rhel_version component to have a base class that the other classes inherit to cut down on the duplicate code. Added the IsRhel9 component check. Added rhel9 checks to the redhat_release combiner. Added the IsRhel9 check for the BootLoaderEntries parser. Added the IsRhel9 check for the corosync_cmapctl_cmd_list datasource, and added IsRhel9 to the collect manifest for the corosync_cmapctl_cmd_list datasource.

### DIFF
--- a/insights/components/rhel_version.py
+++ b/insights/components/rhel_version.py
@@ -1,6 +1,6 @@
 """
-IsRhel6, IsRhel7 and IsRhel8
-===============================
+IsRhel6, IsRhel7, IsRhel8, and IsRhel9
+======================================
 
 An ``IsRhel*`` component is valid if the
 :py:class:`insights.combiners.redhat_release.RedHatRelease` combiner indicates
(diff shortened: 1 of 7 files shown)


insights/components/rhel_version.py
```diff
@@ -11,64 +11,48 @@
...
+            raise SkipComponent("Not RHEL{vers}".format(vers=version))
         self.minor = rhel.minor
 
 
 @component(RedHatRelease)
-class IsRhel7(object):
-    """
-    This component uses ``RedHatRelease`` combiner
-    to determine RHEL version. It checks if RHEL7, if not \
-    RHEL7 it raises ``SkipComponent``.
```
bfahr, line 44: @ryan-blakley all of the child classes need to have docstrings so that they will be fully documented in the rendered documentation.

bfahr (changes requested):

insights/components/rhel_version.py
ryan-blakley: @bfahr Ahh that's a good point I added them back.

insights/specs/default.py
```diff
@@ -143,7 +143,7 @@ class DefaultSpecs(Specs):
     cmdline = simple_file("/proc/cmdline")
     corosync = simple_file("/etc/sysconfig/corosync")
 
-    @datasource(HostContext, [IsRhel7, IsRhel8])
+    @datasource(HostContext, [IsRhel7, IsRhel8, IsRhel9])
```
xiangce, line 146: Sorry for the delay, and Just FYI, the author of this datasource @huali027, now is checking if this command works the same in RHEL 8 and RHEL 9. And will get back to you.

ryan-blakley: That's okay thank you for letting me know, I just checked the package versions and saw that corosync in r9 and r8 were based on the same upstream version. But actually checking on a setup system would be a much better verification.

huali027: @xiangce @ryan-blakley good to me, I found the commands on RHEL9 are the same with RHEL8.

xiangce (approved):
Good to me.

PR #3289, opened 2021-11-16
Update setup.py

ryan-blakley:
Check all that apply:

Removed the client console script as it doesn't work anymore since the launching of the client is actually handle by a file in the insights-client repo. Added the pyyaml to the runtime the way pytest was added to the testing set instead of sys.version_info way it was using.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,4 @@
 import os
-import sys
 from setuptools import setup, find_packages
 
 __here__ = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


ryan-blakley:
@subpop could you review this when you get a chance please? The client entry point doesn't work currently, but we just wanted to make sure removing the client entry doesn't mess with anything on the insights-client side.

subpop (approved):
The client does not use the client entrypoint, so I'd say its safe to remove.

bfahr (approved):

PR #3298, opened 2021-12-06
Add log_response_text flag to log downloads or not in verbose mode

mhuth:
* https://issues.redhat.com/browse/YARA-232
* Don't log the malware signatures download in verbose mode
* It is a big binary blob and may leave artifacts on the command line afterwards

### All Pull Requests:

Check all that apply:

The malware-detection app downloads the malware signatures as a big binary blob.  When running insights-client in verbose mode, this binary blob is displayed on the terminal and can leave trailing text after the command has finished.  There is no real need to display this binary blob and this PR adds a flag to the connection module to toggle displaying the response.text or not when in verbose mode.  It is on by default which maintains the current behaviour, but for malware-detection, it is turned off so the binary blob isn't displayed on the terminal or in the logs.


subpop:
test this please

subpop (approved):

bfahr (approved):

Issue #3313, opened 2021-12-22
The parser "SatellitePostgreSQLQuery" generate wrong data if there is other output before the real rows data 


huali027:
Sometimes, the psql --csv command output the bellow content:
```
logname: no login name
/etc/profile.d/hkuser.sh: line 40: HISTFILE: readonly variable
id,name
1,Puppet_Base
```
Then the parser "SatellitePostgreSQLQuery" generates the following data which is wrong
```
 [{'logname: no login name': '/etc/profile.d/hkuser.sh: line 40: HISTFILE: readonly variable'}, {'logname: no login name': 'id', None: ['name']}, {'logname: no login name': '1', None: ['Puppet_Base']}]
```

xiangce:
Resolved by #3314

PR #3314, opened 2021-12-22
Fix: Enhance parser "SatellitePostgreSQLQuery"

huali027:

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/satellite_postgresql_query.py
+++ b/insights/parsers/satellite_postgresql_query.py
@@ -10,10 +10,8 @@ SatelliteComputeResources - command ``psql -d foreman -c 'select name, type from
 -----------------------------------------------------------------------------------------------------------
 SatelliteSCAStatus - command ``psql -d candlepin -c "select displayname, content_access_mode from cp_owner" --csv``
 -------------------------------------------------------------------------------------------------------------------
-
 SatelliteKatelloEmptyURLRepositories - command ``psql -d foreman -c 'select id, name from katello_root_repositories where url is NULL;' --csv``
 -----------------------------------------------------------------------------------------------------------------------------------------------
-
 SatelliteCoreTaskReservedResourceCount - command ``psql -d pulpcore -c 'select count(*) from core_taskreservedresource' --csv``
 -------------------------------------------------------------------------------------------------------------------------------
 """
@@ -25,12 +23,13 @@ from csv import DictReader
 from insights import parser, CommandParser
 from insights.specs import Specs
 from insights.parsers import SkipException, ParseException
-from insights.parsers import keyword_search
+from insights.parsers import keyword_search, calc_offset
 
 
 class SatellitePostgreSQLQuery(CommandParser, list):
     """
-    Parent class of satellite postgresql table queries.
+    Parent class of satellite postgresql table queries. It can not be used
+    directly, A child class with overriding headers attribute is required.
     It saves the rows data into a list. Each row is saved into a dict.
     The key is the column name, the value is the value of the column.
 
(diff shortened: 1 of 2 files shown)


huali027:
Fixes: https://github.com/RedHatInsights/insights-core/issues/3313

huali027:

clarabez:
Hi, @huali027!
We've identified that setup pipeline is failing for RHEL6 on "QE Pull Request Testing", but it is working as expected for RHEL79, RHEL84 and RHEL90.

This issue has been reported and is being analyzed here: [RHCLOUD-17392](https://issues.redhat.com/browse/RHCLOUD-17392).

insights/parsers/satellite_postgresql_query.py
```diff
@@ -72,15 +72,24 @@ class SatellitePostgreSQLQuery(CommandParser, list):
...
         SkipException: when there isn't data in the table
-        ParseException: when the output isn't in good csv format
+        ParseException: when the output isn't in good csv format or the subclass doesn't define headers
     """
+    headers = []
 
     def parse_content(self, content):
-        if not content or len(content) == 1:
-            raise SkipException("There is no data in the table")
+        if self.__class__.__name__ != 'SatellitePostgreSQLQuery' and not self.headers:
```
xiangce, line 80: When defining the base class like this (use an undefined/empty static member as a required `gurad`), it's not a good idea to instantiate or use it like this. that's to say, this base class should be instantiated due the `self.headers` is never contain a value, unless in its child classes. 

Better remove all the direct tests against the base class. but test them in the child-classes instead.

insights/parsers/satellite_postgresql_query.py
```diff
@@ -139,6 +148,7 @@ class SatelliteAdminSettings(SatellitePostgreSQLQuery):
         >>> table.get_setting('destroy_vm_on_host_delete')
         True
     """
+    headers = ['name', 'value', 'default']
```
xiangce, line 151: This is different with the description of Line-163

insights/parsers/satellite_postgresql_query.py
```diff
@@ -72,15 +72,24 @@ class SatellitePostgreSQLQuery(CommandParser, list):
...
+        ParseException: when the output isn't in good csv format or the subclass doesn't define headers
     """
+    headers = []
 
     def parse_content(self, content):
-        if not content or len(content) == 1:
-            raise SkipException("There is no data in the table")
+        if self.__class__.__name__ != 'SatellitePostgreSQLQuery' and not self.headers:
+            raise ParseException("Please define the headers attribute which contains the column names.")
+        start_index = None
```
xiangce, line 82: the `insights.parsers.calc_offset` can be used to get the `start_index`.

xiangce (changes requested):

insights/parsers/satellite_postgresql_query.py
huali027: Do u mean I don't mention "name" in Line-163? In Line-163, I just explain "default" and "value" must both be queried to determine the final value of the setting.

insights/parsers/satellite_postgresql_query.py
huali027: @xiangce I update the code to move the test in child-classes, please help to review again, thanks.

insights/parsers/satellite_postgresql_query.py
huali027: Updated, please help to check again, thanks.

insights/parsers/satellite_postgresql_query.py
xiangce: By defining the `header`, I think it implicitly means that the header line of the content should contain all of the items in the `header`.  so the L145~L146 is not required anymore.

And the base class should also be updated as:

```
if not all(t in content[start_index] for t in self.headers):
   raise ...
```

xiangce: I just enhanced the `calc_offset` and added a new `require_all` parameter
- https://github.com/RedHatInsights/insights-core/pull/3316
After that, you can use it, like:
```
start_index = calc_offset(content, self.headers, require_all=True)
```

huali027:

xiangce (approved):

PR #3328, opened 2022-01-27
feat: New spec and parser for losetup -l

takayuki-nagata:
Signed-off-by: Takayuki Nagata <tnagata@redhat.com>

### All Pull Requests:

Check all that apply:

This parser is used by a insights rule to check configured loop back devices, its back end files and 'DIO' flags.

### DIFF
--- a/insights/parsers/losetup.py
+++ b/insights/parsers/losetup.py
@@ -1,0 +1,73 @@
+"""
+LoSetup - command ``/usr/sbin/losetup -l``
+==========================================
+
+This parser reads the output of ``/usr/sbin/losetup -l`` into a list of entries.
+Each entry is a dictionary of headers:
+
+* ``NAME`` - the path name of the loop back device (strings)
+* ``SIZELIMIT`` - the data end position of backing file in bytes (integer)
+* ``OFFSET`` - the data start position of backing file in bytes (integer)
+* ``AUTOCLEAR`` - the autoclear flag (boolean)
+* ``RO`` - the read only flag (boolean)
+* ``BACK-FILE`` - the path of the backing file (strings)
+* ``DIO`` - the direct I/O flag (boolean)
+* ``LOG-SEC`` - the logical sector size of the loop device in bytes (integer)
+
+Sample output of ``losetup -l`` command is::
+
+    NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE      DIO LOG-SEC
+    /dev/loop0         0      0         0  0 /root/disk.img   1     512
+
+Examples:
+
+    >>> type(losetup)
+    <class 'insights.parsers.losetup.LoSetup'>
+    >>> len(losetup)
+    1
+    >>> losetup[0]['NAME']
+    '/dev/loop0'
+    >>> losetup[0]['RO']
+    False
+    >>> losetup[0]['DIO']
+    True
+    >>> losetup[0]['NAME']
+    '/dev/loop0'
+    >>> losetup[0]['SIZELIMIT']
+    0
+    >>> losetup[0]['OFFSET']
+    0
+    >>> losetup[0]['AUTOCLEAR']
+    False
+    >>> losetup[0]['RO']
+    False
+    >>> losetup[0]['BACK-FILE']
+    '/root/disk.img'
+    >>> losetup [0]['DIO']
+    True
+    >>> losetup[0]['LOG-SEC']
+    512
+"""
+from insights import CommandParser, parser
+from insights.specs import Specs
+from insights.parsers import parse_delimited_table, SkipException
+
+
+@parser(Specs.losetup)
+class LoSetup(CommandParser, list):
+    """
+    Parses the output of the ``/usr/sbin/losetup -l`` command.
+    """
+
+    def parse_content(self, content):
+        if not content:
+            raise SkipException("Empty output.")
+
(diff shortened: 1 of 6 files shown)


ryan-blakley:
This needs to go through the spec approval process first, I sent an email with the instructions.

insights/parsers/losetup.py
```diff
@@ -0,0 +1,82 @@
...
+    >>> losetup[0]['LOG-SEC']
+    512
+"""
+from insights import CommandParser, parser
+from insights.specs import Specs
+from insights.parsers import parse_delimited_table
+
+
+@parser(Specs.losetup)
+class LoSetup(CommandParser):
```
ryan-blakley, line 57: @takayuki-nagata Could you switch the class to inheriting a list instead of using the older self.data variable format? You can look at the ChronycSources parser in the ntp_sources.py file as an example. Basically just inherit list and do self.extend(parser_delimited_table(content)).

insights/parsers/losetup.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __iter__(self):
+        for entry in self.data:
+            yield entry
+
+    def __getitem__(self, idx):
+        return self.data[idx]
+
+    def parse_content(self, content):
+        self.data = parse_delimited_table(content)
```
ryan-blakley, line 75: Before running parse_delimited_table, please add a check for empty content, and add a test that test empty content. 
``` python
if not content:
     raise SkipException("Empty output.")
```

insights/parsers/tests/test_losetup.py
```diff
@@ -0,0 +1,45 @@
...
+/dev/loop0         0      0         1  0 /root/disk1.img   1     512
+"""
+
+LOSETUP_RHEL7 = """
+NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE
+/dev/loop3      4096    256         0  1 /root/disk3.img
+"""
+
+
+def test_losetup():
```
ryan-blakley, line 16: Please add a doctest, the previous example I mentioned ntp_sources parser, you can reference it's test for doc test.

ryan-blakley (changes requested):

insights/parsers/losetup.py
takayuki-nagata: I have switched the class to inheriting a list, and have changed to storing a result of parser_delimited_table() to itself. In addition, I have added data() as a property.

insights/parsers/losetup.py
takayuki-nagata: I have added the check, and have added a test pattern for empty content.

insights/parsers/tests/test_losetup.py
takayuki-nagata: I have added the doctest.

takayuki-nagata:
@ryan-blakley Thank you for your review. I have resent v2.

insights/parsers/losetup.py
```diff
@@ -0,0 +1,83 @@
...
+                if key in entry:
+                    entry[key] = int(entry[key])
+            for key in ['AUTOCLEAR', 'RO', 'DIO']:
+                if key in entry:
+                    entry[key] = True if entry[key] == '1' else False
+
+    @property
+    def data(self):
+        """
+        Set data as property to keep compatibility
```
xiangce, line 81: @takayuki-nagata - Sorry for the late comment.

Generally, this `self.data` attribute is not required for the newly developed parser, it's recommended for users to use it as `a` `list` parser directly. so, it's not necessary to add this `property` to keep backward compatibility. Would you please remove it here and from the above `Attributes` section in the doc part? Thanks.

takayuki-nagata: @xiangce sure, I have removed the property and have re-sent v3.

psachin:
Hey @takayuki-nagata 

Can you also include the test case for RHEL6 where the command fails with the following error

```
-bash: /usr/sbin/losetup: No such file or directory
```

xiangce:
> 

@psachin - The base class [CommandParser](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/__init__.py#L529) could handle the `No such file or directory` error.

@takayuki-nagata - would you please double check whether it's necessary to support `RHEL6` or not by this Parser (and the rule)? If `RHEL6` is not required, we can just leave it as it was. But if you also want to support `RHEL 6`, please check where the `losetup` locate on it? and add it to the spec like:

```python
# insights/specs/default.py
losetup = first_of(simple_command("/usr/sbin/losetup -l"), simple_command("/RHEL6_path/to/losetup -l"))
# insights/specs/insights_archive.py
# - [Optional if they are different]
losetup = first_file("insights_commands/losetup_-l", "insights_commands/path/to/losetup_-l")
```

takayuki-nagata:
@xiangce In `RHEL6`, the command path is `/sbin/losetup`. It does not support `-l` option, and only support deprecated old format (`-a` option). I will use this parser for a rule with `RHEL8`, and this `LoSetup` parser does not require to support `RHEL6`.

psachin:
> 

@takayuki-nagata If the parser don't need to support RHEL6 we leave it as it is.

xiangce (approved):

ryan-blakley (approved):
Thank you for the updates, everything looks good to me now.

PR #3329, opened 2022-01-27
Improved yum/dnf updates datasource to work on RHEL 8.5+

MichaelMraka:
### All Pull Requests:

Check all that apply:

Current `yum updates` datasource is limited to work on RHEL7.
This PR extends functionality to RHEL 8.5+ systems.
When run on RHEL 8 prior RHEL 8.5 it skips data collection (a bug in dnf ignores cacheonly flag and may cause downloading data over network ).

### DIFF
--- a/insights/specs/datasources/yum_updates.py
+++ b/insights/specs/datasources/yum_updates.py
@@ -5,24 +5,102 @@ import json
 import time
 
 from insights import datasource, HostContext, SkipComponent
-from insights.components.rhel_version import IsRhel7
+from insights.components.rhel_version import IsRhel7, IsRhel8, IsRhel9
 from insights.core.spec_factory import DatasourceProvider
 
-sorted_cmp = None
 try:
+    from functools import cmp_to_key
+
     # cmp_to_key is not available in python 2.6, but it has sorted function which accepts cmp function
     def sorted_cmp(it, cmp):
-        from functools import cmp_to_key
         return sorted(it, key=cmp_to_key(cmp))
 except ImportError:
     sorted_cmp = sorted
 
 
-class UpdatesManager:
+class DnfManager:
+    """ Performs package resolution on dnf based systems """
+    def __init__(self):
+        self.base = dnf.base.Base()
+        self.releasever = dnf.rpm.detect_releasever("/")
+        self.basearch = dnf.rpm.basearch(hawkey.detect_arch())
+        self.packages = []
+        self.repos = []
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        pass
+
+    @staticmethod
+    def pkg_cmp(a, b):
+        if a.name != b.name:
+            return -1 if a.name < b.name else 1
+        vercmp = rpm.labelCompare((str(a.e), a.v, a.r), (str(b.e), b.v, b.r))
+        if vercmp != 0:
+            return vercmp
+        if a.reponame != b.reponame:
+            return -1 if a.reponame < b.reponame else 1
+        return 0
+
+    def sorted_pkgs(self, pkgs):
+        # if package is installed more than once (e.g. kernel)
+        # don't report other installed (i.e. with @System repo) as updates
+        return sorted_cmp([pkg for pkg in pkgs if pkg.reponame != "@System"], self.pkg_cmp)
+
+    def load(self):
+        self.base.conf.read()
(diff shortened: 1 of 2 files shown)


insights/specs/datasources/yum_updates.py
```diff
@@ -18,83 +16,176 @@ def sorted_cmp(it, cmp):
     sorted_cmp = sorted
 
 
-class UpdatesManager:
-    """ Performs package resolution on yum based systems """
-    def __init__(self):
+# Select which manager to use based on the available system libraries.
+try:
+    try:
```
ryan-blakley, line 21: Would it be possible to move the class definitions out of the try catches, instead set a variable? Also I noticed there is some duplicate code in the UpdatesManager class definitions. Would it be possible to have a base class that the classes inherit, that way the duplicate functions and variables can be defined in the base class ex: __init__, __enter__, __exit__, pkg_nevra? Or could you just use one class, and perform checks in each function base on the variable set in the try catches?

insights/specs/datasources/yum_updates.py
```diff
@@ -18,83 +16,176 @@ def sorted_cmp(it, cmp):
...
-        if adv:
-            return adv.get_metadata()['update_id']
-        return None
-
-    @staticmethod
-    def last_update():
-        return 0
-
-
-@datasource(HostContext, [IsRhel7])
```
ryan-blakley, line 97: Just curious why you removed the datasource decorator instead of just removing the IsRhel7 part?

insights/tests/datasources/test_yum_updates.py
```diff
@@ -0,0 +1,10 @@
+from insights.core.spec_factory import DatasourceProvider
+from insights.specs.datasources.yum_updates import yum_updates
+
+
+# Verify that the yum_updates broker correctly executes.
+def test_yum_updates_runs_correctly():
+    broker = {}
+    result = yum_updates(broker)
```
ryan-blakley, line 8: This doesn't test that any of the data is processed correctly right? Would it be possible to add in tests based on mocked output?

ryan-blakley (changes requested):

insights/specs/datasources/yum_updates.py
MichaelMraka: The datasource decorator was added in the past to constrain this datasource to RHEL7 only so I removed it because now we can run it on all systems.
But it turned out it's wrong and need to add `IsRhel8, IsRhel9` instead of removing it ;-).

insights/specs/datasources/yum_updates.py
MichaelMraka: Simplified imports and changed it into 2 classes, one inheriting from the other.

insights/tests/datasources/test_yum_updates.py
MichaelMraka: Ok, updated with a mocked data test.

insights/specs/datasources/yum_updates.py
ryan-blakley: Thank you that looks better.

insights/specs/datasources/yum_updates.py
ryan-blakley: Ah yeah the decorator is required for every datasource.

insights/tests/datasources/test_yum_updates.py
ryan-blakley: Awesome thank you.

ryan-blakley (changes requested):
@MichaelMraka all the changes look good. One more thing since you're modifying this datasource. If you don't mind could you take a look at the following bz, and see if you could easily add a fix for it? https://bugzilla.redhat.com/show_bug.cgi?id=2035118

MichaelMraka:
Hi @ryan-blakley ,
I've actually hit the same bug during my tests. It's fixed by this PR:
```
test-yum-updates.py:126:
        try:
            self.base.doSackSetup()
            self.packages = self.base.pkgSack.returnPackages()
        except yum.Errors.RepoError:
            # RepoError is raised when cache is empty
            pass
```
I'll update bugzilla accordingly.

patchkez:
test me

ryan-blakley (approved):
Everything looks good to me, thank you for the updates.

PR #3331, opened 2022-02-09
Exclude some Specs from IP address obfuscation

Glutexo:
### All Pull Requests:

Check all that apply:

This Pull Request is a fix for [ESSNTL-444](https://issues.redhat.com/browse/ESSNTL-444) / [Bug 2026141](https://bugzilla.redhat.com/show_bug.cgi?id=2026141), introducing a list of Specs excluded from the IP obfuscation by the SOSCleaner. The files generated by some Specs contain package versions consisting of four numeric parts. Those strings are indistinguishable from IPv4 addresses and the SOSCleaner incorrectly obfuscates them. The now invalid package versions case false rule hits further down in the cloud console.

As a hotfix, a hardcoded list of Specs excluded from the IP address obfuscation is introduced. A more proper way would be to have a flag in the Spec that would allow it to exclude itself from the obfuscation, or to provide a custom obfuscation pattern. This solution would however require a change to the Spec architecture.

### DIFF
--- a/insights/client/constants.py
+++ b/insights/client/constants.py
@@ -87,4 +87,3 @@ class InsightsConstants(object):
     rhsm_facts_file = os.path.join(os.sep, 'etc', 'rhsm', 'facts', 'insights-client.facts')
     # In MB
     archive_filesize_max = 100
-    obfuscation_excluded_specs = ("insights.specs.Specs.installed_rpms",)
--- a/insights/client/constants.py
+++ b/insights/client/constants.py
@@ -87,3 +87,4 @@ class InsightsConstants(object):
     rhsm_facts_file = os.path.join(os.sep, 'etc', 'rhsm', 'facts', 'insights-client.facts')
     # In MB
     archive_filesize_max = 100
+    obfuscation_excluded_specs = ("insights.specs.Specs.installed_rpms",)
--- a/insights/contrib/soscleaner.py
+++ b/insights/contrib/soscleaner.py
@@ -32,6 +32,7 @@ import logging
 import tarfile
 import six
 import hashlib
+import json
 
 from insights.util import content_type
 
@@ -77,6 +78,10 @@ class SOSCleaner:
         self.keywords = None
         self.kw_db = dict() #keyword database
         self.kw_count = 0
+
+        # Doing so here to avoid cyclic imports.
+        from insights.client.constants import InsightsConstants as constants
+        self.excluded_specs = constants.obfuscation_excluded_specs
 
     def _skip_file(self, d, files):
         '''
@@ -716,6 +721,21 @@ class SOSCleaner:
             self.logger.exception(e)
             raise Exception("CleanFilesOnlyError: unable to process")
 
+    def _excluded_files(self):
+        '''Collect data files excluded from IP address obfuscation.
+        Their paths are part of the meta_data JSONs.
+        '''
+        excluded_files = []
+        meta_data_path = os.path.join(self.dir_path, "meta_data")
+        for dir_name, file_names in self._walk_report(meta_data_path).items():
+            for file_name in file_names:
+                file_path = os.path.join(dir_name, file_name)
+                with open(file_path) as file:
+                    meta_data = json.load(file)
+                if meta_data["name"] in self.excluded_specs:
+                    excluded_files.append(meta_data["results"]["object"]["relative_path"])
+        return excluded_files
+
     def clean_report(self, options, sosreport): # pragma: no cover
         '''this is the primary function, to put everything together and analyze an sosreport'''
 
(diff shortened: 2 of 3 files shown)


insights/contrib/soscleaner.py
```diff
@@ -78,6 +79,10 @@ def __init__(self, quiet=False):
         self.kw_db = dict() #keyword database
         self.kw_count = 0
 
+        # Doing so here to avoid cyclic imports.
+        from insights.client.constants import InsightsConstants as constants
+        self.ip_obfuscation_blacklist = constants.ip_obfuscation_blacklist
```
subpop, line 84: Couldn't this list be maintained as a static list in `soscleaner.py` directly? Is there any advantage to putting it in `insights/client/constants.py`? That way, we'd avoid this cyclic import, and the tests could avoid having to pull `insights.client` modules into contrib.

insights/client/constants.py
```diff
@@ -87,3 +87,4 @@ class InsightsConstants(object):
     rhsm_facts_file = os.path.join(os.sep, 'etc', 'rhsm', 'facts', 'insights-client.facts')
     # In MB
     archive_filesize_max = 100
+    ip_obfuscation_blacklist = ("insights.specs.Specs.installed_rpms",)
```
subpop, line 90: Can we call it a "deny list" or an "exclude list" instead of a "black list"?

subpop (changes requested):

insights/contrib/soscleaner.py
```diff
@@ -711,6 +716,18 @@ def _clean_files_only(self, files):
             self.logger.exception(e)
             raise Exception("CleanFilesOnlyError: unable to process")
 
+    def _blacklisted_files(self):
```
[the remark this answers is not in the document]
Glutexo: The class is very inconsistent with function comments. Some functions don’t have any documentation at all. But it doesn’t hurt to put one here. 👍🏻

insights/client/constants.py
Glutexo: “deny list” doesn’t sound correct to me as we do not deny access or similar. “exclude list” is much better. I was thinking about an “exempt list”, too.

insights/contrib/soscleaner.py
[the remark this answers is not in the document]
subpop: I know. The lack of documentation across the entire code-base is a problem. But we should avoid making that a bigger problem whenever we can; adding documentation when we do add new functions and methods is one way we can do that. It's the [broken window theory](https://en.wikipedia.org/wiki/Broken_windows_theory) applied to software development.

insights/client/constants.py
subpop: Yea, I think either of those would work.

insights/contrib/soscleaner.py
Glutexo: Although the constants import here is the immediate cause of the error, the real culprit is the flawed module structure.

I chose Constants as the first place to go for hardcoded values. The SOSCleaner is still a natural place for a list of excluded Specs.

subpop: I think `insights.client.constants` should contain only constants that pertain to `insights.client` and its submodules. In this case, `soscleaner` is under `insights.contrib`. Pulling in `insights.client` only for a constant array seems unnecessary to me. We could store the array in the module that needs it. This avoids the cyclic import (in this case), and it avoids furthering the interdependence of the modules in the project. The benefit here (keeping all "constants" in a file called "constants.py") does not outweigh the risk (increased module interdependence and explicit need to import mid-file just to avoid a cyclic import).

insights/contrib/soscleaner.py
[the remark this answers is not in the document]
Glutexo: I am not sure whether abundant comments do more good than harm, but that’s a philosophical question. Here it makes sense as finding file paths in the JSON Specs is non-trivial.

Glutexo:
I implemented the requested changes. I am requesting a new review. Thanks @subpop for your feedback!

insights/tests/test_soscleaner.py
```diff
@@ -1,15 +1,77 @@
...
+from shutil import rmtree
+from tempfile import mkdtemp
 
+from mock.mock import call
 from mock.mock import Mock
+from mock.mock import patch
 from pytest import mark
 
+from insights.client.config import InsightsConfig
+from insights.client.data_collector import CleanOptions
```
subpop, line 15: I was hoping we could reduce this interdependency, but it appears that is going to require more work. For example, why is `CleanOptions`, a structure used by `SOSCleaner`, declared in `insights.client.data_collector`?

subpop (approved):

insights/tests/test_soscleaner.py
Glutexo: I was surprised by that too. And why is it a structure anyway.

PR #3337, opened 2022-02-16
Support downloading malware-detection rules via Satellite

mhuth:
Check all that apply:

* [ x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [ x] Is this PR an enhancement?

This PR allows the malware-detection app to download rules via a Satellite proxy.  The malware-detection rules are on C.R.C but if the system is using a Satellite, then malware-detection needs to use the Satellite URL (and hence its certificates) to initiate the request, which the Satellite will then forward onto C.R.C.

The malware-detection app runs insights-client auto_config to determine if a Satellite is being used by checking the base_url.


subpop (approved):
Seems reasonable to me.

bfahr (approved):

PR #3342, opened 2022-02-21
Revert satellite version enhancement and Enhance "CapsuleVersion" only

huali027:
### All Pull Requests:

Check all that apply:


huali027:
@xiangce 
Last Friday, an advisor rule failed, I found some satellite servers registered to themselves which is supported long time ago. 
And today I run RAT again and find some satellite servers even registering to another satellite server or a capsule. Before I thought it is only done for network-disconnected env. But from RAT result, this happens with network env too.
So I revert the previous changes and keep the same as before for `SatelliteVersion` like only checking "satellite" is installed.
And I enhance "CapsuleVersion" to skip it if it is already a satellite server and remove "foreman" checking since sometimes it is a capsule but "foreman" is there due to historical reasons.
Please help to review, thanks.

xiangce (approved):

PR #3350, opened 2022-03-02
Add a new spec and parser for /etc/insights-client/tags.yaml

tciernik:
Signed-off-by: Tamara Ciernikova <tciernik@redhat.com>

Check all that apply:

### Description of Additions:

New spec `insights_client_tags` and its parser was added to collect data from `/etc/insights-client/tags.yaml`.

### DIFF
--- a/insights/parsers/insights_client_tags.py
+++ b/insights/parsers/insights_client_tags.py
@@ -1,0 +1,52 @@
+"""
+InsightsClientTags - file ``/etc/insights-client/tags.yaml``
+=======================================================================
+"""
+
+from insights import parser
+from insights.core import YAMLParser
+from insights.specs import Specs
+
+
+@parser(Specs.insights_client_tags)
+class InsightsClientTags(YAMLParser):
+    """
+    This class parses the YAML-formatted file `/etc/insights-client/tags.yaml`.
+    The file is created by the user after running this command:
+
+    `insights-client --group=<group-name>`
+
+    The command above creates the file and adds an initial group tag
+    as defined by the `--group` option.
+    Other tags are added manually by the user.
+
+    Sample input data:
+
+        group: eastern-sap
+        location: Boston
+        description:
+        - RHEL8
+        - SAP
+        key 4: value
+
+    Examples:
+        >>> type(tags)
+        <class 'insights.parsers.insights_client_tags.InsightsClientTags'>
+        >>> tags.data
+        {'group': 'eastern-sap', 'location': 'Boston', 'description': ['RHEL8', 'SAP'], 'key 4': 'value'}
+        >>> tags.group
+        'eastern-sap'
+        >>> tags.get('location')
+        'Boston'
+        >>> tags.get('key 1', 'not defined')
+        'not defined'
+    """
+
+    @property
+    def group(self):
+        """
+        Return value of the group tag or None if the group tag is missing.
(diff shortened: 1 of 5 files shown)


insights/parsers/insights_client_tags.py
```diff
@@ -0,0 +1,52 @@
...
+    This class parses the YAML-formatted file `/etc/insights-client/tags.yaml`.
+    The file is created by the user after running this command:
+
+    `insights-client --group=<group-name>`
+
+    The command above creates the file and adds an initial group tag
+    as defined by the `--group` option.
+    Other tags are added manually by the user.
+
+    Sample input data:
```
xiangce, line 23: You may want to use a double colon `::` to put the following data into a literal block 

see: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

xiangce (commented):
The new specs need to go through the spec approval process at first, I'll send the instructions to you.

xiangce:
You should also add its doc entry to the `docs/shared_parsers_catalog/` directory and fix the CI test errors.

insights/parsers/insights_client_tags.py
```diff
@@ -0,0 +1,52 @@
...
+    Other tags are added manually by the user.
+
+    Sample input data::
+
+        group: eastern-sap
+        location: Boston
+        description:
+        - RHEL8
+        - SAP
+        key 4: value
```
jobselko, line 30: I would remove this line because the example shows "real data", so having value as `key 4: value` did not correspond to `group: eastern-sap`, `location: Boston`.

insights/parsers/insights_client_tags.py
```diff
@@ -0,0 +1,52 @@
...
+        >>> tags.group
+        'eastern-sap'
+        >>> tags.get('location')
+        'Boston'
+        >>> tags.get('key 1', 'not defined')
+        'not defined'
+    """
+
+    @property
+    def group(self):
```
jobselko, line 46: Is it necessary to have a `@property` for this case? The attribute can be set directly in `__init__`.

Moreover, it seems this only duplicates already existing value.
See `assert tags.group == 'eastern-sap'` and `assert tags.get("group") == 'eastern-sap'` in tests.

insights/parsers/tests/test_insights_client_tags.py
```diff
@@ -0,0 +1,48 @@
...
+    assert tags.group == 'eastern-sap'
+    assert tags.get('location') == 'Boston'
+    assert tags.get('description') == ['RHEL8', 'SAP']
+    assert tags.get('key 1') is None
+
+    cls = tags.__class__
+    name = ".".join([cls.__module__, cls.__name__])
+    with pytest.raises(ParseException) as e:
+        InsightsClientTags(context_wrap(TAGS_UNPARSABLE))
+    assert "{0} couldn't parse yaml.".format(name) in str(e.value)
```
jobselko, line 42: I would remove this last assert line because the exception message in `YAMLParser` could be changed in the future, and we should not test this behaviour here.

jobselko (changes requested):

tciernik:
A parser for `tags.yaml` already exists in `insights-core`. The parser was not working, but is fixed now. This one is not needed anymore and I am closing this PR.

PR #3353, opened 2022-03-09
feat: New parser for /usr/bin/od -An -t d /dev/cpu_dma_latency

crackcodecamp:

### All Pull Requests:

Check all that apply:

*Add your description here*


crackcodecamp:
@xiangce Please help to review the PR.

insights/parsers/hexdump.py
```diff
@@ -0,0 +1,50 @@
...
+        ['00000000  01 00 00 00                                       |....|', '00000004']
+        >>> hex_dump.hex_value
+        '00 00 00 01'
+        >>> hex_dump.dec_value
+        1
+    """
+
+    def parse_content(self, content):
+        # The self.data returns the raw output of the command.
+        self.data = content
```
xiangce, line 40: `data` is not a good attribute for parsers anymore, since the name is a bit ambiguous and cannot indicate the real structure it stored.
If you want to add an attribute to return the `RAW` content, better rename it to `raw_content` or other.

xiangce (changes requested):
@crackcodecamp 

1. This requires going through the new spec approval process at first.
2. `HexDump` is not a good spec/parser name to me. The command `hexdump` could run against many different target files, better rename the spec and create a new parser based on the base on `HexDump`.
3. And since `hexdump` is a utility command and can be run against different files, you might need to do some updates to the base `HexDump` to handle multiple lines output.

insights/parsers/hexdump.py
xiangce: As discussed, I think it would be better to implement the parser more generically. The `hexdump` is a command to dump any kind of file to the `hex` format. In case the target file (`/dev/cpu_dma_latency` in this case) is different from the examples or polluted, and results in the `hexdump` doesn't output as expected in the `typical output`, the parser will fail. Please research a bit about the output of `hexdump` and refine the parser.

crackcodecamp: Sure.

crackcodecamp:
@xiangce The PR updated as per new spec. Please review.

insights/parsers/hexdump.py
crackcodecamp: Replaced `hexdump` with `od` command.

xiangce (approved):

psachin (approved):
Good to me.

PR #3354, opened 2022-03-09
fix: Fix slowness on RHEL 8 by simplifying looping over pkgs

ryan-blakley:
Check all that apply:

The code had an issue where it was looping over the dictionary updates that was being updated in another loop. So it was checking the beginning pkgs over and over each loop for updates which was un-necessary and caused the slowness. So I simplified the code down to the least amount of looping and variables.


xiangce (approved):
Looks good to me

Issue #3356, opened 2022-03-11
When filters are added for a spec which is collected by glob_file, the parser can not get the files when one of the files does not contain the filtered lines


huali027:
An example is:
rsyslog_conf = RegistryPoint(filterable=True, multi_output=True)
rsyslog_conf = glob_file(["/etc/rsyslog.conf", "/etc/rsyslog.d/*.conf"])
When I tested it, I found the file "/etc/rsyslog.conf" contains the filtered lines, but the file "/etc/rsyslog.d/listen.conf" doesn't contain the filtered lines.
In the created archive, it is like:
`{"exec_time": 6.890296936035156e-05, "ser_time": 0.05234384536743164, "errors": ["Traceback (most recent call last):\n  File \"/var/lib/insights/newest.egg/insights/core/serde.py\", line 195, in dehydrate\n    doc[\"results\"] = marshal(value, root=self.data, pool=self.pool)\n  File \"/var/lib/insights/newest.egg/insights/core/serde.py\", line 105, in marshal\n    return [f(t) for t in v]\n  File \"/var/lib/insights/newest.egg/insights/core/serde.py\", line 85, in serialize\n    \"object\": to_dict(obj, root=root),\n  File \"/var/lib/insights/newest.egg/insights/core/spec_factory.py\", line 1041, in serialize_text_file_provider\n    rc = obj.write(dst)\n  File \"/var/lib/insights/newest.egg/insights/core/spec_factory.py\", line 288, in write\n    p.write(dst)\n  File \"/var/lib/insights/newest.egg/insights/util/subproc.py\", line 155, in write\n    six.reraise(be.__class__, be, sys.exc_info()[2])\n  File \"/var/lib/insights/newest.egg/insights/util/subproc.py\", line 151, in write\n    raise CalledProcessError(rc, self.cmds[0], \"\")\nCalledProcessError: (1, ['grep', '-F', '$ActionQueueFileName\\n)\\n(\\n{\\n}\\nimjournal', '/etc/rsyslog.d/listen.conf'], '')\n"], "name": "insights.specs.Specs.rsyslog_conf", "results": null}`
As a result, the parser can't continue since it doesn't get the collected files.
However, the expected result is the parser can get the "/etc/rsyslog.conf" file since it contains what we wanted.

xiangce:
Thanks for reporting this issue, I do think this is a bug of the core_collection, it should skip the files which can be collected of the `glob_file` when there are some unable to collect.

[...]

PR #3357, opened 2022-03-11
fix: Keep the results once one of them is good

huali027:
* When some spec is collecting several files, we should keep the results
  once one of them is good, and only set results to empty when all of them
  fails with errors
* fix: #3356 

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/core/serde.py
+++ b/insights/core/serde.py
@@ -95,15 +95,37 @@ def deserialize(data, root=None):
 
 
 def marshal(v, root=None, pool=None):
+    data = {
+        'errors': [],
+        'results': None
+    }
+
+    def call_serializer(func, value):
+        result = error = None
+        try:
+            result = func(value)
+        except Exception:
+            error = traceback.format_exc()
+        if result:
+            if isinstance(data['results'], list):
+                data['results'].append(result)
+            else:
+                data['results'] = result
+        if error:
+            data['errors'].append(error)
+
     if v is None:
-        return
+        return data
     f = partial(serialize, root=root)
     if isinstance(v, list):
+        data['results'] = []
         if pool:
-            return list(pool.map(f, v))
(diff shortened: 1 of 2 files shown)


huali027:
fixes  https://github.com/RedHatInsights/insights-core/issues/3356

huali027:

huali027:
> @xiangce Please help to review, thanks.

@xiangce Please hold on to the review, I forgot to handle "pool".

huali027:
@xiangce  Please help to review, thanks.

insights/core/serde.py
```diff
@@ -95,15 +95,37 @@ def deserialize(data, root=None):
 
 
 def marshal(v, root=None, pool=None):
+    data = {
+        'errors': [],
+        'results': None
+    }
```
xiangce, line 101: This `data` makes the code a bit complex, why not just return a tuple from the inner `call_serializer` function?
e.g.
```python
        try:
            return func(value), None
        except Exception:
            return None, traceback.format_exc()
```

huali027: @xiangce
It is to keep consistent with before. When there is one result for the spec, we should set it to the result, instead of a list.
As we discussed,  I found we'll return different types for the `marshal` function in this way. And I have to check the values based on the returned type. 
And in the test file, when we test the `marchal` function, after getting the result, we also need to check the returned value type. So I'm wondering if it is good to return different types for one function.

huali027: @xiangce updated as discussed, please help to review, thanks.

xiangce (approved):

xiangce:
@bfahr - Since this PR updates the basic function of the core, please have a review. Thanks.

bfahr (approved):
I've tested it and it seems to have fixed the problem.

PR #3358, opened 2022-03-11
feat: Add systctl.d spec, parser, and combiner

ryan-blakley:
Check all that apply:

Currently core only collects /etc/sysctl.conf but none of the conf files in the /etc/sysctl.d/ directory. So I added a spec for them. Updated the parsers to use a base class. Added a new combiner for all of the conf files in the /etc/sysctl.d/ directory.


insights/combiners/sysctl_conf.py
```diff
@@ -0,0 +1,56 @@
...
+Sysctl configuration files
+==========================
+"""
+import re
+
+from insights.core.plugins import combiner
+from insights.parsers.sysctl import SysctlDConf
+
+
+@combiner(SysctlDConf)
```
xiangce, line 11: Hi @ryan-blakley, I think it would be better to just add a `SysctlConf` combiner that combines both the `SysctlConf` parser and `SysctlDConf` parsers, and take the `SysctlDConf` parsers in the high priority (take their values as the final values). See the [sysctl-man-page:CONFIGURATION DIRECTORIES AND PRECEDENCE](https://man7.org/linux/man-pages/man5/sysctl.d.5.html)  part.

And I think all the existing rules should also be updated to use the new combiner once it's merged. What do you think?

ryan-blakley: Hey @xiangce so I contemplated adding the `SysctlConf` parser to the combiner also. But I decided not to, because technically `/etc/sysctl.conf` only gets read because the `/etc/sysctl.d/99-sysctl.conf` file is a symlink to it. If that symlink is missing `/etc/sysctl.conf` doesn't get read in. I guess technically we could deprecate the `SysctlConf` parser for the combiner.

Based on that man page link, do you think I need to add a spec for the `/usr/lib/sysctl.d/*.conf` files also, since a third party package could add sysctls via that path? I see there are conf files on my minimal RHEL 7 test system in that directory.

Thank you for mentioning the order part, I will need to add some logic to update the dictionary based on the file name order.

Yes I think all of the rules that use the `SysctlConf` parser should be updated to use the new combiner once merged.

xiangce: Hey @ryan-blakley - Thanks for the sharing. 
1. Regarding the `/etc/sysctl.d/99-sysctl.conf`. Although it's a fact in RHEL 7/8/(and 9) that the `/etc/sysctl.d/99-sysctl.conf` will be created and point to the `/etc/sysctl.conf` by default, it should not be a blocker for the combiner to take it as a source. Since it's a configuration file, it's possible for the users (customers) to remove the symbol link file intentionally or unintentionally, which results in rules' false positive in the specific situation.
And I didn't find this symbol link file in the minimal RHEL 6.10 system which is still supported by Red Hat. So I think we'd better add it to the combiner, after all, even if it's a symbol link to the `/etc/sysctl.conf`, the combiner's result won't be wrong. 
2. I agree with you that we should add a spec for the `/usr/lib/sysctl.d/*.conf` files together and add its parser to the combiner.

Thanks.

ryan-blakley: Ahhh I didn't see that on RHEL 6 that it reads /etc/sysctl.conf, I added it to the combiner. I also added the /usr/lib/sysctl.d spec, parser, and added it to the combiner.

insights/combiners/sysctl_conf.py
```diff
@@ -0,0 +1,71 @@
...
+        '1'
+        >>> "vm.dirty_ratio" in sysctl_conf
+        True
+        >>> sysctl_conf.search("domainname")
+        {'kernel.domainname': 'example.com'}
+    """
+    def __init__(self, sysctl_conf, sysctl_d_confs_usr, sysctl_d_confs_etc):
+        super(SysctlConfs, self).__init__()
+        if sysctl_conf:
+            self.update(sysctl_conf)
```
xiangce, line 47: @ryan-blakley - it seems that the current order doesn't match with the sysctl-manual
```
# man sysctl.d

CONFIGURATION DIRECTORIES AND PRECEDENCE
       Configuration files are read from directories in /etc/, /run/,
       /usr/local/lib/, and /usr/lib/, in order of precedence, as listed
       in the SYNOPSIS section above. Files must have the ".conf"
       extension. Files in /etc/ override files with the same name in
       /run/, /usr/local/lib/, and /usr/lib/. Files in /run/ override
       files with the same name under /usr/.
```

And the `sysctl --system` gives the following output:
```
# sudo sysctl --system
* Applying /usr/lib/sysctl.d/10-default-yama-scope.conf ...
kernel.yama.ptrace_scope = 0
* Applying /usr/lib/sysctl.d/50-coredump.conf ...
kernel.core_pattern = |/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h %e
kernel.core_pipe_limit = 16
* Applying /usr/lib/sysctl.d/50-default.conf ...
kernel.sysrq = 16
kernel.core_uses_pid = 1
kernel.kptr_restrict = 1
net.ipv4.conf.all.rp_filter = 1
net.ipv4.conf.all.accept_source_route = 0
net.ipv4.conf.all.promote_secondaries = 1
net.core.default_qdisc = fq_codel
fs.protected_hardlinks = 1
fs.protected_symlinks = 1
* Applying /usr/lib/sysctl.d/50-libkcapi-optmem_max.conf ...
net.core.optmem_max = 81920
* Applying /usr/lib/sysctl.d/50-pid-max.conf ...
kernel.pid_max = 4194304
* Applying /usr/lib/sysctl.d/60-libvirtd.conf ...
fs.aio-max-nr = 1048576
* Applying /usr/lib/sysctl.d/88-abc.conf ...
kernel.shmmax = 18446744073692774390
* Applying /etc/sysctl.d/99-abc.conf ...
kernel.shmmax = 18446744073692774380
* Applying /etc/sysctl.d/99-sysctl.conf ...
kernel.shmmax = 18446744073692774377
* Applying /etc/sysctl.d/99-xyz.conf ...
kernel.shmmax = 18446744073692774388
* Applying /etc/sysctl.conf ...
kernel.shmmax = 18446744073692774377
```

```
# sudo sysctl -a | grep shmmax
kernel.shmmax = 18446744073692774377
```

It seems that the correct order in the combiner should be:
```
/usr/lib/sysctl.d/*.conf  -> /etc/sysctl.d/*.conf -> /etc/sysctl.conf
```

insights/combiners/tests/test_sysctl_conf.py
```diff
@@ -0,0 +1,81 @@
...
+    assert r['kernel.domainname'] == 'example.com'
+    assert r['kernel.modprobe'] == '/sbin/mod probe'
+    assert 'kernel.domainname.invalid' not in r
+    assert r.search("dirty") == {"vm.dirty_ratio": "15", "vm.dirty_background_ratio": "5"}
+    assert r.search("test") == {}
+
+
+def test_all_sysctl_conf_different_order():
+    conf_1 = sysctl.SysctlConf(context_wrap(SYSCTL_CONF_TEST1, path=SYSCTL3_CONF_FILENAME))
+    conf_2 = sysctl.SysctlDConfUsr(context_wrap(SYSCTL_CONF_TEST2, path=SYSCTL1_CONF_FILENAME))
```
xiangce, line 62: Do you think it would be better to change the `path` of the `SysctlDConfUsr` to a real path of `/usr/lib/sysctl.d/xx-xx.conf`?

insights/combiners/tests/test_sysctl_conf.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+
+SYSCTL1_CONF_FILENAME = "/etc/sysctl.d/10-test.conf"
+SYSCTL2_CONF_FILENAME = "/etc/sysctl.d/20-test.conf"
+SYSCTL3_CONF_FILENAME = "/etc/sysctl.d/30-test.conf"
+SYSCTL4_CONF_FILENAME = "/etc/sysctl.d/99-test.conf"
+
+
+def test_all_sysctl_conf():
+    conf_1 = sysctl.SysctlConf(context_wrap(SYSCTL_CONF_TEST1, path=SYSCTL1_CONF_FILENAME))
```
xiangce, line 48: It would be better that the path of `SysctlConf` should be `/etc/sysctl.conf`.

ryan-blakley: Good catch I definitely missed that, I'll correct the paths based on the parser used.

xiangce (approved):
Looks good to me. Thanks @ryan-blakley

PR #3359, opened 2022-03-14
fix: Switch to reading crontab file rather than run the command

ryan-blakley:
Check all that apply:

Based on the complaints in bz [1759080](https://bugzilla.redhat.com/show_bug.cgi?id=1759080), I switched the crontab specs to read the crontab files instead of running the command since it does user id look ups every time the command is ran. But by switching to simple_file the spec would be skipped if the file doesn't exist. I also added the nova_conf file as a dep for the nova_uid so that the id look up doesn't occur unless nova is installed.


insights/specs/insights_archive.py
```diff
@@ -106,7 +106,6 @@ class InsightsArchiveSpecs(Specs):
     ipv4_neigh = simple_file("insights_commands/ip_-4_neighbor_show_nud_all")
     ipv6_neigh = simple_file("insights_commands/ip_-6_neighbor_show_nud_all")
     iscsiadm_m_session = simple_file("insights_commands/iscsiadm_-m_session")
-    keystone_crontab = simple_file("insights_commands/crontab_-l_-u_keystone")
```
bfahr, line 109: @ryan-blakley since you are changing the source of the data to be collected, we need to leave the archive spec in place, and change it to a `first_file` to ensure that it will be read from the archive regardless of whether the client collected it with the old spec or the new spec.

insights/specs/insights_archive.py
```diff
@@ -182,7 +181,6 @@ class InsightsArchiveSpecs(Specs):
     netstat_s = simple_file("insights_commands/netstat_-s")
     nmcli_conn_show = simple_file("insights_commands/nmcli_conn_show")
     nmcli_dev_show = simple_file("insights_commands/nmcli_dev_show")
-    nova_crontab = simple_file("insights_commands/crontab_-l_-u_nova")
```
bfahr, line 185: Same comment as above.

bfahr (changes requested):
Just the two comments, otherwise looks good.

insights/specs/insights_archive.py
ryan-blakley: ack updated

bfahr (approved):

PR #3364, opened 2022-03-17
Fix: [insights-client] insights-client --status returns 1 when it is unregistered ESSNTL-2001

ahitacat:

### All Pull Requests:

Check all that apply:

`insights-client --status` was exiting good if it was registered, and otherwise. But the correct exit should be bad if a problem happened and good whether it is registered and unregistered.

Follow discussion in [ESSNTL-2001](https://issues.redhat.com/browse/ESSNTL-2001)

Clarification Edit (I copied here @subpop 's comment ):
To clarify, this PR is changing the behavior of insights-client so that regardless of the output of the --status command, the process will exit with a 0 if the process executed successfully. Prior to this change, the process would exit with a 1 if the output was "unregistered" and a 0 if the output was "registered". This change brings the client in more compliance with normal command line utility behaviors.

### DIFF
--- a/insights/client/phase/v1.py
+++ b/insights/client/phase/v1.py
@@ -233,10 +233,9 @@ def post_update(client, config):
     if config.status:
         if reg_check:
             logger.info('This host is registered.')
-            sys.exit(constants.sig_kill_ok)
         else:
             logger.info('This host is unregistered.')
-            sys.exit(constants.sig_kill_bad)
+        sys.exit(constants.sig_kill_ok)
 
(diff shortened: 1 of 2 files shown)


subpop:
To clarify, this PR is changing the behavior of `insights-client` so that regardless of the output of the `--status` command, the process will exit with a `0` if the process executed successfully. Prior to this change, the process would exit with a `1` if the output was "unregistered" and a `0` if the output was "registered". This change brings the client in more compliance with normal command line utility behaviors.

PR #3360, opened 2022-03-16
fix: Enhance parser Grub2Config

wushiqinlou:
### All Pull Requests:

Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [ ] Is this PR to correct an issue?
* [x] Is this PR an enhancement?

### Complete Description of Additions/Changes:

Spec `grub2_cfg` can also exist on RHEL8, so update to make it can be used for RHEL8 systems.

PR #3367, opened 2022-03-25
Revert "fix: Enhance parser Grub2Config"

ryan-blakley:
### All Pull Requests:

Check all that apply:

Reverts RedHatInsights/insights-core#3360, this commit broke my internal Insights-Facts app, and I imagine would cause issues with rule hits. This commit removed the BootLoaderEntries combiner from the GrubConf combiner, which means any rules only utilizing the GrubConf combiner looking for cmdline entries will no longer return anything for RHEL 8+ systems that are using the default blscfg setting. It looks like there is only 1 rule utilizing both combiners, and around 20+ rules that utilizes only the GrubConf combiner and not both. Instead of removing the BootLoaderEntries combiner, there needs to be a check if grub2.boot_entries exist, if so set self.grub = _grub2 otherwise self.grub = grub_bles.

### DIFF
--- a/insights/combiners/grub_conf.py
+++ b/insights/combiners/grub_conf.py
@@ -92,11 +92,11 @@ class BootLoaderEntries(object):
             raise SkipComponent()
 
 
-@combiner([Grub1Config, Grub2Config, Grub1EFIConfig, Grub2EFIConfig],
+@combiner([Grub1Config, Grub2Config, Grub1EFIConfig, Grub2EFIConfig, BootLoaderEntries],
           optional=[InstalledRpms, CmdLine, LsSysFirmware, RedHatRelease])
 class GrubConf(object):
     """
-    Process Grub configuration v1, v2 based on which type is passed in.
+    Process Grub configuration v1, v2, and BLS based on which type is passed in.
 
     Attributes:
         version (int): returns 1 or 2, version of the GRUB configuration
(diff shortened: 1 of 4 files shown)


ryan-blakley:
@xiangce and @wushiqinlou could you please review this. I put my reasoning in the description. If I missed something please let me know.

wushiqinlou:
@ryan-blakley - If blscfg is not enabled and grub2.boot_entries exist, it is not correct to set "self.grub = grub_bles". I think we can enhance this combiner with following logic:

~~~
if blscfg is enabled and grub2.boot_entries exist:
    self.grub = _grub2 
else:
    self.grub = grub_bles
~~~

Please let me know your thought.

ryan-blakley:
@wushiqinlou Right that's what I was saying in the description. I think the below code should work to achieve what you're wanting, and the RedHatRelease combiner could be removed.
```
if grub_bles and not _grub2:
    self.grub = grub_bles
elif grub_bles and _grub2 and "blscfg" in _grub2.get("configs", ""):
    self.grub = grub_bles
else:
    # Check grub version via installed-rpms
    if rpms:
```

wushiqinlou:
@ryan-blakley - For condition "if grub_bles and not _grub2:", if grub.cfg does not exist, we do not know if "blscfg" is enabled or not, it is not accurate to set "self.grub = grub_bles". In this case, it is better to run into condition "if self.grub is None and cmdline:".

ryan-blakley:
@wushiqinlou since we're deviating off the topic of this PR, I'll move my response to your PR. I feel this PR should be merged then your PR can be rebased, and the discussion on how to enhance it can continue there.

xiangce:
@ryan-blakley - Sorry for bringing this trouble to you. we'll merge this PR at first when the `revert` MRs for related rules repo are ready.

xiangce:
@ryan-blakley - The related MRs for rules were ready and merged together with this PR. Please check if your Insights-Facts app gets back again. Thanks.

ryan-blakley:
@xiangce Thank you I was able to rebuild the rpm locally, and confirmed the revert fixed my Insights-Facts app. Once the release is done tomorrow I can push an update for it to use version 269+ of core. Thanks again.

PR #3371, opened 2022-03-29
fix: Fix deprecation warning for using ET.getiterator

ryan-blakley:
Check all that apply:

Using the xml element tree getiterator function has been deprecated for a while it seems. To get rid of the warning I updated to using iter wrapped in a list which is all the getiterator function did.

### DIFF
--- a/insights/specs/datasources/candlepin_broker.py
+++ b/insights/specs/datasources/candlepin_broker.py
@@ -56,7 +56,7 @@ def candlepin_broker(broker):
         if content:
             root = ET.fromstring('\n'.join(content))
             # remove namespace before save to avoid urgly search
-            for node in root.getiterator():
+            for node in list(root.iter()):
                 prefix, has_namespace, postfix = node.tag.rpartition('}')
                 if has_namespace:
                     node.tag = postfix


insights/specs/datasources/candlepin_broker.py
```diff
@@ -56,7 +56,7 @@ def candlepin_broker(broker):
         if content:
             root = ET.fromstring('\n'.join(content))
             # remove namespace before save to avoid urgly search
-            for node in root.getiterator():
+            for node in list(root.iter()):
```
bfahr, line 59: @ryan-blakley since `.iter()` was added in python 2.7, what about python 2.6 support?  I wonder if the existing tests are not covering this code?

ryan-blakley: @bfahr I think 2.6 clients use the ET that's in contrib which is the 2.7 version backported I thought.

bfahr: @ryan-blakley that may be the case, thanks for checking.

bfahr (approved):

PR #3372, opened 2022-03-29
fix: Update bond and bond_dynamic_lb spec

ryan-blakley:
Check all that apply:

Any file under /proc/net/bonding/ is a valid bond interface. The bond interface names don't have to start with bond, so change the glob to collect all file under the directory. Updated the bond_dynamic_lb spec to glob all interface names, since the /sys/class/net/*/bonding/ directory only exist for bond devices.


insights/specs/default.py
```diff
@@ -96,8 +96,8 @@ class DefaultSpecs(Specs):
     azure_instance_plan = simple_command("/usr/bin/curl -s -H Metadata:true http://169.254.169.254/metadata/instance/compute/plan?api-version=2018-10-01&format=json --connect-timeout 5", deps=[IsAzure])
     bios_uuid = simple_command("/usr/sbin/dmidecode -s system-uuid")
     blkid = simple_command("/sbin/blkid -c /dev/null")
-    bond = glob_file("/proc/net/bonding/bond*")
-    bond_dynamic_lb = glob_file("/sys/class/net/bond[0-9]*/bonding/tlb_dynamic_lb")
+    bond = glob_file("/proc/net/bonding/*")
+    bond_dynamic_lb = glob_file("/sys/class/net/*/bonding/tlb_dynamic_lb")
```
bfahr, line 100: @ryan-blakley Please review and possibly add some tests to the parsers for these two specs to ensure that they still work with the broader glob filter.

bfahr (changes requested):

insights/specs/default.py
xiangce: Hi @bfahr, would you please review it again to check if the new commits work for you? Thanks.

xiangce: Hi @bfahr, would you please review it again to check if the new commits work for you? Thanks.

PR #3379, opened 2022-04-08
fix: Httpd tracebacks displaying when the client is ran

ryan-blakley:
Check all that apply:

Fixed a parsing issue when spaces are added after a greater than symbol. Moved the newer httpd conf parsers to the httpd_conf parser file instead of in the combiner. If they're in the combiner exceptions aren't handled properly and are printed to the console instead. Converted the newer parsers to using a base class since they were using duplicate code. Removed the unused functions from the combiner that I think were used when executing it by itself. Fixed several line to long flake errors in the tests.


xiangce (commented):
Hey @ryan-blakley, Since the original `HttpdConf `was deprecated for a very long time already (more than 3 years, I think), I think it's OK to remove them from the parser and rename the new one as `HttpdConf` instead of `Httpd_Conf ` since "Xx_Xx" is not a so good python name style for class. What do you think?

And the same for the deprecated combiners, e.g. the `HttpdConfAll` can also be removed. Would you please remove them?

BTW, 
> Moved the newer httpd conf parsers to the httpd_conf parser file instead of in the combiner. If they're in the combiner exceptions aren't handled properly and are printed to the console instead.

This is really a nice finding and a good point that we should pay attention to in the future. Thanks.

xiangce:
And, I also found the `insights.combiners.httpd_conf._HttpdConf` is widely used in the test of the prodsec rules, maybe it could a choice to keep exposing it in the combiner, to avoid breaking these tests. Otherwise, we should inform the prodsec team to update the tests while Merging this PR.

ryan-blakley:
@xiangce Yeah the name was brought up in our PR review meeting I was going to update it today, but I'm all for removing the deprecated parser and naming the new one the same as the old one. 

Regarding the prodsec test, thanks for checking that I completely forgot, I will reach out to them to try and align an update on their side with us merging this PR.

ryan-blakley:
@xiangce I removed the old deprecated parser and combiner and renamed the moved parser to HttpdConf. Let me know if it looks good now. I also sent an email to the prodsec team about the failing tests.

kgrant-rh:
The relevant Product Security rules and tests have been updated.

bfahr (approved):

vdymna (approved):

PR #3381, opened 2022-04-11
Add new system_user_dirs datasource and parser

jobselko:

### All Pull Requests:

Check all that apply:

We are working on new rules (CVE-2021-35937, CVE-2021-35938, CVE-2021-35939), which need a new parser and data source for checking non-root owned packaged directories.

CI is failing because the datasource uses the rpm module, which is the system package and the current virtual environment does not have it.

### DIFF
--- a/docs/custom_datasources_index.rst
+++ b/docs/custom_datasources_index.rst
@@ -131,6 +131,14 @@ insights.specs.datasources.ssl_certificate
     :show-inheritance:
     :undoc-members:
 
+insights.specs.datasources.system_user_dirs
+-------------------------------------------
+
+.. automodule:: insights.specs.datasources.system_user_dirs
+    :members: system_user_dirs
+    :show-inheritance:
+    :undoc-members:
+
 insights.specs.datasources.user_group
 -------------------------------------
 
--- a/insights/parsers/system_user_dirs.py
+++ b/insights/parsers/system_user_dirs.py
@@ -23,9 +23,9 @@ class SystemUserDirs(Parser):
 
         >>> type(system_user_dirs)
         <class 'insights.parsers.system_user_dirs.SystemUserDirs'>
-        >>> system_user_dirs.data
+        >>> system_user_dirs.packages
         ['ca-certificates', 'kmod', 'sssd-ldap']
     """
 
     def parse_content(self, content):
-        self.data = content
+        self.packages = content
--- a/insights/parsers/system_user_dirs.py
+++ b/insights/parsers/system_user_dirs.py
@@ -1,26 +1,31 @@
 """
 SystemUserDirs - datasource ``system_user_dirs``
 ================================================
+
+Parser for CVE-2021-35937, CVE-2021-35938, and CVE-2021-35939.
+For more information, see the ``system_user_dirs`` datasource.
 """
 
-from insights import JSONParser, parser
+from insights import Parser, parser
 from insights.specs import Specs
 
 
 @parser(Specs.system_user_dirs)
-class SystemUserDirs(JSONParser):
+class SystemUserDirs(Parser):
     """
-    Class for parsing the ``system_user_dirs`` datasource.
+    Class for enabling the data from the ``system_user_dirs`` datasource.
 
     Sample output of this datasource is::
 
-        '{"pcp-testsuite-5.3.3-1.fc33.x86_64": ["/var/lib/pcp/testsuite"]}'
+        ["ca-certificates", "kmod", "sssd-ldap"]
 
     Examples:
 
         >>> type(system_user_dirs)
         <class 'insights.parsers.system_user_dirs.SystemUserDirs'>
         >>> system_user_dirs.data
-        {'pcp-testsuite-5.3.3-1.fc33.x86_64': ['/var/lib/pcp/testsuite']}
+        ['ca-certificates', 'kmod', 'sssd-ldap']
     """
-    pass
+
+    def parse_content(self, content):
+        self.data = content
--- a/insights/parsers/system_user_dirs.py
+++ b/insights/parsers/system_user_dirs.py
@@ -1,0 +1,26 @@
+"""
+SystemUserDirs - datasource ``system_user_dirs``
+================================================
+"""
+
+from insights import JSONParser, parser
+from insights.specs import Specs
+
+
+@parser(Specs.system_user_dirs)
+class SystemUserDirs(JSONParser):
+    """
+    Class for parsing the ``system_user_dirs`` datasource.
+
+    Sample output of this datasource is::
+
+        '{"pcp-testsuite-5.3.3-1.fc33.x86_64": ["/var/lib/pcp/testsuite"]}'
+
+    Examples:
+
+        >>> type(system_user_dirs)
+        <class 'insights.parsers.system_user_dirs.SystemUserDirs'>
+        >>> system_user_dirs.data
+        {'pcp-testsuite-5.3.3-1.fc33.x86_64': ['/var/lib/pcp/testsuite']}
+    """
+    pass
--- a/insights/specs/datasources/system_user_dirs.py
+++ b/insights/specs/datasources/system_user_dirs.py
@@ -1,75 +1,106 @@
 """
-Custom datasource for checking non-root owned packaged directories
+Custom datasource for CVE-2021-35937, CVE-2021-35938, and CVE-2021-35939.
 
-Author: Florian Festi <ffesti@redhat.com>
+Authors: ffesti@redhat.com, jobselko@redhat.com
 """
-import grp
-import json
-import pwd
-import stat
 
-from insights import datasource
+import grp
+import pwd
+
+from insights.core.context import HostContext
 from insights.core.dr import SkipComponent
-from insights.core.spec_factory import DatasourceProvider
+from insights.core.plugins import datasource
+from insights.core.spec_factory import DatasourceProvider, simple_command
+from insights.specs import Specs
+
+
+class LocalSpecs(Specs):
+    """
+    Local spec used only by the system_user_dirs datasource
+    """
+    rpm_args = simple_command(
+        'rpm -qa --qf="[%{=NAME}; %{FILEMODES:perms}; %{FILEUSERNAME}; %{FILEGROUPNAME}\n]"'
+    )
 
 
 def get_shells():
-    with open('/etc/shells') as f:
-        return set((line.strip() for line in f))
+    """
+    Returns all full pathnames of valid login shells without nologins.
+    """
+    with open("/etc/shells") as file:
+        return [line.strip() for line in file if "nologin" not in line]
 
 
 def get_users():
+    """
+    Returns all users with shell specified in get_shells() except for root.
+    """
     shells = get_shells()
-    users = set()
+    users = []
 
     for user in pwd.getpwall():
-        shell = user[6].strip()
-        if shell not in shells or shell in ('/sbin/nologin', '/bin/nologin'):
+        name = user[0]
+        shell = user[6]
+
+        if name == "root" or (shell not in shells):
             continue
-        users.add(user[0])
-    users.discard("root")
+        users.append(name)
     return users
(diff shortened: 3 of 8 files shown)


jobselko:
@bfahr @xiangce Could you please help with reviewing this PR? Thank you very much!

insights/specs/datasources/system_user_dirs.py
```diff
@@ -0,0 +1,79 @@
...
+                    groups.add(entry[0])  # add groups containing a user
+    return groups
+
+
+@datasource()
+def system_user_dirs(_broker):
+    try:
+        import rpm
+    except ImportError:
+        return ""
```
xiangce, line 57: It would be more suitable to raise a `SkipComponent` here, in case the `rpm` module can't be imported successfully, to skip the following steps, and hence the parser won't be triggered.

insights/tests/datasources/test_system_user_dirs.py
```diff
@@ -0,0 +1,48 @@
+"""
+Tests for custom datasource for checking non-root owned packaged directories
+
+Author: Florian Festi <ffesti@redhat.com>
+"""
+import mock
+import rpm
```
xiangce, line 7: I met the same error  as reported in the CI, in my local test.
```
insights/tests/datasources/test_system_user_dirs.py:7: in <module>
    import rpm
E   ModuleNotFoundError: No module named 'rpm'
```
I think you might also need to make a `mock` `rpm.hdr()` in the following mocked `TransactionSet` class.

docs/custom_datasources_index.rst
```diff
@@ -91,6 +91,14 @@ insights.specs.datasources.ssl_certificate
     :show-inheritance:
     :undoc-members:
 
+insights.specs.datasources.system_user_dirs
+--------------------------------------------------
```
xiangce, line 95: Would you please adjust the length of the `---` to the same as the above title?

xiangce (changes requested):

insights/specs/datasources/system_user_dirs.py
jobselko: Yes, I was thinking exactly about this one. Maybe it would be better to move this block to the top of the file?

docs/custom_datasources_index.rst
jobselko: Of course, I overlooked it. Thanks!

insights/tests/datasources/test_system_user_dirs.py
jobselko: My local tests passed because I created a new virtual environment like this: `python3.6 -m venv venv --system-site-packages`. It works even with python3.9.

insights/specs/datasources/system_user_dirs.py
```diff
@@ -0,0 +1,79 @@
...
+                continue
+            users.add(entry[0])
+    users.discard("root")
+    return users
+
+
+def get_groups(users):
+    groups = set()
+
+    with open('/etc/group') as f:
```
ryan-blakley, line 38: Would it be possible to use the grp python module to query group info also?

insights/tests/datasources/test_system_user_dirs.py
xiangce: Okay, but with that in the test, I'm afraid that we can never get the CI being passed for the following PRs.

insights/specs/datasources/system_user_dirs.py
xiangce: It sounds good to me.

insights/tests/datasources/test_system_user_dirs.py
jobselko: Hi, I wrote to the code author, and we will try to improve the tests and remove the `rpm` module from it. Unfortunately, the CI will still fail because the `rpm` module is also used in the code.

insights/specs/datasources/system_user_dirs.py
```diff
@@ -0,0 +1,79 @@
...
+def get_shells():
+    with open('/etc/shells') as f:
+        return set((line.strip() for line in f))
+
+
+def get_users():
+    shells = get_shells()
+    users = set()
+
+    with open('/etc/passwd') as f:
```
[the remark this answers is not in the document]
jobselko: I will look at it, thank you.

insights/specs/datasources/system_user_dirs.py
jobselko: I will look at it, thank you.

insights/tests/datasources/test_system_user_dirs.py
jobselko: Hi @xiangce. I think we are fine with using the `rpm` module in tests. The `mock` module is used here only to mock testing data, and it is not necessary to mock the whole `rpm` module in tests. Do you agree with me?

bfahr: @jobselko the main concern is that we cannot rely on the CI tests due to the `rpm` dependency.  We'll have to test this manually.  The larger problem is that other developers won't be able to run pytest locally without tests failing or worse pytest failing to run due to the raised exceptions when imports fail.

insights/specs/datasources/system_user_dirs.py
```diff
@@ -0,0 +1,75 @@
...
+import stat
+
+from insights import datasource
+from insights.core.dr import SkipComponent
+from insights.core.spec_factory import DatasourceProvider
+
+try:
+    import rpm
+except ImportError:
+    raise SkipComponent
```
bfahr, line 18: @jobselko you'll have to move this import and try/catch down into your datasource.  Its being raised here prior to the core infrastructure invoking the datasource so its not being caught in the same way it would be if raised from the datasource.

insights/tests/datasources/test_system_user_dirs.py
jobselko: @bfahr I apologise for my confusion, but what do you suggest here as a solution?

xiangce: @jobselko - as mentioned above, it could be resolved by modifying the CI script:`.github/workflows/main.yml`. Please have a try according to your steps in local to all py26, py27, and py36.

xiangce:
BTW, @jobselko, since this is a new spec, for new specs to `uploader.json` or new datasource specs directly added to core (like this one), the approval in the assets repo is required. Please open an issue in the assets repo to apply the approval for this spec at the same time.

insights/tests/datasources/test_system_user_dirs.py
jobselko: In a discussion on Slack, we agreed that a better solution is to use the `rpm` command instead of an `rpm` python package. Changes into the `.github/workflows/main.yml` file would be too complicated as the CI runs on an Ubuntu machine. I am closing this thread and will rewrite the code.

jobselko:
@xiangce @ryan-blakley @bfahr Could you please review it again?

Could you please help me with what the MR in `insights-core-assets` should look like?

xiangce (changes requested):
Since this is a `datasource` spec that is not supported by the legacy uploader.json collection, it's not required to modify the `uploader.json`. But per the workflow, you need to raise an issue for it to the core-assets repo (there is an "issue template" too for issues). Thanks.

insights/parsers/system_user_dirs.py
```diff
@@ -0,0 +1,31 @@
...
+Parser for CVE-2021-35937, CVE-2021-35938, and CVE-2021-35939.
+For more information, see the ``system_user_dirs`` datasource.
+"""
+
+from insights import Parser, parser
+from insights.specs import Specs
+
+
+@parser(Specs.system_user_dirs)
+class SystemUserDirs(Parser):
```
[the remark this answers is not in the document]
jobselko: I agree with changing the attribute name. However, I cannot entirely agree with your inheritance recommendation. The `Parser` class represents a parser, not a `list`, and I do not think it should have all the `list` functionality and expose it further. If you can, we can discuss it in more detail, but I would leave it as it is now.

ryan-blakley:
@jobselko regarding the docs failure it looks like it's related to Sphinx 5.1.0 release, I'll create a PR to pin the version to 5.0.2 for now.

insights/parsers/system_user_dirs.py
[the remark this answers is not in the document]
ryan-blakley: @jobselko do you plan on adding other attributes to this parser in the future? If so I understand why you wouldn't want the parser to inherit a list. But if you don't plan on adding any other attributes I agree with @xiangce that this should probably inherit list, instead of having to use an attribute to return the list. That's the common practice we've been following for most new and updated parsers, they either inherit list or dict. You mentioned you don't think it should have all of the list functionality exposed, could you elaborate more on that as I don't understand your concern with that?

xiangce:
@jobselko - The PR that @ryan-blakley raised for the Sphinx latest version's issue was merged to the master, Please Update/Rebase this branch to the latest.

insights/parsers/system_user_dirs.py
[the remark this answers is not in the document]
jobselko: @ryan-blakley Yes, there might be a situation when I need to update the parser in the future (e.g. customer requiring more information about problematic packages).
Regarding the inheritance, I do not have any concerns. I consider it a bad design. You only want to have an attribute storing a list. You do not want to use the rest of the list functionality; inheritance is unnecessary. If you would have a custom list (with some functionality adjusted to your needs), then it makes sense to inherit from `list` or `collections.UserList`. But in your case, you are just trying to replace `self.packages = content` by `self.extend(content)`. It will not make things easier to understand or use. I agree that both solutions work, but I feel that inheritance is not the right approach here.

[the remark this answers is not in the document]
jobselko: To put it more directly: if you enable the parser to use all the functionality from the `list`, then you can, for example, use an addition operation on two parsers and so on. And that could lead to unexpected behaviour later.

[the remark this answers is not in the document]
ryan-blakley: @jobselko Thank you for your explanation, I was just curious if there was a known issue when inheriting certain built-ins since you're on the security team. I just wanted to make sure we weren't doing something incorrectly. I definitely understand your opinion and concerns about the inheritance approach, and I'm good with how you setup the parser.

bfahr:
@jobselko I agree with your preference in this case.  

@ryan-blakley and @xiangce are both correct in that we encourage use of `list` or `dict` as the base class when the parser has only one attribute.  The context for that recommendation is from a time when all parsers had only one attribute `data`, and had to implement a bunch of dunder functions, and later use the [LegacyItemAccess](https://github.com/RedHatInsights/insights-core/blob/9e7ccce78512d1af428a370f83fb9166d5f3ffed/insights/core/__init__.py#L395) mix-in.

Since `data` was not very descriptive of what was being represented by the parser we now discourage use of `data` and encourage use of meaningful names that represent their content.  We also encourage the use of representative base classes when only a single attribute is implemented, but I would say its not mandatory to use the representative base class.  However if you find that you need to implement methods around the attribute that look like `list` or `dict` methods then you should base your Parser on that class.

Bottom line, my recommendation is use of `list` and `dict` base classes for Parsers with one attribute is encouraged but optional, and up to the Parser developer.  @xiangce and @ryan-blakley what do you think?

jobselko:
@bfahr @ryan-blakley Thank you very much for sharing the background you have to consider. I definitely do not want to change your established processes. But generally, I find the inheritance too overused (for reference, see `Composition over inheritance principle` or `Liskov substitution principle`), and I am trying to use it more cautiously.

@ryan-blakley I have taken the liberty to resolve the conversation. I believe this PR is ready for merging.

xiangce (approved):
Thank you @bfahr, @ryan-blakley and @jobselko , I like this kind of open conversation we have in the PR, I believe it would make the insights-core become a better and more open project.

I'll merge it soon today.

PR #3382, opened 2022-04-12
Skip malware-detection tests on RHEL6/python2.6 (not supported)

mhuth:

This is to correct problems encountered when running the malware-detection tests on RHEL6 systems outside the automated PR check environment. Malware detection won't be supported on RHEL6 so no need to check it works in that environment anyway.

### DIFF
--- a/insights/tests/client/apps/test_malware_detection.py
+++ b/insights/tests/client/apps/test_malware_detection.py
@@ -1,5 +1,7 @@
 import os
 import re
+import sys
+
 import pytest
 import yaml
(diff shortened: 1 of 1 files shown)


bfahr (approved):
This looks good to me. @Glutexo I'll wait until you review before merging.

Glutexo:
Although Python 2.6 / RHEL 6 is not supported, adding support here is very easy. But the original problem with the tests is that it uses an external utility and is dependent on an actual state of the machine.

PR #3396, opened 2022-04-22
When insights client is killed the directories in /var/tmp are not removed rhbz#2009773

ahitacat:
### All Pull Requests:

Check all that apply:

[rhbz#200973](https://bugzilla.redhat.com/show_bug.cgi?id=2009773) 
When insights-client is killed with a SIGKILL signal, as does the `oom-killer`, the directories for the archive created in `/var/tmp` aren't removed. That leaves a lot of directories in this path, increasing the size of the `/var/tmp` folder.

This fix adds the following:
- Handle SIGTERM signal
- Archive is created in a directory with the same prefix. Each time a new archive is created, it removes the previous one if it exists.
- `--keep-archive` will be storing the archive outside `/var/tmp`, and is up to each user to manage those stored archives
- Tests for this new performance


insights/client/archive.py
```diff
@@ -223,12 +226,40 @@ def cleanup_tmp(self):
...
+            logger.debug("Deleting previous archive %s", file)
+            shutil.rmtree(file, True)
+
+    def storing_archive(self):
+        keep_archive_dir = '/var/cache/insights-client'
+        if not os.path.exists(keep_archive_dir):
+            try:
+                os.makedirs(keep_archive_dir)
+            except OSError as e:
+                logger.error(e)
```
subpop, line 254: We often get bug reports complaining about stack traces in console output. Maybe we can change this line to `logger.debug`? Does printing the exception here print the entire stack trace or just the exception's class & message?

subpop (changes requested):

insights/client/archive.py
ahitacat: It prints the stack trace but it was a leftover of debugging it, also I have noticed the test wasn't good for this method

xiangce:
@subpop, @ahitacat - May I know the latest status of this PR? Can we merge and release it this week? Thanks.

ahitacat:
@xiangce This PR still needs review.

ahitacat:
@xiangce this PR is reviewed and can be merged

Issue #3395, opened 2022-04-22
LsVarLog parser does not handle group name containing space


rahulXs:
[LsVarLog parser](https://insights-core.readthedocs.io/en/latest/_modules/insights/parsers/ls_var_log.html#LsVarLog) is returning `ValueError` while parsing lines containing group name with space.

```
# Example 1
-rw-r--r--.  1 root   domain users@CR.LOCAL        36548 Jan 13 10:43 Xorg.1.log
-rw-r--r--.  1 root   domain users@CR.LOCAL        28245 Jan 13 10:38 Xorg.1.log.old
# Example 2
-rw-r--r--.  1 root   domain users        40893 Jan 31 16:34 Xorg.1.log
-rw-r--r--.  1 root   domain users        36911 Dec 20 09:49 Xorg.1.log.old
```
Error:
```
File "/opt/venv/lib64/python3.6/site-packages/insights/core/__init__.py", line 1736, in listing_of
    return self.listings[directory]['entries']
  File "/opt/venv/lib64/python3.6/site-packages/insights/core/ls_parser.py", line 228, in __getitem__
    self._load()
  File "/opt/venv/lib64/python3.6/site-packages/insights/core/ls_parser.py", line 199, in _load
    rest = parse_non_selinux(parts[1:])
  File "/opt/venv/lib64/python3.6/site-packages/insights/core/ls_parser.py", line 54, in parse_non_selinux
    result["size"] = int(size)
ValueError: invalid literal for int() with base 10: 'users@CR.LOCAL'
```

psachin:
@rahulXs Thanks for raising this. I have verified this issue.

PR #3406, opened 2022-05-10
Draft: Parse SSSD group listings

psachin:
The SSSD group listings breaks the present implementation by parsing the domain
group as the file-size.

Resolves: #3395

### All Pull Requests:

Check all that apply:

*Add your description here*

### DIFF
--- a/insights/core/ls_parser.py
+++ b/insights/core/ls_parser.py
@@ -140,12 +140,12 @@ def parse_rhel8_selinux(parts):
 
 
 def parse_sssd(parts):
-    links, owner, group1, group2, last = parts
+    links, owner, group, domain, last = parts
     size, rest = last.split(None, 1)
     result = {
         "links": int(links),
         "owner": owner,
-        "group": group1 + " " + group2,
+        "group": group + " " + domain,
         "size": int(size),
         "date": rest[:12],
         "name": rest[13:]
@@ -210,7 +210,7 @@ class Directory(dict):
                 # always have at least two pieces separated by ':'.
                 if ":" in line.split()[4]:
                     rest = parse_rhel8_selinux(parts[1:])
-                elif "@" in line.split()[4]:
+                elif line.split()[4].startswith("users") or "@" in line.split()[4]:
                     parts = line.split(None, 5)
                     rest = parse_sssd(parts[1:])
                 else:
--- a/insights/core/ls_parser.py
+++ b/insights/core/ls_parser.py
@@ -139,6 +139,21 @@ def parse_rhel8_selinux(parts):
     return result
 
 
+def parse_sssd(parts):
+    links, owner, group1, group2, last = parts
+    size, rest = last.split(None, 1)
+    result = {
+        "links": int(links),
+        "owner": owner,
+        "group": group1 + " " + group2,
+        "size": int(size),
+        "date": rest[:12],
+        "name": rest[13:]
+    }
+
+    return result
+
+
 PASS_KEYS = set(["name", "total"])
 DELAYED_KEYS = ["entries", "files", "dirs", "specials"]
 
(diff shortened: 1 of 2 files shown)


xiangce (changes requested):
It looks like this update could not cover the other example mentioned in the issue #3395
```
# Example 2
-rw-r--r--.  1 root   domain users        40893 Jan 31 16:34 Xorg.1.log
-rw-r--r--.  1 root   domain users        36911 Dec 20 09:49 Xorg.1.log.old
```
In example-2, there is no "@" in the group name field "domain users".

And I have to agree that without this specific character "@", it is a difficult issue to solve based on the current "ls_parser" resolution. Maybe it's possible to consider another overall reconstruction plan, e.g. the `insights.parsers.parse_fixed_table`. 
It's just my thoughts, you may have a better resolution.

psachin:
@xiangce 

Thanks for pointing that out. I thought I included both the examples (which I did initially) but then I removed it to test each line individually. Finally I forgot to include the second example. My bad!

I researched a bit and found that this strange user-group occurs when the sssd/krb5  DOMAIN/REALM is configured. From what I observed the owner is `domain users` followed by the `@` and DOMAIN name (`CR.LOCAL` in the example). Hence under the assumption that the group will always start with `users` in the subsequent commit I am checking if the 4th-index `startswith()` `users`.  Hopefully this should take care of the user-group issue for both the scenarios. 

As to not break the current convention I am storing `domain` as an owner and rest as the group.

>   Maybe it's possible to consider another overall reconstruction plan, e.g. the insights.parsers.parse_fixed_table.
It's just my thoughts, you may have a better resolution.

Yeah! I agree that anything new in the `ls_parser` is not easy to implement especially when we have to parse different elements depending upon the  generic and SElinux context. I thought of using `parse_fixed_table` initially but gave-up eventually because before passing the data to `parse_fixed_table` it has to be cleaned for tabled and it still has to parse different numbers of columns based on the generic and SElinux context.

xiangce:
@psachin 
> I researched a bit and found that this strange user-group occurs when the sssd/krb5 DOMAIN/REALM is configured. From what I observed the owner is `domain users` followed by the `@` and DOMAIN name (`CR.LOCAL` in the example). Hence under the assumption that the group will always start with `users` in the subsequent commit I am checking if the 4th-index `startswith()` `users`. Hopefully this should take care of the user-group issue for both the scenarios.

From this official guide document [Adjusting how SSSD interprets full user names](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_authentication_and_authorization_in_rhel/assembly_additional-configuration-for-identity-and-authentication-providers_configuring-authentication-and-authorization-in-rhel#proc_adjusting-how-sssd-interprets-full-user-names_assembly_additional-configuration-for-identity-and-authentication-providers), it seems that `"user"` is just an example used in the doc. And I did a bit of research by using the RAT tool and foune it's true and different `username`s can be used here. For example:
```
/var/log:
total 3834048
drwxr-xr-x. 11 root              root                             4096 May 17 03:10 .
drwxr-xr-x. 20 root              root                              281 Nov 15 15:42 ..
-rw-------.  1 root              utmp                            19200 Apr 30 22:45 btmp-20220501
drwxr-xr-x.  2 chrony            chrony                              6 Jan 10  2019 chrony
-rwxrwx---.  1 lfis@corp.cdw.com it security@corp.cdw.com        56337 May 17 03:15 cron
-rwxrwx---.  1 lfis@corp.cdw.com root                           196293 Apr 24 03:40 cron-20220424
-rwxrwx---.  1 lfis@corp.cdw.com it security@corp.cdw.com       194961 May  1 03:09 cron-20220501
-rwxrwx---.  1 lfis@corp.cdw.com it security@corp.cdw.com       194926 May  8 03:20 cron-20220508
-rwxrwx---.  1 lfis@corp.cdw.com it security@corp.cdw.com       195355 May 15 03:13 cron-20220515
-rw-r--r--.  1 root              root                           101107 Apr 23 04:02 dmesg
```
and 
```
/var/log:
total 571780
-rw-------.  1 root   root                            0 May  8 03:37 spooler-20220515
drwxr-x---.  2 sssd   sssd                         4096 Apr 22 03:17 sssd
-rw-------.  1 root   gls-int-lp-u-linux team  35607509 May 17 02:05 sudo.log
```

So, things become more complex ... may be it's time to think another resolution.

xiangce:
Closing due to it won't fix the issue.

PR #3420, opened 2022-05-24
fix(Compliance): Find policy correctly when there is one datasteam file

marleystipich2:
Check all that apply:

`grep` behaves differently when searching for a string in a single file. The issue is that it doesn't include the filename in the output so for example `grep xccdf_org.ssgproject.content_profile_anssi_bp28_high /usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml` would return `<xccdf-1.2:Profile id="xccdf_org.ssgproject.content_profile_anssi_bp28_high">` but `/usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml:  <xccdf-1.2:Profile id="xccdf_org.ssgproject.content_profile_anssi_bp28_high">` is needed for `insights-client --compliance` to work correctly. This can be fixed by passing the `-H` switch.

### DIFF
--- a/insights/client/apps/compliance/__init__.py
+++ b/insights/client/apps/compliance/__init__.py
@@ -31,6 +31,7 @@ else:
 NONCOMPLIANT_STATUS = 2
 COMPLIANCE_CONTENT_TYPE = 'application/vnd.redhat.compliance.something+tgz'
 POLICY_FILE_LOCATION = '/usr/share/xml/scap/ssg/content/'
+SCAP_DATASTREAMS_PATH = '/usr/share/xml/scap/'
 SSG_PACKAGE = 'scap-security-guide'
 REQUIRED_PACKAGES = [SSG_PACKAGE, 'openscap-scanner', 'openscap']
 
@@ -154,14 +155,14 @@ class ComplianceClient:
         return glob("{0}*rhel{1}*.xml".format(POLICY_FILE_LOCATION, self.os_major_version()))
 
     def find_scap_policy(self, profile_ref_id):
-        grepcmd = 'grep ' + profile_ref_id + ' ' + ' '.join(self.profile_files())
+        grepcmd = 'grep -H ' + profile_ref_id + ' ' + ' '.join(self.profile_files())
         if not six.PY3:
             grepcmd = grepcmd.encode()
         rc, grep = call(grepcmd, keep_rc=True)
         if rc:
             logger.error('XML profile file not found matching ref_id {0}\n{1}\n'.format(profile_ref_id, grep))
             return None
-        filenames = findall('/usr/share/xml/scap/.+xml', grep)
+        filenames = findall(SCAP_DATASTREAMS_PATH + '.+xml', grep)
         if not filenames:
             logger.error('No XML profile files found matching ref_id {0}\n{1}\n'.format(profile_ref_id, ' '.join(filenames)))
             exit(constants.sig_kill_bad)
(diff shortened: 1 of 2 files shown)


insights/tests/client/apps/test_compliance.py
```diff
@@ -304,6 +304,35 @@ def test_find_scap_policy(config, call):
     compliance_client.profile_files = lambda: ['/something']
     assert compliance_client.find_scap_policy('ref_id') == PATH
 
+@patch("insights.client.config.InsightsConfig")
+def test_find_scap_policy_with_one_datastream_file(config):
+    compliance_client = ComplianceClient(config)
+    dir1 = "/usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml"
```
skateman, line 310: I'm not sure if it's a good idea to write to these paths in the testing environment. Can we rather mock the path somewhere in the `find_scap_policy` to something in the `tmp`? Or even better to create fixtures and mock the path there?

skateman:
CI issues are relevant

insights/tests/client/apps/test_compliance.py
marleystipich2: Yeah, I wasn't sure about this either but I didn't know how else to test this line `filenames = findall('/usr/share/xml/scap/.+xml', grep)` in `find_scap_policy` without using the correct path or mocking `findall`. 
But when you say "mock the path" do you mean use a `tmp` path instead of the `/usr/share/xml/scap/ssg/content/` path? I could mock the `findall` method here `filenames = findall('/usr/share/xml/scap/.+xml', grep)` to use the a tmp path instead of the `/usr/share/xml/scap/` path.

skateman: Extract the path into a constant (or a variable?) and try to override it from the test framework.

marleystipich2:
@skateman How does this look now?

insights/client/apps/compliance/__init__.py
```diff
@@ -31,6 +31,7 @@
 NONCOMPLIANT_STATUS = 2
 COMPLIANCE_CONTENT_TYPE = 'application/vnd.redhat.compliance.something+tgz'
 POLICY_FILE_LOCATION = '/usr/share/xml/scap/ssg/content/'
+SCAP_FILE_LOCATION = '/usr/share/xml/scap/'
```
skateman, line 34:
```suggestion
SCAP_DATASTREAMS_PATH = '/usr/share/xml/scap/'
```

insights/tests/client/apps/test_compliance.py
```diff
@@ -305,6 +305,39 @@ def test_find_scap_policy(config, call):
...
 
+@patch("insights.client.config.InsightsConfig")
+def test_find_scap_policy_with_one_datastream_file(config, tmpdir):
+    compliance_client = ComplianceClient(config)
+    dir1 = tmpdir.mkdir('scap')
+    file = dir1.join('test_file.xml')
+    file.write("""
+    <xccdf-1.2:Profile id="xccdf_org.ssgproject.content_profile_anssi_bp28_high">
+    </xccdf-1.2:Profile>
+        """)
```
skateman, line 316: I don't really like the flie creation on the fly, but as far as I see there are no other file-based fixtures in the codebase :disappointed:

skateman (changes requested):
Looks good to me :+1:

insights/tests/client/apps/test_compliance.py
marleystipich2: Right, I saw a few other tests doing something similar and didn't see file-based fixtures either so I figured this was the preferred way to create a file in here

skateman (approved):

PR #3423, opened 2022-05-26
feat: New spec and parser for group_info

xiangce:

### All Pull Requests:

Check all that apply:

New datasource spec and parser for group info

### DIFF
--- a/insights/parsers/user_group.py
+++ b/insights/parsers/user_group.py
@@ -1,0 +1,74 @@
+"""
+Parsers for ``user`` and ``group``
+==================================
+
+Module for processing the ``user`` and ``group`` of Linux.
+
+Parsers provided by this module include:
+
+GroupInfo - command ``getent group <groupname>``
+------------------------------------------------
+
+"""
+from insights import parser, CommandParser
+from insights.parsers import ParseException, SkipException, keyword_search
+from insights.specs import Specs
+
+
+@parser(Specs.group_info)
+class GroupInfo(CommandParser, list):
+    """
+    Class to parse the output of ``getent group <groupname>``
+
+    Typical output of the ``getent group <groupname>`` command is::
+
+        wheel:x:10:admin,tester
+        mem:x:8:
+
+    Examples:
+    >>> type(grp)
+    <class 'insights.parsers.user_group.GroupInfo'>
+    >>> grp[0]['id']
+    10
+    >>> grp[0]['name']
+    'wheel'
+    >>> grp[0]['user_list']
+    ['admin', 'tester']
+    """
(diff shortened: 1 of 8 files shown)


insights/specs/datasources/user_group.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+from insights.specs import Specs
+from insights.core.context import HostContext
+from insights.core.dr import SkipComponent
+from insights.core.plugins import datasource
+from insights.core.filters import get_filters
+
+
+@datasource(HostContext)
+def groups(broker):
```
bfahr, line 16: @xiangce this confused me at first.  Your not really collecting anything here, but just returning the list of filters for this data spec.  What do you think about renaming it to something like `group_filters` instead of `groups`?  Its an unexpected way of using filters, but I believe it provides the same net results.

bfahr (changes requested):

insights/specs/datasources/user_group.py
xiangce: @bfahr - Thanks for the suggestion. I modified it to return the `list` directly instead of a string, and rename it as `group_filter_list`. What do you think?

xiangce: @bfahr - Updated. Thanks for the suggestion.

bfahr: @xiangce that looks good, thanks!

insights/parsers/user_group.py
```diff
@@ -0,0 +1,77 @@
...
+            raise SkipException
+
+        for line in content:
+            try:
+                row = line.split(':')
+                self.append(
+                    dict(
+                        name=row[0],
+                        id=int(row[2]),
+                        user_list=row[-1].split(',') if row[-1] else []
```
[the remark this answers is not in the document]
xiangce: @psachin - I get the name of this column from https://man7.org/linux/man-pages/man5/group.5.html. Anyway, the other columns do not follow it. I'd like to change to `users`.

insights/tests/datasources/test_user_group.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+
+from insights.specs import Specs
+from insights.core import filters
+from insights.specs.datasources.user_group import group_filters
+from insights.core.dr import SkipComponent
+
+
+def setup_function(func):
```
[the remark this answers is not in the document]
xiangce: the `setup_function` doesn't do anything hurt for the following tests, so I think it's not necessary to `teardown` it hear.

psachin (approved):

PR #3424, opened 2022-05-26
New specs for systemd ls output and modification of existing parser

jobselko:

### All Pull Requests:

Check all that apply:

New specs for checking systemd permissions and modification of existing parser. The rest of the parsers will be added in the future.


xiangce (approved):
@jobselko - it looks good to me.

insights/specs/default.py
```diff
@@ -380,13 +380,16 @@ def httpd_cmd(broker):
     ls_ocp_cni_openshift_sdn = simple_command("/bin/ls -l /var/lib/cni/networks/openshift-sdn")
     ls_origin_local_volumes_pods = simple_command("/bin/ls -l /var/lib/origin/openshift.local.volumes/pods")
     ls_osroot = simple_command("/bin/ls -lan /")
-    ls_run_systemd_generator = simple_command("/bin/ls -lan /run/systemd/generator")
+    ls_run_systemd = simple_command("/bin/ls -lanRL /run/systemd")
```
bfahr, line 383: @jobselko you will also need to make the corresponding change the spec name in [insights_archive.py](https://github.com/RedHatInsights/insights-core/blob/0f084dfc6b6f9785194ae6a02cee20ccd1e6ca79/insights/specs/insights_archive.py#L125)

bfahr (changes requested):

insights/specs/default.py
```diff
@@ -380,13 +380,16 @@ def httpd_cmd(broker):
...
     ls_R_var_lib_nova_instances = simple_command("/bin/ls -laR /var/lib/nova/instances")
     ls_sys_firmware = simple_command("/bin/ls -lanR /sys/firmware")
     ls_tmp = simple_command("/bin/ls -la /tmp")
     ls_usr_bin = simple_command("/bin/ls -lan /usr/bin")
     ls_usr_lib64 = simple_command("/bin/ls -lan /usr/lib64")
     ls_usr_lib_systemd = simple_command("/bin/ls -lanRL /usr/lib/systemd")
+    ls_usr_local_lib_systemd = simple_command("/bin/ls -lanRL /usr/local/lib/systemd")
```
ryan-blakley, line 390: @jobselko Could you please remove the specs that the parsers aren't be added yet. If a spec is added to core it needs a parser added with it, otherwise we're collecting something that can't be parsed.

insights/specs/insights_archive.py
```diff
@@ -122,7 +122,7 @@ class InsightsArchiveSpecs(Specs):
     ls_ocp_cni_openshift_sdn = simple_file("insights_commands/ls_-l_.var.lib.cni.networks.openshift-sdn")
     ls_origin_local_volumes_pods = simple_file("insights_commands/ls_-l_.var.lib.origin.openshift.local.volumes.pods")
     ls_osroot = simple_file("insights_commands/ls_-lan")
-    ls_run_systemd_generator = simple_file("insights_commands/ls_-lan_.run.systemd.generator")
+    ls_run_systemd = simple_file("insights_commands/ls_-lanRL_.run.systemd")
```
xiangce, line 125: It would be good to keep compatible with the archives that were already collected with the old spec, please use `first_file` here.
```
ls_run_systemd = first_file("insights_commands/ls_-lanRL_.run.systemd", "insights_commands/ls_-lan_.run.systemd.generator")
```

Thanks.

xiangce: Never mind, since this spec is no longer used by any rules. I think it's OK to change it to match the new command directly.

insights/specs/default.py
xiangce: See the discussion on the spec MR.

PR #3425, opened 2022-05-27
feat: new spec and parser for /etc/sudoers

xiangce:
### All Pull Requests:

Check all that apply:

*Add your description here*

### DIFF
--- a/insights/combiners/sudoers.py
+++ b/insights/combiners/sudoers.py
@@ -7,11 +7,11 @@ Module for combining the parsing results of ``/etc/sudoers`` and
 
 """
 from insights import combiner
-from insights.parsers.sudoers import EtcSudoers
+from insights.parsers.sudoers import SudoersBase, EtcSudoers
 
 
 @combiner(EtcSudoers)
-class Sudoers(object):
+class Sudoers(SudoersBase):
     """
     Class to combiner the ``/etc/sudoers`` and ``/etc/sudoers.d/*``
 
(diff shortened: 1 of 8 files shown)


insights/combiners/sudoers.py
```diff
@@ -0,0 +1,97 @@
...
+
+Module for combining the parsing results of ``/etc/sudoers`` and
+``/etc/sudoers.d/*`` files.
+
+"""
+from insights import combiner
+from insights.parsers.sudoers import EtcSudoers
+
+
+@combiner(EtcSudoers)
```
psachin, line 13: @xiangce Generally we develop combiner out of [two or more components](https://insights-core.readthedocs.io/en/latest/notebooks/Insights%20Core%20Tutorial.html?highlight=combiners#Combiners). Is there are special requirement because of which you decide to create a combiner out of `EtcSudoers` parser alone?

psachin: Besides the code of `get()` & `last()` in the combiner as well as the parser is redundant.

xiangce: The `EtcSudoers` is against a `multi_output=Ture` spec, which results in the parser result being a list. This combiner is going to combine the result. 

No, they are not redundant since they work for different components.

xiangce: @psachin - if you meant the duplicate code of the two functions. I created a base class and removed the duplicates. Please have a look again.

psachin (approved):

PR #3427, opened 2022-05-30
feat: Updated the parser to also return  allow-recursion content

rasrivas-redhat:

### All Pull Requests:

Check all that apply:

* This enhancement to also parse the `allow-recursion` value from the `options` section of the /etc/named.conf.


rasrivas-redhat:
@psachin, @xiangce - Please help to review the PR. Thanks.

insights/parsers/named_conf.py
```diff
@@ -16,31 +16,44 @@
...
+    of the named.conf configurations and class ``NamedConf`` to parse the `include` directives and
+    the `allow-recursion` value from the `options` section.
 
     .. note::
         Please refer to the super-class :py:class:`insights.parsers.named_checkconf:NamedCheckConf`
         for more usage information.
 
     Attributes:
-        includes (list): List of files in 'include' section.
+        includes (list): List of files in 'include' and 'allow-recursion' value from the 'options' section.
```
xiangce, line 27: I think it's a new attribute that is at the same level as the `self.includes`, you'd add it as a new attribute.

insights/parsers/named_conf.py
```diff
@@ -16,31 +16,44 @@
...
+        options_flag = False
         super(NamedConf, self).parse_content(content)
 
         try:
-            for line in [l for l in content if l.strip().startswith('include ') and ';' in l]:
-                includes.append(line.split(';')[0].replace('"', '').split()[1])
+            for line in content:
+                if line.strip().startswith('include ') and ';' in line:
+                    includes.append(line.split(';')[0].replace('"', '').split()[1])
+                if line.strip().startswith('options ') and ';' not in line:
```
xiangce, line 49: I doubt if the check of the `options` section here is required or correct.

Is the `allow-recursion` a unique directive only for the `option` section?

1.  If not, that's it may appear in other sections.
     The current code is incorrect. It's required to check the end of the `options` section, to avoid getting the `allow-recursion` from the other sections which could override the expected one.
2. If yes, it seems that it's not necessary to do this check.

insights/parsers/named_conf.py
```diff
@@ -16,31 +16,44 @@
...
         includes = []
+        allow_recursion_address = []
+        options_flag = False
         super(NamedConf, self).parse_content(content)
 
         try:
-            for line in [l for l in content if l.strip().startswith('include ') and ';' in l]:
-                includes.append(line.split(';')[0].replace('"', '').split()[1])
+            for line in content:
+                if line.strip().startswith('include ') and ';' in line:
```
xiangce, line 47: the `line.strip()` is used times inner the loop, better assign it to a local variable to reduce compute.

```line = line.strip()```

File: insights/parsers/named_conf.py [hunk 4]
```diff
@@ -16,31 +16,44 @@
...
+                    includes.append(line.split(';')[0].replace('"', '').split()[1])
+                if line.strip().startswith('options ') and ';' not in line:
+                    options_flag = True
+                if options_flag and line.strip().startswith('allow-recursion ') and line.endswith(';'):
+                    for addr in line.strip().split('allow-recursion', 1)[1].split('{', 1)[1].split('}', 1)[0].split():
```
xiangce, line 52: How about

```python
allow_recursion_address.extend(i.strip(';') for i in line.split(None, 1)[1].strip('{}; ').split())
```

xiangce (changes requested):

insights/parsers/named_conf.py
rasrivas-redhat: Correct, updated related to it.

insights/parsers/named_conf.py
rasrivas-redhat: Yes, it's a great point. Updated.

insights/parsers/named_conf.py
rasrivas-redhat: Thanks for the suggestion. Updated.

xiangce (approved):

PR #3434, opened 2022-06-02
feat: Add --no-load-default arg to the insights-run command

ryan-blakley:
Check all that apply:

Add --no-load-default arg, if this arg is passed the default plugins aren't loaded. This allows the execution of other projects without loading all the specs from core which can cause exceptions. Re-ordered the arguments so that they're in alphabetical order. Added some empty lines between some of the code below the args so that it's easier to read.


xiangce (approved):
Looks good to me.

PR #3435, opened 2022-06-06
Fix content_type [rhbz#2083665]

ahitacat:
### All Pull Requests:

Check all that apply:

According to documentation, `content-type` can be: gz, bz2, xz or none and should match the compression type of the payload. But using `content-type` this way is causing an error while uploading the archive, as it needs a MIME `content-type` structure.

This fix appends the compression type to the content-type required to upload an archive `application/vnd.redhat.com.advisor.collection+`. Also ensures that the compression of the archive is the same as the content-type provide.

Resolves: rhbz#2083665


LightOfHeaven1994:
@ahitacat checked Compliance with this MR, works fine. Thanks

skateman (approved):
The Seal of Approval

strider (approved):
Looks good to me!

PR #3418, opened 2022-05-19
feat: Add --parallel arg for insights-run

ryan-blakley:
Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [ ] Is this PR to correct an issue?
* [x] Is this PR an enhancement?

### Complete Description of Additions/Changes:
Added parallel execution of rules executed via insights-run. Updated some docstrings that were missing args.

PR #3436, opened 2022-06-07
feat: Support parallelly running for insights-engine

xiangce:

### All Pull Requests:

Check all that apply:

The insights-engine (old) runs rules via the InsightsEvaluators, this PR updates the Evaluators to support running rules in parallel.
- See also #3418 
- BTW, I didn't make the new `parallel` as an attribute of the `Evaluators` class, since I think it is better to pass it via arguments. (The same to the `incremental`, but to avoid breaking the interface, I didn't touch it in this change)


ryan-blakley (approved):
Everything looks good to me.

PR #3443, opened 2022-06-16
Replace xfail with positive test

Glutexo:
### All Pull Requests:

Check all that apply:

Removed the _xfail_ mark and made the false positive an expected result. We do not plan to fix the function. Instead, we introduced an exclusion list that bypasses the substitution function for Specs that would yield false positives.

Xfail’s original intent is to temporarily mark tests written before the code fix. The version number false positive is not the case. The _xfail_ only clutters the pytest output.


ahitacat (approved):
:seal:

PR #3446, opened 2022-06-17
DRAFT fix: The run command loading default plugins

ryan-blakley:
Check all that apply:

I moved the load_default_plugins in the run function to be up under the if print summary, but I didn't think about run being executed without print summary. So this adds it back if not print summary.


ryan-blakley:
This isn't needed anymore closing the PR.

PR #3447, opened 2022-06-21
feat: add "modinfo_filtered_modules" to collect the filtered modules information

huali027:
* Add new spec "modinfo_modules" to collect the filtered modules
* Add another new spec "modinfo_filtered_modules" to run "modinfo filtered_modules" since the result will be filtered if the spec is filterable, and then other info without the filter words will be ignored, which is not expected.
  
### All Pull Requests:

Check all that apply:


huali027:

insights/specs/__init__.py
```diff
@@ -387,7 +387,7 @@ class Specs(SpecSet):
     modinfo_ixgbe = RegistryPoint()
     modinfo_veth = RegistryPoint()
     modinfo_vmxnet3 = RegistryPoint()
-    modinfo = RegistryPoint(multi_output=True)
+    modinfo = RegistryPoint(filterable=True)
```
xiangce, line 390: Since removing the `multi_output=True` will change the usage of the Parser, better to use another symbolink name.

insights/parsers/modinfo.py
```diff
@@ -1,36 +1,12 @@
...
-
-ModInfoEach - Command ``modinfo *``
------------------------------------
-for any module listed by ``lsmod``
-
-ModInfoAll - Command ``modinfo *(all modules)``
------------------------------------------------
-for all modules listed by ``lsmod``
-
+ModInfoAll - Command ``modinfo *(filtered modules)``
```
xiangce, line 2: Since there is a `All` in the name `ModInfoAll`, it makes me think this is the parser for all the modules loaded on this host. How about using another name? like `ModuleInfo`

xiangce (changes requested):
It would be better to explain or give an example of how to use the filter for this spec since the filter does not work with the directly dependent `Specs.modinfo_all`.

xiangce:
@huali027 - please add a short description to the MR in the first comment.

insights/parsers/modinfo.py
huali027: @xiangce, good to me, I'll update it.

huali027:
@xiangce  updated based on the comments, please help to review.

xiangce (approved):
It looks good to me.

psachin (approved):

PR #3451, opened 2022-06-23
feat: Parser for "ls systemd units"

jobselko:
### All Pull Requests:

Check all that apply:

This PR aims to consolidate all `"ls systemd units"` parsers into one file. The combiner will be created afterwards.

MR with relevant changes in insights-core-assets: !437


xiangce:
@jobselko - Would you please merge all these `ls systems units` specs into one? just the `ls_etc`: https://github.com/RedHatInsights/insights-core/blob/bbc2186e17d039e0f6a3e4a8d3eabcfaec4670f7/insights/specs/default.py#L322-L325


In this way, all the `ls` results will be wrapped into one `simple_file` and the combiner you mentioned may be not required.

psachin:
Hey @jobselko I think we should [deprecate](https://insights-core.readthedocs.io/en/latest/api.html?highlight=deprecate#feature-deprecation) the parsers instead of deleting them directly since this is a public repo we don't know whether the parser is already in use. A deprecating warning will alert the user against the alternative parser or the new module where it is migrated (in this case)

jobselko:
@xiangce I reworked it. Thank you!
@psachin I do not think this would be a problem. These parsers are unused (by our teams) and were added around three months ago.

xiangce (approved):
@jobselko it looks good to me

jobselko:
I came across these bugs: #3453, #3454

psachin (approved):

xiangce:
@jobselko - You missed updating the `insights_archive.py` file in this PR

xiangce (approved):

xiangce:
The `python26-test` in CI failed due to an environmental issue. 
~~~
Certificate did not match expected hostname: pypi.python.org.
~~~
I'll ignore it and merge this PR

Issue #3475, opened 2022-07-21
A bug of parser "LvmConfig"


huali027:
The parser raises "E           json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 23 (char 22)" when the content of "/usr/sbin/lvmconfig --type full" is like:
```
global {
	umask=077
	test=0
	units="r"
	si_unit_consistency=1
	suffix=1
	activation=1
	fallback_to_lvm1=0
	format="lvm2"
	proc="/proc"
	etc="/etc"
}
```
It can not parse "umask=077" successfully.

PR #3476, opened 2022-07-21
fix: the parser "LvmConfig" raises exception

huali027:
It fails to parse the content when there is "umask=077" in the configuration.

### All Pull Requests:

Check all that apply:

fixes: https://github.com/RedHatInsights/insights-core/issues/3475

### DIFF
--- a/insights/parsers/lvm.py
+++ b/insights/parsers/lvm.py
@@ -699,7 +699,8 @@ def _lvm_render(o):
     if isinstance(o, dict):
         parts = ['"%s": %s' % (k, _lvm_render(v)) for k, v in o.items()]
         return "{%s}" % ",".join(parts)
-    if isinstance(o, str) and o.isdigit():
+    # for umask=077, it can not be parsed by json directly, transfer it to string first
+    if isinstance(o, str) and o.isdigit() and o.startswith('0') and len(o) > 1:
         return '"%s"' % o
     return "%s" % o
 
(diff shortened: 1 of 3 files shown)


huali027:

insights/parsers/lvm.py
```diff
@@ -699,6 +699,9 @@ def _lvm_render(o):
     if isinstance(o, dict):
         parts = ['"%s": %s' % (k, _lvm_render(v)) for k, v in o.items()]
         return "{%s}" % ",".join(parts)
+    # for umask=077, it can not be parsed by json directly, transfer it to string first
+    if isinstance(o, str) and o.isdigit() and o.startswith('0') and len(o) > 1:
+        return '"%s"' % o
```
xiangce, line 704: From this page https://www.techoism.com/understanding-umask-value-in-linux/, I got 
```
The minimum and maximum UMASK value for a folder is 000 and 777
The minimum and maximum UMASK value for a file is 000 and 666
```

But when the value of the umask is not started with `0`, it won't cause issue #3475 and the value will be stored as `interge type` with the raw digit value. And it will result that for the same key `umask`, there would be different types of `int` and `str` for its values.

I think we should do more research and discussion as well.

huali027: @xiangce How about we transfer the data before storing?

insights/parsers/lvm.py
```diff
@@ -718,10 +711,16 @@ def parse_content(self, content):
...
             elif line[0] == "\t":
                 k, v = line.strip().split("=", 1)
-                dd[key][k] = v
+                # umask=077 will raise exception, and also no need to
+                # transfer it, just keep it as string
+                if k != 'umask':
+                    try:
+                        v = json.loads(v)
+                    except Exception:
+                        raise SkipException("Failed to parse line %s." % line)
```
xiangce, line 720: It looks like the `ParseException` is more suitable here.

xiangce (approved):

psachin (approved):

PR #3481, opened 2022-07-25
Feat: New grub2_editenv_list parser

TZ3070:
Signed-off-by: Xinting Li <xintli@redhat.com>

### All Pull Requests:

Check all that apply:

New parser for command 'grub2-editenv list'


xiangce:
@TZ3070 - Is this ready for review?

For new spec, please raise an MR/issue to the assets repo first to get it approved at first. Thanks.

insights/parsers/grub2_editenv_list.py
```diff
@@ -0,0 +1,80 @@
...
+
+This parser reads the output of command ``grub2-editenv list``, which list the
+current variables in GRUB environment block file. The output without error message
+is laid out in a key=value format similar to an ini file except it doesn't have any
+laid headers.
+"""
+
+from insights import get_active_lines, parser
+from insights.parsers import SkipException
+from insights.parsers.grubenv import GrubEnv
```
ryan-blakley, line 13: @TZ3070 Since this is importing from the grubenv parser file, could you move this parser to that file instead of having a separate file just for this parser since they're both about grubenv.

TZ3070: Updated

insights/parsers/grubenv.py
```diff
@@ -1,14 +1,18 @@
...
 This parser reads the GRUB environment block file. The file is laid out in a
 key=value format similar to an ini file except it doesn't have any headers.
 
 The parser stores the key/value pairs in itself which inherits a dict. This
 file is only used in Grub 2, but in RHEL8 with BLS being the default. There
 were several variables added that are referenced in the
 ``/boot/loader/entries/*.conf`` files.
+
+Grub2EditenvList - command ``grub2-editenv list``
+=================================================
```
xiangce, line 13: For sub-title, `---` should be used.

insights/parsers/grubenv.py
```diff
@@ -90,3 +94,61 @@ def has_tuned_params(self):
...
+        >>> grub2_editenv_list.has_tuned_params
+        True
+        >>> grub2_editenv_list.tuned_params
+        'transparent_hugepages=never'
+        >>> grub2_editenv_list['saved_entry']
+        '295e1ba1696e4fad9e062f096f92d147-4.18.0-305.el8.x86_64'
+    """
+
+    def __init__(self, context):
+        super(Grub2EditenvList, self).__init__(context)
```
xiangce, line 132: It seems "errors" is the only difference between the two parsers, it would be better to add a new base class and make them two inherit from the base.

xiangce (changes requested):

insights/parsers/grubenv.py
TZ3070: @xiangce Updated

insights/parsers/grubenv.py
```diff
@@ -1,76 +1,60 @@
...
+
     def __init__(self, context):
-        super(GrubEnv, self).__init__(context)
+        super(GrubenvBase, self).__init__(context)
 
     def parse_content(self, content):
         if not content:
             raise SkipException("Empty output.")
 
+        self._error = []
```
ryan-blakley, line 44: Not a big deal but since this is a list, could this be changed to self._errors so it's plural.

insights/parsers/grubenv.py
```diff
@@ -1,76 +1,60 @@
...
             key, value = line.split("=", 1)
-
-            # Some keys can have empty values, so just skip them.
             if not value:
                 continue
 
             data[key] = value
 
-        if not data:
+        if (not data) and (not self._error):
```
ryan-blakley, line 57: Could the check on errors be in a sub if statement, that way if there is no data but there were errors it should raise a ParseException, otherwise throw a SkipException. The way it is now, if there is no data parsed but an error line was stored it would return an empty dict when it shouldn't.

insights/parsers/grubenv.py
```diff
@@ -90,3 +74,74 @@ def has_tuned_params(self):
...
+        >>> grub2_editenv_list.has_tuned_params
+        True
+        >>> grub2_editenv_list.tuned_params
+        'transparent_hugepages=never'
+        >>> grub2_editenv_list['saved_entry']
+        '295e1ba1696e4fad9e062f096f92d147-4.18.0-305.el8.x86_64'
+    """
+
+    @property
+    def error(self):
```
ryan-blakley, line 146: Could this be moved to the base class so both classes can use the property? Also could this be changed to errors since it's returning a list, and add an entry to the attributes docstring in the base class.

insights/parsers/grubenv.py
TZ3070: Thanks for your comment. 
For 'GrubEnv' parser, it processes the content in '/boot/grub2/grubenv' file. The file saves environment variables for boot time, and will never contain error messages.
For 'Grub2EditenvList' parser, it parses the output of the 'grub2-editenv list' command.  The list the current variables in environment block file if there are no errors in the file configuration. Otherwise, there will be no data and only error messages in output. If there is no data parsed but an error line was stored, the return is not only an empty dict, but also self.error which is not empty. And error message is one the information we needed.
```
#e.g.
[root@rhel~]# grub2-editenv list
grub2-editenv: error: invalid environment block.
```

From my point of view, if it is just no data parsed, it is not necessary to throw an exception. If there is no data and no error, the parser must throw an exception.

insights/parsers/grubenv.py
TZ3070: Like the comment I left above, the content of 'GrubEnv' parser will never contain error messages. So, the error attribute is no need to be added in base class for each parser to use.

insights/parsers/grubenv.py
TZ3070: @ryan-blakley Updated

insights/parsers/grubenv.py
xiangce: BTW, @TZ3070 - Because this is special processing, please add the above explanation to the code as a comment, to prevent someone from updating the parser in the future without understanding the logic.

xiangce: Oh, my fault. You have already added the comment.

xiangce (approved):

xiangce:
@TZ3070 - since the existing `GrubEnv` and the new one is a bit relevant, would you please make a combiner for them based on the current implementation?

insights/parsers/grubenv.py
```diff
@@ -30,13 +52,8 @@ class GrubEnv(Parser, dict):
         tuned_params=transparent_hugepages=never
         tuned_initrd=
 
-    Attributes:
-        has_kernelopts (bool): Returns True/False depending on if kernelopts key is in the dict.
-        kernelopts (bool): Returns the string of kernelopts from the dict.
-        has_tuned_params (str): Returns True/False depending of if the tuned_params key is in the dict.
-        tuned_params (str): Returns the string of tuned_params from the dict.
+    Examples::
```
xiangce, line 55: The example section does not need a double colon

insights/specs/insights_archive.py
```diff
@@ -80,6 +80,7 @@ class InsightsArchiveSpecs(Specs):
     getconf_page_size = simple_file("insights_commands/getconf_PAGE_SIZE")
     getenforce = simple_file("insights_commands/getenforce")
     getsebool = simple_file("insights_commands/getsebool_-a")
+    grubenv = first_file(["/boot/grub2/grubenv", "/boot/efi/EFI/redhat/grubenv"])
```
xiangce, line 83: Since the spec is changed to command, here the spec should be updated as:
```
grubenv = first_file(["insights_commands/grub2-editenv_list", "/boot/grub2/grubenv", "/boot/efi/EFI/redhat/grubenv"])
```

insights/tests/parsers/test_grubenv.py
```diff
@@ -31,6 +31,10 @@
 ######################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################
```
xiangce, line 31: I believe these kind of lines should be removed after the spec changed to command.

insights/parsers/grubenv.py
```diff
@@ -18,10 +16,34 @@
...
+    being the default. There were several variables added that are referenced in
+    the ``/boot/loader/entries/*.conf`` files.
+
+    Attributes::
+
+        has_kernelopts (bool): Returns True/False depending on if kernelopts key is in the dict.
+        kernelopts (bool): Returns the string of kernelopts from the dict.
+        has_tuned_params (str): Returns True/False depending of if the tuned_params key is in the dict.
+        tuned_params (str): Returns the string of tuned_params from the dict.
+        errors (str): Returns the string of error message.
```
xiangce, line 44: It should be a list.

insights/parsers/grubenv.py
xiangce, line 55: The example section does not need a double colon

xiangce (changes requested):

xiangce (approved):

Issue #3482, opened 2022-07-25
The `/etc/cloud/cloud.cfg` will be also be filtered, expected content is not collected.


xiangce:
When specifying the filters of the `cloud_cfg` spec as
```
add_filter(Specs.cloud_cfg, ['network', 'ssh_deletekeys', 'debug'])
```

The collected file will look like:
```
{"debug": {"output": "/var/log/cloud-init-debug.log"}, "network": null, "ssh_deletekeys": 1}
```

But in fact, the original content is:
```
users:
  - name: demo
    ssh-authorized-keys:
      - key_one
      - key_two
    passwd: $6$j212wezy$7H/1LT4f9/N3wpgNunhsIqtMj62OKiS3nyNwuizouQc3u7MbYCarYeAHWYPYb2FT.lbioDm2RrkJPb9BZMN1O/

ssh_deletekeys: 1

network:
    version: 1
    config:
    - type: physical
        name: eth0
        subnets:
          - type: dhcp
          - type: dhcp6

system_info:
    default_user:
    name: user2
    plain_text_passwd: 'someP@assword'
    home: /home/user2

debug:
    output: /var/log/cloud-init-debug.log
    verbose: true
```

The expected `network` should not be `null`

Issue #3478, opened 2022-07-22
The `cloud_cfg` datasource converts the result to json, but it should be yaml.


crackcodecamp:
In https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/datasources/cloud_init.py#L114 the result is converted into the json but when using this datasource in the rule, it failed with `insights.parsers.ParseException: insights.parsers.cloud_cfg.CloudCfg couldn't parse json.` and skip the condition.

PR #3484, opened 2022-07-27
fix: change cloud_cfg to Yaml and modify the source spec

xiangce:

### All Pull Requests:

Check all that apply:

- fix: #3482
   Change the output of the datasource `cloud_cfg` to Yaml format to keep compatible with the original file that will also be collected by sosreport
- fix: #3478
   Change the source spec to "cloud_cfg_filtered" to avoid it being filtered before the process of datasource
   The filter will be applied to the original spec "cloud_cfg"


psachin:
If not urgent, I want understand this issue with @crackcodecamp before review.

bfahr:
@xiangce I'm concerned about this change because I think it might cause confusion between the the specs `cloud_cfg` which is filtered and `cloud_cfg_filtered` which is not a filtered spec but uses filters from `cloud_cfg`.  I would say that this spec/parser is actually broken since it currently returns json instead of yaml, and because it does not retain the structure of the config file.  Would you consider just fixing the current spec and parser instead of the current approach?

xiangce:
@bfahr - Thanks for reviewing.

Due to the current `filter` policy, the original content of the spec will be filtered per the `allowlist` by force when the spec is marked as `filterable=True` no matter if it's a legacy spec or datasource spec. This caused issue #3482 and resulted in the parser will never get the expected result specified in the `filters`, even after issue #3478 is fixed and the returned content is converted to Yaml.
The new added spec`cloud_cfg_filter` is just a workaround, since currently, we cannot adjust the filter policy for `datasource specs` yet (I think we have discussed this in one of our sync meeting). Does it work if we add adequate comment/docstrings for the usage of the two specs?

xiangce:
@bfahr - I added some notes for the two specs in the docstring, please have a look.

bfahr (approved):

PR #3486, opened 2022-07-29
fix: Update aws specs to use IMDSv2

### LINKED ISSUE #3485: AWS EC2 metadata specs need to support IMDSv2
The current specs to collect metadata from AWS EC2 nodes only supports IMDSv1, and they need to be updated to be compatible with [IMDSv2 sessions](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html).  The problem that this creates is the specs `aws_instance_id_doc` and `aws_instance_id_pkcs7` will execute on an EC2 node with IMDBv2 support only, and they will not generate an error, but no data is returned.  

The specs need to support a session token so that they can work with both IMDBv1 and v2.

bfahr:
### All Pull Requests:

Check all that apply:

* AWS IMDSv2 uses sessions for metadata access which requires the extra step of collecting a token prior to accessing metadata
* Add a new datasource to get the IMDSv2 token
* Modifiy existing specs to use the token to access metadata
* Add datasource to docs
* This change resolves #3485

### DIFF
--- a/insights/specs/datasources/aws.py
+++ b/insights/specs/datasources/aws.py
@@ -1,0 +1,42 @@
+"""
+Custom datasources for aws information
+"""
+from insights.components.cloud_provider import IsAWS
+from insights.core.context import HostContext
+from insights.core.dr import SkipComponent
+from insights.core.plugins import datasource
+from insights.core.spec_factory import simple_command
+from insights.specs import Specs
+
+
+class LocalSpecs(Specs):
+    """ Local specs used only by aws datasources """
+    aws_imdsv2_token = simple_command(
+        '/usr/bin/curl -s -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600" --connect-timeout 5',
+        deps=[IsAWS]
+    )
+
+
+@datasource(LocalSpecs.aws_imdsv2_token, HostContext)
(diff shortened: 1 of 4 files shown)


insights/specs/datasources/aws.py
```diff
@@ -0,0 +1,42 @@
...
+from insights.core.dr import SkipComponent
+from insights.core.plugins import datasource
+from insights.core.spec_factory import simple_command
+from insights.specs import Specs
+
+
+class LocalSpecs(Specs):
+    """ Local specs used only by aws datasources """
+    aws_imdsv2_token = simple_command(
+        '/usr/bin/curl -s -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600" --connect-timeout 5',
```
xiangce, line 15: This update looks good to me and per my test, it works well on AWS instances.

However, my only concern is, that the raw token will be present in the collected file name. I'm not sure if this would increase the attack surface to the customer's EC2 instances, but it sounds like a good idea to limit the token period to a shorter valid period, e.g. `60` seconds, what do you think?

bfahr: @xiangce that is a good suggestion, thank you.  I'll make the change.

xiangce (approved):

PR #3493, opened 2022-08-07
Automatically retry malware-detection downloads & uploads

mhuth:

This PR sets the number of download/upload retries for malware-detection to 5.  This is especially important for uploads because we don't want to fail after one unsuccessful attempt to upload the results archive.  Doing a malware scan could take a long time, and if we are unable to upload the archive, then insights-client fails and deletes the archive, which would waste all that effort in doing the scan, and be an unhappy user experience.  A better user experience would be to at least retry uploading the archive a number of times before failing.  (An even better user experience would be to keep the archive instead of deleting it  but that's another PR for another day.)

### DIFF
--- a/insights/client/apps/malware_detection/__init__.py
+++ b/insights/client/apps/malware_detection/__init__.py
@@ -373,10 +373,12 @@ class MalwareDetectionClient:
                 # Remove extras slashes (/) in the file name and leading double slashes too (normpath doesn't)
                 item = os.path.normpath(item).replace('//', '/')
                 # Assume the item represents a filesystem item
-                if os.path.exists(item):
-                    self.scan_fsobjects.append(item)
-                else:
+                if not os.path.exists(item):
                     logger.info("Skipping missing filesystem_scan_only item: '%s'", item)
+                elif os.path.islink(item):
+                    logger.info("Skipping symlink filesystem_scan_only item: '%s'.  Please use non-symlink items", item)
+                else:
+                    self.scan_fsobjects.append(item)
 
             if self.scan_fsobjects:
                 logger.info("Scan only the specified filesystem item%s: %s", "s" if len(self.scan_fsobjects) > 1 else "",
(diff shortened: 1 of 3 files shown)


subpop:
@Glutexo or @ahitacat, could you take a look at this PR when you get a chance?

ahitacat (approved):
LGTM

ahitacat:
> An even better user experience would be to keep the archive instead of deleting it but that's another PR for another day.

This can be feasible when configuring the app in `client/config.py` by adding the keep archive variable

ahitacat:
test me

mhuth:
Piggybacked in some changes around using symlinks in the filesystem_scan_* options.

mhuth:
@ahitacat @subpop Actually I've been testing using `retries/--retry` and it doesn't seem to be working in the way I was expecting.  In a VM, I run something like `insights-client --retry 3 --verbose` then as its nearing the end of the run and is about to upload the archive, I shutdown the network interface on the VM to simulate a network problem.  At which point I thought insights-client would retry uploading the archive a number of times (whatever retries was set to).  However it seems to quit on the first attempt.

Here is some sample output at the end of the insights-client run with the network interface down:
```
...
2022-08-08 16:29:47,068    DEBUG insights.client.archive Tar File: /var/tmp/insights-archive-qrq8oe1l/insights-rhel86-vm-20220808062921.tar.gz
2022-08-08 16:29:47,116    DEBUG insights.client.archive Deleting: /var/tmp/insights-archive-op7kt6ml/insights-rhel86-vm-20220808062921
2022-08-08 16:29:47,126    DEBUG insights.client.archive Tar File Size: 428812
2022-08-08 16:29:47,126     INFO insights.client.client Uploading Insights data.
2022-08-08 16:29:47,127    DEBUG insights.client.utilities Found /etc/insights-client/machine-id
2022-08-08 16:29:47,127    DEBUG insights.client.connection Uploading /var/tmp/insights-archive-qrq8oe1l/insights-rhel86-vm-20220808062921.tar.gz to https://cert-api.access.redhat.com/r/insights/uploads/fab5afc6-36d6-4ac6-b56a-ec37c45a7b10
2022-08-08 16:29:47,127  NETWORK insights.client.connection POST https://cert-api.access.redhat.com/r/insights/uploads/fab5afc6-36d6-4ac6-b56a-ec37c45a7b10
2022-08-08 16:29:47,128    ERROR insights.client.phase.v1 HTTPSConnectionPool(host='cert-api.access.redhat.com', port=443): Max retries exceeded with url: /r/insights/uploads/fab5afc6-36d6-4ac6-b56a-ec37c45a7b10 (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x7f0604a06dd8>: Failed to establish a new connection: [Errno -2] Name or service not known',))
2022-08-08 16:29:47,129    DEBUG insights.client.archive Deleting: /var/tmp/insights-archive-op7kt6ml
```

I was expecting the last couple of lines to be repeated the number of times `retries/--retry` is set to but that didn't happen.  Do I have the wrong assumptions about `retries/--retry`?  Or am I testing it incorrectly?

mhuth:
@ahitacat @subpop Here is a screenshot showing more examples of the retries value not taking effect.  I added some logging lines to print out the number of retries.  The reproducer was shutting down the network interface during the insights-client run.   Is this expected behaviour?  I would think if retries > 1 then it would retry the upload regardless of the error condition and only after exhausting the retries would it then exit.  However this code hasn't been modified in 4 years so perhaps I am incorrect in my understanding?  

My main motivation for reporting this is, because a malware-detection scan might take a long time to complete, eg hours, then if there were a temporary network issue, insights-client should retry the upload rather than exiting after one attempt.  I can have a go at changing the code if you think my understanding is correct and it really should retry if --retry has been set.

Screenshot from 2022-08-15 09-59-06

ahitacat:
@mhuth this seems like a bug. I have opened a BZ to tackle this issue. https://bugzilla.redhat.com/show_bug.cgi?id=2117686

bfahr:
@ahitacat since you have opened a BZ for the client issue I'll go ahead and merge this.

PR #3496, opened 2022-08-15
feat: New spec  and parser for"auditctl -l"

huali027:

### All Pull Requests:

Check all that apply:

This update is targeting to add a new spec and parser for "auditctl -l". Since there is an existing module `auditctl_status`, and the existing parser didn't follow the latest design. This PR also:
- Mark the old "AuditctlStatus" parser as deprecated
- Move the `auditctl` related parsers to `auditctl` module.

### DIFF
--- a/insights/parsers/auditctl.py
+++ b/insights/parsers/auditctl.py
@@ -4,70 +4,16 @@ AuditCtl - command ``auditctl xxx``
 
 This module contains the following parsers:
 
-AuditdStatus - command ``auditctl -s``
---------------------------------------
 AuditRules - command ``auditctl -l``
 ------------------------------------
+AuditStatus - command ``auditctl -s``
+-------------------------------------
+
 """
 
 from insights import parser, CommandParser
 from insights.parsers import ParseException, SkipException
 from insights.specs import Specs
-
-
-@parser(Specs.auditctl_status)
-class AuditdStatus(CommandParser, dict):
-    """
-    Module for parsing the output of the ``auditctl -s`` command.
-
-    Typical output on RHEL6 looks like::
-
-        AUDIT_STATUS: enabled=1 flag=1 pid=1483 rate_limit=0 backlog_limit=8192 lost=3 backlog=0
-
-    , while on RHEL7 and later, the output changes to::
-
-        enabled 1
-        failure 1
-        pid 947
-        rate_limit 0
-        backlog_limit 320
-        lost 0
-        backlog 0
-        loginuid_immutable 0 unlocked
-
-    Example:
-        >>> type(auds)
-        <class 'insights.parsers.auditctl.AuditdStatus'>
-        >>> "enabled" in auds
-        True
-        >>> auds['enabled']
-        1
-    """
-    def parse_content(self, content):
-        if not content:
-            raise SkipException("Input content is empty.")
-        if len(content) > 1:
-            for line in content:
-                k, v = line.split(None, 1)
-                # Mind the 'loginuid_immutable' on RHEL7
-                if k.strip() == "loginuid_immutable":
-                    self[k.strip()] = v.strip()
-                else:
-                    try:
-                        self[k.strip()] = int(v.strip())
-                    except ValueError:
-                        raise ParseException('Unexpected type in line %s' % line)
-        if len(content) == 1:
-            line = list(content)[0].strip()
-            if line.startswith("AUDIT_STATUS:"):
-                for item in line.split(None)[1:]:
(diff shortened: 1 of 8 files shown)


huali027:
@xiangce  Please help to review.
Since there is already a parser for the command "auditctl status", I put them together. But the file name isn't general for "auditctl" commands. Currently, I'm afraid it will impact the existing parser path, I don't change it. Please suggest if I need to rename the file.

insights/parsers/auditctl_status.py
```diff
@@ -60,3 +67,43 @@ def parse_content(self, content):
...
+        >>> len(audit_rules)
+        9
+        >>> '-w /etc/selinux -p wa -k MAC-policy' in audit_rules
+        True
+
+    Raises:
+        SkipException: When there are not rules.
+    """
+    def __init__(self, context):
+        super(AuditRules, self).__init__(context, extra_bad_lines=['no rules'])
```
psachin, line 102: This looks like a warning message from the command when there are no rules rather than the bad lines. I think we can use this to raise an exception when no rules are found.

insights/parsers/auditctl_status.py
```diff
@@ -36,6 +41,14 @@ class AuditctlStatus(LegacyItemAccess, CommandParser):
         >>> auds['enabled']
         1
     """
+    def __init__(self, context):
+        deprecated(
+            AuditctlStatus,
+            "Please use the :class:`insights.parsers.auditctl.AuditdStatus` instead.",
+            "3.1.25"
```
huali027, line 48: @psachin after discussion with @xiangce , we agree to add a new module to include the parsers about "auditctl" commands, because the original file name is "auditctl_status.py", it's not good to add other "auditctl" commands. And I mark the original one as deprecated. Now the version is wrong, I'm not sure which version I should use, please help here, thanks.

insights/parsers/auditctl_status.py
huali027: @psachin, updated as suggested.

insights/parsers/auditctl_status.py
psachin: 3.1.25 is OK

xiangce: Better introduce a global variable, e.g. "NEXT_MINOR_VER", to avoid such kind of confusion.

huali027: Currently, the exact version list isn't done, so a few parsers are marked as "3.1.25". Sachin said it is a release which is one year later. After the version list comes out, maybe we'll mark different versions for the parser. Maybe in one release, there is only one parser deprecated. In this case, maybe a global variable is useless? Do we need to do it now for "3.1.25"?

xiangce: Sorry for the incomplete comment, that is for @psachin, but not a block for this PR, please go ahead and use the hard-coded `3.1.25` .

> Maybe in one release, there is only one parser deprecated. In this case, maybe a global variable is useless?

No matter how many `component` would be deprecated in the next minor release version, it would be better to use a global variable than the hard-coded versions, from the development and release perspective.

psachin: @xiangce By using custom version the developers can plan the feature deprecation accordingly. The developer might not want to deprecate the feature in NEXT_MINOR_VER. I have drafted a change in the docs with examples to help them in this process[1]. 

[1]  https://github.com/RedHatInsights/insights-core/pull/3500

insights/parsers/auditctl.py
```diff
@@ -0,0 +1,109 @@
...
+        >>> type(auds)
+        <class 'insights.parsers.auditctl.AuditdStatus'>
+        >>> "enabled" in auds
+        True
+        >>> auds['enabled']
+        1
+    """
+    def parse_content(self, content):
+        if not content:
+            raise ParseException("Input content is empty.")
```
xiangce, line 48: This should be a  skip.

insights/parsers/auditctl.py
```diff
@@ -0,0 +1,109 @@
...
+        if len(content) == 1:
+            line = list(content)[0].strip()
+            if line.startswith("AUDIT_STATUS:"):
+                for item in line.split(None)[1:]:
+                    try:
+                        k, v = item.split('=')
+                        self.data[k.strip()] = int(v.strip())
+                    except ValueError:
+                        continue
+
```
xiangce, line 70: Better skip it when get nothing parsered

xiangce (approved):

PR #3504, opened 2022-08-25
Feat: Add spec and parser for cryptsetup luksDump

danzatt:

### All Pull Requests:

Check all that apply:

- Added `cryptsetup_luksDump` spec which collects the output of `cryptsetup luksDump` spec with salt and digest fields filtered out (although luksDump public is designed to contain only public information, salt and digest can potentially be sensitive)
- Added `LUKS_Dump` parser, which parses the output using `parsr` library (it works for both - LUKS1 and LUKS2 output formats)

### DIFF
--- a/docs/custom_datasources_index.rst
+++ b/docs/custom_datasources_index.rst
@@ -101,7 +101,7 @@ insights.specs.datasources.lpstat
 
 
 insights.specs.datasources.luks_devices
----------------------------------
+---------------------------------------
 
 .. automodule:: insights.specs.datasources.luks_devices
     :members: LUKS_block_devices, LocalSpecs, LUKS_data_sources
--- a/docs/custom_datasources_index.rst
+++ b/docs/custom_datasources_index.rst
@@ -104,7 +104,7 @@ insights.specs.datasources.luks_devices
 ---------------------------------------
 
 .. automodule:: insights.specs.datasources.luks_devices
-    :members: LUKS_block_devices, LocalSpecs, LUKS_data_sources
+    :members: Luks_block_devices, LocalSpecs, Luks_data_sources
     :show-inheritance:
     :undoc-members:
 
--- a/docs/custom_datasources_index.rst
+++ b/docs/custom_datasources_index.rst
@@ -99,6 +99,15 @@ insights.specs.datasources.lpstat
     :show-inheritance:
     :undoc-members:
 
+
+insights.specs.datasources.luks_devices
+---------------------------------
+
+.. automodule:: insights.specs.datasources.luks_devices
+    :members: LUKS_block_devices, LocalSpecs, LUKS_data_sources
+    :show-inheritance:
+    :undoc-members:
+
 insights.specs.datasources.malware_detection
 --------------------------------------------
 
--- a/docs/custom_datasources_index.rst
+++ b/docs/custom_datasources_index.rst
@@ -104,7 +104,7 @@ insights.specs.datasources.luks_devices
 ---------------------------------------
 
 .. automodule:: insights.specs.datasources.luks_devices
-    :members: Luks_block_devices, LocalSpecs, Luks_data_sources
+    :members: luks_block_devices, luks_data_sources, LocalSpecs
     :show-inheritance:
     :undoc-members:
 
--- a/insights/parsers/cryptsetup_luksDump.py
+++ b/insights/parsers/cryptsetup_luksDump.py
@@ -36,19 +36,19 @@ class DocParser(object):
         FirstLevelKVPair = FirstIndent >> Key + Value1
         SecondLevelKVPair = SecondIndent >> WithIndent(Key + Value)
 
-        LUKS2SectionName = Key << Char("\n")
-        LUKS2SectionEntry = (FirstLevelKVPair + Many(SecondLevelKVPair).map(dict)).map(self.convert_type)
-        LUKS2Section = LUKS2SectionName + Many(LUKS2SectionEntry).map(dict) << Opt(Many(Char("\n")))
-        LUKS2Body = Many(LUKS2Section, lower=1)
+        Luks2SectionName = Key << Char("\n")
+        Luks2SectionEntry = (FirstLevelKVPair + Many(SecondLevelKVPair).map(dict)).map(self.convert_type)
+        Luks2Section = Luks2SectionName + Many(Luks2SectionEntry).map(dict) << Opt(Many(Char("\n")))
+        Luks2Body = Many(Luks2Section, lower=1)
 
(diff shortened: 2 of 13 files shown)


insights/parsers/cryptsetup_luksDump.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def convert_status(section):
+    section[2]["status"] = section[1]
+    return [section[0], section[2]]
+
+
+value_chars = set(string.printable) - set("\n\r")
+
+FirstLine = Literal("LUKS header information", value="header") << AnyChar.until(Char("\n")) + Opt(Many(Char("\n")))
```
ryan-blakley, line 27: Could you please review the httpd_conf parser, and add all of the parsr logic in a DocParser class like the httpd_conf parser. That way all of the logic is wrapped in a class so it's better organized and has an exceptions catch.

insights/parsers/cryptsetup_luksDump.py
```diff
@@ -0,0 +1,175 @@
...
+        dump(dict of dicts): A top level dict containing the dictionaries
+            representing the header, data segments, keyslots, digests
+            and tokens.
+
+    """
+    def parse_dump(self, text):
+        self.dump = None
+        header = LUKS_Header(text)
+
+        if header["header"]["Version"] == "1":
```
ryan-blakley, line 164: Generally when using parsr, this logic is in the parsr logic using the | symbol. The way you have it setup you're parsing the content twice it looks like.  Example would be something similar to the below.

`Doc = LUKS_Header + (LUKS1Body | LUKS2Body)`

insights/parsers/cryptsetup_luksDump.py
```diff
@@ -0,0 +1,175 @@
...
+            return header | body
+        elif header["header"]["Version"] == "2":
+            header, body = (LUKS_Header + LUKS2Body)(text)
+            return header | body
+
+    def parse_content(self, content):
+        try:
+            self.dump = self.parse_dump("\n".join(content).replace("\t", " "*8) + "\n")
+        except:
+            self.dump = None
```
ryan-blakley, line 175: If there is an exception then a SkipException or ParseException should be raised.

insights/parsers/cryptsetup_luksDump.py
```diff
@@ -0,0 +1,175 @@
+"""
+LUKS_Dump - command ``cryptsetup luksDump``
```
ryan-blakley, line 2: Please add a doc entry under docs/shared_parsers_catalog/ for this parser.

ryan-blakley (changes requested):
Additionally to all my review comments, I approved the PR test to run and it appears there is several flake8 issues. If you could please address those as well.

danzatt:
Thank you for the feedback. Hopefully, I have addressed all the remarks now.

ryan-blakley (changes requested):

docs/custom_datasources_index.rst
```diff
@@ -99,6 +99,15 @@ insights.specs.datasources.lpstat
     :show-inheritance:
     :undoc-members:
 
+
+insights.specs.datasources.luks_devices
+---------------------------------
```
xiangce, line 104: Would you please change these two lines to the same length?

insights/parsers/cryptsetup_luksDump.py
```diff
@@ -0,0 +1,172 @@
...
+============================================================
```
xiangce, line 3: Better keep this `===` line in the same length as the title too.

insights/parsers/cryptsetup_luksDump.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def __call__(self, content):
+        try:
+            return self.Top(content)
+        except Exception:
+            raise ParseException("There was an exception when parsing one of the outputs of cryptsetup luksDump commands.")
+
+
+@parser(Specs.cryptsetup_luksDump)
+class LUKS_Dump(Parser):
```
xiangce, line 72: I think I understand you keep the "LUKS" capitalized, however, per the [PEP 8](https://peps.python.org/pep-0008/#class-names), it would be better to use the `CapsWords` style for the class names to distinguish with the [Type Variable Names](https://peps.python.org/pep-0008/#type-variable-names). Would you mind changing it per `PEP 8`?

danzatt:
I have removed the stray print commands, fixed the length of underline and changed the `LUKS` capitalisation to `Luks`.

insights/parsers/cryptsetup_luksDump.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def __call__(self, content):
+        try:
+            return self.Top(content)
+        except Exception:
+            raise ParseException("There was an exception when parsing one of the outputs of cryptsetup luksDump commands.")
+
+
+@parser(Specs.cryptsetup_luksDump)
+class Luks_dump(Parser):
```
xiangce, line 72: @danzatt - Thanks for the update, but I think you may misunderstand the `CapsWords`. Would you please change the class to `LuksDump`?

danzatt: Sorry, done now.

danzatt:
I have added a new component for checking if `cryptsetup` has tokens support (version at least 2.4.0). If it does, I pass the `--disable-external-tokens` parameter. This is because otherwise 3rd party token plugins' code can be invoked, which can take a potentially long time.

psachin (approved):

xiangce (changes requested):
@danzatt - I like your latest update that added the `dependent component`, it makes the data is collected only on the systems that match the requirements.

However, to ensure the `dependent component` works as expected, all of them should be pre-loaded before the collection.

Please add them to the pre-loaded lists, [here](https://github.com/RedHatInsights/insights-core/blob/master/insights/collect.py#L88), e.g.
```
        - name: insights.parsers.blkid.BlockIDInfo
          enabled: true
        - name: insights.components.blkid.cryptsetup
          enabled: true
```
The `InstalledRpms` was already added ([here](https://github.com/RedHatInsights/insights-core/blob/master/insights/collect.py#L107-L109)) , so no need to added it again.

danzatt:
Sure, done now.

bfahr:
@danzatt since custom datasources run during client collection its best to have as close as possible to 100% test coverage of the code.  You can check this using the command `pytest --cov` and look at the coverage for your datasource.  Its close at 87% but it'd be great if you can add just a few tests to get the coverage up.

```
Name                                                                   Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------------------------------------------------------------------------------
insights/specs/datasources/luks_devices.py                              52      4     26      6    87%   37, 85, 108, 109->112, 125
```

danzatt:
I have added a couple of tests, now the coverage is 100%.

xiangce (approved):

PR #3507, opened 2022-08-26
Added no_proxy autoconfiguration from rhsm conf and tests

ahitacat:

### All Pull Requests:

Check all that apply:

Insights-client wasn't honoring the no_proxy configuration of `/etc/rhsm/rhsm.conf` file.

This PR adds the no_proxy value to the auto-configuration module, to retrieve the value of the rhsm configuration file. It provides some tests related to this.

This fix is related to: [rhbz#2065233]((https://bugzilla.redhat.com/show_bug.cgi?id=2065233)

### DIFF
--- a/insights/client/auto_config.py
+++ b/insights/client/auto_config.py
@@ -55,7 +55,7 @@ def verify_connectivity(config):
         return False
 
 
-def set_auto_configuration(config, hostname, ca_cert, proxy, is_satellite, is_stage):
+def set_auto_configuration(config, hostname, ca_cert, proxy, is_satellite, is_stage, rhsm_no_proxy=None):
     """
     Set config based on discovered data
     """
@@ -63,6 +63,7 @@ def set_auto_configuration(config, hostname, ca_cert, proxy, is_satellite, is_st
     logger.debug("Attempting to auto configure hostname: %s", hostname)
     logger.debug("Attempting to auto configure CA cert: %s", ca_cert)
     logger.debug("Attempting to auto configure proxy: %s", proxy)
+    logger.debug("Attempting to auto configure no_proxy: %s", rhsm_no_proxy)
     saved_base_url = config.base_url
     if ca_cert is not None:
         saved_cert_verify = config.cert_verify
@@ -70,6 +71,8 @@ def set_auto_configuration(config, hostname, ca_cert, proxy, is_satellite, is_st
     if proxy is not None:
         saved_proxy = config.proxy
         config.proxy = proxy
+    if rhsm_no_proxy and rhsm_no_proxy != '':
+        config.no_proxy = rhsm_no_proxy
     if is_satellite:
         # satellite
         config.base_url = hostname + '/r/insights'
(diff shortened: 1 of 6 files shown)


insights/client/auto_config.py
```diff
@@ -55,21 +55,25 @@ def verify_connectivity(config):
...
     saved_base_url = config.base_url
     if ca_cert is not None:
         saved_cert_verify = config.cert_verify
         config.cert_verify = ca_cert
     if proxy is not None:
         saved_proxy = config.proxy
         config.proxy = proxy
+    rhsm_no_proxy = None
+    if rhsm_no_proxy is not None or not '':
+        config.no_proxy = rhsm_no_proxy
```
strider, line 76: The default value of `rhsm_no_proxy` is `None` in the function parameters. So why resetting it to `None` again here? And it looks like we will always pass in the conditional here.

I guess the condition should be this one:

`if rhsm_no_proxy and rhsm_no_proxy != '':`

insights/client/auto_config.py
```diff
@@ -128,6 +132,9 @@ def _try_satellite6_configuration(config):
         rhsm_proxy_port = rhsm_config.get('server', 'proxy_port').strip()
         rhsm_proxy_user = rhsm_config.get('server', 'proxy_user').strip()
         rhsm_proxy_pass = rhsm_config.get('server', 'proxy_password').strip()
+        rhsm_no_proxy = rhsm_config.get('server', 'no_proxy').strip()
+        if rhsm_no_proxy.lower() == 'None' or rhsm_no_proxy == '':
```
strider, line 136: `str.lower()` is supposed to return a copy of the string converted to lowercase and you compare it to `None`.

strider (changes requested):

insights/client/auto_config.py
ahitacat: you are right

insights/client/auto_config.py
ahitacat: oh my mistake, it needs to be compare with `'none' `

insights/client/auto_config.py
```diff
@@ -128,6 +131,9 @@ def _try_satellite6_configuration(config):
         rhsm_proxy_port = rhsm_config.get('server', 'proxy_port').strip()
         rhsm_proxy_user = rhsm_config.get('server', 'proxy_user').strip()
         rhsm_proxy_pass = rhsm_config.get('server', 'proxy_password').strip()
+        rhsm_no_proxy = rhsm_config.get('server', 'no_proxy').strip()
+        if rhsm_no_proxy.lower() == 'none' or rhsm_no_proxy == '':
```
strider, line 135: May be simplified as:

`if rhsm_no_proxy.lower() in ['none', '']:`

But it is just nitpicking here.

strider (approved):
Tested and it works, LGTM.

ahitacat:
test me

PR #3512, opened 2022-09-02
fix: support InstanceType in saphostctrl

xiangce:

### All Pull Requests:

Check all that apply:

* [x] Is this PR to correct an issue?  -> BZ-1969572

From BZ-1969572, the "InstanceType" should be used as the identifier for SAP `function_instance`
- Added the InstanceType to the fixed filter of saphostctrl
- Remove the unused "CreationClassName" from the fixed filter
- Add a full_type function to get the full InstanceType
- check "Solution Manager Diagnostic Agent" for function_instance
- Convert the Parser SAPHostCtrlInstances to a list

### DIFF
--- a/insights/tests/parsers/test_saphostctrl.py
+++ b/insights/tests/parsers/test_saphostctrl.py
@@ -144,6 +144,7 @@ def test_saphostctrl_docs():
 def test_saphostctrl():
     sap = SAPHostCtrlInstances(context_wrap(SAPHOSTCTRL_HOSTINSTANCES_GOOD))
     assert len(sap) == 11
+    assert sap.data[-3]['SapVersionInfo'] == '749, patch 401, changelist 1806777'
     assert sap[-3]['SapVersionInfo'] == '749, patch 401, changelist 1806777'
     assert sorted(sap.instances) == sorted([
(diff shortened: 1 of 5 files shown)


insights/tests/parsers/test_saphostctrl.py
```diff
@@ -142,11 +143,11 @@ def test_saphostctrl_docs():
 
 def test_saphostctrl():
     sap = SAPHostCtrlInstances(context_wrap(SAPHOSTCTRL_HOSTINSTANCES_GOOD))
-    assert len(sap) == 10
-    assert sap.data[-2]['SapVersionInfo'] == '749, patch 401, changelist 1806777'
```
bfahr, line 146: @xiangce it might be good to leave this test in too for `data` since you added it as a property to maintain backward compatibility.

bfahr (changes requested):
Also I noticed some possible rule issues but I think those are related to test data and not structural changes.

insights/tests/parsers/test_saphostctrl.py
xiangce: Fixed.

And fixed the test data in rules per the parser. 

Thanks for checking!

xiangce: @bfahr - The items you pointed out were fixed, I went ahead and merged this PR. For further issues, please let me know.

PR #3514, opened 2022-09-07
feat: New spec "/etc/cron.d/foreman" and parser

huali027:

### All Pull Requests:
New spec "/etc/cron.d/foreman" and its parser

Check all that apply:

### DIFF
--- a/insights/parsers/cron_jobs.py
+++ b/insights/parsers/cron_jobs.py
@@ -1,0 +1,44 @@
+"""
+Cron job files under /etc/cron* but not users crontab
+=====================================================
+
+This module contains the following parsers:
+
+CronForeman - file ``/etc/cron.d/foreman``
+------------------------------------------
+"""
+
+from insights.core import Scannable
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+
+@parser(Specs.cron_foreman)
+class CronForeman(Scannable):
+    """
+    Parse the ``/etc/cron.d/foreman`` file.
+
+    Sample input::
+
+        SHELL=/bin/sh
+        PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+
+        RAILS_ENV=production
+        FOREMAN_HOME=/usr/share/foreman
+
+        # Clean up the session entries in the database
+        15 23 * * *     foreman    /usr/sbin/foreman-rake db:sessions:clear 2>&1 | gawk '{ print strftime("[\%Y-\%m-\%d \%H:\%M:\%S]"), $0 }' >>/var/log/foreman/cron.log
+
+        # Send out recurring notifications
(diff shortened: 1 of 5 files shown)


huali027:
@xiangce Please help to review.

insights/parsers/cron_jobs.py
```diff
@@ -0,0 +1,44 @@
...
+------------------------------------------
+"""
+
+from insights.core import Scannable
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+
+@parser(Specs.cron_foreman)
+class CronForeman(Scannable):
```
bfahr, line 17: @huali027 why not use the [crontab](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/crontab.py) parser here instead of Scannable, and add to the module with the other crontab data parsers?

huali027: @bfahr, the jobs under "/ect/cron.d" is a little different from "crontab", this is from the man doc:
```
Cron checks these files and directories:

/etc/crontab
      system  crontab.   Nowadays  the  file  is empty by default.  Originally it was usually used to run daily, weekly, monthly jobs.  By default these jobs are now run through anacron which reads /etc/anacrontab configuration file.  See anacrontab(5) for more details.

/etc/cron.d/
     directory that contains system cronjobs stored for different users.

/var/spool/cron
    directory that contains user crontables created by the crontab command.
```

for the jobs in "/etc/cron.d/*", the format is like:
```
* * * * * root touch /tmp/file
```
Additionally, the `username` is extra needed in the jobs line. 

Since it is classified in the cron doc and also the format is a little different, I didn't put it in the "crontab" module. Currently, the "crontab" module covers the jobs in the "/var/spool/cron" directory.

And also the new advisor rule does not care about the time, just the command, so I use "Scannable" directly.
Do we need to put it together and enhance the `CrontabL` to support two kinds of formats?

bfahr: @huali027 thank you for the explanation.  What would probably be most helpful would be more examples of how to get the information from the parser.  A developer would really have to dig into the Scannable code to understand how this parser provides data without more examples.

huali027: @bfahr, sure, I'll add more examples.

huali027:
@xiangce Please help to review again, thanks.

xiangce (changes requested):

insights/parsers/cron_jobs.py
```diff
@@ -0,0 +1,120 @@
...
+"""
+
+from insights.core import Parser
+from insights.parsers import get_active_lines, ParseException
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+
+@parser(Specs.cron_foreman)
+class CronForeman(Parser, list):
```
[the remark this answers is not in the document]
huali027: > I believe it's possible to support more users in the future, better make a super-class for it.

updated as suggested, please help to check, thanks.

psachin (approved):

PR #3522, opened 2022-09-13
Remove usage of reregistration and deprecate cli-option

ahitacat:

### All Pull Requests:

Check all that apply:

Working on deprecate --force-reregistration client option. This option will be deprecate in favor of the explicit running of `insights-client --unregister` and `insights-client --register`. This work is part of the improvement of the re-registration safeguard [RHIN-500]

Resolves: #rhbz2071082

### DIFF
--- a/insights/client/config.py
+++ b/insights/client/config.py
@@ -334,7 +334,9 @@ DEFAULT_OPTS = {
     'reregister': {
         'default': False,
         'opt': ['--force-reregister'],
-        'help': 'Forcefully reregister this machine to Red Hat. Use only as directed.',
+        'help': ('This flag is deprecated and it will be removed in a future release.'
+                 'Forcefully reregister this machine to Red Hat.'
+                 'Please use `insights-client --unregister && insights-client --register `instead'),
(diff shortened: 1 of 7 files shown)


clarabez:
LGTM!

insights/client/config.py
```diff
@@ -326,7 +326,7 @@ def _core_collect_default():
     'reregister': {
         'default': False,
         'opt': ['--force-reregister'],
-        'help': 'Forcefully reregister this machine to Red Hat. Use only as directed.',
+        'help': argparse.SUPPRESS,
```
strider, line 329: As we are in deprecation mode, I will not silence that option from `argparse` now. But I would suggest adding the deprecation message, announcing the ineffectiveness of that latter and its future removal.

strider (changes requested):

insights/client/config.py
ahitacat: Agree, I have added a new message

strider (approved):
LGTM

PR #3523, opened 2022-09-14
Removing extraneous space inserted during refactoring commit 894484a.…

mike-kingsbury:
…  Code was moved to a new function, and a space was added to the split.  bz#2126155

### All Pull Requests:

Check all that apply:

* [ x ] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [ x ] Is this PR to correct an issue?

Reverting a change that introduced a space into a .split(',').  The space broke the functionality of the Insights client when using NO_PROXY with multiple items.

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -276,7 +276,10 @@ class InsightsConnection(object):
         if no_proxy_info:
             insights_service_host = urlparse(self.base_url).hostname
             logger.debug('Found NO_PROXY set. Checking NO_PROXY %s against base URL %s.', no_proxy_info, insights_service_host)
-            for no_proxy_host in no_proxy_info.split(', '):
+            # Split the no_proxy entries on ',', then strip any leading and trailing whitespace.  Create a clean list for the
+            # for loop.
+            no_proxy_info = [host.strip() for host in no_proxy_info.split(',')]
+            for no_proxy_host in no_proxy_info:
                 logger.debug('Checking %s against %s', no_proxy_host, insights_service_host)
                 if no_proxy_host == '*':
                     proxies = None
--- a/insights/tests/client/connection/test_init_session.py
+++ b/insights/tests/client/connection/test_init_session.py
@@ -83,6 +83,14 @@ def test_get_no_proxy_env():
         connection.get_proxies()
     assert connection.proxies is None
 
+    with patch.dict(os_environ, {"HTTPS_PROXY": "env.proxy.example.com", "NO_PROXY": "redhat.com,example.com"}, clear=True):
+        connection.get_proxies()
+    assert connection.proxies is None
+
+    with patch.dict(os_environ, {"HTTPS_PROXY": "env.proxy.example.com", "NO_PROXY": "url.com,example.com"}, clear=True):
+        connection.get_proxies()
+    assert connection.proxies == {'https': 'env.proxy.example.com'}
+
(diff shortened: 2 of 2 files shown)


xiangce:
I'm just wondering if we could add a test case for this? It seems an issue that can be detected by a sufficient test.

mike-kingsbury:
> I'm just wondering if we could add a test case for this? It seems an issue that can be detected by a sufficient test.

Give me a couple of minutes, you're correct it should be testable.

xiangce:
@mike-kingsbury - Thanks for the update. It looks good to me.  @ahitacat, please have a look, if it is also good for you, I think we can merge and release today.

ahitacat:
In that case it would fail with `no_proxy` values separated by comma **and** space. Like the one is given in the [bug report](https://bugzilla.redhat.com/show_bug.cgi?id=2065233) This no_proxy function get the the value from env or from rhsm. So let's try to support bot of them.

insights/client/connection.py
```diff
@@ -276,7 +276,7 @@ def get_proxy(self, proxy_info, no_proxy_info, environment):
         if no_proxy_info:
             insights_service_host = urlparse(self.base_url).hostname
             logger.debug('Found NO_PROXY set. Checking NO_PROXY %s against base URL %s.', no_proxy_info, insights_service_host)
-            for no_proxy_host in no_proxy_info.split(', '):
+            for no_proxy_host in no_proxy_info.split(','):
```
ahitacat, line 279:
```suggestion
            for no_proxy_host in no_proxy_info.split(','):
                no_proxy_host.strip()
```
To remove any white spaces added while listing the values.

insights/tests/client/connection/test_init_session.py
```diff
@@ -83,6 +83,14 @@ def test_get_no_proxy_env():
         connection.get_proxies()
     assert connection.proxies is None
 
+    with patch.dict(os_environ, {"HTTPS_PROXY": "env.proxy.example.com", "NO_PROXY": "redhat.com,example.com"}, clear=True):
```
ahitacat, line 86:
```suggestion
    with patch.dict(os_environ, {"HTTPS_PROXY": "env.proxy.example.com", "NO_PROXY": "example.com,redhat.com"}, clear=True):
```
The problem appears when getting values after a comma.

ahitacat (changes requested):
Also it could be useful to add a check under `test_get_no_proxy_rhsm` that tests when `no_proxy` has values separated by comma and space.

~~~
    with patch.dict(os_environ, {"HTTPS_PROXY": "env.proxy.example.com", "NO_PROXY": "example.com, redhat.com"}, clear=True):
        connection.get_proxies()
    assert connection.proxies == {'https': 'env.proxy.example.com'}
 ~~~

insights/client/connection.py
mike-kingsbury: Almost....   .strip() doesn't modify the string, it returns a new string.  Implemented as such.

insights/tests/client/connection/test_init_session.py
mike-kingsbury: I've added a series of new test cases including spaces and different ordering of the domain names.

ahitacat: This looks great :smile:

ahitacat (commented):
This changes looks great and it can be merged. Thanks @mike-kingsbury for open and fix the bug.

ahitacat (approved):

PR #3525, opened 2022-09-15
Add spec and parser for luksmeta command

danzatt:

### All Pull Requests:

Check all that apply:

- added `luks1_block_devices` which depends on `LuksDump` parser and filters just LUKS1 devices and returns their paths (based on the UUID)
- added `luksmeta` spec which runs `luksmeta show` command on every LUKS1 device to show information about custom metadata embedded in the keyslots
- added `LuksMeta` parser to parse the `luksmeta show` command

### DIFF
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -259,7 +259,7 @@ plugins:
         - name: insights.components.cryptsetup
           enabled: true
 
-        - name: insights.parsers.cryptsetup_luksDump.LuksDump
+        - name: insights.combiners.cryptsetup.luks1_block_devices
           enabled: true
 """.strip()
 
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -258,9 +258,6 @@ plugins:
 
         - name: insights.components.cryptsetup
           enabled: true
-
-        - name: insights.combiners.cryptsetup.luks1_block_devices
-          enabled: true
 """.strip()
 
 
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -258,6 +258,9 @@ plugins:
 
         - name: insights.components.cryptsetup
           enabled: true
+
+        - name: insights.parsers.cryptsetup_luksDump.LuksDump
+          enabled: true
 """.strip()
 
 
--- a/insights/combiners/cryptsetup.py
+++ b/insights/combiners/cryptsetup.py
@@ -1,0 +1,37 @@
+"""
+Cryptsetup - combine metadata about LUKS devices
+================================================
+
+Use LuksDump parser to filter just LUKS version 1 devices and return their
+paths (by UUID).
+"""
+
+from insights import SkipComponent
+from insights.core.plugins import combiner
+from insights.parsers.cryptsetup_luksDump import LuksDump
+
+
+@combiner(LuksDump)
+def luks1_block_devices(dumps):
+    """
+    This combiner provides a list of LUKS version 1 encrypted device.
+
+    Sample data returned::
+
+        ['/dev/disk/by-uuid/2a6c383b-f57c-4a77-8da2-a83b2e4aa9c5', '/dev/disk/by-uuid/76f799fd-8aaa-4b25-828c-976db6a54307']
+
+    Returns:
+        list: List of the LUKS version 1 encrypted block devices.
+
+    Raises:
+        SkipComponent: When there is not any LUKS version 1 encrypted block
+        device on the system.
+    """
+
+    luks1_devices = []
+    for luks_device in filter(lambda x: x.dump["header"]["Version"] == "1", dumps):
+        luks1_devices.append("/dev/disk/by-uuid/" + luks_device.dump["header"]["UUID"])
+
+    if not luks1_devices:
+        raise SkipComponent
+    return luks1_devices
--- a/insights/combiners/cryptsetup.py
+++ b/insights/combiners/cryptsetup.py
@@ -2,43 +2,15 @@
 Cryptsetup - combine metadata about LUKS devices
 ================================================
 
-Use LuksDump parser to filter just LUKS version 1 devices and return their
-paths (by UUID). Also combine outputs of LuksDump and LuksMeta parsers (with
-the same UUID) into a single LuksDevice.
+Combine outputs of LuksDump and LuksMeta parsers (with the same UUID) into a
+single dictionary.
 """
 
 import copy
 
-from insights import SkipComponent
 from insights.core.plugins import combiner
 from insights.parsers.cryptsetup_luksDump import LuksDump
 from insights.parsers.luksmeta import LuksMeta
-
-
-@combiner(LuksDump)
-def luks1_block_devices(dumps):
-    """
-    This combiner provides a list of LUKS version 1 encrypted device.
-
-    Sample data returned::
-
-        ['/dev/disk/by-uuid/2a6c383b-f57c-4a77-8da2-a83b2e4aa9c5', '/dev/disk/by-uuid/76f799fd-8aaa-4b25-828c-976db6a54307']
-
-    Returns:
-        list: List of the LUKS version 1 encrypted block devices.
-
-    Raises:
-        SkipComponent: When there is not any LUKS version 1 encrypted block
-        device on the system.
-    """
-
-    luks1_devices = []
-    for luks_device in filter(lambda x: x.dump["header"]["Version"] == "1", dumps):
-        luks1_devices.append("/dev/disk/by-uuid/" + luks_device.dump["header"]["UUID"])
-
-    if not luks1_devices:
-        raise SkipComponent
-    return luks1_devices
 
 
 @combiner(LuksDump, optional=[LuksMeta])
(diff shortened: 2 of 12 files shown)


insights/parsers/luksmeta.py
```diff
@@ -0,0 +1,79 @@
...
+
+        >>> parsed_result.keyslots[0].metadata is None
+        True
+
+        >>> parsed_result.keyslots[1].metadata
+        'cb6e8904-81ff-40da-a84a-07ab9ab5715e'
+
+
+    Attributes:
+        keyslots(dict): A list of 8 KeyslotSpecification objects, describing
```
xiangce, line 64: I think this `keyslots` should be a list.

insights/parsers/luksmeta.py
```diff
@@ -0,0 +1,79 @@
...
+        every LUKS keyslot. The KeyslotSpecification contains the index, state
```
xiangce, line 65: To make sure the docstring can be shown correctly after being compiled, please add one more intention from the 2nd line, e.g.
```
keyslots(list): A list of 8 KeyslotSpecification objects, describing
      and metadata fileds. Metadata field stores the UUID of the application
      that has stored metadata in the keyslot.
```

insights/parsers/luksmeta.py
```diff
@@ -0,0 +1,79 @@
...
+        and metadata fileds. Metadata field stores the UUID of the application
+        that has stored metadata in the keyslot.
+    """  # noqa
+
+    def __init__(self, context):
+        self.keyslots = [None] * 8
```
xiangce, line 71: Does it make sense to "skip" this parser by raising a `SkipException` when there is nothing needed to parse?

insights/specs/datasources/luks_devices.py
```diff
@@ -123,3 +124,28 @@ def luks_data_sources(broker):
         return datasources
 
     raise SkipComponent
+
+
+@datasource(LuksDump, HostContext)
```
xiangce, line 129: Since the `LuksDump` is required for this datasource, please add it to the `insights/collect.py` too.

insights/parsers/luksmeta.py
```diff
@@ -0,0 +1,79 @@
+"""
+luksmeta - command ``luksmeta show``
+====================================
+This class provides parsing for the output of luksmeta <device_name>.
```
xiangce, line 4: it should be `luksmeta -d <device_name>` here in the doc?

insights/parsers/luksmeta.py
xiangce: BTW, if it works for you, it's recommended to mix the parser with the built-in `list` and remove this single attribute, when it's a simple list.


```
class LuksMeta(Parser, list):

    self[index] = ...

```

xiangce (changes requested):
I also find this [KCS-3617741](https://access.redhat.com/solutions/3617741), please check if this error `"Device is not initialized (/dev/mapper/vg_luks-lv_luks)"` should be handled or not in this parser. Thanks.

danzatt:
I have gone through the remarks and fixed them now.

insights/parsers/luksmeta.py
xiangce: You may miss this one.

xiangce (approved):
Thanks for the update, it looks good to me now.

insights/parsers/luksmeta.py
danzatt: Fixed now

insights/parsers/luksmeta.py
```diff
@@ -0,0 +1,98 @@
...
+        'cb6e8904-81ff-40da-a84a-07ab9ab5715e'
+    """  # noqa
+
+    def parse_content(self, content):
+        if len(content) >= 1 and "Device is not initialized" in content[0]:
+            raise SkipComponent
+
+        # LUKS1 contains exactly 8 keyslots
+        if len(content) != 8:
+            raise SkipComponent
```
bfahr, line 92: @danzatt your comment implies that there must be 8 keyslots.  Is it an error if there are not exactly 8?

insights/specs/datasources/luks_devices.py
bfahr: @xiangce @danzatt why not just use parsers and a collector instead of adding a datasource that has to run on a customer systems?

bfahr, line 129: @xiangce @danzatt why not just use parsers and a collector instead of adding a datasource that has to run on a customer systems?

xiangce: @danzatt, yeap, what @bfahr suggested makes sense to me. It seems possible to run the `luksmeta -d` command against all the `/dev/disk/by-uuid/*` devices. And then get what you wanted (uuids where ["header"]["Version"] == "1" in LuksDump) via a combiner.

insights/specs/default.py
```diff
@@ -369,6 +370,7 @@ class DefaultSpecs(Specs):
     lspci_vmmkn = simple_command("/sbin/lspci -vmmkn")
     lsscsi = simple_command("/usr/bin/lsscsi")
     lsvmbus = simple_command("/usr/sbin/lsvmbus -vv")
+    luksmeta = foreach_execute(cryptsetup.luks1_block_devices, "luksmeta show -d %s")
```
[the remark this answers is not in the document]
xiangce: BTW, please use the full command path for the `luksmeta` command (/usr/bin/luksmeta) in the spec, since the insights-client may not get the correct PATH environment.


And, Would you please modify the following two specs to use the full path (/usr/sbin/cryptsetup)? https://github.com/RedHatInsights/insights-core/blob/534888a20d32c256dd1443fe8b281c9927e4054d/insights/specs/datasources/luks_devices.py#L43-L44

And sorry for missing this in the review of that PR.

insights/collect.py
```diff
@@ -251,6 +251,9 @@
 
         - name: insights.components.cryptsetup
           enabled: true
+
+        - name: insights.combiners.cryptsetup.luks1_block_devices
```
xiangce, line 255: If It's okay to run the `luksmeta` against all the `by-uuid` devices, this component is not required and is not needed to be `pre-loaded`.

danzatt:
Yes, indeed it should be simpler to run the `luksmeta` command against all devices. I have changed the PR now and removed the `luks1_block_devices` dependency. (I fixed the `cryptsetup`/`luksmeta` paths too.)

insights/combiners/cryptsetup.py
```diff
@@ -0,0 +1,80 @@
...
+
+import copy
+
+from insights import SkipComponent
+from insights.core.plugins import combiner
+from insights.parsers.cryptsetup_luksDump import LuksDump
+from insights.parsers.luksmeta import LuksMeta
+
+
+@combiner(LuksDump)
```
xiangce, line 18: @danzatt - from my perspective, It seems this combiner function is not so required. Firstly, it's based on the single parser, if really want such an interface to return the "luks1 block devices", you can totally implement it in the `LuksDump` parser itself. Or, you can move this check into the `LuksDevices` combiner as a double safety-check here L77: `if luks_metas and uuid in luksmeta_by_uuid:`

insights/combiners/cryptsetup.py
```diff
@@ -0,0 +1,80 @@
...
+                luksmeta_by_uuid[luks_meta["device_uuid"].lower()] = luks_meta
+
+        for luks_dump in luks_dumps:
+            uuid = luks_dump.dump["header"]["UUID"].lower()
+            luks_dump_copy = copy.deepcopy(luks_dump.dump)
+
+            if luks_metas and uuid in luksmeta_by_uuid:
+                luks_dump_copy["luksmeta"] = luksmeta_by_uuid[uuid]
+
+            self.append(luks_dump_copy)
```
xiangce, line 80: Better add a check here: when got nothing from the combination (e.g. no luks devices and the list is empty), better raise a SkipComponent to skip the combiner.

danzatt:
Is there anything I can do to get this merged?

insights/combiners/cryptsetup.py
```diff
@@ -0,0 +1,80 @@
...
+    luks1_devices = []
+    for luks_device in filter(lambda x: x.dump["header"]["Version"] == "1", dumps):
+        luks1_devices.append("/dev/disk/by-uuid/" + luks_device.dump["header"]["UUID"])
+
+    if not luks1_devices:
+        raise SkipComponent
+    return luks1_devices
+
+
+@combiner(LuksDump, optional=[LuksMeta])
```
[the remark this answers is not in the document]
xiangce: I just checked your latest update to the test of this combiner, it now looks good to me. Sorry for the late reply, I was just back from a long leave.

bfahr (approved):

insights/combiners/cryptsetup.py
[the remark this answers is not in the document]
danzatt: Great, thank you for the review.

PR #3526, opened 2022-09-16
Add combiner rhel for edge

wushiqinlou:
### All Pull Requests:

Check all that apply:

Create a combiner to determine if it is an edge computing system.


insights/combiners/rhel_for_edge.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, rpms, cmdline, units, redhatrelease):
+        self.is_edge = False
+        self.is_automated = False
+
+        if 'rpm-ostree' in rpms and 'yum' not in rpms and 'ostree' in cmdline and "red hat enterprise linux release" in redhatrelease.raw.lower():
+            self.is_edge = True
+            if units.is_running("rhcd.service"):
+                self.is_automated = True
+
+    def gen_edge_return_data(self, update_target_packages, install_target_packages, target_repos):
```
xiangce, line 61: @wushiqinlou - I'm a bit confused with this member function. It just sets a dict according to the given arguments, is it required to be an interface to the combiner? If so, please add a docstring and test for it.

insights/combiners/rhel_for_edge.py
```diff
@@ -0,0 +1,70 @@
...
+    Attributes:
+        is_edge (bool): True when it is an edge computing system
+        is_automated (bool): True when the the edge computing system is configured to use automated management
+    """
+
+    def __init__(self, rpms, cmdline, units, redhatrelease):
+        self.is_edge = False
+        self.is_automated = False
+
+        if 'rpm-ostree' in rpms and 'yum' not in rpms and 'ostree' in cmdline and "red hat enterprise linux release" in redhatrelease.raw.lower():
```
xiangce, line 56: Per the current implementation, this combiner will always get triggered and return something. Whether it's Okay to just skip this combiner by raising a "SkipComponent` when a system is not an `edge computing system`? which could save a bit of memory on the server.

Meanwhile, take this combiner an optional `condition` in the rule, only when it's got fire, it's an `edge computing system. What do you think? @wushiqinlou

skontar: I would prefer to not raise `SkipComponent`. I think that should be raised only if invalid data are provided to the `combiner`? The docstring says "This class should be raised by components that want to be taken out of dependency resolution." If the combiner gets valid data and can make decision if system is or is not Edge system, I do not see a reason why it should remove itself from dependency resolution instead of providing expected answer?

xiangce: It also makes sense to me that skips the combiner only when cannot make the decision due to the lack of required parsers.
I just thought from the Advisor rules perspective, the results are the same, and "nothing" will be done for the advisor rule (when the rule also works for `edge`): no matter when either it's not an `edge` or cannot identify if it's an `edge`.

I have no problem making the combiner always get triggered if it's hard for the ProdSec rules to work in the same way as the Advisor rules.

skontar: You are right that rules in this case do not care about distinguishing between "not edge" and "do not know". The reason why I do not like it is that when debugging rules you may have hard time finding out if there is a bug in combiner or if really so many systems are "not edge". 

Rules could work with `SkipComponent` by basically writing a workaround in condition (`if combiner is None: this actually means False and you better hope it is not None all the time due to bug`), I think that approach is unexpected from the rule writer perspective and semantically incorrect.

Prodsec rules can handle it either way.

xiangce: I agree with you.

insights/combiners/rhel_for_edge.py
wushiqinlou: Remove the function as it is not strong dependency usage.

insights/combiners/rhel_for_edge.py
```diff
@@ -0,0 +1,58 @@
...
+
+            Connection status for test.localhost:
+            - Connected to Red Hat Subscription Management
+            - The Red Hat connector daemon is active
+
+    If a system can upload insights archive, it must be connected to Red Hat Subscription Management, rhcd service running
+    means an edge computing system is configured to use automated management. Note: it is able to run rhcd service on the
+    edge system created corkpit edge image, "is_automated" is only for front-end resolution surface, it is used when
+    customers determine that the image is from console.
+
```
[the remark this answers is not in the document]
wushiqinlou: Updated the description

xiangce (approved):

insights/combiners/rhel_for_edge.py
```diff
@@ -0,0 +1,64 @@
...
+from insights.parsers.cmdline import CmdLine
+from insights.parsers.installed_rpms import InstalledRpms
+from insights.parsers.systemd.unitfiles import ListUnits
+from insights.parsers.redhat_release import RedhatRelease
+
+
+@combiner(InstalledRpms, CmdLine, ListUnits, RedhatRelease)
+class RhelForEdge(object):
+    """
+    Combiner for checking if the system is an edge computing systems.
```
jobselko, line 21: I think that `Sphinx` will have a problem parsing the comments. Have you tried to generate the docs via `make html_debug`?
Also, I have a problem understanding the docstring; maybe it would be worth reworking it a bit.

insights/combiners/rhel_for_edge.py
```diff
@@ -0,0 +1,64 @@
...
+        >>> rhel_for_edge_obj.is_edge
+        True
+        >>> rhel_for_edge_obj.is_automated
+        True
+    """
+
+    def __init__(self, rpms, cmdline, units, redhatrelease):
+        self.is_edge = False
+        self.is_automated = False
+        if 'rpm-ostree' in rpms and 'yum' not in rpms and 'ostree' in cmdline and "red hat enterprise linux release" in redhatrelease.raw.lower():
```
jobselko, line 61: The long "if statement" is hard to read. What do you think of this variant?

```
    def __init__(self, rpms, cmdline, units, redhatrelease):
        self.is_edge = False
        self.is_automated = False

        edge_rpm = 'rpm-ostree' in rpms and 'yum' not in rpms
        edge_cmdline = 'ostree' in cmdline
        edge_rhel = 'red hat enterprise linux release' in redhatrelease.raw.lower()

        if edge_rpm and edge_cmdline and edge_rhel:
            self.is_edge = True
            if units.is_running("rhcd.service"):
                self.is_automated = True
```

insights/tests/combiners/test_rhel_for_edge.py
```diff
@@ -0,0 +1,110 @@
...
+kernel-4.18.0-305.19.1.el8_4.x86_64
+""".strip()
+
+CONTENT_INSTALLED_RPMS_EDGE = """
+kernel-4.18.0-305.19.1.el8_4.x86_64
+rpm-ostree-2021.5-2.el8.x86_64
+""".strip()
+
+
+def test_rhel_for_edge_true_1():
```
jobselko, line 45: You have five tests for testing the same structure/behaviour. I think that usage of `@pytest.mark.parametrize` would be helpful here. (I also think the variable names could be shorter, e.g. without the `CONTENT_` prefix.)

All of your five tests can be written as follows:
```
@pytest.mark.parametrize("rpms, cmdline, systemctl, rhel, is_edge, is_automated", [
    (CONTENT_INSTALLED_RPMS_EDGE, CMDLINE_EDGE, CONTENT_SYSTEMCTL_LIST_UNITS_NO_AUTOMATED,
     CONTENT_REDHAT_RELEASE_RHEL, True, False),
    (CONTENT_INSTALLED_RPMS_EDGE, CMDLINE_EDGE, CONTENT_SYSTEMCTL_LIST_UNITS_AUTOMATED,
     CONTENT_REDHAT_RELEASE_RHEL, True, True),
    (CONTENT_INSTALLED_RPMS_RHEL, CMDLINE_EDGE, CONTENT_SYSTEMCTL_LIST_UNITS_AUTOMATED,
     CONTENT_REDHAT_RELEASE_RHEL, False, False),
    (CONTENT_INSTALLED_RPMS_EDGE, CMDLINE_RHEL, CONTENT_SYSTEMCTL_LIST_UNITS_AUTOMATED,
     CONTENT_REDHAT_RELEASE_RHEL, False, False),
    (CONTENT_INSTALLED_RPMS_EDGE, CMDLINE_EDGE, CONTENT_SYSTEMCTL_LIST_UNITS_AUTOMATED,
     CONTENT_REDHAT_RELEASE_COREOS, False, False)
])
def test_rhel_for_edge(rpms, cmdline, systemctl, rhel, is_edge, is_automated):
    install_rpms = InstalledRpms(context_wrap(rpms))
    cmdline = CmdLine(context_wrap(cmdline))
    list_units = ListUnits(context_wrap(systemctl))
    redhat_release = RedhatRelease(context_wrap(rhel))

    result = RhelForEdge(install_rpms, cmdline, list_units, redhat_release)
    assert result.is_edge is is_edge
    assert result.is_automated is is_automated
```

Let me know if you have any questions regarding the example I wrote.

jobselko (commented):
Hi @wushiqinlou, I have a few comments, but probably nothing important. You can ignore them and resolve these discussions if you think they are unnecessary.

In the end, I am not the one who approves this PR :)

insights/combiners/rhel_for_edge.py
wushiqinlou: @jobselko - We checked the docs via `html_debug`. Could you please point out what kind of problem in detail?

insights/tests/combiners/test_rhel_for_edge.py
wushiqinlou: @jobselko - I think current test format is OK, you can raise a PR to update the test format which you prefer after this PR gets merged.

insights/combiners/rhel_for_edge.py
jobselko: I am sorry, you are right. For some reason, I had a missing empty line behind `::`, but in the current version, the formatting is ok.

skontar (approved):

PR #3528, opened 2022-09-20
fix: make SAPHostCtrlInstances compatible with old archives

xiangce:
- fix: #3530
- Generate the InstanceType as per the InstanceName, since it is not collected by the old archives.

### All Pull Requests:

Check all that apply:

- Since the old archives didn't collect the InstanceType lines, when applying the new logic of #3512 against the old archives an exception will be thrown and the `Sap` combiner gets fail.
- This PR fixes this and makes the parse compatible with the old archives.

### DIFF
--- a/insights/parsers/saphostctrl.py
+++ b/insights/parsers/saphostctrl.py
@@ -108,8 +108,12 @@ class SAPHostCtrlInstances(CommandParser, list):
             inst[fields[0]] = fields[2]
 
             if fields[0] == 'InstanceName':
+                short_type = fields[2].strip('0123456789')
+                # Back forward compatible: a short InstanceType was from the InstanceName
+                if 'InstanceType' not in inst:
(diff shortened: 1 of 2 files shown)


psachin (approved):

insights/parsers/saphostctrl.py
```diff
@@ -108,8 +108,12 @@ def parse_content(self, content):
             inst[fields[0]] = fields[2]
 
             if fields[0] == 'InstanceName':
+                short_type = fields[2].strip('0123456789')
+                # Back forward compatible: a short InstanceType was from the InstanceName
+                if 'InstanceType' not in inst:
```
psachin, line 113: These lines are missed in the coverage.

xiangce: updated

PR #3539, opened 2022-10-04
fix: Add ability to return exceptions during insights collect

vdymna:
### All Pull Requests:

Check all that apply:

* Modified how exceptions are handleled in the `insights.core.serde.marshal` function, so error tracebacks are captured and serialized to a meta-data file in the archive (as before) and also exceptions are added to the broker for later retrieval.

* Added `_parse_broker_exceptions` and modified `collect` function in the `insights.collect` module, so all broker exceptions are analyzed after core collection and exception types that are defined in `EXCEPTIONS_TO_REPORT` set will be included as a dict and returned from the `collect()` function, as a second value in the tuple. This is needed so insights-client can analyze those errors and direct the collection logic flow differently if necessary.

### DIFF
--- a/insights/client/core_collector.py
+++ b/insights/client/core_collector.py
@@ -52,7 +52,12 @@ class CoreCollector(DataCollector):
         manifest = collect.default_manifest
         if hasattr(self.config, 'manifest') and self.config.manifest:
             manifest = self.config.manifest
-        collected_data_path = collect.collect(manifest=manifest, tmp_path=self.archive.tmp_dir, rm_conf=core_blacklist, client_timeout=self.config.cmd_timeout)
+        collected_data_path, exceptions = collect.collect(
+            manifest=manifest,
+            tmp_path=self.archive.tmp_dir,
+            rm_conf=core_blacklist,
+            client_timeout=self.config.cmd_timeout
+        )
 
         # update the archive dir with the reported data location from Insights Core
         if not collected_data_path:
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -260,6 +260,11 @@ plugins:
           enabled: true
 """.strip()
 
+EXCEPTIONS_TO_REPORT = set([
+    OSError
+])
+"""Exception types that should be reported on after core collection."""
+
 
 def load_manifest(data):
     """ Helper for loading a manifest yaml doc. """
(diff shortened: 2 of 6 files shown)


vdymna:
@subpop please take a look at this when you have a chance.

vdymna:
@Glutexo and @ahitacat can you please take a look at this PR, since this is going to affect core collection, and will require some changes in insights-client.

ahitacat (changes requested):
Hi Vitaliy, I like this idea.
However the change is not complete, I have tried to run it and this new collection broke insights-client. Are you willing to fill in the missing parts by yourself?
Feel free to ask if you run into unclarities with the Client part of the codebase.

vdymna:
@ahitacat yes that is correct, this PR introduces breaking changes because the return signature of the `collect()` function in the `insights.collect` is changing, this why I'm trying to bring this to everyone's attention. This insights-core PR will require corresponding changes in the insights-client codebase. At the bare minimum, the `CoreCollector` class has to be changed [here](https://github.com/RedHatInsights/insights-core/blob/127edf4291e0c97d2b2d702df9b6b35f530c98d4/insights/client/core_collector.py#L55). 
Since our team does not make changes to the insights-client, I have not made any of those corresponding changes.

If you would like, I can go ahead and make the changes to the `CoreCollector` [code](https://github.com/RedHatInsights/insights-core/blob/127edf4291e0c97d2b2d702df9b6b35f530c98d4/insights/client/core_collector.py#L55), by adding that new variable that will hold exceptions returned by the `collect()` function? So the code can just run without failing.
However, I think a final solution will require a holistic approach from the insights-client team to first define and then implement what exactly the client should do or not do (archive wise) when various errors are occurring during the spec collection (like OSErrors for example).

vdymna:
@ahitacat I've made the mentioned change, please take a look again.

bfahr (approved):

vdymna:
@xiangce wanted to say to please not to merge this PR yet, I'm waiting on the insights-client team to make sure they have looked at it and approved it. Thanks.

xiangce:
@vdymna - Acked. Thanks for the information.

ahitacat:
test me

ahitacat (approved):

PR #3540, opened 2022-10-05
Registration check unregisters when it is not connected

ahitacat:

### All Pull Requests:

Check all that apply:

Insights-client is in a faulty state when the system is unregistered from rhsm and it is configured with `legacy_upload=True`. With no legacy option set, insights-client unregisters the system when it receives a response with a status greater than 400, that means rhsm could not be reached for many reasons. With legacy option it doesn't registers the system. This makes inconsistent the behavior of insights-client when checking the registration status.

Actual results:
```
> sudo insights-client --status
System is registered locally via .registered file. Registered at 2022-08-31T12:29:02.150601
Insights API says this machine is NOT registered.
```
Results with `legacy_upload=False`
```
 >sudo insights-client --status
This host is registered.
```

Fix results:
With `legacy_upload=True`
`registration_check()` method will only unregisters the system locally if API tells is not registered.
```
> sudo insights-client --status
System is registered locally via .registered file. Registered at 2022-10-06T08:04:15.668125
Insights API says this machine is NOT registered.
System unregistered locally via .unregistered file
```
Results with `legacy_upload=False`
`registration_check()` method will only unregisters the system locally if API tells is not registered, not every time the API could not be reached.
```
 >sudo insights-client --status
This host is registered.
```

Resolves: rhbz#2123776


Glutexo (approved):
The if-else statements are not very clear, but that’s the legacy of the already convoluted code. I will revisit this more thoroughly to reduce the amount of technical debt. But I don’t want to block the pull request any longer.

PR #3547, opened 2022-10-11
feat(client): add --manifest argument

subpop:
Add a top-level --manifest argument. This argument expects a path to an
insights collection manifest file. If the argument is not None, the
manifest is used instead of the default manifest to collect data.

Fixes: https://issues.redhat.com/browse/ESSNTL-3540

### All Pull Requests:

Check all that apply:

*Add your description here*

### DIFF
--- a/insights/client/core_collector.py
+++ b/insights/client/core_collector.py
@@ -51,7 +51,11 @@ class CoreCollector(DataCollector):
 
         manifest = collect.default_manifest
         if hasattr(self.config, 'manifest') and self.config.manifest:
-            manifest = self.config.manifest
+            if self.config.app is None:
+                with open(self.config.manifest, 'r') as f:
+                    manifest = f.read()
+            else:
+                manifest = self.config.manifest
         collected_data_path, exceptions = collect.collect(
             manifest=manifest,
             tmp_path=self.archive.tmp_dir,
--- a/insights/client/config.py
+++ b/insights/client/config.py
@@ -127,6 +127,14 @@ DEFAULT_OPTS = {
         'group': 'actions',
         'dest': 'app'
     },
+    'manifest': {
+        'default': None,
+        'opt': ['--manifest'],
+        'help': 'Collect using the provided manifest',
+        'action': 'store',
+        'group': 'actions',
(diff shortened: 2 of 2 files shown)


insights/client/core_collector.py
```diff
@@ -51,7 +52,8 @@ def run_collection(self, conf, rm_conf, branch_info, blacklist_report):
 
         manifest = collect.default_manifest
         if hasattr(self.config, 'manifest') and self.config.manifest:
-            manifest = self.config.manifest
+            with open(self.config.manifest, 'r') as f:
```
ahitacat, line 55: When `--collector` is used it sets the variable to the content of a content manifest. In this case, this line will fail as config.manifest is not a file.
I'd suggest to check if config.app is not None before open the manifest as a file. 

Also to check if config.manifest is a file, we would want to handle the exception if the manifest is wrong specified.

ahitacat (changes requested):

insights/client/core_collector.py
subpop: Does it make sense to catch the Exception if we don't have any action to take? Shouldn't we let the Exception go unhandled, so that callers of `run_collection` can catch it and handle it accordingly?

subpop:
I changed the logic a bit. Now it keeps the first condition untouched, and only reads the contents of a file if `self.app` is None.

insights/client/core_collector.py
```diff
@@ -51,7 +51,11 @@ def run_collection(self, conf, rm_conf, branch_info, blacklist_report):
 
         manifest = collect.default_manifest
         if hasattr(self.config, 'manifest') and self.config.manifest:
-            manifest = self.config.manifest
+            if self.app is None:
```
ahitacat, line 54:
```suggestion
            if self.config.app is None:
```

ahitacat (changes requested):

insights/client/core_collector.py
subpop: Oh duh. Thank you. I should not be allowed to write Python.

ahitacat (approved):
LGTM

PR #3553, opened 2022-10-14
[WIP] Allow --retries to work with network exceptions

mhuth:

Check all that apply:

This is an idea to fix https://bugzilla.redhat.com/show_bug.cgi?id=2117686

The insights-core connection.py code doesn't really handle network exceptions.  Callers of the connection._http_request code just expect to see status_codes containing errors rather than handling any network exceptions.  This PR catches network exceptions in the _http_request method and converts it into a requests.Response object with an error status code that can be handled further up the call stack by functions that just look at the status_code.  However, rather than creating a *real* response object, I create a Mock Response object, because we didn't really get a response, but the functions up the call stack only really handle responses, not exceptions.  And so it is with the --retries logic, it only really works with responses and status_codes.  

With this in mind, rather than rewriting all this network stack stuff to raise/handle exceptions and potentially break existing functionality, I just create a fake response object for the network exception and give it a generic 502 Bad Gateway response code.  

Cop out? ... indeed.  Ugly? ... perhaps.  Simple(r)? ... yep.  Seems to work? ... seems to ...

### DIFF
--- a/insights/client/client.py
+++ b/insights/client/client.py
@@ -314,6 +314,7 @@ def _legacy_upload(config, pconn, tar_file, content_type, collection_duration=No
     logger.info('Uploading Insights data.')
     api_response = None
     for tries in range(config.retries):
+        logger.debug("Upload attempt %d of %d ...", tries + 1, config.retries)
         upload = pconn.upload_archive(tar_file, '', collection_duration)
 
         if upload.status_code in (200, 201):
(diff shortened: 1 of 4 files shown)


mhuth:
Closing

mhuth:
Closing this in favour of https://github.com/RedHatInsights/insights-core/pull/3558

PR #3558, opened 2022-10-18
Handle upload exceptions allowing --retries to work properly

mhuth:

### All Pull Requests:

Check all that apply:

This is to fix https://bugzilla.redhat.com/show_bug.cgi?id=2117686

The insights-core connection.py._http_request() code doesn't handle network exceptions. Callers of this function just get status_codes containing errors rather than any network exception handling. This PR catches network exceptions in the _http_request() method and raises them up the call stack until they are caught in the client.py.upload() and _legacy_upload() functions.  This allows the --retries option to work correctly because before, --retries didn't work if there was a network exception.

### DIFF
--- a/insights/tests/client/test_client.py
+++ b/insights/tests/client/test_client.py
@@ -9,7 +9,7 @@ from insights.client.archive import InsightsArchive
 from insights.client.config import InsightsConfig
 from insights import package_info
 from insights.client.constants import InsightsConstants as constants
-from mock.mock import patch, Mock, call
+from mock.mock import patch, Mock, call, ANY
 from pytest import mark
 from pytest import raises
 
@@ -313,7 +313,8 @@ def test_reg_check_unregistered_unreachable():
 @patch('insights.client.client.InsightsConnection.upload_archive',
        return_value=Mock(status_code=500))
 @patch('insights.client.os.path.exists', return_value=True)
-def test_upload_500_retry(_, upload_archive):
+@patch("insights.client.client.logger")
+def test_upload_500_retry(logger, _, upload_archive):
 
     # Hack to prevent client from parsing args to py.test
     tmp = sys.argv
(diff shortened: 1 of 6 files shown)


mhuth:
Logs with --retries 3 and network exceptions (disabled the NIC in the VM):
Screenshot from 2022-10-18 18-39-03

ahitacat (commented):
This LGTM good job I have some questions about the test that I want to @Glutexo to take a look at them.

insights/tests/client/test_client.py
```diff
@@ -329,10 +330,50 @@ def test_upload_500_retry(_, upload_archive):
...
+        logger.error.assert_called_with("All attempts to upload have failed!")
     finally:
         sys.argv = tmp
 
 
+@patch('insights.client.client.constants.sleep_time', 0)
+@patch('insights.client.client.InsightsConnection.upload_archive')
+@patch('insights.client.os.path.exists', return_value=True)
+@patch("insights.client.client.logger")
+def test_upload_exception_retry(logger, path_exists, upload_archive):
```
ahitacat, line 343: Sometimes a preceding underscore is used to mark an argument with no usage in the function.
```def test_upload_exception_retry(logger, _path_exists, upload_archive):```

File: insights/tests/client/test_client.py [hunk 2]
```diff
@@ -329,10 +330,50 @@ def test_upload_500_retry(_, upload_archive):
...
         assert upload_archive.call_count == retries
+        logger.error.assert_any_call("Upload attempt %d of %d failed! Reason: %s", 1, config.retries, ANY)
+        logger.error.assert_called_with("All attempts to upload have failed!")
     finally:
         sys.argv = tmp
 
 
+@patch('insights.client.client.constants.sleep_time', 0)
+@patch('insights.client.client.InsightsConnection.upload_archive')
+@patch('insights.client.os.path.exists', return_value=True)
```
ahitacat, line 341: I'm a bit afraid to mock the path.exists method, as it is use in a lot of places of this code. I have experience some unwanted failures because of that. If this is used to check if a file exists I would suggest to use @contexmanager to create a temporary file to check against.

insights/tests/client/test_client.py
mhuth: I removed the mock for os.path.exists, so the path_exists argument wasn't needed anymore.

insights/tests/client/test_client.py
mhuth: Good idea.  I removed the os.path.exists mock and used a contextmanager for managing the temporary archive file.  In my defence though, the surrounding methods mock the os.path.exists method :)

mhuth: Actually, the tests didn't like the contextmanager for the temporary archive because the file didn't exist.  Instead I've moved the os.path.exists mock to only surround the upload() call, thus limiting its scope, instead of using it for the whole test.  Anyway, as I mentioned before, I was just copying what the other tests had done.

ahitacat (approved):
:seal:

mhuth:
> seal

Thanks for the review/approval @ahitacat .  Unfortunately I don't have permissions to merge this PR.  Are you able to do that?  Thanks.

insights/tests/client/test_client.py
ahitacat: oh, you can created as this: https://github.com/RedHatInsights/insights-core/blob/master/insights/tests/client/test_client.py#L70, I'm using it to create the registration test up in this test

ahitacat:
> > seal
> 
> Thanks for the review/approval @ahitacat . Unfortunately I don't have permissions to merge this PR. Are you able to do that? Thanks.

No, I don't have that power

PR #3562, opened 2022-10-24
feat: Add parser container_inspect

wushiqinlou:

### All Pull Requests:

Check all that apply:

This PR is used to collect the output of "docker|podman inspect <container ID>" for new advisor rule.


wushiqinlou:
@xiangce - Please review.

insights/specs/datasources/containers_inspect.py
```diff
@@ -0,0 +1,94 @@
...
+    Returns:
+        list: item is JSON string containing filtered information.
+
+    Raises:
+        SkipComponent: When the filter/path does not exist or any exception occurs.
+    """
+
+    def _find(d, tag, path):
+        if tag in d:
+            yield d[tag]
```
xiangce, line 61: I don't think it's correct to make this `_find` a generator, since it may miss the keys which is at the end of the `filters` but locate in the top of the `json`.

insights/specs/datasources/containers_inspect.py
```diff
@@ -0,0 +1,94 @@
...
+                    yield i
+    try:
+        filters = get_filters(Specs.containers_inspect_vars)
+        contents = []
+        for item in broker[LocalSpecs.containers_inspect_data_raw]:
+            engine = item.cmd.split()[0].split("bin/")[-1]
+            contents.append((item.content, engine))
+        if contents and filters:
+            total_results = []
+            for content in contents:
```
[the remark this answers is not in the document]
xiangce: A full version which fixed the `_find` generator:
```python
    def _get_kvs(key, _dict):
        ret = {}
        for k, v in _dict.items():
            if key == k:
                ret.update({k: v})
            elif isinstance(v, dict):
                ret.update(_get_kvs(key, v))
        return ret

    try:
        filters = get_filters(Specs.containers_inspect_vars)
        total_results = []
        if filters:
            for item in broker[LocalSpecs.containers_inspect_data_raw]:
                raw_data = json.loads(''.join(item.content))[0]
                engine, _, container_id = item.cmd.split(None)
                filter_result=dict(Id=container_id, engine=os.path.basename(engine))
                for filt in filters:
                    filter_result.update(_get_kvs(filt, raw_data))
                total_results.append(filter_result) if filter_result else None
```

insights/parsers/containers_inspect.py
```diff
@@ -0,0 +1,45 @@
...
+                "engine": "podman",
+                "HostConfig": {"Privileged": false}}
+            }
+        ]
+
+    Attributes:
+        data (list): A list containing the parsed information
+
+    Examples:
+        >>> str(inspect_containers.data[0]["Id"])
```
xiangce, line 36: Since the usage of the filter to this parser is different from normal specs, please add an example here.

xiangce (changes requested):
Besides the following comments, the "podman_inspect.py" should also be deprecated.

insights/specs/datasources/containers_inspect.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Raises:
+        SkipComponent: When the filter/path does not exist or any exception occurs.
+    """
+
+    def _find(d, tag, path):
+        if tag in d:
+            yield d[tag]
+        for k, v in d.items():
+            if isinstance(v, dict):
```
xiangce, line 63: The `v` could be a `list`.

wushiqinlou:
@xiangce - I updated this PR according to your comments, please check.

wushiqinlou:
New collected archive:

~~~
# cat insights_containers/insights_containers

[{"engine": "podman", "Config|Cmd": ["sleep", "100000"], "Image": "acf3e09a39c95d354539b6591298be0b0814f5d74e95e722863241192b9a079b", "Id": "81b8de0d16ce", "HostConfig|Privileged": false}, {"engine": "podman", "Config|Cmd": ["sleep", "100000"], "Image": "acf3e09a39c95d354539b6591298be0b0814f5d74e95e722863241192b9a079b", "Id": "327202b625d0", "HostConfig|Privileged": false}, {"engine": "docker", "Config|Cmd": ["sleep", "100000"], "Image": "acf3e09a39c95d354539b6591298be0b0814f5d74e95e722863241192b9a079b", "Id": "706c223d3f58", "HostConfig|Privileged": false}, {"engine": "docker", "Config|Cmd": ["sleep", "100000"], "Image": "acf3e09a39c95d354539b6591298be0b0814f5d74e95e722863241192b9a079b", "Id": "a7c6b05025f4", "HostConfig|Privileged": false}, {"engine": "docker", "Config|Cmd": ["sleep", "1000000"], "Image": "acf3e09a39c95d354539b6591298be0b0814f5d74e95e722863241192b9a079b", "Id": "c7efee959ea8", "HostConfig|Privileged": true}]
~~~

xiangce (approved):
Approved pending the approval of new specs

insights/specs/datasources/containers_inspect.py
```diff
@@ -0,0 +1,107 @@
...
+                for i in _find(v, tag, path):
+                    path.append(k)
+                    yield i
+    try:
+        filters = get_filters(Specs.containers_inspect_vars)
+        contents = []
+        for item in broker[LocalSpecs.containers_inspect_data_raw]:
+            engine = item.cmd.split()[0].split("bin/")[-1]
```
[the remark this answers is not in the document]
psachin: How about using `pathlib` instead which is better than `os.path..`  [PEP 428](https://peps.python.org/pep-0428/)?

insights/specs/datasources/container/containers_inspect.py
```diff
@@ -0,0 +1,91 @@
...
+                        if key in val:
+                            val = val[key]
+                        else:
+                            break
+                    # If the filtered key does not exist, skip it
+                    if val == raw_data:
+                        continue
+                    filter_result[item] = val
+                total_results.append(filter_result)
+            if total_results:
```
psachin, line 87: Can you add tests for these lines?

wushiqinlou: @psachin - Generally this condition is always True, I added this condition in case there is unexpected situation that `total_results` is `False`, I will add a comment to explain it.

xiangce: Hm... the `False` branch of this line will never reach, since the `required` component `LocalSpecs.containers_inspect_data_raw` guaranteed the `loop` of L69 must be trapped at least once. Then the `total_results` must include at least one `filter_results`.

However, I think it's worth leaving this check here, as an `explicit` after assertion.

insights/specs/datasources/containers_inspect.py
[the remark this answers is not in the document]
xiangce: I didn't see 'pathlib' is broad used in the core, for now, I believe we'd better keep `os.path`  as it works as expected.

PR #3572, opened 2022-10-28
fix: missing call to a RPM format generation function

danzatt:
The refactoring made in commit 4438d91b3979557bc2019838c716c3b1378ec36a missed a call to a function. Instead of passing results of the function generated by _make_rpm_formatter, it passed the generated function directly to the string substitution.

### All Pull Requests:

Check all that apply:

This resulted in calling the rpm command with wrong parameters and rpm not finding any packages:
```
$ insights-inspect insights.parsers.cryptsetup_luksDump.LuksDump -D
...
DEBUG:insights.util.subproc:Executing: [['timeout', '-s', '15', '30', '/bin/rpm', '-qa', '--qf', '<function _make_rpm_formatter.<locals>.inner at 0x7f71f3af29e0>']]
...
```

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -61,6 +61,9 @@ def _make_rpm_formatter(fmt=None):
     return inner
 
 
+format_rpm = _make_rpm_formatter()
+
+
 class DefaultSpecs(Specs):
     # Dep specs that aren't in the registry
     block_devices_by_uuid = listdir("/dev/disk/by-uuid/", context=HostContext)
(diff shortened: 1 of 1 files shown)


insights/specs/default.py
```diff
@@ -61,6 +61,9 @@ def inner(idx=None):
     return inner
 
 
+format_rpm = _make_rpm_formatter()
```
ryan-blakley, line 64: @danzatt good catch on this, I'm not sure how this slipped passed my diff output in the previous commit. But the issue is actually with the inner function in _make_rpm_formatter. If you could instead remove the inner function since it's not actually used at all for it's purpose that I can tell, and instead replace it with `return "\{" + ",".join(fmt) + "\}\n"`.

xiangce (approved):
@danzatt - Thanks for catching this issue and fixing it so quickly. I will submit an off-cycle release for this.

insights/specs/default.py
xiangce: Sorry I missed your comment @ryan-blakley and just released @danzatt's version in a rush. Now I raised anther PR according to Ryan's comment.

PR #3573, opened 2022-10-28
feat: Add timeout to datasources

ryan-blakley:
Check all that apply:

* Added timeout using signals in the datasource invoke function, when the timeout is hit it raises a TimeoutException.

### DIFF
--- a/insights/core/dr.py
+++ b/insights/core/dr.py
@@ -424,6 +424,21 @@ def get_dependency_specs(component):
     return get_requires(component) + get_at_least_one(component)
 
 
+def get_spec_name(c):
+    """
+    Query the actual registry point spec name, by looping through the
+    dependent components and if ".Specs" is in the name and return it.
+    """
+    for cmp in get_dependents(c):
+        name = get_name(cmp)
+        if ".Specs" in name:
+            return name
+        ret = get_spec_name(cmp)
+        if ret:
+            return ret
+    return get_name(c)
+
+
 def get_subgraphs(graph=None):
     """
     Given a graph of possibly disconnected components, generate all graphs of
--- a/insights/core/plugins.py
+++ b/insights/core/plugins.py
@@ -27,6 +27,7 @@ may return:
 from __future__ import print_function
 
 import logging
+import signal
 import traceback
 
 from pprint import pformat
@@ -41,6 +42,11 @@ log = logging.getLogger(__name__)
 
 class ContentException(dr.SkipComponent):
     """ Raised whenever a :class:`datasource` fails to get data. """
+    pass
+
+
+class TimeoutException(Exception):
+    """ Raised whenever a :class:`datasource` hits the set timeout value. """
     pass
 
 
@@ -81,21 +87,36 @@ class datasource(PluginType):
     Decorates a component that one or more :class:`insights.core.Parser`
     subclasses will consume.
     """
+    filterable = False
     multi_output = False
     raw = False
-    filterable = False
+
+    def _handle_timeout(self, signum, frame):
+        raise TimeoutException("Datasource spec {ds_name} timed out after {secs} seconds!".format(
+            ds_name=dr.get_name(self.component), secs=self.timeout))
 
     def invoke(self, broker):
+        # Grab the timeout from the decorator, or use the default of 120.
+        self.timeout = getattr(self, "timeout", 120)
+
+        signal.signal(signal.SIGALRM, self._handle_timeout)
+        signal.alarm(self.timeout)
         try:
             return self.component(broker)
         except ContentException as ce:
             log.debug(ce)
-            broker.add_exception(self.component, ce, traceback.format_exc())
+            broker.add_exception(dr.get_spec_name(self.component), ce, traceback.format_exc())
             raise dr.SkipComponent()
         except CalledProcessError as cpe:
             log.debug(cpe)
-            broker.add_exception(self.component, cpe, traceback.format_exc())
+            broker.add_exception(dr.get_spec_name(self.component), cpe, traceback.format_exc())
             raise dr.SkipComponent()
+        except TimeoutException as te:
+            log.debug(te)
+            broker.add_exception(dr.get_spec_name(self.component), te, traceback.format_exc())
+            raise dr.SkipComponent()
+        finally:
+            signal.alarm(0)
 
 
 class parser(PluginType):
--- a/insights/core/serde.py
+++ b/insights/core/serde.py
@@ -115,13 +115,21 @@ def marshal(comp, broker, root=None, pool=None):
 
     if isinstance(v, list):
         repeat = len(v)
+
+        # Some datasources return a list of strings, not a list of objects.
+        # So check if the first list entry doesn't have a serializer.
+        if repeat and not get_serializer(v[0]):
+            return None, None
+
         if pool:
             data = list(pool.map(call_serializer, [ser_func] * repeat, v, [exc_func] * repeat))
         else:
             data = list(map(call_serializer, [ser_func] * repeat, v, [exc_func] * repeat))
+
         results = [i[0] for i in data if i[0]]
         errors = [i[1] for i in data if i[1]]
         return results, errors
+
     return call_serializer(ser_func, v, exc_func)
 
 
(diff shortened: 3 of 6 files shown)


insights/core/dr.py
```diff
@@ -705,6 +710,8 @@ def __init__(self, *deps, **kwargs):
         tags.extend(kwargs.get("tags", []) or [])
         self.tags = set(tags)
 
+        self.timeout = kwargs.get("timeout", 120)
```
xiangce, line 713: Do you think it's necessary to define a global variable for the default timeout for datasources?

xiangce (commented):
@ryan-blakley - do you think it's necessary to add the `TimeoutException` errors into the `./meta_data/` directory of the insights-archive? It would be helpful for us to pay attention to which datasources caused the `TimeoutException` errors.

ryan-blakley:
@xiangce I agree they should probably be written to the meta_data directory, I'll look into why that's not happening.

insights/core/dr.py
ryan-blakley: I didn't think so I based it off of how simple_command has a default timeout set, and has the parameter option to set it manually. I didn't want to have to pass in the global variable every time a datasource is defined, because we could easily forget to add the timeout parameter when adding new ones. But if you think that logic isn't sound, I can definitely add a global variable like how your PR was setup.

xiangce: No worries, I think your implement that applying the `120` timeout to each datasource by default and adjusting it only when necessary is better.

xiangce:
> @xiangce I agree they should probably be written to the meta_data directory, I'll look into why that's not happening.

I think we should add the `insights.specs.datasources` to the `persist` section of the `default_mainfest` and fix this issue #3578 as well

ryan-blakley:
> I think we should add the `insights.specs.datasources` to the `persist` section of the `default_mainfest` and fix this issue #3578 as well

@xiangce so do you want me to add in the fix from your PR #3579, or keep them separate? I tested your fix and it works with the fix in this PR.

xiangce:
> > I think we should add the `insights.specs.datasources` to the `persist` section of the `default_mainfest` and fix this issue #3578 as well
> 
> @xiangce so do you want me to add in the fix from your PR #3579, or keep them separate? I tested your fix and it works with the fix in this PR.
@ryan-blakley - I think they are two different points, so I merged #3579 and closed #3578. Please rebase with it.

insights/core/plugins.py
```diff
@@ -96,6 +109,18 @@ def invoke(self, broker):
...
             raise dr.SkipComponent()
+        except TimeoutException as te:
+            log.debug(te)
+            # For some reason when this exception fires, the component is
+            # insights.specs.default.DefaultSpecs.<function> instead of the
+            # actual registry point spec name. So modify the key of the
+            # exception entry to fix it, so it can be accessed properly
+            # during dehydration.
+            broker.add_exception("insights.specs.Specs.{0}".format(
+                dr.get_name(self.component).split('.')[-1]), te, traceback.format_exc())
```
xiangce, line 120: Here, for datasources, the `self.component` would be `insights.specs.datasources.<module>.<function>`, based on the current code, the key inserted to the `broker.exceptions` is `insights.specs.Specs.<function>`.

But this `insights.specs.Specs.<function>` will never be a valid name when doing the dehydration.  https://github.com/RedHatInsights/insights-core/blob/e713e3f5814da3580622ad919d836bc5dd7b7056/insights/core/serde.py#L202

The `dehydrate()` will only dehydrate the specs defined under the `insights.specs.Specs` module according to the "persist" items declared in the `default_mainfest` (currently, `insights.specs.Specs.<specs>` only)

We may need to get the original `spec` (defined in the `insights.specs.Specs.<spec>`) of this `self.component` and depends on the `self.component` via itself. e.g.:
```python
            def _get_spec(comp):
                comps = dr.get_dependents(comp)
                for cmp in comps:
                    sp_name = dr.get_name(cmp).split('.')
                    if len(sp_name) == 4 and sp_name[2] == "Specs":
                        return cmp
                    return _get_spec(cmp)

            broker.add_exception(dr.get_name(_get_spec(self.component)), te, traceback.format_exc())
```

xiangce: Ideally, to make it strict, I think we may need to ensure it is consistent with the "persist", e.g. only insert the `traceback` when it's declared in the `persist`  (`insights.specs.Specs` by default).

insights/core/serde.py
```diff
@@ -196,27 +196,34 @@ def dehydrate(self, comp, broker):
...
-                "name": name,
-                "exec_time": broker.exec_times.get(c),
-                "errors": errors
-            }
+            errors = [broker.tracebacks[e] for e in broker.exceptions.get(comp, [])]
+
+            # For datasource timeouts the key is the name not the component, so
+            # check for the name in the exception dict.
+            if broker.exceptions.get(name):
+                errors.extend([broker.tracebacks[e] for e in broker.exceptions.get(name, [])])
```
xiangce, line 209: Maybe nitpicking, but how about?
```
exs = broker.exceptions
errors = [broker.tracebacks[e] for e in exs.get(comp, exs.get(name, []))]
```

insights/core/plugins.py
ryan-blakley: So the issue I was running into was the exceptions weren't being dehydrated in serde when datasources are called directly. This is because the key it was using to check was insights.specs.Specs.system_user_dirs for example, but the exception entry key was the function itself. So at first I changed the entry key when adding the exception to the name of the component, but that came out to be insights.specs.default.DefaultSpecs.system_user_dirs. I didn't know of a way to query the actual spec name. But thank you for the get_dependents function, I altered the _get_spec suggested function a little bit and using it works much better than what I had.

xiangce: The current `_get_spec_name` works well when the name of datasource is the same as the `symbolic_name` of the spec, e.g. the `system_user_dirs`. 
https://github.com/RedHatInsights/insights-core/blob/9a6b5c37ed9f69d798946922c8cfbbf50e7e0a0d/insights/specs/__init__.py#L702
https://github.com/RedHatInsights/insights-core/blob/9a6b5c37ed9f69d798946922c8cfbbf50e7e0a0d/insights/specs/default.py#L617

But it won't work when they are different, e.g. the `du_dirs` :

https://github.com/RedHatInsights/insights-core/blob/9a6b5c37ed9f69d798946922c8cfbbf50e7e0a0d/insights/specs/__init__.py#L150
https://github.com/RedHatInsights/insights-core/blob/9a6b5c37ed9f69d798946922c8cfbbf50e7e0a0d/insights/specs/default.py#L175

In this case of `du_dirs`, the `_get_spec_name` returns `insights.specs.default.DefaultSpecs.du_dirs`, but `insights.specs.Specs.du_dirs` is expected.

And it seems the `du_dirs` is not the only case.

ryan-blakley: Good catch, I fixed it by returning the name of the component if ".Specs" isn't found. For some reason datasources called from a spec factory function like foreach_execute never had the weird key issue.

insights/core/serde.py
ryan-blakley: Good nitpick, I was trying to figure out a way to better optimize it, and I hadn't thought about putting the second get call in the default of the first. Thanks, I updated it.

insights/core/plugins.py
```diff
@@ -96,6 +124,12 @@ def invoke(self, broker):
             log.debug(cpe)
             broker.add_exception(self.component, cpe, traceback.format_exc())
```
xiangce, line 125: Better apply the `inner` and `recursive`ly `_get_spec_name` to each of the 3 `*Exception`s here in the invoke to avoid generating `meta_data` for the local specs that do not need to be dehydrated.

insights/core/plugins.py
```diff
@@ -81,11 +87,33 @@ class datasource(PluginType):
     Decorates a component that one or more :class:`insights.core.Parser`
     subclasses will consume.
     """
+    filterable = False
     multi_output = False
     raw = False
-    filterable = False
+
+    def _get_spec_name(self):
```
xiangce, line 94: I think it would be better to modify this function as an `inner` function of the `invoke` and change it to a `recursive` function to get its final `insights.specs.Specs` spec.

xiangce (commented):
The current implementation still cannot entirely prevent the client from generating `meta_data` for the local specs of some datasource specs.

E.g.:

```
ls insights-localhost-20221114032906/meta_data/insights.specs.datasources.*
insights-localhost-20221114032906/meta_data/insights.specs.datasources.awx_manage.LocalSpecs.awx_manage_check_license_data_raw.json
insights-localhost-20221114032906/meta_data/insights.specs.datasources.candlepin_broker.LocalSpecs.candlepin_broker_input.json
insights-localhost-20221114032906/meta_data/insights.specs.datasources.cloud_init.LocalSpecs.cloud_cfg_input.json
insights-localhost-20221114032906/meta_data/insights.specs.datasources.dir_list.du_dir_list.json
insights-localhost-20221114032906/meta_data/insights.specs.datasources.lpstat.LocalSpecs.lpstat_v.json
```

I think the following changes could be made to achieve that.

insights/core/plugins.py
ryan-blakley: Updated adding to each exception, also removed the persist addition.

insights/core/plugins.py
ryan-blakley: Ahhh I did forget the recursive part sorry about that. Updated, and moved it to the dr module as it is needed to be called in the serde module during hydration.

xiangce (approved):
Great! It looks good to me. Thanks @ryan-blakley

PR #3574, opened 2022-10-31
fix: remove the inner functions of the _make_rpm_formatter

xiangce:

### All Pull Requests:

Check all that apply:

Remove the inner function as commented [here](https://github.com/RedHatInsights/insights-core/pull/3572/files#r1008107573) by Ryan.

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -52,19 +52,13 @@ def _make_rpm_formatter(fmt=None):
             '"buildhost":"%{BUILDHOST}"',
             '"sigpgp":"%{SIGPGP:pgpsig}"'
         ]
-
-    def inner(idx=None):
-        if idx:
-            return "\{" + ",".join(fmt[:idx]) + "\}\n"
-        else:
-            return "\{" + ",".join(fmt) + "\}\n"
-    return inner
-
-
-format_rpm = _make_rpm_formatter()
+    return "\{" + ",".join(fmt) + "\}\n"
 
 
 class DefaultSpecs(Specs):
+    # Required functions
+    rpm_format = _make_rpm_formatter()
+
     # Dep specs that aren't in the registry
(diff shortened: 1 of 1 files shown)


psachin:
Good to me now.

psachin (approved):

insights/specs/default.py
```diff
@@ -52,19 +52,13 @@ def _make_rpm_formatter(fmt=None):
...
-    return inner
-
-
-format_rpm = _make_rpm_formatter()
+    return "\{" + ",".join(fmt) + "\}\n"
 
 
 class DefaultSpecs(Specs):
+    # Variables that need to be pre-loaded
+    rpm_format = _make_rpm_formatter()
```
ryan-blakley, line 60: @xiangce could you move this under the below deps specs section? I meant for that section to be for these types of specs that are essentially just variables used by real specs.

insights/specs/default.py
```diff
@@ -73,6 +64,7 @@ class DefaultSpecs(Specs):
     httpd_pid = simple_command("/usr/bin/pgrep -o httpd")
     openshift_router_pid = simple_command("/usr/bin/pgrep -n openshift-route")
     ovs_vsctl_list_br = simple_command("/usr/bin/ovs-vsctl list-br")
+    rpm_format = _make_rpm_formatter()
```
bfahr, line 67: @xiangce since this is not actually a spec would it make more sense for `rpm_format`  to be moved outside of the class and renamed `_rpm_format` instead?

xiangce: @bfahr - Moving it outside of the DefaultSpecs makes sense too. Do you also think that we should move the `etc_and_sub_dirs` outside as e.g. `_etc_and_sub_dirs`?

xiangce: Oops, my bad, forgot to check the flake8 result in my local. Now it's fixed. Please check it again when free.

PR #3586, opened 2022-11-07
Update system_user_dirs datasource

jobselko:
### All Pull Requests:

Check all that apply:

This PR should fix the complexity issue and also update the code logic, which was not entirely accurate.

### DIFF
--- a/insights/specs/datasources/system_user_dirs.py
+++ b/insights/specs/datasources/system_user_dirs.py
@@ -18,7 +18,7 @@ class LocalSpecs(Specs):
     Local spec used only by the system_user_dirs datasource.
     """
     rpm_args = simple_command(
-        'rpm -qa --qf="[%{=NAME}; %{FILEMODES:perms}; %{FILEUSERNAME}; %{FILEGROUPNAME}\n]"',
+        'rpm -qa --nosignature --qf="[%{=NAME}; %{FILENAMES}; %{FILEMODES:perms}; %{FILEUSERNAME}; %{FILEGROUPNAME}\n]"',
         signum=signal.SIGTERM
     )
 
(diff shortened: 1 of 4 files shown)


insights/specs/datasources/system_user_dirs.py
```diff
@@ -72,33 +74,48 @@ def system_user_dirs(broker):
...
+                dirs[path_name] = pkg_name
+
+    # Stores a package if its associated file is in a writable directory
+    for line in content:
+        _, path_name, perms, _, _ = line.split("; ")
+        if perms[0] != "d":
+            dir_name = path_name.rsplit('/', 1)[0]
+
+            if dir_name in dirs:
+                packages.add(dirs[dir_name])
```
xiangce, line 118: From Ryan's script's result, the `loop` is the most time-consuming part on that host, I think it's not so good to iterate the content twice, how about changing to an equivalent set-comprehension to a built-in query against the dict?
```python
    dir_package = {}
    dirs = set()
    packages  = set()
    # Stores a writeable directory with its package
    for line in content:
        pkg_name, path_name, perms, user, group = line.split("; ")
        if perms[0] == "d":
            user_w = user in users and perms[2] == "w"
            group_w = group in groups and perms[5] == "w"
            others_w = perms[8] == "w"
            dirs.add(path_name.rsplit('/', 1)[0])
            if user_w or group_w or others_w:
                dir_package[path_name] = pkg_name

    # Stores a package if its associated file is in a writable directory
    packages.update(dir_package[dir] for dir in dirs if dir in dir_package)
```

ffesti: This code is not correct although the idea has some merit. The second loop looks at all files not only dirs. So dirs.add line needs to be outside the if perms[0] == "d": block.

xiangce: Yeap, If you are right, then [the original code](https://github.com/RedHatInsights/insights-core/pull/3586/files#diff-62ba94046189d470d928eedfbab6784c6b0cc238e80c112289fb484791bbbe09R111-R115) must be wrong. My update strictly followed the original code and they must be equivalent.  @jobselko, If you think it's okay to update per my comment, please pay attention to the comment from @ffesti

xiangce: Oops, my bad, I didn't note that the two `if` statement is opposite.

The correct code should be:

```
    dir_package = {}
    dirs = set()
    packages  = set()
    # Stores a writeable directory with its package
    for line in content:
        pkg_name, path_name, perms, user, group = line.split("; ")
        if perms[0] == "d":
            user_w = user in users and perms[2] == "w"
            group_w = group in groups and perms[5] == "w"
            others_w = perms[8] == "w"
            if user_w or group_w or others_w:
                dir_package[path_name] = pkg_name
        else:
            dirs.add(path_name.rsplit('/', 1)[0])

    # Stores a package if its associated file is in a writable directory
    packages.update(dir_package[dir] for dir in dirs if dir in dir_package)
```

Sorry about that.

xiangce (approved):
Thanks @jobselko

PR #3587, opened 2022-11-08
feat: New spec "/etc/fapolicyd/rules.d/*.rules" and parser

huali027:
feat: New spec "/etc/fapolicyd/rules.d/*.rules" and parser

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/fapolicyd_rules.py
+++ b/insights/parsers/fapolicyd_rules.py
@@ -13,12 +13,20 @@ class FapolicydRules(LogFileOutput):
     """
     Parse the content of ``/etc/fapolicyd/rules.d/*.rules`` file.
 
+    .. note::
+
+        The rules do not require to get the parsed result currently.
+        It just need to check if it contains specific lines, so use
+        :class:`insights.core.LogFileOutput` as the base class.
+
     Sample input::
 
         deny_audit perm=any pattern=ld_so : all
         deny_audit perm=any pattern=ld_preload : all
 
     Examples:
+        >>> from insights.parsers.fapolicyd_rules import FapolicydRules
+        >>> FapolicydRules.last_scan('ld_so_deny_audit_test', 'deny_audit perm=any pattern=ld_so : all')
         >>> type(fapolicyd_rules)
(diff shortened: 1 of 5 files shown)


huali027:

insights/parsers/fapolicyd_rules.py
```diff
@@ -0,0 +1,27 @@
...
+========================================================
+"""
+
+from insights import parser
+from insights.core import LogFileOutput
+from insights.specs import Specs
+
+
+@parser(Specs.fapolicyd_rules)
+class FapolicydRules(LogFileOutput):
```
xiangce, line 12: Generally this PR looks good to me, but since the base class `LogFileOutput` is designed for log-like files, could you please add corresponding explanation in the doc for this parser about why using the `LogFileOutput` as base?

insights/tests/parsers/test_fapolicyd_rules.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def test_udev_rules():
+    FapolicydRules.last_scan('test_deny_audit_lo_so1', 'deny_audit perm=any pattern=ld_so : all')
+    result = FapolicydRules(context_wrap(CONTENT, path='/etc/fapolicyd/rules.d/30-patterns.rules'))
+    assert result.test_deny_audit_lo_so1.get('raw_message') == 'deny_audit perm=any pattern=ld_so : all'
+
+
+def test_doc():
+    FapolicydRules.last_scan('ld_so_deny_audit_test', 'deny_audit perm=any pattern=ld_so : all')
```
xiangce, line 23: It would be better move this line to the `examples`, to make the example a bit clear.

insights/parsers/fapolicyd_rules.py
psachin: @huali027 I can understand the reason why `LogFileOutput` is used to make the scanning easier However it will make more sense to use the generic `Parser` with `parse_content` to collect lines. But for now this is ok.

psachin (approved):

PR #3592, opened 2022-11-15
feat: New spec "timedatectl" and the parser

huali027:
feat: New spec "timedatectl" and the parser

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/date.py
+++ b/insights/parsers/date.py
@@ -8,11 +8,11 @@ Date - command ``date``
 -----------------------
 DateUTC - command ``date --utc``
 --------------------------------
-TimeDateCtlStatusParser - command ``timedatectl status``
---------------------------------------------------------
+TimeDateCtlStatus - command ``timedatectl status``
+--------------------------------------------------
 """
 
-from insights.parsers import ParseException
+from insights.parsers import ParseException, SkipException
 import six
 import sys
 from datetime import datetime
@@ -110,7 +110,7 @@ class DateUTC(DateParser):
 
 
 @parser(Specs.timedatectl_status)
-class TimeDateCtlStatusParser(CommandParser, dict):
+class TimeDateCtlStatus(CommandParser, dict):
     """
     Class to parse the ``timedatectl status`` command output.
     It saves the infomartion in each line into a dict.
@@ -165,14 +165,22 @@ class TimeDateCtlStatusParser(CommandParser, dict):
     }
 
     def parse_content(self, content):
-        colon_index = -1
         dict_key = None
         warning_start = True
+        non_blank_line = None
+        for line in content:
+            if line.strip():
+                non_blank_line = line
+                break
+        if non_blank_line is None:
+            raise SkipException('No data in the output.')
+        try:
+            colon_index = non_blank_line.index(':')
+        except ValueError:
+            raise ParseException('No colon found, the line %s is not in expected format.' % line)
         for line in content:
             if not line.strip():
                 continue
-            if colon_index == -1:
-                colon_index = line.index(':')
             if line[colon_index] == ':':
                 key = line[:colon_index].strip()
                 value = line[colon_index + 1:].strip()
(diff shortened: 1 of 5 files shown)


huali027:
@xiangce @psachin Please help to review, thanks.

insights/specs/default.py
```diff
@@ -638,6 +638,7 @@ class DefaultSpecs(Specs):
     testparm_v_s = simple_command("/usr/bin/testparm -v -s")
     thp_enabled = simple_file("/sys/kernel/mm/transparent_hugepage/enabled")
     thp_use_zero_page = simple_file("/sys/kernel/mm/transparent_hugepage/use_zero_page")
+    timedatectl = simple_command('/usr/bin/timedatectl')
```
bfahr, line 641: @huali027 could you use the command `timedatectl show --all` instead to produce machine readable form and simplify the parser?

bfahr: @huali027 this spec requires an entry in `insights/specs/insights_archive.py`.

bfahr (changes requested):

insights/specs/default.py
huali027: > @huali027 could you use the command `timedatectl show --all` instead to produce machine readable form and simplify the parser?
@bfahr `timedatectl show` is only supported on RHEL8 and later.

insights/parsers/date.py
```diff
@@ -107,6 +89,116 @@ class DateUTC(DateParser):
...
+        >>> ctl_info['system_clock_synchronized']
+        'yes'
+        >>> ctl_info['local_time']
+        datetime.datetime(2022, 11, 14, 23, 4, 6)
+    """
+    date_format = '%a %Y-%m-%d %H:%M:%S'
+
+    # unify the different names in rhel7 and rhel8
+    key_mapping = {
+        'ntp_synchronized': 'system_clock_synchronized'
```
huali027, line 164: @xiangce As we discussed, I found only one name can be unified. For "NTP enabled" on RHEL7, it means the enabled status, but on RHEL8, it is "NTP service", which means the running status.

insights/parsers/date.py
```diff
@@ -107,6 +89,116 @@ class DateUTC(DateParser):
...
+    }
+
+    def parse_content(self, content):
```
psachin, line 167: @huali027 This is complex to understand. How about using [parsr](https://insights-core.readthedocs.io/en/latest/api_index.html?highlight=parsr#module-insights.parsr) here?

huali027: > @huali027 This is complex to understand. How about using [parsr](https://insights-core.readthedocs.io/en/latest/api_index.html?highlight=parsr#module-insights.parsr) here?

@psachin The output is colon aligned except for the "Warning" part, and some key has multiple lines. I think `parsr` won't make it easier too.

psachin:
@huali027 Is this ready for review again because there is some discussion related to the key mapping which is not addressed?

xiangce (approved):

psachin:
@huali027 

```bash
---------- coverage: platform linux, python 3.6.15-final-0 -----------
Name                       Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------
insights/parsers/date.py      73     52     38      0    28%   48-58, 167-213
----------------------------------------------------------------------
TOTAL                         73     52     38      0    28%

```

insights/parsers/date.py
```diff
@@ -107,6 +88,124 @@ class DateUTC(DateParser):
...
+                # this line is also the content of the previous key
+                self[dict_key] += ' ' + line.strip()
+            elif line.lstrip().startswith('Warning:'):
+                warning_start = True
+                self['warning'] = line.split(':')[1].strip()
+            else:
+                if warning_start:
+                    self['warning'] += ' ' + line.strip()
+                else:
+                    raise ParseException('Unexpected format of line %s.' % line)
```
huali027, line 211: @xiangce I found we don't need to check if dict is empty here. Since the previsous checking can make it's not empty.
Please help to review.

psachin:
Looks good to me now.

psachin (approved):

PR #3597, opened 2022-11-16
Rename system_user_dirs to rpm_pkgs

jobselko:

### All Pull Requests:

Check all that apply:

Renaming should ensure that we have new data to check.


insights/specs/datasources/system_user_pkgs.py
```diff
@@ -70,7 +70,7 @@ def get_groups(users):
 
 
 @datasource(LocalSpecs.rpm_args, HostContext)
-def system_user_dirs(broker):
+def system_user_pkgs(broker):
```
ryan-blakley, line 73: I was actually going to request if we could rename this datasource(ds), so thank you for initiating this. One thing, could you rename the function to something like pkgs_with_writable_dirs for example to better describe what the function is doing instead of it having the same name as the ds. 

The next suggestion is optional I was just going to throw it out there. Could we maybe change the filename to rpm_pkgs.py or pkgs.py as well? That way if we ever need to add more rpm/pkgs related ds functions in the future we can put them in this file.

jobselko: I agree with both renaming, but I think that adding new stuff into this module would make it unclear (but that is a question for the future).

ryan-blakley: Awesome thank you for the change, everything looks good to me

ryan-blakley (approved):

xiangce (commented):
This change looks good to me, just left some trifle comments, please have a look .

jobselko:
@xiangce Thank you very much for double-checking!

xiangce (approved):

PR #3598, opened 2022-11-16
feat: Add timeout to datasources

ryan-blakley:
### All Pull Requests:

Check all that apply:

Added timeout using signals in the datasource invoke function, when the timeout is hit it raises a TimeoutException. Added code in the dehydrate to properly dehydrate caught exceptions.

### DIFF
--- a/insights/core/dr.py
+++ b/insights/core/dr.py
@@ -424,6 +424,39 @@ def get_dependency_specs(component):
     return get_requires(component) + get_at_least_one(component)
 
 
+def is_registry_point(component):
+    return type(component).__name__ == "RegistryPoint"
+
+
+def get_registry_points(component):
+    """
+    Loop through the dependency graph to identify the corresponding spec registry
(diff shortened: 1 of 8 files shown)


insights/core/dr.py
```diff
@@ -424,6 +424,26 @@ def get_at_least_one(comp):
     return get_requires(component) + get_at_least_one(component)
 
 
+def get_spec_name(c):
```
xiangce, line 427: Just a suggestion, but it would be better to add a unit test and coverage test for this new function

And, it seems the `False branch of L441` (441->436) will never be covered.

ryan-blakley: Good suggestion, that was something I intended to do and forgot. So thank you for reminding me.

xiangce (approved):
Good to me

xiangce:
@bfahr, @lhuett, @vdymna - would you mind having a look again at this PR? Thanks.

bfahr (approved):

PR #3601, opened 2022-11-17
fix: allow legacy upload to stage

subpop:
Allow the value of legacy_upload, as set in the config file, to remain
regardless of whether or not the client is connecting to the stage
environment.

Related: ESSNTL-3782


subpop:
This is not a solution.

PR #3602, opened 2022-11-17
Rename downloaded temp malware rules file

mhuth:

### All Pull Requests:

Check all that apply:

This is a rather weak attempt at hiding the downloaded malware rules file, but its slightly less obvious than what's currently being used.  This is a temporary fix until I can get something better happening, eg having the rules file in memory and written to disk as needed.


xiangce (approved):
looks good to me.

PR #3603, opened 2022-11-18
Do not union the mount points

chenlizhong:
For IBM GPFS, there is internal mount concept. If a file system is only internally mounted, it does not show up in mount or df output.
  Reference: https://www.ibm.com/support/pages/gpfs-internal-and-external-mount-states

### All Pull Requests:

Check all that apply:

*Add your description here*

### DIFF
--- a/insights/combiners/mounts.py
+++ b/insights/combiners/mounts.py
@@ -43,13 +43,13 @@ class Mounts(object):
 
     def __init__(self, binmount, procmounts, mountinfo):
         self._mounts = {}
-        all_mount_points = set()
+        all_mount_points = []
         if binmount:
-            all_mount_points = set(binmount.mounts.keys())
+            all_mount_points = binmount.mounts.keys()
         elif procmounts:
-            all_mount_points = set(procmounts.mounts.keys())
+            all_mount_points = procmounts.mounts.keys()
         elif mountinfo:
-            all_mount_points = set(mountinfo.mounts.keys())
+            all_mount_points = mountinfo.mounts.keys()
 
         for mount_point in all_mount_points:
             this_binmount = binmount.get_dir(mount_point) if binmount else None
--- a/insights/combiners/mounts.py
+++ b/insights/combiners/mounts.py
@@ -48,7 +48,7 @@ class Mounts(object):
             all_mount_points = set(binmount.mounts.keys())
         elif procmounts:
             all_mount_points = set(procmounts.mounts.keys())
-        else:
+        elif mountinfo:
             all_mount_points = set(mountinfo.mounts.keys())
 
         for mount_point in all_mount_points:
--- a/insights/combiners/mounts.py
+++ b/insights/combiners/mounts.py
@@ -13,7 +13,7 @@ Examples:
     >>> type(mounts)
     <class 'insights.combiners.mounts.Mounts'>
     >>> len(mounts)
-    19
+    15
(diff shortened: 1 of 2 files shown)


chenlizhong:
internal mount is no meaning for non-GPFS, so do not need to union it.

insights/combiners/mounts.py
```diff
@@ -43,11 +43,14 @@ class Mounts(object):
 
     def __init__(self, binmount, procmounts, mountinfo):
         self._mounts = {}
+        all_mount_points = set()
```
ryan-blakley, line 46: I'm not sure I understand this change, this is a combiner that's supposed to combine all the mount points discovered in each parser that exists.  This change is no longer combining the mounts, and is instead only setting all_mount_points to which ever parser is found in an order. If you could explain further the reasoning behind this change it might help me understand.

Also you referenced the gpfs filesystem, but that file system doesn't appear to be supported in RHEL. So I'm not sure why it's being brought into core, but if it needs to be and gpfs mounts are having an issue. I would recommend creating a separate combiner/parser specifically for gpfs mounts. So that it doesn't break any rules that use this combiner currently.

chenlizhong: The gpfs has internal and external mounts. Below kind of internal mounts are just in /proc/mounts and /proc/self/mountinfo.  These internal mounts are not in the output of command mount.  We do not need to analyze the internal mounts. So the code changes is to use the output of mount only when it is available.  We do not support gpfs, but the gpfs internal mounts impact the combiner for the original code(Many internal mount points will be unioned)


apps_fs /data/ire/apps gpfs rw,seclabel,noatime 0 0
apps_fs /apps/ampl gpfs rw,seclabel,noatime 0 0
apps_fs /apps/anaconda gpfs rw,seclabel,noatime 0 0
apps_fs /apps/Gauss14 gpfs rw,seclabel,noatime 0 0
apps_fs /apps/gauss16 gpfs rw,seclabel,noatime 0 0
apps_fs /apps/GeoDa gpfs rw,seclabel,noatime 0 0
apps_fs /apps/ibm gpfs rw,seclabel,noatime 0 0
apps_fs /apps/intel gpfs rw,seclabel,noatime 0 0
apps_fs /apps/iveware gpfs rw,seclabel,noatime 0 0
apps_fs /apps/matlab gpfs rw,seclabel,noatime 0 0
apps_fs /apps/R gpfs rw,seclabel,noatime 0 0
apps_fs /apps/SAS gpfs rw,seclabel,noatime 0 0
apps_fs /apps/scripts gpfs rw,seclabel,noatime 0 0
apps_fs /apps/shared gpfs rw,seclabel,noatime 0 0
apps_fs /apps/stata16 gpfs rw,seclabel,noatime 0 0

chenlizhong: One example here, One system with gpfs service running on RHEL linux, there are more than 40000+ items in /proc/mounts, this leads to this combiner costs 25mins to  parse the content.

xiangce (commented):
Based on the logic of this combiner, and the rules that use it, it seems that the existence of this combiner is a bit tasteless. Let's have a further talk about this to check if it's okay to just remove this combiner.

xiangce:
@chenlizhong - do you want to change this PR to remove this combiner or raise another PR to remove it?

chenlizhong:
@xiangce
I will close this one and create one new MR to remote mounts combiner.

PR #3604, opened 2022-11-22
Speed up `keyword_search` by storing pre-processed data

PaulWay:

### All Pull Requests:

Check all that apply:

Call analysis by @dkuc has shown the `key_match()` function is called many times, often with the same values.  This work attempts to speed up the function by:

* transforming the row keys explicitly in a dict, rather than implicitly by reconstructing the entire row data with transformed keys
* caching the transformed row keys
* pre-preparing the keyword arguments into a list of search terms
* evaluating equality of row data quickly (the common case).
* returning no rows early if the sought key does not exist in the row data (after transform)

These speed up the 'keyword_search' from 1.693 seconds of total time to 0.301 seconds in processing a small archive.


psachin:
@PaulWay Is it possible to add tests related to this?

PaulWay:
@psachin - not really sure; the lru_cache decorator does provide a 'cache' method to check how the cache is being used and to reset it if necessary.  But the real test is: does everything work the same way it always has - only faster?  I'm not sure there's any point in testing the 'faster' assertion :smile: but all the existing tests make sure it works the same way.

xiangce:
It seems the CI environments got broken, hence the pipelines failed. However, I do get the following test errors in my local:
```
========================================================================= test session starts =========================================================================
platform linux -- Python 3.9.14, pytest-7.1.3, pluggy-1.0.0
...
test_lspci_vmmkn ___________________________________________________________________________

    def test_lspci_vmmkn():
        lspci_vmmkn = LsPciVmmkn(context_wrap(LSPCI_VMMKN))
        lspci = LsPci(None, lspci_vmmkn)
        assert sorted(lspci.pci_dev_list) == ['00:00.0', '00:02.0', '00:03.0', '00:16.0', '00:19.0', '00:1b.0']
        assert lspci.search(Device='155a', Vendor='8086') == [
            {
                'Slot': '00:19.0', 'Class': '0200', 'Vendor': '8086',
                'Device': '155a', 'SVendor': '17aa', 'SDevice': '2214',                                                                                                                'Rev': '04', 'Driver': 'e1000e', 'Module': ['e1000e'],
            }                                                                                                                                                                  ]
>       assert lspci.search(Dev_Details__contains='I218') == []
E       AssertionError: assert [{'Class': '0...': '09', ...}] == []
E         Left contains one more item: {'Class': '0600', 'Device': '0a04', 'Driver': 'hsw_uncore', 'Rev': '09', ...}
E         Use -v to get more diff
```

@PaulWay - could you please have a look and fix the flake8 errors as well?

PaulWay:
Hi @xiangce - yep, I see similar errors.  I'm working on this now.  Dicts being unhashable and seeming to change their `id()`s is not helping me here :smiling_face_with_tear:

PaulWay:
Are we still having problems with the CI environment?

xiangce:
> Are we still having problems with the CI environment?

Yeap, this may relate to this issue: https://github.com/actions/setup-python/issues/162, @chenlizhong mentioned it in #3606 , And I asked him to raise a new PR to fix it. Please wait for a while

PaulWay:
OK, finally we get the tests mostly running successfully?

insights/parsers/netstat.py
```diff
@@ -478,6 +478,7 @@ def parse_content(self, content):
         self.data = dict((s.name, s._merge_data_index()) for s in sections)
         self.lines = dict((s.name, s.lines) for s in sections)
         self.datalist = dict((s.name, s.datalist) for s in sections)
+        self.dataobjs = dict((s.name, s) for s in sections)
```
xiangce, line 481: Could you please rename this attribute as `self._dataobjs` to indicate it's a private attribute? since it's only used internally? Or, you may add it to the above `Attributes` section.

PaulWay: Sure thing - done.

xiangce (approved):
It looks good to me. But since this is a kind of basic change, I'd like to wait for feedback from @bfahr  or @ryan-blakley.

bfahr:
@xiangce I'm going to run the times and errors report using this change so that we can try to measure any performance improvement for this change.

xiangce:
@PaulWay - @bfahr is now testing this PR with the insights data and comparing it to the original code to check for performance improvements. But recently, the DataHub met some issues due to maybe some outage during the holiday, so the test is not completed yet. Sorry about replying late.

PaulWay:
Just squashed everything down into a single commit for neatness (there were several rewrites of the code that confused the rebasing).

PaulWay:
@RedHatInsights/qe - what's the problem with this test?  I can't get any information on it.

PaulWay:
Hi all - I've redone this PR on top of current master, as rebasing it was causing too much trouble :-)

insights/parsers/__init__.py
```diff
@@ -498,38 +506,90 @@ def keyword_search(rows, **kwargs):
...
+        txkeys = getattr(parent, txform_cache_attr)
+    else:
+        # Store the translation from the search key to the key in the data.
+        # We start with the reverse transform - the key, to the transformed
+        # key.  This avoids doing the string replacement for every key in
+        # every row.  Then we invert that to map from the search term (with
+        # spaces and dashes replaced by underscores) to the key in the row
+        # data.
+        rev_txkeys = dict()
+        for row in rows:
```
xiangce, line 551: Is it necessary to iterator all `rows`? Generally, one line is okay, as each row should have the same `keys`. Imagine the content is huge, this `for` loop would be time-consuming.
Or, if you are worried about different keys in rows, how about adding one local cache to fast it?:

```python
_raw_keys = set()
for row in rows:
    if _raw_keys.issuperset(row.keys()):
        continue
    _raw_keys.update(row.keys())
for key in _raw_keys:
    if key not in rev_txkeys:
        rev_txkeys[key] = key.replace(' ', '_').replace('-', '_')
```

insights/parsers/__init__.py
```diff
@@ -498,38 +506,90 @@ def keyword_search(rows, **kwargs):
...
             if matcher not in matchers:
                 # put key back the way we found it, matcher fn unchanged
-                key = key + '__' + matcher
-            else:
-                matcher_fn = matchers[matcher]
-        return key in my_row and matcher_fn(my_row[key], value)
-
-    data = []
+                data_key = search_keyword
+                matcher = 'equals'
```
xiangce, line 576: Here, your code definitely looks good, this is just kind of nitpick.
How about using the `partition()` to simplify the code as:
```python
for search_keyword, value in kwargs.items():
   data_key, _, matcher = search_keyword.partition('__')
   if matcher not in matchers:
        matcher = 'equals'
   ...
```

insights/parsers/__init__.py
```diff
@@ -498,38 +506,90 @@ def keyword_search(rows, **kwargs):
...
+        # sure that the search will never match...
+        if data_key not in txkeys:
+            return []
+        search_terms.append((
+            txkeys[data_key], matcher, matchers[matcher], value
+        ))
+
+    def key_match(data_key, matcher, matcher_fn, value):
+        if matcher == 'equals':
+            return data_key in row and row[data_key] == value
```
xiangce, line 587: CMIIW, I think the `matcher_fn` of `equals` (`lambda s, v: s == v`), would also work in the way of L588. Why not just remove this check and return L588 directly?
```python
def key_match(data_key, matcher, matcher_fn, value):
    return data_key in row and matcher_fn(row[data_key], value)
```

xiangce (changes requested):
Looks like the global cache `keyword_search_transformed_row` is not used and can be removed. And please take a look at my other comments. Thanks.

insights/parsers/__init__.py
```diff
@@ -498,38 +506,90 @@ def keyword_search(rows, **kwargs):
...
+        return data_key in row and matcher_fn(row[data_key], value)
+
+    data = list()
+    # Match on the translated row, return the original row
     for row in rows:
-        if all(map(lambda kv: key_match(row, kv[0], kv[1]), kwargs.items())):
+        if all(key_match(*term) for term in search_terms):
```
dkuc, line 593: @PaulWay  Wouldn't it make sense to turn this `all` into a regular loop so that you can drop out early?

My thinking is that the expensive `key_match` is called for every term in `search_term`.  This is unnecessary because you can stop after the first False value is found.

insights/parsers/__init__.py
```diff
@@ -454,21 +455,28 @@ def parse_delimited_table(table_lines,
     return r
 
 
-def keyword_search(rows, **kwargs):
+# A cache for the transformed keys of the row dict for a keyword_search.
+# This needs to be outside keyword_search because it needs to persist across
+# uses of keyword_search.
+keyword_search_transformed_row = dict()
```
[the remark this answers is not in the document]
PaulWay: Good catch - yes, this was replaced by the `_transform_cache` in the parent object.  I've removed that.

insights/parsers/__init__.py
PaulWay: Hmmm.  Interesting approach!  It's faster too AFAICS, so that's what we'll do.

PaulWay: The thing we're trying to handle is the weird case that a column might be named 'foo__bar', which looks like a 'foo__startswith' matcher invocation.  I can see the use of the `contains` and `startswith` and even `endswith` in the plugins, so these are being used; I can also see plenty of uses of search without the matcher invocation.  I can't see any uses of `__` in a keyword that isn't followed by a matcher.  So we want to optimise for searches with no '__', then finding the matcher as quickly as possible, and then finally the slow path of a keyword that has '__' in it but no matcher.

insights/parsers/__init__.py
PaulWay: Well, `all` will drop out early if any item fails, so I think it's going to already do that.

insights/parsers/__init__.py
PaulWay: I tested that and it was slower.  While it's nicer to have the single line of code, I made a judgement call that we do almost all of our comparisons using the `equals` matcher so making that as fast as possible was a priority.

xiangce (commented):
And with introducing the `row_keys_change`, test of some advisor rules is broken again.  It should be because the `InputData` to the corresponding Parsers are not so rigorous in tests (different test files used different `rows` for the same Spec/Parser). However, it's worth double-checking them carefully.

insights/parsers/__init__.py
dkuc: You're correct.  It was the map that you removed that was causing the full evaluation.  My mistake.

PaulWay:
Hi @xiangce - yeah, I was kind of expecting that.  I'll update the core tests to verify any fixes.

PaulWay:
Now going through the plugin tests and working out which parsers need `row_keys_change` set.

It does feel to me like maybe the parser could just parse the keys it should have, rather than the `keyword_search` function having to guess them...  But that probably gets out of sync with the content if we don't also cause an error when the actual search keyword is not one of the recognised keywords.

xiangce (approved):
After a couple of tests in my local, this change won't break any rules anymore. Although I didn't see significant performance improvement (due to the amount of data is too small) with the tests against archives, I think it's good to go.

PR #3605, opened 2022-11-22
fix: fix errors in ethtool

chenlizhong:

### All Pull Requests:

Check all that apply:

Fix errors:
insights.parsers.ethtool.CoalescingInfo ValueError("invalid literal for int() with base 10: 'n/a'")
insights.parsers.ethtool.TimeStamp ParseException('bad line: \tsoftware-receive')
insights.parsers.ethtool.TimeStamp ParseException('bad line: \thardware-transmit')
insights.parsers.ethtool.Ring ValueError("invalid literal for int() with base 10: 'n/a'"

### DIFF
--- a/insights/parsers/ethtool.py
+++ b/insights/parsers/ethtool.py
@@ -52,6 +52,7 @@ if sys.version_info.major == 3:
     # Python 3
     unicode = str
 
+
 def extract_iface_name_from_path(path, name):
     """
     Extract the 'real' interface name from the path name.  Basically this
--- a/insights/parsers/ethtool.py
+++ b/insights/parsers/ethtool.py
@@ -48,7 +48,7 @@ from ..parsers import ParseException
 from .. import parser, LegacyItemAccess, CommandParser
 from insights.specs import Specs
 
-if sys.version_info.major == 3:
+if sys.version_info[0] == 3:
     # Python 3
     unicode = str
 
--- a/insights/parsers/ethtool.py
+++ b/insights/parsers/ethtool.py
@@ -42,11 +42,15 @@ TimeStamp - command ``/sbin/ethtool -T {interface}``
 
 import os
 import re
+import sys
 from collections import namedtuple
 from ..parsers import ParseException
 from .. import parser, LegacyItemAccess, CommandParser
 from insights.specs import Specs
 
+if sys.version_info.major == 3:
+    # Python 3
+    unicode = str
 
 def extract_iface_name_from_path(path, name):
     """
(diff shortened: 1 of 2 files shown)


insights/parsers/ethtool.py
```diff
@@ -501,9 +501,10 @@ def parse_content(self, content):
         for line in content[2:]:
             if line.strip():
                 (key, value) = [s.strip() for s in line.split(":", 1)]
-                value = int(value)
-                self.data[key] = value
-                setattr(self, key.replace("-", "_"), value)
+                if 'n/a' not in value:
```
psachin, line 504: Since there is not description I am assuming that handling `n/a` is causing the error. Can you please add some test to cover that changes?

xiangce: Yeap, @chenlizhong could you please add a relevant description for the issue and tests of the update?

xiangce: BTW, please check these checkboxes if already match. And , we have agreed to use the [Coventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) for the insights-core to make it a better public and open-sourced project. Here is a cheat sheet for your reference as well: https://cheatography.com/albelop/cheat-sheets/conventional-commits/

chenlizhong: Update tests and will add one comment to show fixing errors.

chenlizhong:
@psachin @xiangce @ryan-blakley 
This MR is to fix below errors:

> insights.parsers.ethtool.CoalescingInfo       ValueError(\"invalid literal for int() with base 10: 'n/a'\")
> insights.parsers.ethtool.TimeStamp             ParseException('bad line: \\tsoftware-receive')
> insights.parsers.ethtool.TimeStamp             ParseException('bad line: \\thardware-transmit')
> insights.parsers.ethtool.Ring                         ValueError(\"invalid literal for int() with base 10: 'n/a'\")

psachin (approved):

insights/parsers/ethtool.py
ryan-blakley, line 504: @chenlizhong Looking at the exceptions that you posted, would it be better to check the type of the value variable and make sure it's capable to be passed to int? Like the below. That way if another case of the variable ends up being a string it will be skipped as well.

```
if value.isnumeric():
    value = int(value)
    self.data[key] = value
    setattr(self, key.replace("-", "_"), value)
```

ryan-blakley (changes requested):

insights/parsers/ethtool.py
chenlizhong: Thanks. Updated the code.

ryan-blakley: Awesome thank you it looks good to me, but there is one failed test due to python 2.6's sys.version_info not having named variables.

xiangce (approved):

PR #3619, opened 2022-11-30
Keep the code to delete previously-named malware rules file

mhuth:
- just in case the previously-named rules file isn't deleted for some reason

### All Pull Requests:

Check all that apply:

Just some small changes to some recent malware-detection commits I made.  Main one was making sure the old downloaded rules file is still deleted.  I've renamed the file, but its possible (although rare) that the old file could be maintained on the customer's system and isn't deleted as we've changed the file name now that we are deleting.  This commit makes sure the old file name is still deleted as well.


xiangce (approved):
Good to me

PR #3625, opened 2022-12-12
Delete old malware rules files from /var/tmp as well

mhuth:

### All Pull Requests:

Check all that apply:

Some customers have /var/tmp instead of /tmp and unfortunately the logic for deleting old malware rules files fails in that case. This PR adds /var/tmp as a location to try deleting old rules files from.  Why do we want to delete old malware rules files ... because they cause false positives in matching systems.


xiangce (approved):
Good to me

PR #3636, opened 2022-12-16
feat: Add env override to CommandOutputProvider

ryan-blakley:
### All Pull Requests:

Check all that apply:

Added the ability to override env variables in CommandOutputProvider for specs, along with updating each function that inherited from CommandOutputProvider. Added an override for the yum command to unset the LC_ALL env variable. Fixed an issue where self.create_env() was being called multiple times instead of once. Fixed some pep issues.

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -303,7 +303,8 @@ class CommandOutputProvider(ContentProvider):
     """
     Class used in datasources to return output from commands.
     """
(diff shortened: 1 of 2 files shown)


ryan-blakley:
@xiangce Not sure if this is the best method, but this change fixes the issue mentioned in bz 2091084, and possibly the others related to character issues.

xiangce (approved):
@ryan-blakley - I think this is a good workaround to the character issue and thanks for adding the interface to `override` the default evn.

Issue #3637, opened 2022-12-20
The redhat release combiner parses oracle linux wrongly


chenlizhong:
Below content is from Oralce linux. The combiner identify it as RHEL 8.4

$ uname -a
Linux myhostname 4.18.0-305.19.1.el8_4.x86_64 #1 SMP Wed Sep 19 15:42:05 PDT 2021 x86_64 x86_64 x86_64 GNU/Linux

$ cat /etc/redhat-release
Red Hat Enterprise Linux release 8.3 (Ootpa)

PR #3638, opened 2022-12-20
fix: enhance redhat_release combiner to exclude non-RHEL linux

chenlizhong:

### All Pull Requests:

Check all that apply:

To fix the issue https://github.com/RedHatInsights/insights-core/issues/3637
Use the /etc/os-release to check it is RHEL or not.

### DIFF
--- a/insights/combiners/redhat_release.py
+++ b/insights/combiners/redhat_release.py
@@ -4,8 +4,8 @@ Red Hat Release
 
 Combiner for Red Hat Release information. It uses the results of
 the ``Uname`` parser and the ``RedhatRelease`` parser to determine the release
-major and minor version.  ``Uname`` is the preferred source of data.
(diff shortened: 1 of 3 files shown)


xiangce:
@chenlizhong - Per my research, there are a couple of cases that the customer modifies the `/etc/os-release` on purpose to ensure it works with some other applications. For example, they only leave such a line for the `VERSION_ID`:
```
# os-release
VERSION_ID=6.10%
```

When running rules against these hosts, this PR will raise the `SkipCompoent` and skip this combiner, resulting in the rules that depend on this combiner, missing requirements. Maybe cause complaint from customers.

Since the #3640 provides a new combiner that could identify the OS a bit more accurately, we could use that PR for this issue.

chenlizhong:
Yes. we can close this one.

PR #3600, opened 2022-11-17
Add datasource to get jboss versions

chenlizhong:
Signed-off-by: Chen lizhong <lichen@redhat.com>

### All Pull Requests:

Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [ ] Is this PR to correct an issue?
* [ ] Is this PR an enhancement?

### Complete Description of Additions/Changes:

*Add your description here*

PR #3639, opened 2022-12-21
feat: add JbossRuntimeVersions parser

chenlizhong:

### All Pull Requests:

Check all that apply:

Add parser for datasource jboss_runtime_versions

https://github.com/RedHatInsights/insights-core/pull/3600

### DIFF
--- a/insights/parsers/jboss_version.py
+++ b/insights/parsers/jboss_version.py
@@ -3,14 +3,13 @@ JBoss version
 =============
 Provide information about the versions of all running Jboss on a system.
 """
-
-from collections import namedtuple
 import json
 
+from collections import namedtuple
 from insights import Parser, parser
 from insights.specs import Specs
 
-# define nametuple to store the property of version
+# define namedtuple to store the property of version
 _VersionNameTuple = namedtuple("_VersionNameTuple", ["file_path", "product", "version", "code_name", "major", "minor", "release"])
 
 
--- a/insights/parsers/jboss_version.py
+++ b/insights/parsers/jboss_version.py
@@ -1,45 +1,60 @@
 """
-JBoss version - File ``$JBOSS_HOME/version.txt``
-================================================
-
-This module provides plugins access to file ``$JBOSS_HOME/version.txt``
-
-Typical content of file ``$JBOSS_HOME/version.txt`` is::
-
-    Red Hat JBoss Enterprise Application Platform - Version 6.4.3.GA
-
-This module parses the file content and stores data in the dict ``self.parsed``.
-The version info can also be got via ``obj.major`` and ``obj.minor``, etc.
-
-Examples:
-    >>> jboss_version.file_path
-    '/home/test/jboss/jboss-eap-6.4/version.txt'
-    >>> jboss_version.raw
-    'Red Hat JBoss Enterprise Application Platform - Version 6.4.3.GA'
-    >>> jboss_version.major
-    6
-    >>> jboss_version.minor
-    4
-    >>> jboss_version.release
-    3
-    >>> jboss_version.version
-    '6.4.3'
-    >>> jboss_version.code_name
-    'GA'
+JBoss version
+=============
+Provide information about the versions of all running Jboss on a system.
 """
 from .. import Parser, parser
 from ..specs import Specs
+import json
+from insights.tests import context_wrap
 
 
 @parser(Specs.jboss_version)
 class JbossVersion(Parser):
-    """Parses the content of file ``$JBOSS_HOME/version.txt``."""
+    """
+    JBoss version - File ``$JBOSS_HOME/version.txt``
+    ================================================
+
+    This class is to access to file ``$JBOSS_HOME/version.txt``
+
+    Typical content of file ``$JBOSS_HOME/version.txt`` is::
+
+        Red Hat JBoss Enterprise Application Platform - Version 6.4.3.GA
+
+    This class parses the file content and stores data in the dict ``self.parsed``.
+    The version info can also be got via ``obj.major`` and ``obj.minor``, etc.
+
+    Examples:
+        >>> jboss_version.file_path
+        '/home/test/jboss/jboss-eap-6.4/version.txt'
+        >>> jboss_version.raw
+        'Red Hat JBoss Enterprise Application Platform - Version 6.4.3.GA'
+        >>> jboss_version.major
+        6
+        >>> jboss_version.minor
+        4
+        >>> jboss_version.release
+        3
+        >>> jboss_version.version
+        '6.4.3'
+        >>> jboss_version.code_name
+        'GA'
+    """
 
     def parse_content(self, content):
         self.raw = content[0]
         product, _, version_name = [v.strip() for v in content[0].partition("- Version")]
-        version, code_name = version_name.strip().rsplit(".", 1)
-        major, minor, release = version.split(".")
+        if " GA" in version_name:
+            # handle Red Hat JBoss Web Server - Version 5.6 GA
+            version = version_name.split(' GA')[0]
+            code_name = "GA"
+            updated_version = version + ".0"
+            major, minor, release = updated_version.split('.')[0:3]
+        else:
+            # add empty code name for Red Hat Data Grid - Version 7.3.0
+            version_name = version_name.strip() + "."
+            major, minor, release, code_name = version_name.split(".")[0:4]
+            version = '.'.join([major, minor, release])
         self.parsed = {
             "product": product,
             "version": version,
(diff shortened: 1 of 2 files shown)


insights/parsers/jboss_version.py
```diff
@@ -1,45 +1,60 @@
...
+            version = version_name.split(' GA')[0]
+            code_name = "GA"
+            updated_version = version + ".0"
+            major, minor, release = updated_version.split('.')[0:3]
+        else:
+            # add empty code name for Red Hat Data Grid - Version 7.3.0
+            version_name = version_name.strip() + "."
+            major, minor, release, code_name = version_name.split(".")[0:4]
+            version = '.'.join([major, minor, release])
         self.parsed = {
```
xiangce, line 58: Since you modified the body of the `parse_content`, could you please 
1. Rename the `self.parsed` to `self._parsed` as a private member (may need to add one more property `product`)
2. Or add a docstring for it to make it a full public attribute?

insights/parsers/jboss_version.py
```diff
@@ -73,3 +88,35 @@ def release(self):
...
+         >>> all_jboss_versions.versions[0].major
+         7
+         >>> all_jboss_versions.versions[0].minor
+         3
+         >>> all_jboss_versions.versions[0].release
+         0
+     """
+
+    def parse_content(self, content):
+        self.versions = []
```
xiangce, line 119: In this case, it would be better to make the combiner as a list e.g. `JbossRuntimeVersions(Parser, list):`, instead of adding a new `self.versions` attribute.

insights/tests/parsers/test_jboss_version.py
```diff
@@ -34,11 +34,55 @@ def test_jboss7():
...
+    assert jboss_v.release == 0
+
+
 def test_jboss_version_doc_examples():
     env = {
         'JbossVersion': JbossVersion,
         'jboss_version': JbossVersion(context_wrap(JBOSS_6,
-                                                   path='/home/test/jboss/jboss-eap-6.4/version.txt'))
+                                                   path='/home/test/jboss/jboss-eap-6.4/version.txt')),
+        'JbossRuntimeVersions': JbossRuntimeVersions,
```
xiangce, line 83: This line is not required. The same to L80

insights/parsers/jboss_version.py
```diff
@@ -1,45 +1,60 @@
...
+import json
+from insights.tests import context_wrap
 
 
 @parser(Specs.jboss_version)
 class JbossVersion(Parser):
-    """Parses the content of file ``$JBOSS_HOME/version.txt``."""
+    """
+    JBoss version - File ``$JBOSS_HOME/version.txt``
+    ================================================
```
xiangce, line 16: It's not a good idea to add such a doc title here.

insights/parsers/jboss_version.py
```diff
@@ -73,3 +88,35 @@ def release(self):
     def code_name(self):
         """string: code name of this running JBoss progress."""
         return self.parsed["code_name"]
+
+
+@parser(Specs.jboss_runtime_versions)
+class JbossRuntimeVersions(Parser):
+    """
+     Versions of multiple running JBoss - File ``data/insights_commands/jboss_versions``
+     ===================================================================================
```
xiangce, line 97: The same comment as above.

xiangce (changes requested):

insights/parsers/jboss_version.py
chenlizhong: Remove them.

insights/tests/parsers/test_jboss_version.py
chenlizhong: Remove them

insights/parsers/jboss_version.py
chenlizhong: I update it as iterator.  As list has add, del etc operations, this class does not need these. So just use the iterator.

insights/parsers/jboss_version.py
chenlizhong: Updated.

chenlizhong:
@xiangce @ryan-blakley Please reveiw

ryan-blakley (changes requested):

insights/parsers/jboss_version.py
```diff
@@ -1,46 +1,58 @@
...
-    >>> jboss_version.code_name
-    'GA'
+JBoss version
+=============
+Provide information about the versions of all running Jboss on a system.
 """
 from .. import Parser, parser
 from ..specs import Specs
+import json
+from insights.tests import context_wrap
```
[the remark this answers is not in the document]
chenlizhong: Updated

insights/parsers/jboss_version.py
```diff
@@ -49,27 +61,76 @@ def parse_content(self, content):
...
+         3
+         >>> all_jboss_versions[0].release
+         0
+     """
+
+    def parse_content(self, content):
+        self._versions = []
+        jboss_version_dict = json.loads(' '.join(content))
+        for j_path, version_content in jboss_version_dict.items():
+            self._versions.append(JbossVersion(context_wrap(version_content, path=j_path)))
```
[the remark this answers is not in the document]
chenlizhong: The content is {jboss_path: content of version.txt} json. It is no meaningful to use JSONParser to parser the json as the main part is content of version.txt. The class JbossRuntimeVersions stands for running versions of JBoss. So current codes are reflect the logic.  I updated the code to use Context class instead of tests/context_wrap.

insights/parsers/jboss_version.py
```diff
@@ -1,46 +1,60 @@
...
+
+from insights import Parser, parser
+from insights.specs import Specs
+from insights.core.context import Context
 
 
 @parser(Specs.jboss_version)
 class JbossVersion(Parser):
-    """Parses the content of file ``$JBOSS_HOME/version.txt``."""
+    """
```
bfahr, line 16: @chenlizhong since the spec `jboss_version` is no longer being collected this parser should be deprecated.  What do you think about pulling the parse_content out into a function and then just calling that function from both the deprecated parser and the new parser?

insights/parsers/jboss_version.py
```diff
@@ -49,27 +63,78 @@ def parse_content(self, content):
...
+         0
+     """
+
+    def parse_content(self, content):
+        self._versions = []
+        jboss_version_dict = json.loads(' '.join(content))
+        for j_path, version_content in jboss_version_dict.items():
+            lines = version_content.strip().splitlines()
+            new_context = Context(content=lines, path=j_path)
+            self._versions.append(JbossVersion(new_context))
```
bfahr, line 125: See my comment above, this could just be a call to the separate parse function, no need to create `Content` here.

insights/parsers/jboss_version.py
```diff
@@ -49,27 +63,78 @@ def parse_content(self, content):
...
 
     @property
     def code_name(self):
         """string: code name of this running JBoss progress."""
-        return self.parsed["code_name"]
+        return self._parsed["code_name"]
+
+
+@parser(Specs.jboss_runtime_versions)
+class JbossRuntimeVersions(Parser):
```
bfahr, line 98: Since you are implementing a list here why not just base this class on the `list` type and get all of the functionality for free?  Then you could remove all of the `__fxns__` to make this look like a list.

bfahr (changes requested):

insights/parsers/jboss_version.py
chenlizhong: OK. will update to list

insights/parsers/jboss_version.py
chenlizhong: This jboss_version is to parse <jboss_home>/version.txt. Insights-client does not collect it. But It is in JDR archive. I'd prefer to keep it.  The new parser (JbossRuntimeVersions) is just to parse multiple version.txt files.  One version.txt is corresponding to one instance of JbossVersion. So it is nature to use the JbossVersion class directly and do not need the seperate function of parse_content.

bfahr: I had not checked the JDR archive specs.  In that case I agree we should keep it.

insights/parsers/jboss_version.py
bfahr: Since we are keeping `JBossVersion` parser it would be cleaner to pull out the content parsing so that you don't have to create a fake context wrapper to call the parser here don't you think?  I don't believe we have any other dependencies between parsers like this do we?

insights/parsers/jboss_version.py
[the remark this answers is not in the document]
chenlizhong: Updated

insights/parsers/jboss_version.py
chenlizhong: Updated

insights/parsers/jboss_version.py
chenlizhong: Updated

insights/parsers/jboss_version.py
chenlizhong: Updated

xiangce (approved):

insights/parsers/jboss_version.py
```diff
@@ -1,75 +1,127 @@
...
-    'GA'
+JBoss version
+=============
+Provide information about the versions of all running Jboss on a system.
 """
-from .. import Parser, parser
-from ..specs import Specs
+
+from collections import namedtuple
+import json
```
ryan-blakley, line 8: This is a nitpick, but could you flip these and add a space between them like below? Everything else looks good to me.

```
import json

```

xiangce: @ryan-blakley - lizhong takes PTO, I update these imports according to your comments. And will merge it soon.

PR #3644, opened 2023-01-03
chore: remove unused spec lsinitrd_kdump_image

TZ3070:
Signed-off-by: Xinting Li <xintli@redhat.com>

### All Pull Requests:

Check all that apply:

There is no insights rule using spec lsinitrd_kdump_image and remove it.


xiangce:
@TZ3070 - in general, we don't delete any existing parsers/combiners directly, instead, we should `deprecate` it firstly.

See https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/ovs_vsctl_list_bridge.py

TZ3070:
@xiangce ACK

PR #3647, opened 2023-01-09
Create timeout signal for hostcontext only

lhuett:
### All Pull Requests:

Check all that apply:

* Since timeout signals are only needed when HostContext is used we only need to set the signal in that case, logic was change to skip when using any other context than Hostcontext.
* Changes were made to the pytest module to add HostContext object to the broker in present tests
* Additional pytests were added to test when not HostContext

### DIFF
--- a/insights/core/plugins.py
+++ b/insights/core/plugins.py
@@ -34,6 +34,7 @@ from pprint import pformat
 from six import StringIO
 
 from insights.core import dr
+from insights.core.context import HostContext
 from insights.util.subproc import CalledProcessError
 from insights import settings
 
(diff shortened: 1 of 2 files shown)


insights/core/plugins.py
```diff
@@ -99,8 +100,9 @@ def invoke(self, broker):
         # Grab the timeout from the decorator, or use the default of 120.
         self.timeout = getattr(self, "timeout", 120)
 
-        signal.signal(signal.SIGALRM, self._handle_timeout)
-        signal.alarm(self.timeout)
+        if HostContext in broker:
```
xiangce, line 103: @lhuett - do you think it's reasonable to also move the definition of the `self.timeout` (L101) into the `if` clause? since it's only used in this case.

lhuett: @xiangce Good point, I can do that.

Issue #3650, opened 2023-01-12
InstalledRpms parser is implementing logic by try-except  and checking for each line


chenlizhong:
Two issues are in below code:

1. It is using from_json  to load first, after failed, use the from_line
2. Do above  for each line. if the file has 1000 lines, these will occur 1000 times

https://github.com/RedHatInsights/insights-core/blob/c62d7dd1659625e250b957b80774b96585b21288/insights/parsers/installed_rpms.py#L256-L272

Suggestion:
Do file  content checking before the  loop to determine from_json or from_line at the beginning.

xiangce:
resolved by https://github.com/RedHatInsights/insights-core/pull/3651

PR #3651, opened 2023-01-12
fix: check the content first in class InstalledRpms

chenlizhong:
Signed-off-by: Chenlizhong <lichen@redhat.com>

### All Pull Requests:

Check all that apply:

To fix the issue https://github.com/RedHatInsights/insights-core/issues/3650

Check the content first to determine from_json or from_list, this will improve the performance.

### DIFF
--- a/insights/parsers/installed_rpms.py
+++ b/insights/parsers/installed_rpms.py
@@ -17,7 +17,7 @@ import six
 import warnings
 
 from ..util import rsplit
-from .. import parser, get_active_lines, CommandParser
+from .. import parser, CommandParser
 from .rpm_vercmp import rpm_version_compare
 from insights.specs import Specs
 from insights import ContainerParser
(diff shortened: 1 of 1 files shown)


insights/parsers/installed_rpms.py
```diff
@@ -253,24 +253,21 @@ def __init__(self, *args, **kwargs):
 
     def parse_content(self, content):
         packages = defaultdict(list)
-        for line in get_active_lines(content, comment_char='COMMAND>'):
-            if line.startswith('error:') or line.startswith('warning:'):
-                self.errors.append(line)
+        all_active_lines = get_active_lines(content, comment_char='COMMAND>')
```
xiangce, line 256: This `get_active_lines` will **go through the whole file** to remove all the `comment_char` that appears in each line. For this parser, the input content is from the `rpm` command, there will never be multiple such `comment_char` comments.

The `"COMMAND>"` comment is for the `foreman-debug` archive only and it will always be the first line of the input content if it's a foreman archive.

So, better do not to use this `get_active_lines` function that iterates the who `installed_rpms` spec which is always a bit large. (save time aggressively :)

```python
if content and "COMMAND>" in content[0]:
    del content[0]  # or content = content[1:]
```

xiangce (commented):
A more aggressive suggestion, do not to use the `get_achive_lines` which will iterator all the lines but do any valuable things for this parser.

insights/parsers/installed_rpms.py
chenlizhong: Updated

xiangce (approved):
bravo! I like this

insights/parsers/installed_rpms.py
```diff
@@ -253,24 +253,21 @@ def __init__(self, *args, **kwargs):
 
     def parse_content(self, content):
         packages = defaultdict(list)
-        for line in get_active_lines(content, comment_char='COMMAND>'):
+        if content and "COMMAND>" in content[0]:
+            content = content[1:]
```
xiangce, line 257: better raise a "SkipComponent" when content is empty before and after this removal

xiangce (commented):
Sorry for missing this in the first review.
In fact, it's not big deal, but I think it's worth checking the empty content instead of `rasing out of index` exceptions

insights/parsers/installed_rpms.py
```diff
@@ -17,7 +17,7 @@
 import warnings
 
 from ..util import rsplit
-from .. import parser, get_active_lines, CommandParser
+from .. import parser, CommandParser
```
xiangce, line 20: and since you updated the code here, could you please update the other `relative` importing to the full importing and sort them in alphabet order?

xiangce (approved):

insights/parsers/installed_rpms.py
chenlizhong: updated

xiangce:
This will break tests of many existing rules, skip merging it in this release cycle.

insights/parsers/installed_rpms.py
```diff
@@ -253,24 +253,24 @@ def __init__(self, *args, **kwargs):
 
     def parse_content(self, content):
         packages = defaultdict(list)
-        for line in get_active_lines(content, comment_char='COMMAND>'):
+        content = [line for line in content if line.strip()]  # in testing file, there are empty line
```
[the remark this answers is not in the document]
chenlizhong: Updated

xiangce:
MR was raised to fix the broken tests, this PR is ready

PR #3656, opened 2023-01-13
Move remaining exceptions to the new exceptions module

ryan-blakley:
### All Pull Requests:

Check all that apply:

Moved the remaining exceptions that were defined in various other modules to the new separate exceptions module. That way all exceptions are defined in their own module. Updated all imports to reference the new path.

### DIFF
--- a/docs/api_index.rst
+++ b/docs/api_index.rst
@@ -30,6 +30,14 @@ insights.core.dr
     :members:
     :exclude-members: requires, optional, metadata, group, tags
 
+insights.core.exceptions
+------------------------
+
+.. automodule:: insights.core.exceptions
+    :members:
+    :show-inheritance:
+    :undoc-members:
+
 insights.core.filters
 ---------------------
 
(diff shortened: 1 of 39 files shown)


insights/core/exceptions.py
```diff
@@ -1,3 +1,49 @@
+class AutologyError(Exception):
```
xiangce, line 1: @ryan-blakley - do you think it's necessary to add a doc-entry/index for this `insights.core.exceptions` module in the `docs`?

insights/util/autology/datasources.py
```diff
@@ -7,11 +7,11 @@
 import inspect
 import insights.core.spec_factory
 import insights.specs
+
+from insights.core.exceptions import AutologyError
 from insights.specs.default import DefaultSpecs as InDefaultSpecs
 from insights.specs.insights_archive import InsightsArchiveSpecs as InArchiveSpecs
 from insights.specs.sos_archive import SosSpecs as InSosSpecs
-from . import AutologyError
```
bfahr, line 13: @ryan-blakley I don't think its necessary to move this exception.  Since this is a utility module, its not used by core for normal processing.

docs/api_index.rst
```diff
@@ -30,6 +30,14 @@ insights.core.dr
     :members:
     :exclude-members: requires, optional, metadata, group, tags
 
+insights.core.exceptions
+---------------------
```
xiangce, line 34: nitpick :), you may want to adjust the length of this line.

ryan-blakley: Yeah it turns out to be the reason it's not building, I'll correct it thanks.

PR #3661, opened 2023-01-19
Make malware-detection app more resilient to unexpected errors

mhuth:
* And make log messages more informative when things go wrong

### All Pull Requests:

Check all that apply:

The malware-detection app can fail silently when things go wrong unexpectedly.  This also causes it to not upload results archives leaving customers confused as to why it failed and why they don't see any results in the UI.  This PR adds a few more try/except blocks around code sections where problems could occur and adds more helpful error messages ... however it doesn't fix the problem so I've still go more work to do there :/  

It also displays an error when the malware-detection app component unexpectedly fails as a whole which at least gives us a starting point for debugging the problem.

### DIFF
--- a/insights/client/apps/malware_detection/__init__.py
+++ b/insights/client/apps/malware_detection/__init__.py
@@ -187,6 +187,7 @@ class MalwareDetectionClient:
         self.add_metadata = self._get_config_option('add_metadata', False)
 
         self.matches = 0
+        self.potential_matches = 0
 
     def run(self):
         # Start the scans and record the time they were started
@@ -202,7 +203,11 @@ class MalwareDetectionClient:
 
             # Write a message to user informing them if there were matches or not and what to do next
             if self.matches == 0:
-                logger.info("No rule matches found.\n")
+                if self.potential_matches == 0:
+                    logger.info("No rule matches found.\n")
+                else:
+                    logger.info("Rule matches potentially found but problems encountered parsing them, so no match data to upload.")
+                    logger.info("Please contact support.\n")
             else:
                 logger.info("Found %d rule match%s.", self.matches, 'es' if self.matches > 1 else '')
                 if not self.test_scan:
@@ -805,7 +810,12 @@ class MalwareDetectionClient:
                 logger.debug("Unable to scan %s: %s", toplevel_dir, cpe.output.strip())
                 continue
 
-            self.parse_scan_output(output.strip())
+            try:
+                self.parse_scan_output(output.strip())
+            except Exception as e:
+                self.potential_matches += 1
+                logger.exception("Rule match(es) potentially found in %s but problems encountered parsing the results: %s.  Skipping ...",
+                                 toplevel_dir, str(e))
 
             dir_scan_end = time.time()
             logger.info("Scan time for %s: %d seconds", toplevel_dir, (dir_scan_end - dir_scan_start))
@@ -872,7 +882,12 @@ class MalwareDetectionClient:
                 logger.debug("Unable to scan process %s: %s", scan_pid, cpe.output.strip())
                 continue
 
-            self.parse_scan_output(output)
+            try:
+                self.parse_scan_output(output)
+            except Exception as e:
+                self.potential_matches += 1
+                logger.exception("Rule match(es) potentially found in process %s but problems encountered parsing the results: %s.  Skipping ...",
+                                 scan_pid, str(e))
 
             pid_scan_end = time.time()
             logger.info("Scan time for process %s: %d seconds", scan_pid, (pid_scan_end - pid_scan_start))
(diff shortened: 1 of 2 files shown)


insights/client/apps/malware_detection/__init__.py
```diff
@@ -863,7 +875,13 @@ def scan_processes(self):
...
                 continue
 
-            self.parse_scan_output(output)
+            try:
+                self.parse_scan_output(output)
+            except Exception as e:
+                self.potential_matches += 1
+                logger.error("Rule match(es) potentially found in process %s but problems encountered parsing the results: %s.  Skipping ...",
+                             scan_pid, str(e))
+                traceback.print_exc()
```
bfahr, line 884: @mhuth do you really want to print this to STDOUT?

insights/client/apps/malware_detection/__init__.py
```diff
@@ -970,11 +988,16 @@ def skip_string_data_lines(string_data_lines):
...
+                try:
+                    # Add extra data to each rule match, beyond what yara provides
+                    # Eg, for files: line numbers & context, checksums; for processes: process name
+                    # TODO: find more pythonic ways of doing this stuff instead of using system commands
+                    metadata_func = self._add_file_metadata if source_type == 'file' else self._add_process_metadata
+                    metadata_func(rule_match['matches'])
+                except Exception as e:
+                    logger.error("Error adding metadata to rule match %s in %s %s: %s.  Skipping ...",
+                                 rule_name, source_type, source, str(e))
+                    traceback.print_exc()
```
bfahr, line 1000: Same comment as above.

insights/specs/datasources/malware_detection.py
```diff
@@ -20,12 +23,17 @@ def malware_detection_app(broker):
...
-                raise SkipComponent
-        except Exception:
-            raise SkipComponent
+                raise SkipComponent("No scan results were produced")
+        except SkipComponent as msg:
+            raise SkipComponent("Skipping malware-detection app: {0}".format(str(msg)))
+        except Exception as err:
+            err_msg = "Unexpected exception in malware-detection app: {0}".format(str(err))
+            logger.error(err_msg)
+            traceback.print_exc()
```
bfahr, line 38: Same comment as above.

insights/client/apps/malware_detection/__init__.py
```diff
@@ -796,7 +802,13 @@ def scan_filesystem(self):
                 logger.debug("Unable to scan %s: %s", toplevel_dir, cpe.output.strip())
                 continue
 
-            self.parse_scan_output(output.strip())
+            try:
+                self.parse_scan_output(output.strip())
+            except Exception as e:
+                self.potential_matches += 1
```
ryan-blakley, line 808: I'm curious what the potential_matches is used for, I see it was added in this PR but isn't used by anything added in the PR? Unless I missed something.

insights/client/apps/malware_detection/__init__.py
mhuth: @bfahr I figured it could be one less troubleshooting step that needed to be done if something completely unexpected went wrong.  But I can remove it if its not very insights-client-y

insights/client/apps/malware_detection/__init__.py
mhuth: @ryan-blakley well, if we hit an exception in that code we are in processing the yara output, so it means we had rule matches but we won't know exactly what they were because of the exception.  So potential_matches is just used to triggger a message to the user that although we don't have any specific scan matches to report to you, we potentially found matches.  And you should probably contact support to help troubleshoot the problem further.

Anyway, that's the reasoning behind potential_matches.

ryan-blakley: Okay thanks for explaining, I just wanted to make sure it wasn't something that was meant to be used but got removed.

insights/client/apps/malware_detection/__init__.py
bfahr: @mhuth I just thought it might be better to expose it in a log message so it would be captured with other run information.

mhuth: Ah ok, good point Bob.  I'll update the PR so it logs the tracebacks

mhuth: @bfahr I'm using logger.exception(<msg>) now to print a traceback to stderr, so it gets displayed on the command line and in the logs.  Thanks for picking this up.

insights/client/apps/malware_detection/__init__.py
mhuth: Now using logger.exception

insights/specs/datasources/malware_detection.py
mhuth: Now using logger.exception

bfahr (approved):

chenlizhong (approved):
Looks good to me

PR #3662, opened 2023-01-19
Deprecate SkipException to help avoid confusion

ryan-blakley:
Check all that apply:

Mark SkipException deprecated to help avoid confusion as the dirty parser difference isn't really needed. Replaced all internal references with SkipComponent.

### DIFF
--- a/docs/exception_model.rst
+++ b/docs/exception_model.rst
@@ -82,7 +82,7 @@ any exceptions in the data (“dirty parser”). This allows rules that don’t 
 exceptions to rely on only the first parser, and those rules will not run if valid data
 is not present.  If the dirty parser identifies errors in the data then it will save
 information regarding the errors for use by rules.  If no errors are found in the data
-then the dirty parser will raise :py:class:`insights.core.exceptions.SkipException`
+then the dirty parser will raise :py:class:`insights.core.exceptions.SkipComponent`
 to indicate to the engine that it should be removed from the dependency hierarchy.
 
 Other Exceptions from Parsers
@@ -99,15 +99,13 @@ types aren’t important and such checks may limit expressiveness and flexibilit
 Parsers should not use the assert statement in place of error handling code.
 Asserts are for debugging purposes only.
 
-SkipComponent and SkipException
-===============================
+SkipComponent
+=============
 
 Any component may raise `SkipComponent` to signal to the engine that
 nothing is wrong but that the component should be taken out of dependency
 resolution. This is useful if a component's dependencies are met but it's
 still unable to produce a meaningful result.
-:py:class:`insights.core.exceptions.SkipException` is a specialization of this for the
-dirty parser use case above, but it's treated the same as `SkipComponent`.
 
 Exception Recognition by the Insights Engine
 ============================================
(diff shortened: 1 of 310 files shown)


bfahr (changes requested):

docs/exception_model.rst
```diff
@@ -84,6 +84,7 @@ is not present.  If the dirty parser identifies errors in the data then it will
 information regarding the errors for use by rules.  If no errors are found in the data
 then the dirty parser will raise :py:class:`insights.core.exceptions.SkipException`
 to indicate to the engine that it should be removed from the dependency hierarchy.
+**SkipException will be deprecated after version 3.2.25.**
```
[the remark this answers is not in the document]
ryan-blakley: @bfahr I updated the docs let me know if it looks better now.

insights/core/exceptions.py
```diff
@@ -81,10 +84,20 @@ def __init__(self, content_type):
...
     Exception that should be thrown from parsers that are explicitly
     written to look for errors in input data.  If the expected error
     is not found then the parser should throw this exception to
     signal to the infrastructure that the parser's output should not be
     retained.
     """
-    pass
+    def __init__(self):
+        deprecated(
+            SkipComponent,
```
xiangce, line 99: Oops, I missed this one, here is should be `SkipException`

insights/core/exceptions.py
```diff
@@ -81,10 +84,20 @@ def __init__(self, content_type):
...
+            "Please use the :class:`insights.core.exceptions.SkipComponent` instead.",
+            "3.2.25"
+        )
+        super(SkipComponent, self).__init__()
```
xiangce, line 103: The same comment as above.
And the `Exception Message` should also be handled here, in case the customized `msg` is passed in.  e.g.`SkipException(msg)`

xiangce (changes requested):

xiangce:
BTW, this PR breaks the tests of gss and prodsec rules, I raised a MR to resolve the test of gss rules and asked ProdSec team to resolve the prodsec rules as well.

insights/core/exceptions.py
ryan-blakley: Ahh good catch sorry about that, fixed.

insights/core/exceptions.py
ryan-blakley: I added handling the msg, I missed that sorry about that.

ryan-blakley:
@xiangce Another good catch I tested the other projects but forgot about gss-rules. In addition to your PR, I corrected the issue by adding a back import in insights/core/__init__.py for SkipException. All tests pass for me now on all projects that I know about.

bfahr (approved):

jholecek-rh:
@ryan-blakley Thank you. The PR looks good. Our rule repository won't be affected. We haven't been using `SkipException` exceptions for over a year now. All tests pass.

xiangce (approved):

insights/core/exceptions.py
xiangce: NEVER mind, it always happens :)

PR #3663, opened 2023-01-19
feat: Add arg to capture skips in the broker

ryan-blakley:
Check all that apply:

Added the --show-skips cli arg to trigger SkipComponents to be captured in the broker for troubleshooting purposes.

### DIFF
--- a/insights/__init__.py
+++ b/insights/__init__.py
@@ -266,9 +266,8 @@ def _load_context(path):
     return dr.get_component(path)
 
 
-def run(component=None, root=None, print_summary=False,
-        context=None, inventory=None, print_component=None):
-
+def run(component=None, root=None, print_summary=False, context=None, inventory=None, print_component=None,
+        store_skips=False):
     args = None
     formatters = None
 
@@ -293,6 +292,8 @@ def run(component=None, root=None, print_summary=False,
         p.add_argument("--context", help="Execution Context. Defaults to HostContext if an archive isn't passed.")
         p.add_argument("--no-load-default", help="Don't load the default plugins.", action="store_true")
         p.add_argument("--parallel", help="Execute rules in parallel.", action="store_true")
+        p.add_argument("--show-skips", help="Capture skips in the broker for troubleshooting.", action="store_true",
+                       default=False)
         p.add_argument("--tags", help="Expression to select rules by tag.")
 
         class Args(object):
@@ -385,6 +386,10 @@ def run(component=None, root=None, print_summary=False,
         graph = dr.COMPONENTS[dr.GROUPS.single]
 
     broker = dr.Broker()
+    if args:
+        broker.store_skips = args.show_skips
+    else:
+        broker.store_skips = store_skips
 
     if args and args.bare:
         ctx = ExecutionContext()  # dummy context that no spec depend on. needed for filters to work
--- a/insights/core/dr.py
+++ b/insights/core/dr.py
@@ -809,6 +809,7 @@ class Broker(object):
             :func:`time.time`. For components that produce multiple instances,
             the execution time here is the sum of their individual execution
             times.
+        store_skips (bool): Weather to store skips in the broker or not.
     """
     def __init__(self, seed_broker=None):
         self.instances = dict(seed_broker.instances) if seed_broker else {}
@@ -816,6 +817,7 @@ class Broker(object):
         self.exceptions = defaultdict(list)
         self.tracebacks = {}
         self.exec_times = {}
+        self.store_skips = False
 
         self.observers = defaultdict(set)
         if seed_broker is not None:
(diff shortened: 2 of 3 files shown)


insights/__init__.py
```diff
@@ -385,6 +386,7 @@ class Args(object):
         graph = dr.COMPONENTS[dr.GROUPS.single]
 
     broker = dr.Broker()
+    broker.cli_args = args
```
jholecek-rh, line 389: I am not sure about this. The CLI is only one way to run the `insights-core` engine. There are many tools out there which do it programmatically. I doesn't feel right that these tools should build an `insights-core` CLI arguments representation to configure the `Broker`.

How about changing the `Broker` API to be configured like this

```suggestion
    broker = dr.Broker()
    broker.store_skips = args.show_skips
```

or like this (unless there are reasons against changing the signature of `Broker.__init__`)

```suggestion
    broker = dr.Broker(store_skips=args.show_skips)
```

insights/core/dr.py
```diff
@@ -1045,8 +1048,12 @@ def run_components(ordered_components, components, broker):
...
             log.warning(pe)
             broker.add_exception(component, pe, traceback.format_exc())
-        except SkipComponent:
-            pass
+        except SkipComponent as sc:
+            if broker.cli_args and broker.cli_args.show_skips:
+                log.warning(sc)
+                broker.add_exception(component, sc, traceback.format_exc())
+            else:
+                pass
```
jholecek-rh, line 1056: Is it enough to add it only here? [`insights.core.plugins.parser`](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/plugins.py#L171) duplicates this logic. Would it make sense to add it there as well?

I am not sure if the logic is duplicated elsewhere. I didn't find anything but it doesn't necessarily mean it isn't. :shrug:

Btw. Would you agree that it might be worth finding a way to avoid this code duplication (not in this PR).

insights/__init__.py
ryan-blakley: So I included the args in the broker because I didn't want to add another arg to I think 8-9 functions that get called prior to dr.run finally getting called. I saw that all of those functions passed the broker, so I figured it was easier to pass the args within the broker. This way in the future if we need to add another arg in a similar way they can all be accessed. Also I will submit another PR to remove the parallel args from those various functions, to instead access the arg through this new method. The cli args in the broker aren't required in order to setup the broker, so it wouldn't effect any other project out there that uses the broker but not the cli.

jholecek-rh: Ok, my suggestions were off because I didn't realize that the parameter wasn't for the broker, but for code that worked with the broker. I think I still have a point that strong ties to the CLI are not desirable.

A single configuration object accessed directly when code needs a configuration value instead of passing individual values around is a good idea, imho. It makes making use of new configuration parameters easier no matter where in the code they are needed.

I am not sure about using the `Broker` class as a vehicle to get the configuration object where it is needed. The class would have two completely independent responsibilities which is generally not a good idea. Also, CLI arguments are passed to code that doesn't use the broker (even before the broker object is created, although the order could be changed). Would you pass the configuration through the broker only to code that needs the broker or you would start passing the broker to code that needs configuration values but not the broker?

If you are worried about the verbosity of passing a dedicated configuration object around, would constraints allow for a global `insights.config` module? It would avoid the need to pass the configuration object in arguments, but it is not threading-friendly if the configuration needs to change after initialization.

Either way, I still think I have a point about binding the configuration to CLI arguments so tightly.  What would a tool need to do to configure the `show_skips` parameter with the proposed implementation if it didn't invoke analysis using `insights.run()` (e.g. because it does too much in some regards and is not flexible enough in other regards).

The `args` object will always define attributes for all CLI arguments with default values supplied through the CLI arguments definition. Code using the object doesn't need to worry about attributes not being defined or having undefined values when a specific default value is always supplied. However, downstream applications don't have easy means to construct an object with the same properties. An object with just one attribute that the application wants to set would not do, would it?

ryan-blakley: I updated the code to use a bool in the broker for storing skips, I also added a parameter to the run function to allow projects that don't use the cli to trigger storing skips. Let me know if this works better for your project.

jholecek-rh: Thanks. The current version works for me. I think it the best we can do without major refactoring (it would be nice if the broker parameter didn't depend on code that uses the broker to implement the behavior).

If you want to simplify it a little, it isn't necessary to add the `store_skips` parameter to the `run()` function. I see that function as implementing the insights CLI, so having the function process the CLI argument is enough. Applications not using the `run()` function can set the parameter on the broker directly.

xiangce (approved):
The current version looks good to me.

insights/__init__.py
ryan-blakley: Awesome thanks for the suggestions.

Regarding the run function it is actually used by several other projects that don't use the cli part, it only parses cli arguments if the print_summary parameter is passed to it. So I added the parameter there just in case those projects ever want to store skips as well.

jholecek-rh: Ack. Makes sense. Thanks!

PR #3665, opened 2023-01-25
Exclude malware-detection rules files in /var/tmp (and other locations)

mhuth:

### All Pull Requests:

Check all that apply:

Customers who have /tmp linked to /var/tmp will get false positives from the downloaded rules file in /var/tmp because we only check for the rules file in /tmp.  /facepalm ... D'oh ... crikey ... sheesh ... good lord ....  Anyway, this PR makes sure we check /tmp, /var/tmp, /usr/tmp and tempfile.gettempdir() for the downloaded rules file so we can ignore it in all those locations.

### DIFF
--- a/insights/client/apps/malware_detection/__init__.py
+++ b/insights/client/apps/malware_detection/__init__.py
@@ -7,7 +7,7 @@ from sys import exit
 import logging
 from glob import glob
 from datetime import datetime
-from tempfile import NamedTemporaryFile
+from tempfile import NamedTemporaryFile, gettempdir
 try:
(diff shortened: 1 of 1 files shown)


mhuth:
Customer confirmed that this patch fixed the problem for them, so it would be good to get this out as soon as we can.

ahitacat (approved):
LGTM

mhuth:
Thanks @ahitacat, how do I get this merged as neither of us have permissions to merge it :/

bfahr (approved):

PR #3529, opened 2022-09-20
fix: bz-2091084: used LC_ALL=C.UTF-8 instead of LC_ALL=C

xiangce:
Signed-off-by: Xiangce Liu <xiangceliu@redhat.com>

### All Pull Requests:

Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [x] Is this PR to correct an issue?
* [ ] Is this PR an enhancement?

### Complete Description of Additions/Changes:

fix: [bz-2091084](https://bugzilla.redhat.com/show_bug.cgi?id=2091084)

PR #3669, opened 2023-01-31
fix: use LC_ALL=C.UTF-8 for subscription-manager

ptoscano:
Signed-off-by: Pino Toscano <ptoscano@redhat.com>

### All Pull Requests:

Check all that apply:

The default environment for `simple_command` includes `LC_ALL=C`, which means ASCII (and not unicode); in case the language of the system is set to a non-English locale that uses unicode characters (e.g. non-Latin1 languages), then `subscription-manager` will fail to output that with encoding issues (e.g. [1]).

As a simple solution, enforce an unicode English locale for `subscription-manager`, which should avoid those encoding issues.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2074745

### DIFF
--- a/insights/specs/default.py
+++ b/insights/specs/default.py
@@ -600,8 +600,10 @@ class DefaultSpecs(Specs):
     sshd_config = simple_file("/etc/ssh/sshd_config")
     sshd_config_perms = simple_command("/bin/ls -lH /etc/ssh/sshd_config")
     sssd_config = simple_file("/etc/sssd/sssd.conf")
-    subscription_manager_facts = simple_command("/usr/sbin/subscription-manager facts")
-    subscription_manager_id = simple_command("/usr/sbin/subscription-manager identity")  # use "/usr/sbin" here, BZ#1690529
+    subscription_manager_facts = simple_command("/usr/sbin/subscription-manager facts",
+                                                override_env={"LC_ALL": "C.UTF-8"})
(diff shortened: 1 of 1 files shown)


ptoscano:
Another possibility here could be the modify `SAFE_ENV` directly to be unicode, applying the same for every command run:
```diff
diff --git a/insights/core/spec_factory.py b/insights/core/spec_factory.py
index 5d433f92..03a9c197 100644
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -32,7 +32,7 @@ SAFE_ENV = {
         "/usr/sbin",
         "/usr/share/Modules/bin",
     ]),
-    "LC_ALL": "C",
+    "LC_ALL": "C.UTF-8",
 }
 """
 A minimal set of environment variables for use in subprocess calls
```
Maybe it could be better, and avoid the need to care about specific applications?

ptoscano:
> Another possibility here could be the modify `SAFE_ENV` directly to be unicode, applying the same for every command run:

This was attempted by #3529, which apparently was withdrawn.

xiangce (approved):

xiangce:
> This was attempted by #3529, which apparently was withdrawn.

Thanks @ptoscano , let's adopt the new resolution of #3636, it's more flexible and won't affect the other commands which are now running well.

PR #3670, opened 2023-02-01
fix: Enhance datasource kernel_module_filters to check the loaded modules

wushiqinlou:

### All Pull Requests:

Check all that apply:

The issue is https://github.com/RedHatInsights/insights-core/issues/3671

My first resolution is to add `keep_rc=True` into the spec:

~~~
modinfo_filtered_modules = command_with_args('modinfo %s', kernel_module_list.kernel_module_filters, `keep_rc=True`)
~~~
After updating, the spec can collect all of the output like this:

~~~
modinfo: ERROR: Module cvblk not found.
modinfo: ERROR: Module drbd not found.
modinfo: ERROR: Module emcpioc not found.
modinfo: ERROR: Module filepath not found.
filename:       /lib/modules/4.18.0-240.22.1.el8_3.x86_64/kernel/drivers/net/ethernet/intel/i40e/i40e.ko.xz
version:        2.8.20-k
...
~~~

However, the problem is that `modinfo: ERROR: Module XXX not found` can break the normal format:

~~~
alias:          pci:v00008086d000015ABsv*sd*bc*sc*i*
almodinfo: ERROR: Module krg_621845_imRH8_K2S_smp64 not found.
modinfo: ERROR: Module oracleafd not found.
ias:          pci:v00008086d000015B0sv*sd*bc*sc*i*
alias:          pci:v00008086d000015AAsv*sd*bc*sc*i*
...
name:           oracleasm
vermagic:       4.18.0-2modinfo: ERROR: Module oracleoks not found.
modinfo: ERROR: Module twnotify not found.
modinfo: ERROR: Module veeamsnap not found.
40.14.1.el8_3.x86_64 SMP mod_unload modversions
sig_id:         PKCS#7
~~~

We can modify the parser `KernelModulesInfo` to fix this issue, however, the broke format is not fixed, I am afraid that it is hard to avoid all the situations.

The second resolution is to update the datasource to only check the loaded modules:

~~~
cat insights/specs/datasources/kernel_module_list.py

@datasource(LsMod, HostContext)
def kernel_module_filters(broker):
    """
    Return a string of a list of modules from the spec filter,
    separated with space.
    """
    filters = sorted((get_filters(Specs.modinfo_modules)))
    if filters:
        new_modules = []
        for item in filters:
            if item in broker[LsMod]:
                new_modules.append(item)
        return ' '.join(new_modules)
    raise SkipComponent
~~~

### DIFF
--- a/insights/specs/datasources/kernel_module_list.py
+++ b/insights/specs/datasources/kernel_module_list.py
@@ -5,10 +5,11 @@ from insights.core.context import HostContext
 from insights.core.exceptions import SkipComponent
 from insights.core.filters import get_filters
 from insights.core.plugins import datasource
+from insights.parsers.lsmod import LsMod
 from insights.specs import Specs
 
 
-@datasource(HostContext)
+@datasource(LsMod, HostContext)
 def kernel_module_filters(broker):
     """
     Return a string of a list of modules from the spec filter,
@@ -16,5 +17,9 @@ def kernel_module_filters(broker):
     """
     filters = sorted((get_filters(Specs.modinfo_modules)))
     if filters:
-        return ' '.join(filters)
+        new_modules = []
+        for item in filters:
+            if item in broker[LsMod]:
+                new_modules.append(item)
+        return ' '.join(new_modules)
     raise SkipComponent
--- a/insights/specs/datasources/kernel_module_list.py
+++ b/insights/specs/datasources/kernel_module_list.py
@@ -17,9 +17,11 @@ def kernel_module_filters(broker):
     """
     filters = sorted((get_filters(Specs.modinfo_modules)))
     if filters:
-        new_modules = []
+        loaded_modules = []
         for item in filters:
             if item in broker[LsMod]:
-                new_modules.append(item)
-        return ' '.join(new_modules)
+                loaded_modules.append(item)
+        if loaded_modules:
+            return ' '.join(loaded_modules)
+        raise SkipComponent
     raise SkipComponent
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -204,8 +204,8 @@ plugins:
         - name: insights.components.virtualization.IsBareMetal
           enabled: true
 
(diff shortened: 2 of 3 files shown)


insights/specs/datasources/kernel_module_list.py
```diff
@@ -5,16 +5,21 @@
 from insights.core.exceptions import SkipComponent
 from insights.core.filters import get_filters
 from insights.core.plugins import datasource
+from insights.parsers.lsmod import LsMod
 from insights.specs import Specs
 
 
-@datasource(HostContext)
+@datasource(LsMod, HostContext)
```
xiangce, line 12: If a parser is required by a datasource, the data source should be `pre-loaded` while the collection, please add it to the "plugins.configs" section of "insights.collect.default_manifest" module, and set its `enable` as true.

insights/specs/datasources/kernel_module_list.py
```diff
@@ -5,16 +5,21 @@
...
     separated with space.
     """
     filters = sorted((get_filters(Specs.modinfo_modules)))
     if filters:
-        return ' '.join(filters)
+        new_modules = []
+        for item in filters:
+            if item in broker[LsMod]:
+                new_modules.append(item)
+        return ' '.join(new_modules)
```
xiangce, line 24: Better raise a `SkipComponent` when no modules are found here.

xiangce (changes requested):

insights/specs/datasources/kernel_module_list.py
wushiqinlou: @xiangce - Updated, please check

insights/specs/datasources/kernel_module_list.py
wushiqinlou: @xiangce - Updated, please check

xiangce (approved):

chenlizhong (approved):
Looks good to me

PR #3680, opened 2023-02-20
feat: New spec "/etc/sos.conf" and its parser

huali027:
New spec "/etc/sos.conf" and its parser

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/sos_conf.py
+++ b/insights/parsers/sos_conf.py
@@ -3,11 +3,8 @@ SosConf - file ``/etc/sos.conf`` or ``/etc/sos/sos.conf``
 =========================================================
 """
 from insights.core import IniConfigFile
-from insights.core.filters import add_filter
 from insights.core.plugins import parser
 from insights.specs import Specs
-
-add_filter(Specs.sos_conf, "[")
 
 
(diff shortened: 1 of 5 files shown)


huali027:

insights/parsers/sos_conf.py
```diff
@@ -0,0 +1,43 @@
+"""
+SosConf - file ``/etc/sos.conf`` or ``/etc/sos/sos.conf``
+=========================================================
+"""
+from insights.core import IniConfigFile
+from insights.core.filters import add_filter
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+add_filter(Specs.sos_conf, "[")
```
xiangce, line 10: TBH, it's not recommended to add the filters of the spec here in the parser, especially for a spec that is not so common as the `ps aux/ps auxww`. How about moving the filters to the rule's side?

huali027: @xiangce yes, works for me. I'll update it.

xiangce (approved):

chenlizhong (approved):

PR #3684, opened 2023-02-21
Update rpm_pkgs datasource and create its new parser

jobselko:
### All Pull Requests:

Check all that apply:

This PR udpates datasource logic to provide additional information about packages. The original `RpmPkgs` parser was replaced by `RpmPkgsWritable` parser.

### DIFF
--- a/insights/parsers/rpm_pkgs.py
+++ b/insights/parsers/rpm_pkgs.py
@@ -2,7 +2,7 @@
 RpmPkgs - datasource ``rpm_pkgs``
 =================================
 
-Parser for CVE-2021-35937, CVE-2021-35938, and CVE-2021-35939.
+Parsers for CVE-2021-35937, CVE-2021-35938, and CVE-2021-35939.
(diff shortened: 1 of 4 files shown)


jobselko:
@xiangce Hi, do you think it is necessary to rename the datasource when updating the logic inside?

xiangce (commented):
@jobselko - TBH, your updates look good to me. However, it breaks the original usage of the Parsers, potentially breaking the tests that depend on it in other projects. Although it seems we might be the only user of it, for a public open-source project, we'd better not change it in this way. Could you please mark the old `RpmPkgs` as deprecated instead of renaming it and add a new parser for the new return value?

And I don't think it's necessary to do the same to the datasource, since it's used internally only.

jobselko:
@xiangce Hi, I have updated the datasource to keep the original logic, marked `RpmPkgs` as deprecated, and created a new parser `RpmPkgsWritable`.

chenlizhong (approved):

PR #3688, opened 2023-03-01
feat: New spec and parser for sys_block_queue_stable_writes

JoySnow:
### All Pull Requests:

Check all that apply:

New spec and parser for glob file: `/sys/block/*/queue/stable_writes`

### DIFF
--- a/insights/parsers/sys_block.py
+++ b/insights/parsers/sys_block.py
@@ -38,7 +38,7 @@ class StableWrites(Parser):
             raise ParseException('Error: {0}'.format(content if content else 'empty file'))
         try:
             self._stable_writes = int(content[0].strip())
-            self._device = self.file_path.split('/')[3]
+            self._device = self.file_path.split('/')[-3]
         except ValueError:
             raise ParseException("Error: unparsable content: {0}".format(content[0]))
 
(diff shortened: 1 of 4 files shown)


JoySnow:
Hi @xiangce , could you help review the changes?

xiangce (commented):
The current code looks good to me, but I have a concern about if the `block device` (`*`) in the path matters? Do you think it's good to add one more property to return the current `device name` (get from the file_path)?

JoySnow:
Hi @xiangce , Added a `device` property for device name . Please help review the update.

insights/parsers/sys_block.py
```diff
@@ -36,10 +38,16 @@ def parse_content(self, content):
             raise ParseException('Error: {0}'.format(content if content else 'empty file'))
         try:
             self._stable_writes = int(content[0].strip())
+            self._device = self.file_path.split('/')[3]
```
xiangce, line 41: Since the file_path would be a absolute path in the runtime, it would be better to use `[-3]` here instead of `[3]`

JoySnow: Yes, good point.

xiangce (approved):

PR #3689, opened 2023-03-01
Fix/gpg verification new

grunwmar:
### All Pull Requests:

Check all that apply:

Pull request **fix/gpg-validation-new**
<https://issues.redhat.com/browse/RHELPLAN-119658>

**New approach**
As `python-gpg` is not in package dependencies for `insights-client` I decided to turn back to solution to *call gpg as subprocess* but with parameter `--homedir` which value (path) is created by `tempfile.TemporaryDirectory()`. 

Test are testing cartesian product of two sets of options; results {valid egg/asc pair, invalid egg/asc pair} and methods {`InsightsClient.verify(path)`, `InsightsUploadConf.validate_gpg_sig(path, sig=None)`, `InsightsUploadConf.validate_gpg_sig(path, sig=path+".asc")`}

**note:**
It is worth to note, that during the test it is needed to stop system’s `insights-client` using command `sudo systemctl stop insights-client-results.path` to prevent creation of `/root/.gnupg` by it and then to not interfere with observation.

**original approach**
~~For *insights-client* I created validation method which uses `python-gpg` package to access system `gpgme` package directly w/o cli/subprocess interphase and imported it to `src/insights_client/__init__.py` where it replaced original implementation. And by using tempfile to create a temporary home dir for gnupg. **No more true!**~~


insights/client/__init__.py
```diff
@@ -304,13 +304,15 @@ def verify(self, egg_path, gpg_key=constants.pub_gpg_path):
 
         # if a valid egg path and gpg were received do the verification
         if egg_path and gpg_key:
-            cmd_template = '/usr/bin/gpg --verify --keyring %s %s %s'
-            cmd = cmd_template % (gpg_key, egg_path + '.asc', egg_path)
+            homedir = tempfile.mkdtemp()
+            cmd_template = '/usr/bin/gpg --homedir %s --verify --keyring %s %s %s '
```
Glutexo, line 308: Is the space at the end of the string intentional?

Glutexo, line 308: Is the space at the end of the string intentional?

PR #3692, opened 2023-03-04
Draft: Allow listdir to return relative paths

jholecek-rh:
### All Pull Requests:

Check all that apply:

This PR is motivated by the need to get access to the ConfigMap names [gathered|https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md#configmaps] by the Insights Operator.

The `listdir` spec factory returned only basenames of listed files and directories. However, the paths of files and directories may contain information that is not available anywhere else and that rules may need to access.

The PR adds a `path_format` argument to the `listdir` spec factory that allows the spec factory to generate a data source that returns paths relative to the execution context root.

I considered adding an argument to the existing spec factory and introducing a new spec factory (derived from the existing one). I chose to add an argument because I think it provides a better developer experience (one configurable spec factory vs multiple very similar spec factories).

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -694,19 +694,30 @@ class listdir(object):
         context (ExecutionContext): the context under which the datasource
             should run.
         ignore (str): regular expression defining paths to ignore.
+        path_format: format of the returned paths; supported values are "baseline" (default, returns
+            only basenames) and "relative" (returns paths relative to the execution context root)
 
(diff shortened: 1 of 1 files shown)


jholecek-rh:
I came to the [conclusion](https://github.com/RedHatInsights/insights-core/pull/3693) that extending `listdir` is not the way to go. I will prepare a PR that adds a new `listglob` spec factory.

PR #3694, opened 2023-03-04
feat: Split listdir to listdir and listglob spec factories

jholecek-rh:
### All Pull Requests:

Check all that apply:

The exact behavior of specs created using the listdir spec factory was difficult to predict because it depended on a number of subtle details:

- ContentException was raised instead of an empty list in some cases, an empty list was returned in other cases.
- The ignore filter was applied to returned values in some cases and to absolute paths in other cases.
- The specs returned only basenames of files and directories even for a complex glob (e.g. `some/*/file`) where returning relative paths makes more sense.

A dedicated `listglob` spec factory can have a more predictable behavior while keeping the implementation simple. In most cases, users will know if they want to list a directory or a glob pattern. An amphibious spec factory is not necessary.
    
This MR:

- Removes support for globs from `listdir` (no known specs rely on this functionality)
- Updates `listdir` to raise `ContentException` only if the specificed path is not an existing directory; an empty list will be returned for empty directories (SkipComponent will be raised one step later for known listdir specs; the change should not affect any rules).
- Adds a `listglob` spec factory with behavior optimized for listing globs

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -690,14 +690,15 @@ class listdir(object):
     path.
 
     Args:
-        path (str): directory or glob pattern to list.
+        path (str): directory to list.
         context (ExecutionContext): the context under which the datasource
             should run.
-        ignore (str): regular expression defining paths to ignore.
+        ignore (str): regular expression defining names to ignore.
 
     Returns:
-        function: A datasource that returns the list of files and directories
-            in the directory specified by path
+        function: A datasource that returns a sorted list of file and directory
+            names in the directory specified by path. The list will be empty when
+            the directory is empty or all names get ignored.
     """
 
     def __init__(self, path, context=None, ignore=None, deps=[]):
@@ -712,11 +713,11 @@ class listdir(object):
         ctx = _get_context(self.context, broker)
         p = os.path.join(ctx.root, self.path.lstrip('/'))
         p = ctx.locate_path(p)
-        result = sorted(os.listdir(p)) if os.path.isdir(p) else sorted(glob(p))
-
-        if result:
-            return [os.path.basename(r) for r in result if not self.ignore_func(r)]
-        raise ContentException("Can't list %s or nothing there." % p)
+        try:
+            result = os.listdir(p)
+        except OSError as e:
+            raise ContentException(str(e))
+        return sorted([r for r in result if not self.ignore_func(r)])
 
 
 class simple_command(object):
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -720,6 +720,31 @@ class listdir(object):
         return sorted([r for r in result if not self.ignore_func(r)])
 
 
+class listglob(listdir):
+    """
+    List paths matching a glob pattern.
+
+    Args:
+        pattern (str): glob pattern to list.
+        context (ExecutionContext): the context under which the datasource
+            should run.
+        ignore (str): regular expression defining paths to ignore.
+
+    Returns:
+        function: A datasource that returns the list of paths that match
+            the given glob pattern. The list will be empty when nothing matches.
+    """
+    def __call__(self, broker):
+        ctx = _get_context(self.context, broker)
+        p = os.path.join(ctx.root, self.path.lstrip('/'))
+        p = ctx.locate_path(p)
+        result = glob(p)
+        # generator expression; we don't need the full list at this step
+        result = (os.path.relpath(r, start=ctx.root) for r in result)
+        result = sorted([r for r in result if not self.ignore_func(r)])
+        return result
+
+
 class simple_command(object):
     """
     Execute a simple command that has no dynamic arguments
--- a/insights/tests/core/spec_factory/test_listdir.py
+++ b/insights/tests/core/spec_factory/test_listdir.py
@@ -1,0 +1,82 @@
+import os
+import pytest
+import tempfile
+
+from insights.core import dr
+from insights.core.context import HostContext
+from insights.core.spec_factory import listdir
+
+
+@pytest.fixture
+def sample_directory(scope="module"):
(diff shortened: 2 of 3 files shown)


insights/core/spec_factory.py
```diff
@@ -719,6 +719,31 @@ def __call__(self, broker):
...
+    """
+    def __call__(self, broker):
+        ctx = _get_context(self.context, broker)
+        p = os.path.join(ctx.root, self.path.lstrip('/'))
+        p = ctx.locate_path(p)
+        result = sorted(glob(p))
+        # generator expression; we don't need the full list at this step
+        result = (os.path.relpath(r, start=ctx.root) for r in result)
+        result = [r for r in result if not self.ignore_func(r)]
+        return result
```
xiangce, line 744: How about raising a ContentException when there is nothing matched, just like the `listdir`?

xiangce: I just saw your comment about this point in #3693, but I still recommend you to raise a ContentException here, since the empty `list` is not worth keeping and taking to the parser side. The final result for the parser would be the same: avoid invalid triggers (otherwise, you need to check empty by yourself). And the `ContentException` would be used for debugging/troubleshooting since it could be get from the `broker.exceptions`. Unless there is a rule that needs to check the empty result as a condition

jholecek-rh: The rule that I need it for will hit if `listglob` returns an empty list. The rule will take additional measures to make sure the list is not empty by accident (e.g. a data collection error). 

Also note that there will be no `@parser` component involved. This datasource already provides structured data. Moreover, `@parser` components cannot work with datasources like `listdir` and `listglob`. The existing `listdir` specs are used as providers in `foreach_*` spec factories. Mine will be used by a "parser" implemented as `@combiner`.

As for the change that you are proposing:

The only case when there are little doubts about raising `ContentException` is when `listdir` is supposed to list a directory which doesn't exists.

In all other cases when `listdir` or `listglob` would return an empty list, the datasource does not know if it is due to the list being genuinely empty or due to a data collection error. There are two reasons for raising `ContentException` in this case:
1. To indicate the option of a data collection error, enforcing rules to take extra steps (specifying an optional dependency) before using the "empty result".
2. To optimize performance (based on the assumption that most applications will be interested in the **presence** of an item in the result).

I would argue that reason 1 is giving rule developers a false sense of security. If a rule is interested in the **absence** of an item in the result, the fact that `listdir` or `listglob` returned a non-empty list does not mean the list is complete. The list can be still incomplete due to data collection errors and the rule should use an independent channel to verify if is safe to use it (e.g. [gathers.json](https://github.com/openshift/insights-operator/blob/master/docs/insights-archive-sample/insights-operator/gathers.json) for Insights Operator archives.

I would argue that reason 2 is negligible. There might be a little overhead processing the components, but empty list means zero iterations of any processing code otherwise.

The `listdir` does not behave consistently like this and can return an empty list (see #3693). There are two options to align `listdir` and `listglob` behavior. Both options require `listdir` changes (I can prepare the changes; the changes do not break existing specs).

1. `listdir` does  not support globs; it raises `ContentException` when the specified path is not a directory. `listdir` and `listglob` can return an empty list.
2. `listdir` and `listglob` always raise `ContentException` instead of returning an empty list. (Whether `listdir` supports globs is not relevant in this case.)

I favor option 1; I think it provides a more flexible API and it matches my immediate use case better.

I can settle for option 2 if `listdir` is changed accordingly. I would still remove glob support from `listdir`.

WDYT?

jholecek-rh: There's another option: add an `allow_empty` argument to the two spec factories. This would allow individual specs to declare if empty list is a valid result or not.

jholecek-rh: We agreed separately on option 1. I'm closing this thread.

jholecek-rh:
@xiangce @ryan-blakley  I've updated the MR to do the whole change of splitting `listdir` into `listdir` and `listglob` as discussed yesterday. Could you take a look again, please?

insights/core/spec_factory.py
```diff
@@ -712,11 +712,39 @@ def __call__(self, broker):
...
+        try:
+            result = os.listdir(p)
+        except OSError as e:
+            raise ContentException(str(e))
+        return sorted([r for r in result if not self.ignore_func(r)])
 
-        if result:
-            return [os.path.basename(r) for r in result if not self.ignore_func(r)]
-        raise ContentException("Can't list %s or nothing there." % p)
+    def _locate_path(self, broker):
```
xiangce, line 721: Do you want to remove this unused function?

insights/core/spec_factory.py
```diff
@@ -712,11 +712,39 @@ def __call__(self, broker):
...
+
+    Args:
+        pattern (str): glob pattern to list.
+        context (ExecutionContext): the context under which the datasource
+            should run.
+        ignore (str): regular expression defining paths to ignore.
+
+    Returns:
+        function: A datasource that returns the list of paths that match
+            the given glob pattern.
```
xiangce, line 737: It would be better if a description like `Empty list will be returned when nothing gets matched`.

xiangce (commented):
TBH, your new update looks great to me, other than the following tiny comments.

insights/core/spec_factory.py
jholecek-rh: Indeed. Sorry for this omission and thanks for spotting it.

insights/core/spec_factory.py
jholecek-rh: Makes sense. Docstrings updated for both `listdir` and `listglob`. Is it OK like this?

xiangce (approved):
@jholecek-rh - It looks good to me. Thanks.

ryan-blakley (approved):
Everything looks good to me.

xiangce:
test me

PR #3695, opened 2023-03-05
Fix malware SSL error fix - make it more flexible

mhuth:
### All Pull Requests:

Check all that apply:

The malware SSL error fix lacked flexibility in detecting the SSL error.  This PR uses a regex to determine if the SSL error has occurred.  It probably needs a better way of determining an SSL error occurred other than string matching the SSL error message, but this seems to work well enough (for the time being).

### DIFF
--- a/insights/client/apps/malware_detection/__init__.py
+++ b/insights/client/apps/malware_detection/__init__.py
@@ -702,7 +702,7 @@ class MalwareDetectionClient:
                     logger.error("Unable to download rules from %s: %s", self.rules_location, str(e))
                     exit(constants.sig_kill_bad)
 
-                if 'CERTIFICATE_VERIFY_FAILED' in str(e):
+                if re.search('SSL.*verify.failed', str(e), re.IGNORECASE):
                     # SSL error occurred, possibly due to a custom CA cert, try using a CA bundle for cert verification
                     self.insights_config.cert_verify = self._get_config_option('ca_cert', ca_cert)
                     logger.debug("Trying cert_verify value %s ...", self.insights_config.cert_verify)
--- a/insights/tests/client/apps/test_malware_detection.py
+++ b/insights/tests/client/apps/test_malware_detection.py
@@ -461,23 +461,26 @@ class TestGetRules:
         # Because we've set retries to 3, the last attempted retry will wait 4 seconds (= 2**2)
         log_mock.debug.assert_called_with("Trying again in %d seconds ...", 4)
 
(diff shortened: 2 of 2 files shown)


insights/client/apps/malware_detection/__init__.py
```diff
@@ -702,7 +702,7 @@ def _get_rules(self):
                     logger.error("Unable to download rules from %s: %s", self.rules_location, str(e))
                     exit(constants.sig_kill_bad)
 
-                if 'CERTIFICATE_VERIFY_FAILED' in str(e):
+                if re.search('SSL.*verify.failed', str(e), re.IGNORECASE):
```
chenlizhong, line 705: How about use the exception type (urllib3.exceptions.SSLError) to make judgement instead of string searching?

mhuth: I'm not sure if one of the machines is returning an SSL error though.  These are the different error messages they are getting:

`2023-02-15 09:21:39,243    DEBUG insights.client.apps.malware_detection Unable to download rules from https://m01eassmg0007.nix.m01.no.myatea.net:443/redhat_access/r/insights/platform/malware-detection/v1/test-rule.yar: HTTPSConnectionPool(host='m01eassmg0007.nix.m01.no.myatea.net', port=443): Max retries exceeded with url: /redhat_access/r/insights/platform/malware-detection/v1/test-rule.yar (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:897)'),))
`
... that one is quite clearly an SSLError exception type.

However this one ...
`2023-03-03 09:59:19,419    DEBUG insights.client.apps.malware_detection Unable to download rules from https://m01eassmg0007.nix.m01.no.myatea.net:443/redhat_access/r/insights/platform/malware-detection/v1/test-rule.yar: ("bad handshake: Error([('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')],)",)
`
... is less obvious that the exception is of type SSLError.

That's why I went with string pattern matching.  The regex matches both errors and the customer has tested my regex solution and its working for them.  I'm less confident check if the exeception type is an SSLError will work in the second case above.

chenlizhong: I think the two error are both 'requests.exceptions.SSLError', but I am not sure and can not test this. So let us use the searching method first.

chenlizhong (approved):
Maybe all error are requests.exceptions.SSLError. But I can not test these. So approve first. You can use the searching method now.

PR #3708, opened 2023-03-14
feat: New datasource and parser for leapp-report.json

xiangce:
### All Pull Requests:

Check all that apply:


insights/specs/datasources/leapp.py
```diff
@@ -0,0 +1,61 @@
...
+        'inhibitor': ['title', 'summary', ('detail', 'remediations')]
+    }
+    try:
+        content = broker[LocalSpecs.leapp_report_raw].content
+        report = json.loads('\n'.join(content))
+        if isinstance(report, dict):
+            results = []
+            for entry in report.get('entries', []):
+                for flag, keys in valid_keys.items():
+                    if flag in entry['flags']:
```
andywaltlova, line 45: You couldn't know and it occurred to me only this morning, leapp tool is deprecating the 'flags' key  in favor of new 'groups' ('flags' + 'tags'), **the schema with 'flags' will still be default for all customers**,  but to make it work in future, could this be changes in a way that if 'flags' key is not present then the parser will check 'groups'  key?

meaning it would work  for this entry
```
{
    .... other data
    'flags': ['inhibitor']
    .... other data
}
``` 
and also this entry  (only difference is that in this second one 'flags' and 'tags' keys are not present, instead 'groups' key is
```
{
    .... other data
    'groups': ['inhibitor']
    .... other data
}
``` 

the mentioned change is

Wecros: Just sharing hack I used in other script this morning 
```py
entry_flags = entry.get("flags", entry.get("groups", []))
```

xiangce: Thanks for the information. Updated.

PR #3710, opened 2023-03-15
Fix logs of check registration status for legacy upload

ahitacat:
### All Pull Requests:

Check all that apply:

This method gets the response to check if the client is registered or not. The commit checks if the response has failed and logs that using the `handle_fail_rcs` method.

Resolves: ESSNTL-4445


xiangce (approved):
Looks good to me

PR #3711, opened 2023-03-15
Feat: New parser ls_rsyslog_errorfile

TZ3070:
### All Pull Requests:

Check all that apply:

Add new parser ls_rsyslog_errorfile and datasource rsyslog_errorfile

### DIFF
--- a/insights/combiners/os_release.py
+++ b/insights/combiners/os_release.py
@@ -259,8 +259,7 @@ class OSRelease(object):
             ret.update(data)
             return ret
 
-        self._is_rhel = True
-        self._release = 'Red Hat Enterprise Linux'
+        self._release = 'RHEL'
         self._reasons = {}
         _dmesg = dmesg.linux_version if dmesg else dmesg
         if not list(filter(None, [uname, _dmesg, rpms])):
@@ -270,12 +269,10 @@ class OSRelease(object):
                 ret = _from_os_release(osr) if osr else dict()
                 ret.update(_from_redhat_release(rhr)) if rhr else None
                 if ret.get('other_linux', 'RHEL') != 'RHEL':
-                    self._is_rhel = False
                     self._release = ret['other_linux']
                     self._reasons = {'reason': 'NON-RHEL: os-release/redhat-release'}
             else:
                 # Nothing means NON-RHEL
-                self._is_rhel = False
                 self._release = 'Unknown'
                 self._reasons = {'reason': 'Nothing available to check'}
         else:
@@ -288,7 +285,6 @@ class OSRelease(object):
                                 result, _from_installed_rpms(rpms, uname)))
             # 'other_linux' means NON-RHEL
             if 'other_linux' in result and result['other_linux'] != 'RHEL':
-                self._is_rhel = False
                 self._release = result.pop('other_linux')
                 self._reasons = result
 
@@ -297,7 +293,7 @@ class OSRelease(object):
         """
         Returns True if it's RHEL, False for NON-RHEL.
         """
-        return self._is_rhel
+        return self._release == 'RHEL'
 
     @property
     def release(self):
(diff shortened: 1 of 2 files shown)


TZ3070:
@xiangce Please have a review

insights/parsers/ls_rsyslog_errorfile.py
```diff
@@ -0,0 +1,33 @@
...
+from insights.core.plugins import parser
+from insights.specs import Specs
+from insights.core import CommandParser, ls_parser
+
+
+@parser(Specs.ls_rsyslog_errorfile)
+class LsRsyslogErrorfile(CommandParser):
+    def parse_content(self, content):
+        ls_data = ls_parser.parse(content, '').get('')
+        self.data = ls_data.get('entries')
```
xiangce, line 33: It's not recommended to use `self.data` as an attribute since the `data` is ambiguous in indicating both the "type" and data content. Could you mind changing the attribute name a bit reasonable, e.g `entries` ?

insights/specs/__init__.py
```diff
@@ -332,6 +332,7 @@ class Specs(SpecSet):
     ls_ocp_cni_openshift_sdn = RegistryPoint()
     ls_origin_local_volumes_pods = RegistryPoint()
     ls_osroot = RegistryPoint()
+    ls_rsyslog_errorfile = RegistryPoint(multi_output=True)
```
xiangce, line 335: It's not so good from the using perspective to mark it as `multi_output=True`, since it's from a `datasource` better change it to a single file.

insights/specs/datasources/rsyslog_errorfile.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@datasource(RsyslogAllConf, HostContext)
+def errorfiles(broker):
+    """
+    This datasource provides the list of the rsyslog errorfile paths.
+
+    Sample data returned::
+
+        ['/var/log/omelasticsearch.log', '/var/log/oversized.log']
```
xiangce, line 18: Following up on the above comment, here return a string separated by a whitespace, like `/var/log/omelasticsearch.log /var/log/oversized.log`, then the output can be wrapped into one single file.

xiangce (changes requested):

insights/parsers/ls_rsyslog_errorfile.py
```diff
@@ -0,0 +1,38 @@
+"""
+LsRsyslogErrorfile - command ``ls -ln <path>``
+==============================================
+
+The ``ls -ln <path>`` command provides information for the listing of the
+specified directory.
+
+Attributes:
+    data (dict): Dictionary of keys with values in dict.
```
xiangce, line 9: I believe this should be `entries`, And please move these `attributes` .... and `examples` into the Parsers docstring instead of the module's docstring.

insights/parsers/ls_rsyslog_errorfile.py
```diff
@@ -0,0 +1,38 @@
...
+
+@parser(Specs.ls_rsyslog_errorfile)
+class LsRsyslogErrorfile(Parser):
+    def parse_content(self, content):
+        parsed_content = []
+        for line in content:
+            if 'No such file or directory' not in line:
+                parsed_content.append(line)
+        ls_data = ls_parser.parse(parsed_content, '').get('')
+        self.entries = ls_data.get('entries')
```
xiangce, line 38: Please Skip this parser when there is nothing parsable (in case all are `No such file or directory`)

insights/specs/datasources/rsyslog_errorfile.py
xiangce, line 1: I suggest renaming this datasource module to `rsyslog_confs`, in case, there will be some other datasources depend on the`RsyslogAllConf` command, we don't need to add a new python module file.

xiangce:
And please fix the conflicts.

Issue #3731, opened 2023-04-06
Collect AWS public IP and Hostname


ezr-ondrej:
Public IP and Hostname is not known to the instance itself on AWS.
This is kept in its metadata[1] and it needs to be collected from there.

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html

lzap:
Please close this issue, thanks!

PR #3732, opened 2023-04-06
feat: collect AWS public IPv4 as cannonical fact

ezr-ondrej:
### All Pull Requests:

Check all that apply:

This tries to add an AWS public IP address, that is not known to the machine itself.

Issue: #3731

Refs ESSNTL-4588

This is yet a draft, I'd love to hear if my approach would work here, so far I've pulled just IP as it's easier do demonsrate and agree upon my intended approach :)

### DIFF
--- a/insights/util/canonical_facts.py
+++ b/insights/util/canonical_facts.py
@@ -17,6 +17,7 @@ from insights.parsers.azure_instance import AzureInstanceID, AzureInstanceType
 from insights.parsers.dmidecode import DMIDecode
 from insights.parsers.gcp_instance_type import GCPInstanceType
 from insights.parsers.installed_rpms import InstalledRpms
+from insights.parsers.ip import AWSPublicIPv4
 from insights.parsers.rhsm_conf import RHSMConf
 from insights.parsers.subscription_manager import SubscriptionManagerFacts
 from insights.parsers.yum import YumRepoList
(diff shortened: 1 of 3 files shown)


insights/util/canonical_facts.py
```diff
@@ -123,22 +124,27 @@ def _filter_falsy(dict_):
...
+    if aws_ipv4:
+        ips_list.append(aws_ipv4.data)
+
     facts = dict(
         insights_id=valid_uuid_or_None(_safe_parse(insights_id)),
         machine_id=valid_uuid_or_None(_safe_parse(machine_id)),
         bios_uuid=valid_uuid_or_None(_safe_parse(bios_uuid)),
         subscription_manager_id=valid_uuid_or_None(submanid.data if submanid else None),
-        ip_addresses=ips.data if ips else [],
+        ip_addresses=ips_list,
```
subpop, line 147: The canonical facts JSON schema is not (to my knowledge) strictly defined anywhere, but it is assumed to include a certain set of keys (in other words, it follows a de facto standard). There are other implementations in code that (for instance, [rhc](https://github.com/RedHatInsights/rhc/blob/main/canonical_facts.go#L112) and in [cloud-connector](https://github.com/RedHatInsights/cloud-connector/blob/90a7ec8ae8cc5a201a0fd05f9afd4f565ced4a9a/internal/cloud_connector/protocol/types.go#L42-L51)) that assume the current schema is adhered to. I suggest you take care when making changes to canonical facts. This change is safe, since you don't appear to be altering the key strings at all, but the above history of canonical facts is important to keep in mind.

xiangce:
@ezr-ondrej - This PR added a new Spec to the core collection, for any new specs, it's required to get it approved before collecting it. Could you please go to the insights-core-assets Gitlab repo to get it approved? I added the link of the repo to the ESSNTL-4588, please have a look.

insights/util/canonical_facts.py
kruai: FYI, here's the list of canonical facts: https://consoledot.pages.redhat.com/docs/dev/services/inventory.html#_host_deduplication

PR #3734, opened 2023-04-11
fix: update ubuntu image

ahitacat:
### All Pull Requests:

Check all that apply:

Updated ubuntu image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,20 +34,28 @@ jobs:
 
   python26-test:
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
+    container: ubuntu:18.04
 
     steps:
+    - name: Install sudo package
+      run: apt update && apt install sudo
+    - name: Install add-apt-repository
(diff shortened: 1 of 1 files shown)


mhuth:
*sigh* why are we using Ubuntu images and not Red Hat? :cry:

subpop:
> _sigh_ why are we using Ubuntu images and not Red Hat? cry

I've wondered about this too, but I'd say that change should be a separate PR from this one. This one is about fixing a broken test. We could explore the idea of switching to UBI images or CentOS Stream images in another PR.

tiran:
* `test_with_zip` seems to fail because the `unzip` binary is not installed
* `test_magic_plan_b` fails to load the `magic_open` C function pointer from libmagic with ctypes. This might be a problem with libmagic, ctypes, or libffi.
* The `test_oscap_*` seems to fail because it cannot find `rpm` binary (Ubuntu?).

ahitacat:
I'm closing this PR as it was continued in https://github.com/RedHatInsights/insights-core/pull/3766

PR #3752, opened 2023-04-14
Clarify logic when normalizing snippets

subpop:
Fixes a bug in the normalizeSnippet function that can inadvertently process elements in a CommendedSeq twice when iterating over elements in the sequence.

Fixes: ESSNTL-4354


ahitacat (changes requested):
The changes looks good to me.  @subpop can you add some new tests to avoid possible regressions?

subpop:
Finally got a unit test reproducing the bug. The "snippet" object has to be one returned by `loadPlaybookYaml` in order for the `normalizeSnippet` function to double-append strings in a sequence.

ahitacat (approved):
This looks good to me

PR #3755, opened 2023-04-18
feat: collect built-in metadata in 'data' via datasource

### LINKED ISSUE #3739: `meta_data` JSON file generated incorrect for some built-in file
Reported by @r14chandra:

The following issue is detected in puptoo's log:

```
" File \"/usr/local/lib/python3.8/site-packages/src/puptoo/process/profile.py\", line 781, in postprocess\n if facts[\"system_profile\"].get(\"display_name\"):\n", "KeyError: 'system_profile'\n"]}
```

This would be a regression bug introduced by the #3723 

The `meta_data/insights.specs.Specs.display_name.json` file was generated incorrectly even when there is no `display_name` in the archive. This resulted in the following check of  puptoo https://github.com/RedHatInsights/insights-puptoo/blob/master/src/puptoo/process/profile.py#L485 being True.

xiangce:
- also fix #3739 
- new built-in metadata files will be generated in the `data` dir for core collection
- mark the MetadataProvider as `deprecated` and remove it from 3.5.0
- and keep the old built-in files for backward compatibility
- see INSGHTCORE-203
### All Pull Requests:

Check all that apply:

- most of the built-in metadata are collected via specs
   and their old collections in the client are also kept
- the `blacklist_report` is not implemented in datasource yet
   since it depends on the #3679


insights/core/spec_factory.py
```diff
@@ -156,41 +156,6 @@ def __repr__(self):
         return '%s("%r")' % (self.__class__.__name__, self.path)
 
 
-class MetadataProvider(FileProvider):
-    """
-    Class used for insights-core built-in files.  These files should not
-    be filtered, redacted or blocked.
```
bfahr, line 162: @xiangce do you need to include some logic to make sure that the `tags` datasources are not redacted?

xiangce: @bfahr - With the new implementation (in #3679), these  built-in files will be excluded explicitly [here](https://github.com/RedHatInsights/insights-core/blob/2beef363db473b88221d590aea1df97194d6b150/insights/core/spec_cleaner.py#L44)

And in further, they can also be excluded by adding the `not_obfuscate` option [here](https://github.com/RedHatInsights/insights-core/pull/3679/files#diff-11e9e8667252cdaf31413d23bdfbe835b1b0eed3a307700ff69a4266e72b472dR5-R11)

Please let me know if this work for you on this point.

xiangce:
@chenlizhong and @JoySnow - As this PR for now, doesn't do changes in the client part,  it's ready. please kindly review this PR.

chenlizhong (approved):
The code is good to me

JoySnow (approved):

PR #3756, opened 2023-04-20
fix: DnfModuleList parser and collect dnf_module_list

psegedy:
### All Pull Requests:

Check all that apply:

DnfModuleList parser stores only the last module:stream which is incorrect, to solve this we should
- store all streams for module
- assign list of profiles to individual streams
- assign summary to individual streams
- parse [e] [x] [d] [a] [i] module stream and profile flags
- collect `dnf module list` output

This is going to be a breaking change but I haven't found any app using this parser in RedHatInsights org. It will be used by vulnerability-engine and patchman-engine to fix evaluation of CVEs/errata for modular packages by using this parser in insights-puptoo https://github.com/RedHatInsights/insights-puptoo/pull/272

### DIFF
--- a/insights/parsers/dnf_module.py
+++ b/insights/parsers/dnf_module.py
@@ -17,42 +17,90 @@ from insights.parsers import parse_fixed_table
 from insights.specs import Specs
 
 
+class Profile(object):
+    """
+    An object for the dnf module:stream profile
+
+    Attributes:
+        profile (Optional[str]): Profile of the dnf module:stream
+        default (bool): Default flag of the dnf module:stream profile
+        installed (bool): Installed flag of the dnf module:stream profile
+    """
+    def __init__(self, profile):
+        self.profile = profile.split(" ")[0]
+        self.default = "[d]" in profile
+        self.installed = "[i]" in profile
+
+
+class DnfModuleStream(object):
+    """
+    An object for the dnf module:stream
+
+    Attributes:
+        stream (str): Name of the module stream
+        profiles (list): List of profiles of the dnf module:stream
+        summary (str): Summary of the dnf module:stream
+        default (bool): Default flag of the dnf module:stream
+        enabled (bool): Enabled flag of the dnf module:stream
+        disabled (bool): Disabled flag of the dnf module:stream
+        installed (bool): Installed flag of the dnf module:stream
+        active (Optional[bool]): Active flag of the dnf module:stream
+    """
+    def __init__(self, data=None, cmd="list"):
+        data = data or {}
+        self._cmd = cmd
+        self._stream = data.get('Stream', '')
+        self.stream = self._stream.split(" ")[0]
+        self.profiles = [Profile(p.strip()) for p in data.get('Profiles', '').split(',') if p.strip()]
+        self.summary = data.get('Summary', '')
+        self.default = "[d]" in self._stream
+        self.enabled = "[e]" in self._stream
+        self.disabled = "[x]" in self._stream
+        self.installed = "[i]" in self._stream
+        self.active = self.enabled
+        if self._cmd == "info":
+            # only `dnf module info` shows `[a]` flag
+            self.active = "[a]" in self._stream
+
+
 class DnfModuleBrief(object):
     """
     An object for the dnf modules listed by ``dnf module list`` command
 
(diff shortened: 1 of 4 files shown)


insights/parsers/dnf_module.py
```diff
@@ -17,22 +17,37 @@
...
+    An object for the dnf module:stream
+
+    Attributes:
+        name (str): Name of the module stream
+        profiles (list): List of profiles of the dnf module:stream
+        summary (str): Summary of the dnf module:stream
+    """
+    def __init__(self, data=None):
+        data = data or {}
+        self.name = data.get('Stream', '')
```
vkrizan, line 31: What do you think about parsing the special bracket flags in this parser?

psegedy: Yeah, I wanted to do it but then I decided to only fix the bug instead of refactoring the whole module

insights/parsers/dnf_module.py
```diff
@@ -17,42 +17,90 @@
...
 class DnfModuleBrief(object):
     """
     An object for the dnf modules listed by ``dnf module list`` command
 
     Attributes:
         name (str): Name of the dnf module
-        stream (str): Stream of the dnf module
-        profiles (list): List of profiles of the dnf module
-        summary (str): Summary of the dnf module
+        streams (list): List of streams of the dnf module
```
vkrizan, line 72: There might be a compatibility issue, when changing from `stream` to `streams`.

insights/parsers/dnf_module.py
```diff
@@ -93,8 +154,24 @@ def parse_content(self, content):
                         trailing_ignore=['Hint:', 'Error:'])
         if not data:
             raise SkipComponent('Nothing need to parse.')
-        for m in data:
-            self.update({m['Name']: DnfModuleBrief(m)})
+
+        modules_streams = {}
+        for ms in data:
+            if ms['Name'] in modules_streams:
+                modules_streams[ms['Name']].add_stream(ms)
```
vkrizan, line 161: Is `Name` key guaranteed to be present?

vkrizan (approved):
Looks great!

insights/parsers/dnf_module.py
psegedy: it is a breaking change which is not compatible at all, previously it was `stream` string, now it is a list

insights/parsers/dnf_module.py
psegedy: yes, it is always printed by `dnf module list`

insights/parsers/dnf_module.py
xiangce: TBH, this PR changes the Interface of the existing parser, that's to say, it would break the rules/services that depend on it. However, as you mentioned, this spec was removed from the collection and was not collected by Insights for a long time, and it fixes some issues which were incorrect. It looks good to me

xiangce (approved):

insights/specs/default.py
```diff
@@ -176,6 +176,7 @@ class DefaultSpecs(Specs):
     dmsetup_status = simple_command("/usr/sbin/dmsetup status")
     dnf_conf = simple_file("/etc/dnf/dnf.conf")
     dnf_modules = glob_file("/etc/dnf/modules.d/*.module")
+    dnf_module_list = simple_command("/usr/bin/dnf module list")
```
xiangce, line 179: Please check the comment regarding this command on the MR to the core-assets repo. Thanks.

xiangce: And I also forget this point, that it's required to add a corresponding item for this `dnf_module_list` to the `insights_archive.py` file as well (to support the old version insights client)

You can use the `mangle` command provided by the insights-core to get the simple_file name of this spec. e.g.
```
# mangle "/usr/bin/dnf module list"
insights_commands/dnf_module_list
```

Thanks.

psegedy: updated default.py to `dnf_module_list = simple_command("/usr/bin/dnf -C --noplugins module list")` and add `dnf_module_list = simple_file("insights_commands/dnf_-C_--noplugins_module_list")` to insights_archive.py

xiangce:
test me

PR #3764, opened 2023-04-25
feat: Add parsers for IPA config files

tiran:
### All Pull Requests:

Check all that apply:

* [X] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [ ] Is this PR to correct an issue?
* [X] Is this PR an enhancement?

### Complete Description of Additions/Changes:

The new parsers in ``insights.parsers.ipa_conf`` can be used to retrieve information if an IPA (FreeIPA, RHEL IdM) installation.

``insights.parsers.ipa_conf.IPAConfig`` parses the IPA client and server config file ``/etc/ipa/default.conf``. The config file contains information such as domain, realm, and server URIs. The parser implements the same fallback logic as IPA's internal config parser.

``insights.parsers.ipa_conf.IPAServerSysRestoreState`` parses IPA server config file ``/var/lib/ipa/sysrestore/sysrestore.state``. It can be used to check, whether a system is an IPA server.

Related to *HMS-20* and *HMS-1285*

PR #3767, opened 2023-04-26
feat: add parser and combiner for IPA

tiran:
### All Pull Requests:

Check all that apply:

``insights.parsers.ipa_conf.IPAConfig`` parses the IPA client and server config file ``/etc/ipa/default.conf``. The config file contains information such as domain, realm, and server URIs. The parser implements the same fallback logic as IPA's internal config parser.

``insights.combiners.ipa.IPA`` combines information from ``IPAConfig``, ``SSSDConfig``, and ``InstalledRpms`` to detect whether a system is an IPA server.

See: HMS-1285

This is an alternative take on PR #3764. It does not have spec and parser for ``/var/lib/ipa/sysrestore/sysrestore.state``. Instead it uses a combiner to check ``IPAConfig``, ``SSSD_Config``, and ``InstalledRpms`` whether a system is an IPA server. In addition it raises a ``ParserException`` when ``/etc/ipa/default.conf`` is invalid.

`/etc/ipa/default.conf` is readable by insights-client
```
# ls -laZ /etc/ipa/default.conf
-rw-r--r--. 1 root root unconfined_u:object_r:etc_t:s0         279 Mar 14 11:20 /etc/ipa/default.conf
```

### DIFF
--- a/insights/combiners/ipa.py
+++ b/insights/combiners/ipa.py
@@ -69,13 +69,13 @@ class IPA(object):
                 "ipa_server_mode", "false"
             )
             if (
-                self._server_rpm
+                self._server_rpm and
                 # all servers are also clients
-                and self.is_client
+                self.is_client and
                 # only servers use LDAPI (LDAP over Unix socket)
-                and self._ipa_conf.ldap_uri.startswith("ldapi://")
+                self._ipa_conf.ldap_uri.startswith("ldapi://") and
                 # SSSD domain must be in server mode
-                and server_mode.lower() == "true"
+                server_mode.lower() == "true"
             ):
                 self._is_server = True
             else:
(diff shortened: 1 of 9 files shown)


insights/combiners/ipa.py
```diff
@@ -0,0 +1,73 @@
...
+        """Is the host an IPA client?"""
+        if not self._client_rpm:
+            return False
+        cfg = self.sssd_domain_config
+        if not cfg.get("id_provider") == "ipa":
+            return False
+        return True
+
+    @property
+    def is_server(self):
```
xiangce, line 59: All other `property`s look good to me, however, this one (and the `is_client`) looks a bit heavy, could you mind introducing one more internal member to do the calculation in advance (e.g. in the `__init__`)?  (Since the insights-core needs to support various py versions, it's not so flexible to allow to use the `cache_property`)

tiran: The `is_client` and `is_server` properties aren't doing any computationally heavy workload. It's just a bunch of boolean checks, dict lookups with string comparison and one `str.startswith()`. If you like I can add manual caches to the properties.

xiangce (approved):

PR #3771, opened 2023-04-27
feat: new option to show hit result only for insights-run

xiangce:
### All Pull Requests:

Check all that apply:

This is for INSGHTCORE-197
- It added a new option `--no-details` for the `insights-run`, 
   only the FAIL/PASS result will be outputted with this option.
   This option only works for the `text` format.


jholecek-rh:
@xiangce Thank you for responding to the user feedback so quickly!

I'm thinking if a new built-in formatter, e.g. `-f text_short`, wouldn't be better than a global CLI argument which is applies only to a specific formatter.  I would reserve CLI  arguments for limiting the scope globally (respected automatically by all formatters) and controlling the amount of debugging information (`-D`, `-t`, `-d` switches). WDYT?

xiangce:
> I'm thinking if a new built-in formatter, e.g. `-f text_short`, wouldn't be better than a global CLI argument which is applies only to a specific formatter. I would reserve CLI arguments for limiting the scope globally (respected automatically by all formatters) and controlling the amount of debugging information (`-D`, `-t`, `-d` switches). WDYT?

Nope, I don't think so. the `--no-details` can also be a global option. it can be reused in the future when other formats e.g. JSON and HTML also want such a `summary` output.

jholecek-rh:
@xiangce Right, I misunderstood how the CLI arguments work. There's probably space for improving clarity, but let's save that for another day. :slightly_smiling_face:

xiangce:
test me

Issue #3799, opened 2023-05-24
Uname padding release section of version seems backwards when missing "el" portion


bostrt:
It seems like the release portion of a version is being padded backwards (when you don't specify `el7`) which results in bad results when comparing versions:

Below, I believe `a` is older than `b` but the comparison reports differently. 

The last command shows the `lv_release` is padding what seems to be backwards with leading zeros. 

```
$ insights --version
insights-core-3.1.25-1
$ insights shell

In [2]: a = Uname.from_uname_str("Linux example 3.10.0-1127.19.1.el7.x86_64 #1 SMP Wed Jun 15 08:55:08 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux")

In [3]: b = Uname.from_kernel("3.10.0-1161")

In [9]: a < b
Out[9]: False

In [10]: a > b
Out[10]: True

In [15]: b
Out[15]: <Uname 'version: 3.10.0; release: 1161; rel_maj: 1161; lv_release: 0.0.0.0.1161'>
```

If I specify `el7` you can see the ordering of the `lv_release` changes and the comparison works as one would expect. However, based on the documentation I doesn't seem like the `el7` should be required: https://github.com/RedHatInsights/insights-core/blob/ea549f1ad52adecbe352a924e614f60d43fb4f10/insights/parsers/uname.py#L28-L36

```
In [20]: b = Uname.from_kernel("3.10.0-1161.el7")

In [21]: b
Out[21]: <Uname 'version: 3.10.0; release: 1161.el7; rel_maj: 1161; lv_release: 1161.0.0.0.el7'>

In [24]: a > b
Out[24]: False

In [25]: a < b
Out[25]: True
```

cc @ryan-blakley @vdymna

PR #3800, opened 2023-05-24
fix: uname pad release without el segment

bostrt:
Changes behavior of Uname parser's `pad_release` such that a missing "el" segment of a release version results in the padded version having version numbers properly arranged.

### All Pull Requests:

Check all that apply:

The current behavior of the Uname parser's `pad_release` function is to unconditionally take the last segment of release version and place it at the end of padded version. This results in padded releases like below when no `el` segment is included (e.g. `el7`). 

- `0.0.0.0.1161`
- `1160.0.0.0.1`

This PR resolves this issue by first checking if the release string to be padded has a period to indicate multiple parts. If not, the release parts are simply joined with _trailing_ zeroes (instead of leading). If the release string _does_ have multiple parts and the last part contains alpha characters then we assume its an `el` part or some other product identifier. The resulting padded release is padded appropriately with the `el` part at the end.

cc @ryan-blakley 

Resolves https://github.com/RedHatInsights/insights-core/issues/3799

### DIFF
--- a/insights/parsers/uname.py
+++ b/insights/parsers/uname.py
@@ -579,7 +579,7 @@ def pad_release(release_to_pad, num_sections=4):
     ``LooseVersion`` comparisons will be correct.
 
     Release versions with less than num_sections will
-    be padded in front of the last section with zeros.
+    padded with zeros where missing versions segments are.
 
     For example ::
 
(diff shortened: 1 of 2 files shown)


ryan-blakley:
@bostrt it looks like there is a flake8 issue.

ryan-blakley (approved):
Looks good to me.

xiangce (approved):

PR #3801, opened 2023-05-25
fix: correctly identify final segment of kernel version in uname

bostrt:
Resolves issue with Uname parser where the last segment of kernel version was always assumed to be the architecture if it didn't contain "el". A check is added to see if the last segment is a digit and if so we assume its not architecture.

Further resolves https://github.com/RedHatInsights/insights-core/issues/3799 and complements PR https://github.com/RedHatInsights/insights-core/pull/3800

### All Pull Requests:

Check all that apply:

When the Uname parser is given a partial kernel version (e.g. `3.10.0-1161.1`) the final segment of the version is incorrectly interpreted as the architecture iff it does not contain the letters `el` (e.g. `3.10.0-1161.1` would have `.1` interpreted as the architecture). 

This PR adds an additional check in `parse_content` to further identify if the last segment of version is just digits or not. If the last segment of version does not contain `el` _and_ not only composed of digits then we assume its an architecture. 

cc @ryan-blakley

### DIFF
--- a/insights/parsers/uname.py
+++ b/insights/parsers/uname.py
@@ -253,7 +253,12 @@ class Uname(CommandParser):
             data['kernel'] = uname_parts[2]
             data['name'] = uname_parts[0]
             data['nodename'] = uname_parts[1]
-        has_arch = "el" not in data['kernel'].split(".")[-1]
+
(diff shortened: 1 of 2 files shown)


ryan-blakley (approved):
Looks good to me.

PR #3822, opened 2023-06-16
Add malware-detection tests that use real yara

mhuth:
- To catch any potential changes in output/behaviour in newer yara releases

### All Pull Requests:

Check all that apply:

With newer versions of yara, I want to be sure that there are no changes in behaviour/output that might affect how the malware detection app uses yara.  This PR adds malware-detection app tests that run when an actual yara binary is detected on the test system.  It will skip the tests if yara isn't installed (which will probably be the case most of the time).  But still, having these extra tests will be useful for detecting changes in behaviour in newer yara versions.

### DIFF
--- a/insights/tests/client/apps/test_malware_detection.py
+++ b/insights/tests/client/apps/test_malware_detection.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


mhuth:
There might be some repetition in the tests now.  The real yara tests are taken from some old unittests I had when I was first getting the malware-detection code into insights-core and I was still using real yara.  The real yara was causing problems so we ended up faking yara but I'm worried that yara might change how it works in future and our fake yara tests wouldn't capture that.  Using real yara tests would.  But if yara isn't installed on the test system, only the fake yara tests are run and the real yara tests are skipped.

Anyway, LMK if you see any problems with this approach.  Good to see the tests work in our automated test environment, so that's encouraging.

mhuth:
I've updated the PR to consolidate some of the repeated tests.  Also, now the tests involving real yara are ignored altogether rather than being skipped if yara is not installed on the system running the tests.  This prevents a whole bunch of 'skipped test' messages from appearing.

xiangce (approved):
T.B.H. the updates of the test file look a bit complex to me, but since all required CI tests are PASSED, it's good for me.

Ps. We'll skip the release of insights this week, please wait for a while, this PR will not be pushed to prod very soon.

PR #3818, opened 2023-06-12
fix: python2.6 pipeline

ahitacat:
### All Pull Requests:

Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [x] Is this PR to correct an issue?
* [ ] Is this PR an enhancement?

### Complete Description of Additions/Changes:

This commit adds:

-  a new package python2.6-gdbm
- noninteractive mode

PR #3823, opened 2023-06-19
fix: Build python26 instead of install

xiangce:
### All Pull Requests:

Check all that apply:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,25 +39,29 @@ jobs:
     container: ubuntu:18.04
 
     steps:
+    - uses: actions/checkout@v3
     - name: Install sudo package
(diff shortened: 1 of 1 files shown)


xiangce:
@ahitacat, Inspired by your PR #3818, I raised this version that builds the py26 from the original source code, please kindly review.

PS. py27 is also removed, I'm trying to fix it in #4017, let's ignore its failure here.

ahitacat (approved):
This looks great, I wanted to build python2.6 this approach is far better than mine :smile_cat: .

I would like to check the possibility of use centOS container instead of building python2.6. That option could save us the building part, and could be more stable during time.

PS: Is sad that py2.7 was also removed. It will have the same approach.

PR #3827, opened 2023-06-21
fix: Using pip for python2.7 instead of python3x

fellipeh:
### All Pull Requests:

Check all that apply:

Since CentOS 7 doesn't have pre-installed python 3.x we need to use this url `https://bootstrap.pypa.io/pip/2.7/get-pip.py` to get the right pip version, to avoid error on docker build process


xiangce (approved):
Thanks for fixing this issue, it works well per my test

PR #3828, opened 2023-06-27
Feat: new spec "/usr/bin/iris list" and its parser 

wushiqinlou:
### All Pull Requests:

Check all that apply:

Create a new parser for command "iris list", which shows the general information of iris instance

### DIFF
--- a/insights/parsers/iris_list.py
+++ b/insights/parsers/iris_list.py
@@ -1,0 +1,59 @@
+"""
+IrisList - Command ``/usr/bin/iris list``
+=========================================
+
+This parser reads the output of ``/usr/bin/iris list``.
+"""
+
+from insights.core import CommandParser
+from insights.core.exceptions import SkipComponent
+from insights.core.plugins import parser
+from insights.parsers import get_active_lines
+from insights.specs import Specs
+
+
+@parser(Specs.docker_info)
+class IrisList(CommandParser, dict):
+    """
+    Parse the output of the ``/usr/bin/iris list`` command.
+
+    Sample Input::
+
+        Configuration 'IRIS'   (default)
+            directory:    /intersystems
+            versionid:    2023.1.0.235.1com
+            datadir:      /intersystems
+            conf file:    iris.cpf  (SuperServer port = 1972, WebServer = 52773)
+            status:       running, since Tue Jun 27 01:55:25 2023
+            state:        ok
+            product:      InterSystems IRIS
+
+    Examples:
+        >>> iris_info['name']
+        'IRIS'
+        >>> iris_info['status']
+        'running, since Tue Jun 27 01:55:25 2023'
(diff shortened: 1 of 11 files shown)


insights/parsers/iris_list.py
xiangce, line 1: how about renaming the file as `iris.py`, in case there are different options need to collect about `iris` in the future, we can wrap them in the same module.

insights/parsers/iris_list.py
```diff
@@ -0,0 +1,59 @@
...
+        'running, since Tue Jun 27 01:55:25 2023'
+        >>> iris_info.is_running
+        True
+    """
+
+    def parse_content(self, content):
+        if not content:
+            raise SkipComponent('No content.')
+
+        for line in get_active_lines(content):
```
xiangce, line 44: `get_active_lines` is not so recommended for command output. I think here we can use the `content` directly. and skip the empty lines, e.g. 
```
if not line.strip():
    continue
```

insights/parsers/iris_list.py
```diff
@@ -0,0 +1,59 @@
...
+
+            if line.startswith('Configuration'):
+                self['name'] = line.split()[1].replace("'", "")
```
xiangce, line 47: I think you may want to use `strip('\'"')` instead of `replace()`

insights/parsers/iris_list.py
```diff
@@ -0,0 +1,59 @@
...
+
+            if ":" in line:
```
xiangce, line 49: Better use `elif` here, or add a `continue` in the above if clause.

insights/parsers/iris_list.py
```diff
@@ -0,0 +1,59 @@
...
+                key, value = line.split(":", 1)
+                self[key.strip()] = value.strip()
+
```
xiangce, line 52: I think it would be better to add a check at the end of the `parse_content`, e.g.
```
if len(self) == 0:
    raise SkipComponent
```

By doing so, the first `conent empty` check can also be ignored.

insights/parsers/iris_list.py
```diff
@@ -0,0 +1,59 @@
...
+    @property
+    def is_running(self):
+        """Return True when the iris instance is running, and False when it is down"""
+        if self.get('status', "").startswith('running'):
```
xiangce, line 56: it's ok to `return self.get('status', "").startswith('running')`  directly.

xiangce (changes requested):
And need to add a doc entry file for this new python module in the `docs/shared_parsers_catalog` folder.

wushiqinlou:
@xiangce - Updated according to your comment, please check.

xiangce (approved):

xiangce:
test me

PR #3832, opened 2023-06-29
Improve parsers/redhat_release to detect CentOS correctly

PaulWay:
### All Pull Requests:

Check all that apply:

* [N] Is this PR an enhancement?

The RedhatRelease parser already does a basic check of the operating system to set flags
for whether this is a RHEL, CentOS, Fedora or unknown system.  We need to test that these
are handled correctly within the RedHatRelease combiner.

Part of this work simplifies the short product code detection in the
RedhatRelease parser, which we now advertise as a property.

Additionally, RHEV 3.5 now correctly detects as a RHEL operating system.


xiangce (commented):
@PaulWay - This change looks good to me, but the update of the combiner will break some tests of the GSS-Rules, since these tests take the `RedHatRelease` for RHEL only. We'd better fix them before merging this PR.

PaulWay:
Good point @xiangce - I'll look through the rules and write up a PR for that.

insights/combiners/redhat_release.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, uname, rh_rel):
             self.major = uname.redhat_release.major
             self.minor = uname.redhat_release.minor
             self.rhel = '{0}.{1}'.format(self.major, self.minor)
-        elif rh_rel and rh_rel.is_rhel:
+        elif rh_rel:
```
xiangce, line 77: Hi @PaulWay - Other than making an MR to the GSS rules, it would be more suitable to keep this `RedHatRelease` combiner for RHEL only, As this combiner is designed for RHEL system only, See the docstring in L40

ryan-blakley: I agree with Xiangce, I wanted to say something about this previously but forgot. I feel like detection for centos, fedora, oracle, etc should be done in the os-release parser not the redha-release parser/combiner. Since os-release is used by every distro, and provides a lot of information that the redhat-release file doesn't provide I feel it's better fitted.

PR #3833, opened 2023-06-29
feat: new Ls Parsers for 'ls' commands

### LINKED ISSUE #3635: FileListing: `.` in file_path of the top dir was converted to slash '/'
Hey @mohitkumarrh - I think this is a bug in the cooperation between the insights core vs the base module `ls_parser`, the insights core used the following `mangle` logic and replaced all of the `/` to `.` as the filename in the insights-archive:
https://github.com/RedHatInsights/insights-core/blob/8656cccc1cbe78e208ac5d1336503b33d447ef99/insights/util/mangle.py#L40
But it didn't record which `.` is a `mangled` dot from the slash.

However, the `FileListing` converted all the `.` back to the `/` and didn't consider whether it's an original `.` in the `file_path`:
https://github.com/RedHatInsights/insights-core/blob/8656cccc1cbe78e208ac5d1336503b33d447ef99/insights/core/__init__.py#L1722

So the path `/var/lib/pcp/pmdas/openmetrics/config.d` in the parser result was stored as `/var/lib/pcp/pmdas/openmetrics/config/d` by mistake.

Since recently, we met a couple of issues in the `ls_parser`, I'm planning to do some enhancement on it. 
Could you please wait for a while?

Sorry about the inconvenience.

_Originally posted by @xiangce in https://github.com/RedHatInsights/insights-core/issues/3620#issuecomment-1354458782_

xiangce:
- See INSGHTCORE-217
- Mark old Ls Parsers as deprecated and will be removed from 3.4.0
- Update the ls_parser to pre-load the `Directory`
- All new `ls_*` specs require a filterable `ls_*_dirs` to collect the listing
   target directories from rules. And a `_non_existing_` dir will be
   appended to the targets, when there is only 1 target being collected.
   By doing so, the #3635 can be fixed.
- fix #3635
   

### All Pull Requests:

Check all that apply:


JoySnow (approved):
@xiangce , the changes are good to me.

ryan-blakley:
@xiangce I apologize for missing this PR last week. But one thing I noticed this morning is the parsers for specs ls_boot, ls_dev, and ls_sys_firmware shouldn't be deprecated since they're still sos_archive specs. I don't know if those specs/parsers are used in gss-rules or not, but I do use the LsBoot parser in my  insights-facts project. Also wouldn't deprecating/removing the old parsers break non-core collection archives in the future?

xiangce:
> @xiangce I apologize for missing this PR last week. But one thing I noticed this morning is the parsers for specs ls_boot, ls_dev, and ls_sys_firmware shouldn't be deprecated since they're still sos_archive specs. I don't know if those specs/parsers are used in gss-rules or not, but I do use the LsBoot parser in my insights-facts project. Also wouldn't deprecating/removing the old parsers break non-core collection archives in the future?

Oh, I didn't expect that either. For sosreports/gss-rules, I think we can keep the existing parsers, or move them from the insights-core to the gss-rules repo. But first of all, I'll remove the deprecation warnings from these parsers used by gss-rules first, and keep these specs in the `sos_archive.py` only (but remove them from the `default.py` and `insights_archive.py`.
For your insights-facts, may I know if it is possible to change to use the new `LSlanR` spec/parser? otherwise, I will also revert the deprecation of this spec too. Please let me know your thoughts.

PR #3849, opened 2023-07-12
chore: discard the deprecation of LsBoot, LsDev and LsSysFirmware

xiangce:
- They are used by gss rules and other apps
   See the latest comment to #3833

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/ls_boot.py
+++ b/insights/parsers/ls_boot.py
@@ -5,40 +5,12 @@ LsBoot - command ``ls -lanR /boot``
 The ``ls -lanR /boot`` command provides information for the listing of the
 ``/boot`` directory.
 
-See the ``FileListing`` class for a more complete description of the
-available features of the class.
+See :class:`insights.parsers.ls.FileListing` for more information.
 
-Sample directory listing::
-
-    /boot:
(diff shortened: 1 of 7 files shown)


insights/parsers/ls_boot.py
```diff
@@ -38,20 +38,19 @@
     True
 """
 
-from .. import FileListing, parser, CommandParser
+from insights import CommandParser, FileListing, parser
```
ryan-blakley, line 41: @xiangce Can you use the new FileListing [function](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/ls.py#L44) that was added for these parsers? If not the deprecation would need to also be removed from the old FileListing [function](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/__init__.py#L1644).

xiangce: @ryan-blakley - Yeap, I missed this. Fixed in the new commit.

Thanks for reviewing.

xiangce:
test me

xiangce:
test me

ryan-blakley (approved):

PR #3856, opened 2023-07-18
fix: Malware fix to handle cert_verify set in conf file

mhuth:
### All Pull Requests:

Check all that apply:

When downloading the malware signatures file from stage, the insights-config.conf file usually has the entry cert_verify = False.  This caused malware-detection to fail to download the signatures file because it wasn't treating False as a string.  This fix treats False as a string and converts it to a boolean before trying to use it. *sigh*


xiangce (approved):
Looks good to me

mhuth:
Thanks @xiangce.  You'll have to merge it too because I don't have permissions to do that

PR #3872, opened 2023-08-04
fix: OSRelease when both os_release and redhat_release avaiable

xiangce:
- Identify non-rhel when one of them is non-rhel

### All Pull Requests:

Check all that apply:

- fix the following issue:
   E.g. when `os_release` is non-rhel, but `redhat_release` is rhel, it will be identified as RHEL incorrectly.


chenlizhong (approved):
Looks good to me

PR #3878, opened 2023-08-09
feat: improve temp directories

ahitacat:
Check all that apply:

This commit improve the use of temporary directories:

- The temporary directory will have a random name: `/var/tmp/insights-client-<random characters>`.
- It checks that the directory is not being in use before using it. If it existed, it will clean that directory and create a new one.
- Remove unused temporary directory and the function that used it: `archive_tmp_directory`.
- Change the cleanup mechanisms to be align with this change.
- Change the tests to be align with this change.
- Refactor the constants used.

### DIFF
--- a/insights/client/archive.py
+++ b/insights/client/archive.py
@@ -2,7 +2,6 @@
 Handle adding files and preparing the archive for upload
 """
 from __future__ import absolute_import
-import glob
 from signal import SIGTERM, signal
 import sys
 import time
@@ -31,7 +30,6 @@ class InsightsArchive(object):
         config          - an InsightsConfig object
         tmp_dir         - a temporary directory in /var/tmp
         archive_dir     - location to collect archive data inside tmp_dir
-        archive_tmp_dir - a temporary directory to write the final archive file
         archive_name    - filename of the archive and archive_dir
         cmd_dir         - insights_commands directory inside archive_dir
         compressor      - tar compression flag to use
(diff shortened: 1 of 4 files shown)


ahitacat:
@ptoscano Can you take a look at this?

insights/client/archive.py
```diff
@@ -253,11 +242,14 @@ def cleanup_previous_archive(self):
...
-        for file in glob.glob(archive_glob):
-            os.path.join('', file)
-            logger.debug("Deleting previous archive %s", file)
-            shutil.rmtree(file, True)
+        try:
+            for directory in os.listdir(constants.insights_tmp_path):
+                directory_path = os.path.join(constants.insights_tmp_path, directory)
+                if directory.startswith(constants.insights_tmp_prefix):
+                    logger.debug("Deleting previous archive %s", directory_path)
+                    shutil.rmtree(directory_path, True)
```
ptoscano, line 250: As I understand, there may be more than one `insights-client` run at the same time; in that case, because of this code there may be this situation:
- `insights-client` 1 starts: it creates its own `/var/tmp/insights-client-FOO` temporary directory
- `insights-client` 2 starts as well: `cleanup_previous_archive()` cleanups all the existing `/var/tmp/insights-client*` stuff including `/var/tmp/insights-client-FOO`
- `insights-client` 1 dies in sadness

What I'd do is removing only the bits older than 24h: this way even slow `insights-client` run will not suddenly get their own temporary directories removed under their feet, yet still cleaning old leftovers

ptoscano (commented):
Generally LGTM (will need @xiangce review as well); I left one note about potential conflicts between instances.

insights/client/archive.py
xiangce: How about adding a restriction that only one instance of "insights-client" is allowed to run at a time?

ahitacat: I am not sure if that is a good idea. If we added so, it will limit the following situations:
1. `insights-client` is running the scheduled collection and the administrator wants to check `insights-client` status. 
2. The administrator want to run `collection` and also `malware`.
3. 
I think that a better option is to allow this kind of situation, that was the first idea (before change the things in the temporary files)

ahitacat:
# Last changes pushed:
- The cleanup mechanism checks if the directories found with the determined prefix are older than 24 hours. Those that are older are removed.
- New tests.

ptoscano:
Thanks for the updates! IMHO LGTM now, I'll leave it to @xiangce for the final review.

xiangce (approved):
It looks good to me too. And I tested the py26 on my local, it's PASSED.

PR #3893, opened 2023-08-24
⚠️ Exclude yum_log from IP obfuscation

Glutexo:
~~⚠️ WARNING: This change could lead to a leak of sensitive data. Do not merge until its safety is confirmed.~~ Update: Although this change may in theory be insecure, there are no real examples of such a leek. See my comment below.

### All Pull Requests:

Check all that apply:

The yum.log file only logs messages on actual package installation, not connection or other repository information. As such, it cannot contain sensitive information including IP addresses. Thus, it’s safe to exclude it from IP address obfuscation, causing SOSCleaner not to treat long version numbers as IP addresses. Packages with long versions will no longer trigger false rule hits in the Platform.

Fixes [ESSNTL-4697](https://issues.redhat.com/browse/ESSNTL-4697)


xiangce (approved):
Although I don't think the newly added tests could reflect the `excluding` of `yum_log`, I believe that adding it to the `exclude_files` works as expected.

Glutexo:
You are right, @xiangce! The test can be put in a separate pull request, it is not related to the code change. The exact list of excluded specs is not tested, thus there will be no test changes in this pull requests.

Glutexo:
I added a warning to the description and converted the PR to a Draft to make sure it’s not merged before we know for sure the patch does not introduce leakage of sensitive data.

xiangce:
Hi @Glutexo , I marked this PR as `ready` according to the comments of INSGHTCORE-253 and will release it in the next release, Please correct me, If I understand incorrectly. Thanks.

Glutexo:
Created a follow-up https://github.com/RedHatInsights/insights-core/pull/3941.

PR #3905, opened 2023-09-04
fix: improve temporary directory

ahitacat:
insights-client was using /var/tmp to store the egg release file.

This commit changes the temporary directory to /var/cache/insights-client.

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/client/constants.py
+++ b/insights/client/constants.py
@@ -79,8 +79,9 @@ class InsightsConstants(object):
     cached_branch_info = os.path.join(default_conf_dir, '.branch_info')
     pidfile = os.path.join(os.sep, 'var', 'run', 'insights-client.pid')
     insights_tmp_path = os.path.join(os.sep, 'var', 'tmp')
+    cache_dir = os.path.join(os.sep, 'var', 'cache', 'insights-client')
     insights_tmp_prefix = 'insights-client'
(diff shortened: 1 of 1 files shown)


ptoscano:
Seems OK; should the new `cache_dir` "constant" be used in `InsightsArchive` instead of its own `self.keep_archive_dir` variable?

ptoscano:
ACK; I'll leave this to @xiangce then for further review.

insights/client/constants.py
```diff
@@ -79,8 +79,9 @@ class InsightsConstants(object):
     cached_branch_info = os.path.join(default_conf_dir, '.branch_info')
     pidfile = os.path.join(os.sep, 'var', 'run', 'insights-client.pid')
     insights_tmp_path = os.path.join(os.sep, 'var', 'tmp')
+    cache_dir = os.path.join(os.sep, 'var', 'cache', 'insights-client')
     insights_tmp_prefix = 'insights-client'
```
xiangce, line 83: This `insights_tmp_prefix` became an unused variable then, I assume that you will remove it or use it in the `cache_dir` in your mentioned update. :)

xiangce (approved):

insights/client/constants.py
ahitacat: oh, no, it is used in the code and it can be re-used in the cache_dir constant :smile:

PR #3920, opened 2023-09-19
fix: use get_dependencies in get_registry_points

xiangce:
- and handle "DefaultSpecs" specially in the method
- see INSGHTCORE-243 for details

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/core/dr.py
+++ b/insights/core/dr.py
@@ -414,7 +414,7 @@ def is_registry_point(component):
     return type(component).__name__ == "RegistryPoint"
 
 
-def get_registry_points(component, module='insights.specs.Specs'):
+def get_registry_points(component, module="insights.specs.Specs"):
     """
     Loop through the dependency graph to identify the corresponding spec registry
     points for the component. This is primarily used by datasources and returns a
@@ -449,7 +449,7 @@ def get_registry_points(component, module='insights.specs.Specs'):
     # Process spec of DefaultSpecs
     spec_name = get_simple_name(component)
     if spec_name:
-        spec_mod = get_component_by_name('{0}.{1}'.format(module, spec_name))
+        spec_mod = get_component_by_name("{0}.{1}".format(module, spec_name))
         reg_points.add(spec_mod) if spec_mod else None
     return reg_points
 
--- a/insights/core/dr.py
+++ b/insights/core/dr.py
@@ -414,7 +414,7 @@ def is_registry_point(component):
     return type(component).__name__ == "RegistryPoint"
 
 
-def get_registry_points(component):
+def get_registry_points(component, module='insights.specs.Specs'):
     """
     Loop through the dependency graph to identify the corresponding spec registry
     points for the component. This is primarily used by datasources and returns a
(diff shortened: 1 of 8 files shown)


insights/core/dr.py
```diff
@@ -428,18 +428,29 @@ def get_registry_points(component):
         (set): A list of the registry points found.
     """
     reg_points = set()
-
+    # Process RegistryPoint spec of Specs
     if is_registry_point(component):
         reg_points.add(component)
-    else:
-        for dep in get_dependents(component):
```
vdymna, line 435: @xiangce yes, looks like you are right, `get_dependents` is moving in the wrong direction in the graph.

insights/core/dr.py
```diff
@@ -414,7 +414,7 @@ def is_registry_point(component):
     return type(component).__name__ == "RegistryPoint"
 
 
-def get_registry_points(component):
+def get_registry_points(component, module='insights.specs.Specs'):
```
vdymna, line 417: @xiangce just a little nitpick, but looks like most of the strings here in the module defined with `"` double quotas and you seem to be defaulting to `'`.

xiangce: @vdymna - yup, we'd better keep the same style in the same module. I updated them. Thanks.

insights/core/dr.py
```diff
@@ -428,18 +428,29 @@ def get_registry_points(component):
...
-                dep_reg_pts = get_registry_points(dep)
+                dep_reg_pts = get_registry_points(dep, module=module)
                 if dep_reg_pts:
                     reg_points.update(dep_reg_pts)
-
+        if reg_points:
+            # Component must have (at least) one RegistryPoint, unless ->
+            return reg_points
+        # it's a spec of DefaultSpecs
+    # Process spec of DefaultSpecs
```
xiangce, line 449: Oops, I think the following process is incorrect, although allows passing a customized `module` to get the wanted module where the "RegistryPoints" are defined, there is no chance in the "dr.run" to modify it :(. It needs more research here.

xiangce: Hi @vdymna and @ryan-blakley - I modified the method again regarding the above issue. For `datasource`s, it needs to iterate  **UP** the `dependents`, but for `combiner/parsers` (high level components), it needs to iterate **DOWN** to the `dependencies`. Please help review it one more time. Thanks.

ryan-blakley (approved):
Looks good to me.

insights/core/dr.py
```diff
@@ -414,7 +414,14 @@ def is_registry_point(component):
     return type(component).__name__ == "RegistryPoint"
 
 
-def get_registry_points(component):
+def is_datasource(component):
```
vdymna, line 417: @xiangce maybe it still make sense to have this function in `dr.py` but I've noticed that `insights/core/plugins.py` has [is_datasource](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/plugins.py#L370) function?

xiangce: @vdymna - Using the `plugins.is_datasource` would cause a `circular import issue`, since the `insights.core.plugins` imported the `insights.core.dr` ([here](https://github.com/RedHatInsights/insights-core/blob/master/insights/core/plugins.py#L37)). And this is also the reason why it's implemented via checking the `comp_type.__name__ == 'datasource'` instead of `type(component) == datasource`.

vdymna: Gotcha, yeah I thought maybe that was the case.

PR #3921, opened 2023-09-20
Download malware rules into /var/lib/insights

mhuth:

This PR renames the malware rules file that is downloaded onto hosts to be scanned and puts it in /var/lib/insights.  This is to make the file more obvious as to its purpose and less likely for it to be flagged as malware itself.


xiangce (approved):
The directory change looks good to me.

PR #3923, opened 2023-09-25
feat: add do_filter option to run_test for rule test

xiangce:
- user can skip the check of filterable specs by
   setting "do_filter=False" when calling run_test 
   if the rule targets do not need to filter
- see INSGHTCORE-254

### All Pull Requests:

Check all that apply:


xiangce:
@chenlizhong and @JoySnow - please help review when free.

Issue #3852, opened 2023-07-13
Allow the gpg keyring to be configured


abadger:
The gpg keyring used to verify that a playbook to be executed by a worker is currently hardcoded to a keyring file that we install as part of insights-core.  This is good for security in production but it would be nice if there was a way to change this value for development and testing of the workers. In that use case, a developer wants to be able to write a script, sign it, have their worker process it as if it came from rhcd, and then verify that the script returned the expected output.  If the only valid key is what we ship, then they would have to have their test scripts signed by the production keyring (which has other, negative security implications as well...) to execute that test.

PR #3930, opened 2023-10-02
RHEL-2480: Do not create /root/.gnupg/ directory by accident

m-horky:
### All Pull Requests:

Check all that apply:

GPG creates a directory `$HOME/.gnupg/` every time it performs some operation. When run under root, but not manually (e.g. via subscription-manager Cockpit plugin), it tries to create and write to this directory, which causes SELinux denials.

This patch utilizes the `--homedir` argument GPG supports in order to move the home directory to a temporary directory for the time of the transaction. After the GPG action is performed, the directory is cleaned up.

You can verify this patch by performing the following actions:

1/ Create your own GPG key pair
  ```
  ~ $ gpg --full-generate-key
  ~ $ gpg --list-secret-keys --keyid-format long
  ~ $ (key identifier is on line `sec`, after the slash)
  ~ $ KEYID=9F2B2565A7CC100E  # (for example)
  ~ $ gpg --armor --export $KEYID > $HOME/.gnupg/my-key.pub.gpg
  ```
2/ Ensure you have created the egg zip file
  ```
  ~/insights-core $ bash build_client_egg.sh
  ```
3/ Sign the egg
  ```
  ~/insights-core $ gpg --detach-sign -u $KEYID --armor insights.zip
  ```
4/ Replace the official GPG key with you own.
5/ Run insights-client without `--no-gpg` or `BYPASS_GPG=True`.

---

Support for additional GPG keys was dropped from this PR. The `crypto.py` still supports it, this way it will be easier to add possibility to load multiple GPG keys later. See #3852 for example use case.

### DIFF
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -3,15 +3,14 @@ import json
 import os
 import logging
 import tempfile
-import shlex
 import shutil
 import sys
 import atexit
-from subprocess import Popen, PIPE
 from requests import ConnectionError
 
 from .. import package_info
 from . import client
+from . import crypto
 from .constants import InsightsConstants as constants
 from .config import InsightsConfig
 from .auto_config import try_auto_configuration
@@ -268,7 +267,7 @@ class InsightsClient(object):
         else:
             logger.debug("Egg update disabled")
 
-    def verify(self, egg_path, gpg_key=constants.pub_gpg_path):
+    def verify(self, egg_path):
         """
             Verifies the GPG signature of the egg.  The signature is assumed to
             be in the same directory as the egg and named the same as the egg
@@ -283,46 +282,42 @@ class InsightsClient(object):
         if egg_path and not os.path.isfile(egg_path):
             the_message = "Provided egg path %s does not exist, cannot verify." % (egg_path)
             logger.debug(the_message)
-            return {'gpg': False,
-                    'stderr': the_message,
-                    'stdout': the_message,
-                    'rc': 1,
-                    'message': the_message}
-        if self.config.gpg and gpg_key and not os.path.isfile(gpg_key):
-            the_message = ("Running in GPG mode but cannot find "
-                            "file %s to verify against." % (gpg_key))
-            logger.debug(the_message)
-            return {'gpg': False,
-                    'stderr': the_message,
-                    'stdout': the_message,
-                    'rc': 1,
-                    'message': the_message}
+            return {
+                'gpg': False,
+                'stderr': the_message,
+                'stdout': the_message,
+                'rc': 1,
+                'message': the_message,
+            }
 
         # if we are running in no_gpg or not gpg mode then return true
         if not self.config.gpg:
-            return {'gpg': True,
-                    'stderr': None,
-                    'stdout': None,
-                    'rc': 0}
+            return {
+                'gpg': True,
+                'stderr': None,
+                'stdout': None,
+                'rc': 0,
+            }
 
         # if a valid egg path and gpg were received do the verification
-        if egg_path and gpg_key:
-            cmd_template = '/usr/bin/gpg --verify --keyring %s %s %s'
-            cmd = cmd_template % (gpg_key, egg_path + '.asc', egg_path)
-            logger.debug(cmd)
-            process = Popen(shlex.split(cmd), stdout=PIPE, stderr=PIPE)
-            stdout, stderr = process.communicate()
-            rc = process.returncode
-            logger.debug("GPG return code: %s" % rc)
-            return {'gpg': True if rc == 0 else False,
-                    'stderr': stderr,
-                    'stdout': stdout,
-                    'rc': rc}
-        else:
-            return {'gpg': False,
-                    'stderr': 'Must specify a valid core and gpg key.',
-                    'stdout': 'Must specify a valid core and gpg key.',
-                    'rc': 1}
+        if egg_path:
+            result = crypto.verify_gpg_signed_file(
+                file=egg_path, signature=egg_path + ".asc",
+                keys=[constants.pub_gpg_path],
+            )
+            return {
+                'gpg': result.ok,
+                'rc': result.return_code,
+                'stdout': result.stdout,
+                'stderr': result.stderr,
+            }
+
+        return {
+            'gpg': False,
+            'stderr': 'Must specify a valid core and gpg key.',
+            'stdout': 'Must specify a valid core and gpg key.',
+            'rc': 1,
+        }
 
     def install(self, new_egg, new_egg_gpg_sig):
         """
--- a/insights/client/collection_rules.py
+++ b/insights/client/collection_rules.py
@@ -6,15 +6,14 @@ import hashlib
 import json
 import logging
 import six
-import shlex
 import os
 import requests
 import yaml
 import stat
 from six.moves import configparser as ConfigParser
 
-from subprocess import Popen, PIPE, STDOUT
 from tempfile import NamedTemporaryFile
+from . import crypto
 from .constants import InsightsConstants as constants
 
 APP_NAME = constants.app_name
@@ -139,22 +138,15 @@ class InsightsUploadConf(object):
         Validate the collection rules
         """
         logger.debug("Verifying GPG signature of Insights configuration")
+
         if sig is None:
             sig = path + ".asc"
-        command = ("/usr/bin/gpg --no-default-keyring "
-                   "--keyring " + constants.pub_gpg_path +
-                   " --verify " + sig + " " + path)
-        if not six.PY3:
-            command = command.encode('utf-8', 'ignore')
-        args = shlex.split(command)
-        logger.debug("Executing: %s", args)
-        proc = Popen(
-            args, shell=False, stdout=PIPE, stderr=STDOUT, close_fds=True)
-        stdout, stderr = proc.communicate()
-        logger.debug("STDOUT: %s", stdout)
-        logger.debug("STDERR: %s", stderr)
-        logger.debug("Status: %s", proc.returncode)
-        if proc.returncode:
+
+        result = crypto.verify_gpg_signed_file(
+            file=path, signature=sig,
+            keys=[constants.pub_gpg_path],
+        )
+        if not result.ok:
             logger.error("ERROR: Unable to validate GPG signature: %s", path)
             return False
         else:
--- a/insights/client/crypto.py
+++ b/insights/client/crypto.py
@@ -1,0 +1,206 @@
+import errno
+import os.path
+import logging
+import shutil
+import tempfile
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+
+class GPGCommandResult(object):
+    """Output of a GPGCommand.
+
+    Attributes:
+        ok (bool): Result of an operation.
+        return_code (int): Return code of the operation.
+        stdout (str): Standard output of the command.
+        stderr (str): Standard error of the command.
+        _command (GPGCommand | None):
+            An optional reference to the GPGCommand object that created the result
+    """
+    def __init__(self, ok, return_code, stdout, stderr, command):
+        self.ok = ok
+        self.return_code = return_code
+        self.stdout = stdout
+        self.stderr = stderr
+        self._command = command
+
+    def __str__(self):
+        return (
+            "<{cls} ok={ok} return_code={code} stdout={out} stderr={err}>"
+        ).format(
+            cls=self.__class__.__name__,
+            ok=self.ok, code=self.return_code,
+            out=self.stdout, err=self.stderr,
+        )
+
+
+class GPGCommand(object):
+    """GPG command run in a temporary environment.
+
+    Attributes:
+        command (list(str)): The command to be executed.
+        keys (list(str)): List of paths to GPG public keys to check against.
+
+        _home (str): Path to the temporary GPG home directory.
+        _raw_command (list(str)): The last invoked command.
+    """
+
+    def __init__(self, command, keys):
+        self.command = command
+        self.keys = keys
+
+        self._home = None
+        self._raw_command = None
+
+    def __str__(self):
+        return "<{cls} _home={home} _raw_command={raw}>".format(
+            cls=self.__class__.__name__,
+            home=self._home,
+            raw=self._raw_command
+        )
+
+    def _setup(self):
+        """Prepare GPG environment.
+
+        Returns (bool):
+            `True` if public GPG keys were imported into temporary environment,
+            `False` if there was an error.
+        """
+        self._home = tempfile.mkdtemp()
+
+        logger.debug("setting up gpg in the temporary environment")
+        for key in self.keys:
+            result = self._run(["--import", key])
+            if not result.ok:
+                logger.debug("failed to import key '{key}': {result}".format(
+                    key=key, result=result
+                ))
+                return False
+
+        return True
+
+    def _cleanup(self):
+        """Clean up GPG environment."""
+        # Try to delete the temporary directory GPG used. As discovered by the
+        # convert2rhel team, we need to handle race conditions:
+        # https://github.com/oamg/convert2rhel/blob/23aadbf0df58c79a8910847d345fcd4092f4656f/convert2rhel/utils.py#L867-L893
+
+        # GPG writes a temporary socket file for the gpg-agent into the home
+        # directory. Sometimes it removes the socket file after `rmtree()` has
+        # determined it should be deleted but before the actual deletion occurs.
+        # This will cause a FileNotFoundError/OSError.
+        # When we encounter that, try to run `rmtree()` again.
+        for _ in range(0, 5):
+            try:
+                shutil.rmtree(self._home)
+            except OSError as exc:
+                if exc.errno == errno.ENOENT:
+                    # We are trying to remove a file that has already been
+                    # removed by gpg-agent itself. Ignore it.
+                    continue
+                # Some other error has happened, let it bubble up.
+                raise
+            else:
+                # We have successfully removed everything. This `break` will
+                # prevent the log statement below from being run.
+                break
+        else:
+            # We called `rmtree()` five times and it failed each time. We cannot
+            # do more without knowing more.
+            logger.debug(
+                "could not clean up temporary gpg directory "
+                "'{path}'".format(path=self._home)
+            )
+
+    def _run(self, command):
+        """Run the actual command.
+
+        Returns (CommandResult): The result of the shell command.
+        """
+        self._raw_command = ["/usr/bin/gpg", "--homedir", self._home] + command
+        process = subprocess.Popen(
+            self._raw_command,
+            stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+        )
+        stdout, stderr = process.communicate()
+
+        result = GPGCommandResult(
+            ok=process.returncode == 0,
+            return_code=process.returncode,
+            stdout=stdout.decode("utf-8"),
+            stderr=stderr.decode("utf-8"),
+            command=self,
+        )
+
+        if result.ok:
+            logger.debug("gpg command {command}: ok".format(command=command))
+        else:
+            logger.debug(
+                "gpg command {command} returned non-zero code: {result}".format(
+                    command=command, result=result,
+                )
+            )
+
+        return result
+
+    def evaluate(self):
+        """Run the command.
+
+        Returns (CommandResult): The result of the shell command.
+        """
+        try:
+            ok = self._setup()
+            if not ok:
+                return
+
+            logger.debug("running gpg in the temporary environment")
+            return self._run(self.command)
+        finally:
+            self._cleanup()
+
+
+def verify_gpg_signed_file(file, signature, keys):
+    """
+    Verify a file that was signed using GPG.
+
+    Parameters:
+        file (str): A path to the signed file.
+        signature (str): A path to the detached signature.
+        keys (list(str)):
+            List of paths to GPG public keys on the filesystem to check against.
+
+    Returns (CommandResult): Evaluated GPG command.
+    """
+    if not os.path.isfile(file):
+        logger.debug(
+            "cannot verify signature of '{file}', file does not exist".format(
+                file=file
+            )
+        )
+        raise OSError(errno.ENOENT, "File '{file}' not found".format(file=file))
+
+    if not os.path.isfile(signature):
+        logger.debug((
+            "cannot verify signature of '{file}', "
+            "signature '{signature}' does not exist"
+        ).format(file=file, signature=signature))
+        raise OSError(
+            errno.ENOENT,
+            "Signature '{sig}' of file '{file}' not found.".format(
+                sig=signature, file=file
+            )
+        )
+
+    gpg = GPGCommand(["--verify", signature, file], keys)
+    logger.debug("starting gpg verification process for '{file}'".format(file=file))
+
+    result = gpg.evaluate()
+
+    if result.ok:
+        logger.debug("signature verification of '{file}' passed".format(file=file))
+    else:
+        logger.debug("signature verification of '{file}' failed".format(file=file))
+
+    return result
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,13 @@
+[extend]
+useDefault = true
+
+
+[allowlist]
+description = "Repository-specific configuration"
+
+paths = [
+    # Tests for `crypto.py` contain public and private GPG keypair. They were
+    # generated specifically for this usecase. The owner of the key is:
+    # insights-core (Signing key for unit testing) <insights-core@example.org>
+    "insights/tests/client/test_crypto.py",
+]
--- a/insights/tests/client/test_crypto.py
+++ b/insights/tests/client/test_crypto.py
@@ -1,0 +1,187 @@
+import os.path
+import shutil
+import subprocess
+import tempfile
+
+from insights.client import crypto
+
+
+# GPG2 exports keys in a way gpg1.4 cannot read. Since gpg1.4 is present in the
+# Ubuntu image that is used in CI to test Python 2.6, we have to use the old
+# version, so it is recognised by all CI branches.
+#
+# $ gpg --gen-key            # gpg1.x
+# $ gpg --full-generate-key  # gpg2.x
+# $ gpg --list-secret-keys --keyid-format long
+# $ KEYID=xxx (line `sec`, after the `/` character)
+# $ gpg --armor --export $KEYID
+# $ gpg --armor --export-secret-keys $KEYID
+
+GPG_PUBLIC_KEY = """-----BEGIN PGP PUBLIC KEY BLOCK-----
+Version: GnuPG v1
+
+mQENBGUlPEcBCACXhinLd4KiyQ9CWXM+gpgo0HMuTdESTlVVDYRjm/ebazephXq7
+0hhAhXallAQJkaXFPuLETumwFYPx60agUDWsUie8gk3TLE+ejuTxoda0Yo6rehsD
+zds1ptHQLzar00SFUlfiJXnMaobXLeNjDSglkynQC4uQODouDa1jkSRfNbDJOJYd
+IBbPBIaJ3PlNrdVutPcx4yIZM1siqMZ9k0g2iYPWyp0ceuP0jpCTPc5TRRP5pLuz
+INRJzhfJ+oPtbFt8Y4s1xEZ0Kfy0sA8Awk3VJqzoCU3ILBU5cGquQDdNggxTRSPV
+X5Z4nW2zjN9lXAkwr29NL1rA1jjLt+c5upYZABEBAAG0SGluc2lnaHRzLWNvcmUg
+KFNpZ25pbmcga2V5IGZvciB1bml0IHRlc3RpbmcpIDxpbnNpZ2h0cy1jb3JlQGV4
+YW1wbGUub3JnPokBOAQTAQIAIgUCZSU8RwIbAwYLCQgHAwIGFQgCCQoLBBYCAwEC
+HgECF4AACgkQn+6Vu126uXAHdgf/Q8a7Jruhyn+EDLL94gAc6kXubvVArVe9Rdpo
+HwG7cj8wUa/7zd7FUcYJuz6bbebgmmlRwFf1CeodGURFpfwf1dkiKV5QqobeXhRp
+srrkeLlMR8bZVFSCvU+oOETpJfMo6feDI/tQrOcIpxrd/cEu2XSQ1JBM/+8NbQiU
+Ma1cWHOmQJ1OD7jWOllUq5hDs1vtUzPORUsYe1V1Dcx89gdlhfc1cc30yoLzDNqu
+0Abn34CAthUkr0sWplzltXOY+Za1kOmQVlaLQ0W3tq8BDi79bdHpTtD/g9QO++Nc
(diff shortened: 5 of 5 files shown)


abadger:
I'm not an insights committer but from past experience working with the gpgme library and, more recently, signature checking in convert2rhel, I support using `--homedir`.  It's the only way that gnupg provides for applications to isolate their signature checking from the user's own gpg setup.

I'll make some comments about the usage of gpg here but the basic questions of whether insights will do this and whether the design fits into the code base will need to be done by someone on the insights team.

insights/client/crypto.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if process.returncode > 0 and stdout:
+        logger.debug("stdout is not empty:\n%s", stdout.strip())
+    if process.returncode > 0 and stderr:
+        logger.debug("stderr is not empty:\n%s", stderr.strip())
+
+    return stdout, stderr, process.returncode
+
+
+def verify_gpg_signed_file(file, signature):
```
abadger, line 42: I would still pass in the key to use here.  It makes it easier to test and makes the function more flexible in case the code needs to use different keys in the future.

insights/client/crypto.py
```diff
@@ -0,0 +1,122 @@
...
+        home (str): Path to GPG's home directory.
+        command (str): Actual arguments passed to the GPG binary.
+
+    Returns (str, str, int):
+        Standard output, standard error and the return code.
+    """
+    command = (
+        "/usr/bin/gpg --homedir {home} {command}"
+    ).format(home=home, command=command)
+    process = subprocess.Popen(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
abadger, line 27: It's always better to use a list to begin with rather than trying to split a string when calling subprocess.  That way things like spaces in filenames are treated correctly.
```suggestion
    command = (
        "/usr/bin/gpg", "--homedir", home, command
    )
    process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```

insights/client/crypto.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    if not os.path.isfile(file):
+        message = "cannot verify signature of '{file}', file does not exist".format(file=file)
+        logger.debug(message)
+        return {
+            'gpg': False,
+            'stdout': '',
+            'stderr': '',
+            'rc': 1,
+        }
```
abadger, line 64: * Unless it's a standard throughout the insights code base not to use exceptions for errors, it would be better to raise an exception for errors like this.
* I only see one caller that is using the stdout, stderr, and rc. (in `client/__init__.py:InsightsClient.verify()`  I would check whether it's okay to only use those values for error cases (where they can be attributes of the exception) and then change the API of this function to return a simple boolean True (gpg ran and the signature was valid) |False (gpg ran and the signature was invalid).
 * In fact, a quick look at `client/__init__.py` seems to show a bug: https://github.com/RedHatInsights/insights-core/pull/3930/files#diff-5ef20b67881931ced25261142697d774de208b9c1c2571da00294e3a84a76038R263  is expecting a boolean True or False already so `verify()` in this if statement: `if (egg_paths and self.verify(egg_paths['core'])['gpg']):` will always be truthy in both the current code and your revision.

insights/client/crypto.py
```diff
@@ -0,0 +1,122 @@
...
+        Standard output, standard error and the return code.
+    """
+    command = (
+        "/usr/bin/gpg --homedir {home} {command}"
+    ).format(home=home, command=command)
+    process = subprocess.Popen(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    raw_stdout, raw_stderr = process.communicate()
+    stdout, stderr = raw_stdout.decode("utf-8"), raw_stderr.decode("utf-8")
+
+    logger.debug(command)
```
abadger, line 31:
```suggestion
    base = [
        "/usr/bin/gpg", "--homedir", home
    ]
    base.extend(command)
    command = base
    
    process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    raw_stdout, raw_stderr = process.communicate()
    stdout, stderr = raw_stdout.decode("utf-8"), raw_stderr.decode("utf-8")

    logger.debug(" ".join(command))
```

insights/client/crypto.py
```diff
@@ -0,0 +1,122 @@
...
+            'gpg': False,
+            'stdout': '',
+            'stderr': message,
+            'rc': 1,
+        }
+
+    keys = [constants.pub_gpg_path]
+    _extra_key = os.getenv("EXTRA_GPG_KEY")
+    if _extra_key:
+        keys.append(_extra_key)
```
abadger, line 81: I would be hesitant to add extra keys here.  The keys to use are a security consideration so adding additional keys from a generically named environment variable in the "library" functions doesn't seem right.  I think it would make more sense to integrate this into whatever is setting `constants.pub_gpg_path` (if you're treating `constants` as a type of application context) or into configuration or other command line parsing so that the determination of keys is all done in a single, logical location.

Even if you decide not to do it there, I think this belongs somewhere higher in the call chain rather than in this function so that this function doesn't need to know anything about the assumptions of the application or caller.

insights/client/crypto.py
```diff
@@ -0,0 +1,122 @@
...
+            ).format(file=file, key=key)
+            logger.debug(message)
+            return {
+                'gpg': False,
+                'stdout': '',
+                'stderr': message,
+                'rc': 1,
+            }
+
+    home = tempfile.mkdtemp()
```
abadger, line 96: If the minimum version of Python is high enough, you can use `NamedTemporaryDirectory` as a context manager to automatically delete this directory: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

If not, you should probably use a `try:` `finally:` with this directory and call `shutil.rmtree()` from the finally block.  Note that I've had issues with gpg creating its own temporary files in its home directory and then `shutil.rmtree()` ends up racing with gpg's own deletion of those files (resulting in `shutil.rmtree()` erroring out because the file it was going to delete has already been deleted by gpg) so you may need to protect against that here as well.

insights/client/crypto.py
```diff
@@ -0,0 +1,122 @@
...
+
+    logger.debug("importing the public keys into temporary environment")
+    for key in keys:
+        stdout, stderr, rc = _call_gpg(home, "--import {key}".format(key=key))
```
abadger, line 100: To go along with the change to how command is parsed in the `_call_gpg()` function:
```suggestion
        stdout, stderr, rc = _call_gpg(home, ["--import",key])
```

insights/client/crypto.py
```diff
@@ -0,0 +1,122 @@
...
+            return {
+                'gpg': False,
+                'stdout': stdout,
+                'stderr': stderr,
+                'rc': rc,
+            }
+
+    logger.debug("verifying signature of '{file}'".format(file=file))
+    stdout, stderr, rc = _call_gpg(
+        home, "--verify {signature} {file}".format(signature=signature, file=file)
```
abadger, line 113:
```suggestion
        home, ["--verify", signature, file]
```

insights/client/crypto.py
abadger: Pino says that this code has to be compatible with python-2.6+ so the `try:` `finally:` is the way to go here.  

For reference, here's my proposal for a similar `try:` `finally:` in convert2rhel code that tries to deal with the race condition I mentioned: https://github.com/oamg/convert2rhel/pull/922/files#diff-ed1ca2e4a8b27f1bee13d608562d4e94eceb2972790eccf4d84f366b750f221d  (Note: It uses a `for-else` in addition to the more common `except-else`)

Glutexo:
A little quick note for the steps to reproduce:

Key generation fails if _pinentry_ is not installed.

```sh
$ gpg --full-generate-key
…
gpg: agent_genkey failed: No pinentry
Key generation failed: No pinentry
```

This can be mitigated by, obviously, installing _pinentry_.

```sh
# dnf install pinentry
$ gpg --full-generate-key 
…
public and secret key created and signed.
…
```

Glutexo:
The public key that Insights Client uses to verify the egg by default, _/etc/insights-client/redhattools.pub.gpg_ is in the binary OpenPGP format. If you use a GPG key in an armored ASCII format, GPG verification fails:

```
WARNING: GPG verification failed.  Not loading egg: insights.zip
```

To fix this, remove the `--armor` argument from the `gpg --export` command:

```sh
$ gpg --export $KEYID > $HOME/.gnupg/my-key.pub.gpg
```

m-horky:
> The public key that Insights Client uses to verify the egg by default, _/etc/insights-client/redhattools.pub.gpg_ is in the binary OpenPGP format. If you use a GPG key in an armored ASCII format, GPG verification fails:
> 
> ```
> WARNING: GPG verification failed.  Not loading egg: insights.zip
> ```

@Glutexo On which system did  you get this? It worked fine for me.

Glutexo:
I wanted to verify that I signed the egg with my own key correctly, so I can be sure that the changes from this PR don’t interfere. There are two ways to do so that I want to share here:

* Change the hard-coded [InsightsConstants.pub_gpg_path](https://github.com/RedHatInsights/insights-core/blob/a9d4c14e995f88af6d3594c81420ddc8d0ff4153/insights/client/constants.py#L62C8-L62C8). This requires rebuilding the egg.
* Replace the public key in _/etc/insights-client/redhattools.pub.gpg_.

The latter is easier, because it doesn’t require rebuilding the egg. Also because the absolute path is composed from a base path and a file name, it is less trivial to point the constant to your key. Because the public key is in an _insights-client_ folder, I’d assume no other software uses it and thus nothing other than the Insights Client should break.

Make sure you backup the original _redhattools.pub.gpg_. Note that either way of replacing the key makes the Core fail when verifying the official (downloaded and bundled) eggs signed with Red Hat’s real key.

Glutexo:
I’d like to note that you can omit the `--keyid-format long` argument for `gpg --export` and use the long key printed on the second line of the _sec_ section.

```
$ gpg --list-secret-keys
/home/insights/.gnupg/pubring.kbx
---------------------------------
sec   rsa3072 2023-11-06 [SC]
      BAF1C340B5DEA32A392D8CA8CA7F5625928C2F5D  # ← This is the long key.
uid           [ultimate] Štěpán Tomsa (Insights Client development test) <stomsa@redhat.com>
ssb   rsa3072 2023-11-06 [E]
```

In this case, it is _BAF1C340B5DEA32A392D8CA8CA7F5625928C2F5D_.

Glutexo:
@m-horky It’s a plain instalation of RHEL 9.2 with GPG (2.3.3), libcrypt (1.10.0) and Insights Client (3.1.7) from the installation DVD. See the commands below: if I export the key with `--armor`, GPG verification fails; it succeeds if I don’t use the argument.

```sh
$ ls -a /etc/insights-client
…
lrwxrwxrwx. 1 root root      49 Nov  7 00:26 redhattools.pub.gpg -> /home/insights/gnupg/insights-client-test.pub.gpg
…
$ gpg --list-secret-keys
…
sec   rsa3072 2023-11-06 [SC]
      BAF1C340B5DEA32A392D8CA8CA7F5625928C2F5D
…
$  gpg --armor --export BAF1C340B5DEA32A392D8CA8CA7F56
25928C2F5D > gnupg/insights-client-test.pub.gpg
$ sudo EGG=insights-core/insights.zip BYPASS_GPG=True insights-client --offline
WARNING: GPG verification failed.  Not loading egg: insights.zip
WARNING: GPG verification failed.  Not loading egg: /etc/insights-client/rpm.egg
$ gpg --export BAF1C340B5DEA32A392D8CA8CA7F56
25928C2F5D > gnupg/insights-client-test.pub.gpg
$ sudo EGG=insights-core/insights.zip BYPASS_GPG=True insights-client --offline
Starting to collect Insights data for localhost.localdomain
…
```

Glutexo (changes requested):
I spotted some questionable parts, that I think need to be fixed or at least defended. Because of that I chose “Request change”.

What I what to challenge, too, is that this pull request introduces quite more changes that could be reviewed and merged separately:

* the actual `--homedir` argument for _gnupg_;
* the new `EXTRA_GPG_KEY` environment variable;
* the new encapsulating _crypto_ module;
* the new tests.

Only the first point actually fixes the original bug. Although I like the other changes, they are more of a refactoring. As I [point out](https://github.com/RedHatInsights/insights-core/pull/3930#issuecomment-1803706631) in one of my previous comments, it is possible to use an extra key even without adding explicit support for one.

insights/client/__init__.py
```diff
@@ -306,23 +296,13 @@ def verify(self, egg_path, gpg_key=constants.pub_gpg_path):
...
-                    'stderr': stderr,
-                    'stdout': stdout,
-                    'rc': rc}
-        else:
-            return {'gpg': False,
-                    'stderr': 'Must specify a valid core and gpg key.',
-                    'stdout': 'Must specify a valid core and gpg key.',
-                    'rc': 1}
+        if egg_path:
+            return crypto.verify_gpg_signed_file(file=egg_path, signature=egg_path + ".asc")
```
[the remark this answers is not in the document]
m-horky: You can filter for specific commits in the **Files changed** tab. I prefer not introducing dumb merge conflicts when the work can be expressed as several consecutive (yet self-contained) changes.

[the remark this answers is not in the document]
m-horky: It felt small enough. I'll split it into smaller commits.

Glutexo (changes requested):

insights/client/__init__.py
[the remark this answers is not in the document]
Glutexo: I would rather prefer multiple smaller pull requests, not because I want to nitpick, but for the pure sake of practicality: the diffs are clearer and it is easier to review the patch. Also, when there are multiple changes, having them in separate PRs allows not only to comment on them separately, but also to merge the indepentent parts faster. Yes, there may be merge conflicts, but that’s nothing to avoid or be afraid of.

m-horky:
I've tried to significantly simplify the code. It was growing so much I will have to follow-up in later PR to add support for the extra PGP keys (with further fixes).

insights/client/crypto.py
```diff
@@ -0,0 +1,157 @@
...
+                sig=signature, file=file
+            )
+        )
+
+    gpg = GPG(["--verify", signature, file], keys)
+    logger.debug("starting gpg verification process for '{file}'".format(file=file))
+
+    gpg.evaluate()
+
+    if gpg.ok:
```
abadger, line 152: A code cleanliness note: `subprocess.Popen` follows the model that you are using here where you do some setup, run a method, and then read the attributes from the object but it's not a very good API idiom to emulate because it's hard to get right, not very flexible, and can lead to bugs down the line.  As an example, if you use `subprocess.Popen.communicate()` twice, it will only run the program once.  The second run will just return cached return information.  On the other hand, your implementation will actually run the program again.  This is a subtle difference that can lead to bugs if future programmers are not careful because the attributes that hold the return values could be modified in that case.

What you're using this for could be done by a function which just returns the values instead of setting them as attributes.  I think the reason you used an object was to separate the generic use of  `--homedir` from the specific command that you want to use.  Some better idioms that python provides for that:

* The main `gpg_run(["--verify", signature, file], key)` simply calls some helper functions for the generic parts.
* A Callable object where you can customize some behaviour of the function when you instantiate the Callable and then it acts as a normal function when you call it.  This is probably closest to the model you have now.  The big difference is that it leads you to return the values the function generates instead of placing them into object attributes.
* A decorator:  You can encapsulate the generic parts into a decorator which sets up the homedir and adds it to the command passed in, runs the function with the modified command, and then cleans up afterwards.

I just thought I'd point this out since copying `subprocess.Popen's` API model was a mistake I sometimes made until I thought about why it wasn't common to see that idiom in other places and realized what the problems were.

m-horky: Yeah. I seemed like an OK idea at first, but looking back a day later it already seemed confusing :) I've split it into two separate objects. I didn't do it at first because it doesn't seem like a common things that's done in this repo, but it improved readability in a significant way.

insights/client/crypto.py
```diff
@@ -0,0 +1,157 @@
...
+            self._run(["--import", key])
+
+            if not self.ok:
+                logger.debug("failed to import key {}".format(key))
+                return False
+
+        return True
+
+    def _cleanup(self):
+        shutil.rmtree(self._home)
```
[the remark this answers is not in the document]
m-horky: Thanks for linking that.

Glutexo (changes requested):

insights/client/__init__.py
```diff
@@ -283,46 +282,42 @@ def verify(self, egg_path, gpg_key=constants.pub_gpg_path):
...
-                    'rc': rc}
-        else:
-            return {'gpg': False,
-                    'stderr': 'Must specify a valid core and gpg key.',
-                    'stdout': 'Must specify a valid core and gpg key.',
-                    'rc': 1}
+        if egg_path:
+            result = crypto.verify_gpg_signed_file(
+                file=egg_path, signature=egg_path + ".asc",
+                keys=[constants.pub_gpg_path],
```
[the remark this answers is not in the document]
m-horky: How's that strange? The placement makes sense to me here.

Previously we agreed to pull the GPG key management out of the crypto module. Since we know we may want to support multiple GPG keys, I think the current approach fully describes our current and future needs. Putting the keys into function signature would be ugly; especially when doing such an important stuff as signature verification is.

In my eyes, the `crypto.py` does self-contained management of cryptographic verification and should be treated as such; I dislike putting more stuff into generic files like `utilities.py`. Especially since it is used on two different places it makes sense to be put into its own module.

And in general, we shouldn't be optimizing for _amount_ of changes. The PR has two self-contained changes that do a single thing, as well as the code itself does. It doesn't make sense to make half-baked solution only to change it in the next PR. The first commit prepares the grounds, without affecting the broken functionality. The second commit fixes the wrong behavior by using the new tested solution. That's exactly what you just described.

m-horky:
FYI, you can limit the view to only one commit. There's a filter in the **Files changed** tab, or you can directly click through the commits on the **Commits** tab, they will take you to the same place. You can still comment on them as in the default view in which they are all merged into one big diff :) 
image

Glutexo:
Wow! I didn’t know that. Thank you! ❤️

Glutexo (commented):
I am thinking whether I like to have two separate commits for

- introducing the  _crypto_ module (https://github.com/RedHatInsights/insights-core/pull/3930/commits/a9c7a9709a8e32c031e2abc83c9c0896f167de13)
- and making it being actually used (https://github.com/RedHatInsights/insights-core/pull/3930/commits/05c49b98caad98c7293fb6541ae682e3b286461d).

If there was a separate pull request for the refactoring, which I would strongly prefer, the two commits are separated exaclty as they should be. Here, however, having more separate changes in a single PR, I feel slight temptation to use the commits to group changes that form a mergeable patch.

I learned how to filter changes by commits and it is possible to select a range of commits, too. The commit subjects are clear enough to do the filtering easily, so there is no need to change. Just sharing my thoughts out loud.

Glutexo (changes requested):

insights/client/crypto.py
```diff
@@ -0,0 +1,205 @@
...
+        self._command = None
+
+    def __str__(self):
+        return (
+            "<{cls} ok={ok} return_code={code} stdout={out} stderr={err}>"
+        ).format(
+            cls=self.__class__.__name__,
+            ok=self.ok, code=self.return_code,
+            out=self.stdout, err=self.stderr,
+        )
```
Glutexo, line 35: This looks like a [dataclass](https://docs.python.org/3/library/dataclasses.html). Dataclasses were introduced in Python 3.7 and Insights Client needs to support Python 2.6, but there are _namedtuples_ that do exactly what this class implements.

```suggestion
from collections import namedtuple
CommandResult = namedtuple("CommandResult", ["ok", "return_code", "stdout", "stderr", "command"])
```

I still doubt, however, that this abstraction is actually needed. Although the structure looks nice and clean, it increases complexity. It may be enough to just work with native Python shell command handling.

Moreover, it looks pretty low-level for something that should live in this specifict _crypto_ module. There are other shell command calls in the codebase that may make use of this structure, if we wanted to keep it.

insights/client/crypto.py
```diff
@@ -0,0 +1,205 @@
...
+        stdout (str): Standard output of the command.
+        stderr (str): Standard error of the command.
+        _command: An optional reference to the object that created the result
+    """
+    def __init__(self, ok, return_code, stdout, stderr):
+        self.ok = ok
+        self.return_code = return_code
+        self.stdout = stdout
+        self.stderr = stderr
+        self._command = None
```
Glutexo, line 26: This attribute that is not set by a constructor argument, but rather by an assignment on an existing object is something that a [namedtuple](https://docs.python.org/2.6/library/collections.html#collections.namedtuple) can’t do. A [dataclass](https://docs.python.org/3/library/dataclasses.html) would, but we can’t use it. I’d rarely see mutating existing objects as a good practice, and doing so only not to have an ommitable value doesn’t seem like something that’d deserve an exception.

insights/client/collection_rules.py
```diff
@@ -139,22 +138,15 @@ def validate_gpg_sig(self, path, sig=None):
...
-            args, shell=False, stdout=PIPE, stderr=STDOUT, close_fds=True)
-        stdout, stderr = proc.communicate()
-        logger.debug("STDOUT: %s", stdout)
-        logger.debug("STDERR: %s", stderr)
-        logger.debug("Status: %s", proc.returncode)
-        if proc.returncode:
+
+        result = crypto.verify_gpg_signed_file(
+            file=path, signature=sig,
+            keys=[constants.pub_gpg_path],
```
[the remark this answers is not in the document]
m-horky: The "infrastructure" to load multiple keys already exists in the crypto module. This will make adding the support for extra GPG keys much easier. I don't see what the difference between a single path and a list of paths is.

insights/client/crypto.py
m-horky: I wouldn't call this a mutation, I understand it as an extended context of the command. It is a private attribute, which implicitly assumes you know what you are doing.
But sure, I've altered the initializer to take a command argument as well.

insights/client/crypto.py
m-horky: I wouldn't call this an increased complexity. It just wraps the data, instead of passing them around in an anonymous dictionary.
I've renamed it to GPGCommandResult to make its existence more clear.

insights/client/__init__.py
```diff
@@ -268,7 +267,7 @@ def update(self):
         else:
             logger.debug("Egg update disabled")
 
-    def verify(self, egg_path, gpg_key=constants.pub_gpg_path):
+    def verify(self, egg_path):
```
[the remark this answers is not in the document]
Glutexo: The most correct approach would then be:

```suggestion
    def verify(self, egg_path, gpg_key):
```

But that would require moving the default value to the calling module, increasing the amount of changes. And since I believe we **should** aim for the smallest possible amount, I’d be most in favor of keeping the original code:

```suggestion
    def verify(self, egg_path, gpg_key=constants.pub_gpg_path):
```

This would increase an inconsistency with the _validate_gpg_sig_ function that doesn’t take a key argument, and references _constants.pub_gpg_path_ in its body. I am slighlty inclined towards not changing the signature, as it would feel more as a refactoring in order to increase consistency.

I can, however, live with the argument being moved out. Our goal is to have the code merged, not to endlessly discuss every tiny bit.

insights/client/crypto.py
Glutexo: _GPGCommandResult_ is definitely better than just plain _CommandResult_. Crypto ≠ GPG, so the module name is not enough and the GPG prefix is great. 👍🏻

I am not sure about the capitalization: whether to use _GPGCommandResult_ or _GpgCommandResult_. I researched on the matter and there is no consensus. My personal taste would be _Gpg_, but lot of big projects use _GPG_. Let’s find out how acronyms are capitalized in the existing parts of the Client and stick to that.

This addressed only the naming. I have more thoughts on the actual code. I‘ll post them separately.

insights/client/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+        self._command = command
+
+    def __str__(self):
+        return (
+            "<{cls} ok={ok} return_code={code} stdout={out} stderr={err}>"
+        ).format(
+            cls=self.__class__.__name__,
+            ok=self.ok, code=self.return_code,
+            out=self.stdout, err=self.stderr,
+        )
```
Glutexo, line 36: Even though this is _only_ a data structure, there is still a lot of complexity.

m-horky:
Today's small update: I've fixed the "type hint" of `GPGCommandResult._command` (used to be incorrectly `str`).

m-horky:
(Nothing changed, I only rebased it over the latest master)

insights/client/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+        stderr (str): Standard error of the command.
+        _command (str | None):
+            An optional reference to the GPGCommand object that created the result
+    """
+    def __init__(self, ok, return_code, stdout, stderr, command):
+        self.ok = ok
+        self.return_code = return_code
+        self.stdout = stdout
+        self.stderr = stderr
+        self._command = command
```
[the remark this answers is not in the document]
Glutexo: This is what a [namedtuple](https://docs.python.org/2.6/library/collections.html?highlight=namedtuple#collections.namedtuple) could look like:

```
GPGCommandResult = namedtuple("GPGCommandResult", ["ok", "return_code", "stdout", "stderr", "command"])
result = GPGCommandResult(true, 0, "out", "", "echo out")
```

You can see that a single line definition (plus one import) do exactly what the entire class does.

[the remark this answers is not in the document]
m-horky: Yes, it's an additional context.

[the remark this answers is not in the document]
m-horky: But we'd lose the documentation of the fields. The `class` construct allows us to properly describe what each value means, without having to rely on the name. Sure, the names are clear and descriptive, but the docstring provides type information and potentially additional context you just won't get from a named tuple.

insights/client/crypto.py
m-horky: It's not that complex, it's just a format string ;)

insights/client/collection_rules.py
```diff
@@ -6,15 +6,14 @@
...
-import shlex
 import os
 import requests
 import yaml
 import stat
 from six.moves import configparser as ConfigParser
 
-from subprocess import Popen, PIPE, STDOUT
 from tempfile import NamedTemporaryFile
+from . import crypto
```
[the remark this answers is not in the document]
m-horky: Well, this allows you to mock the function more easily. If we imported the function itself, Python would create a new object instead of a reference, forcing us to track down how it is implemented. This way there is only one source for the existence of the function implementation.
Plus including the module name makes it easier for the reader to understand where the function is coming from and any additional context it may have. You don't get that from a bare function.

Overall, that is subjective, we may not find an agreement on this detail.

insights/client/crypto.py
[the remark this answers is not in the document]
Glutexo: It is possible to add the field documentation using inheritance:

```python
class GPGCommandResult(namedtuple("GPGCommandResult", ["ok", "return_code", "stdout", "stderr", "command"]))
    """Output of a GPGCommand.

    Attributes:
        ok (bool): Result of an operation.
        return_code (int): Return code of the operation.
        stdout (str): Standard output of the command.
        stderr (str): Standard error of the command.
        command (str | None):
            An optional reference to the GPGCommand object that created the result
"""
```

Unfortunately, the `__doc__` attribute is not writable in Python 2, so it’s not possible to do

```python
GPGCommandResult.__doc__ = """Output of a GPGCommand.

    Attributes:
        ok (bool): Result of an operation.
        return_code (int): Return code of the operation.
        stdout (str): Standard output of the command.
        stderr (str): Standard error of the command.
        command (str | None):
            An optional reference to the GPGCommand object that created the result
```

It is rather clumsy, but still pretty straightforward and I’d say, Pythonic.

Here’s a nice article on named tuples on Real Python: https://realpython.com/python-namedtuple/

However, the Client codebase doesn’t contain many dataclass-like types and even any argument documentation for functions and methods is scarce. That’s not to say it’s a bad idea to have it (although I am not a big fan of typing), just omitting it here wouldn’t make it any worse.

insights/client/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+import logging
+import shutil
+import tempfile
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+
+class GPGCommandResult(object):
+    """Output of a GPGCommand.
```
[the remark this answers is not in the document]
m-horky: It's fair the first line does not provide any more information over the class name. However, it is a good practice to add a one-liner describing the class, and this seems enough. Further information is available in the full docstring.

insights/client/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+        else:
+            logger.debug(
+                "gpg command {command} returned non-zero code: {result}".format(
+                    command=command, result=result,
+                )
+            )
+
+        return result
+
+    def evaluate(self):
```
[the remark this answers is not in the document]
m-horky: That's not something that is common in the code. I'd rather be explicit.

insights/client/crypto.py
```diff
@@ -0,0 +1,206 @@
...
+    Attributes:
+        ok (bool): Result of an operation.
+        return_code (int): Return code of the operation.
+        stdout (str): Standard output of the command.
+        stderr (str): Standard error of the command.
+        _command (GPGCommand | None):
+            An optional reference to the GPGCommand object that created the result
+    """
+    def __init__(self, ok, return_code, stdout, stderr, command):
+        self.ok = ok
```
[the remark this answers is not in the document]
m-horky: Yes, however the evaluation may not be only done based on return code, some GPG command may require further inspection of the captured output, for example. These may not be the same.

insights/client/crypto.py
Glutexo: A namedtuple comes with `__str__` and `__repr__` providing almost exactly the same string.

```python
>>> from collections import namedtuple
>>> GPGCommandResult = namedtuple("GPGCommandResult", ["ok", "return_code", "stdout", "stderr", "command"])
>>> result = GPGCommandResult(True, 0, "out", "", "echo out")
>>> print "Result: %s" % (result,)
GPGCommandResult(True, 0, 'out', '', 'echo out')
```

Although the format string itself is not that complex, the complexity I was refering to is the fact that it is completely redundant. An 8-line method of a custom class inevitably increases the code complexity, even though is as simple as this one. If you encounter it, you still need to read it and understand what it does and why it is there.

Glutexo (approved):
The only thing I was really worried about was the command that didn’t have its arguments properly escaped. This is now fixed and every other comments I have only target the code style. I give my ✅ then, so we can merge.

As could be seen from my comments, I find the code too complex in regard to how simple fix it delivers. A minimal, almost one-line, patch could do the trick. And I believe anything else is a refactoring or tech debt crushing that is a separate effort.

Glutexo:
@ptoscano, @ahitacat Did I remember correctly that you said that this pull request should not be merged right now? I can click the button, just assure me we want it. Thanks!

ahitacat:
Yes, let's wait for QE verification before merging it. I have added the label that we used when the PR are in that status.

Glutexo:
Apparently, I don’t even have the necessary rights to merge. I remember having them.

PR #3933, opened 2023-10-11
test: not raise Exception when Component is filtered in parent Component

xiangce:
- $subject, and when no more additional filters are required in the current context.
- It introduced a global variable `COMPONENT_FILTERED_PARSERS` in `tests`
   to allow rules to skip the duplicated and unnecessary "add_filter"
- It supports when the filterable Component is not used in the context (but only its combiner),
   devs do not need to do duplicated "add_filter"
- When the filterable Component is also imported, which indicates
   additional filters might be required, Exceptions should be thrown properly
- For Combiners, e.g. `Ps`, it shouldn't add to `COMPONENT_FILTERED_PARSERS`,
   since "add_filter" is always required for `Ps` per context

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/tests/__init__.py
+++ b/insights/tests/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import copy
+import importlib
 import inspect
 import itertools
 import json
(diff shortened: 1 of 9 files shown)


xiangce:
@JoySnow @chenlizhong  - please kindly review

wushiqinlou:
@xiangce - LGTM

insights/tests/__init__.py
```diff
@@ -154,11 +163,24 @@ def run_test(component, input_data,
...
+                        # The parser is NOT imported again in the rule
+                        parser = dr.get_component_by_name(parser)
+                        filtered.update(dr.get_registry_points(parser))
+        return filtered
+
     if do_filter and filters.ENABLED:
-        mod = component.__module__
-        sup_mod = '.'.join(mod.split('.')[:-1])
+        mod = dr.get_module_name(component)
+        sup_mod = dr.get_base_module_name(component)
```
JoySnow, line 180: Hi @xiangce , for this `sup_mod` getting line,  I see the new code is mean to get the mod itself instead of the it's `sup`. For example, 
```
mod:  insights.plugins.combiner_rule_wo_filterable_parser
sup_mod (in old code):  insights.plugins
sup_mod (in new code):  combiner_rule_wo_filterable_parser
```
Do you mean such changes?

xiangce: Good catching! Thanks @JoySnow - it's fixed in the new commit.

JoySnow (approved):

Issue #3936, opened 2023-10-16
Cannot load "version_info" when the collecting egg version is old like 3.0.246


xiangce:
For the archives that are collected with old EGGs, e.g. `3.0.246`, the insights cannot load the built-in files e.g. version_info correctly:

```
❯ cat /tmp/insights/insights-gcsbd6hydrsb550.cedelgroup.com-20231014224508/version_info
{"core_version": "3.0.246-1", "client_version": "3.1.7"}

❯ insights-cat insights.specs.Specs.version_info /tmp/insights/insights-gcsbd6hydrsb550.cedelgroup.com-20231014224508
...
Traceback (most recent call last):
  File "/Users/liuxc/work/insights/insights-core/insights/core/plugins.py", line 98, in invoke
    return self.component(broker)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 624, in __call__
    return self.kind(ctx.locate_path(self.path), root=ctx.root, ds=self, ctx=ctx)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 180, in __init__
    super(MetadataProvider, self).__init__(relative_path, root, ds, ctx)
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 138, in __init__
    self.validate()
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 200, in validate
    super(MetadataProvider, self).validate()
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 154, in validate
    raise ContentException("%s does not exist." % self.path)
insights.core.exceptions.ContentException: /tmp/insights/insights-gcsbd6hydrsb550.cedelgroup.com-20231014224508/data/version_info does not exist.

Traceback (most recent call last):
  File "/Users/liuxc/work/insights/insights-core/insights/core/plugins.py", line 98, in invoke
    return self.component(broker)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 624, in __call__
[...]

PR #3937, opened 2023-10-16
fix: compatible support built-in files for old archives

xiangce:
- fix:#3936

### All Pull Requests:

Check all that apply:


JoySnow (approved):
@xiangce I can reproduce the issue in my local dev env after reinstalling pip pkgs. And the fix works well.

Issue #3939, opened 2023-10-17
The base class "FileListing" of all parsers for "ls -xxx" specs raise error when parsing the output of "ls -lZ /dev" 


huali027:
It raises "ValueError: invalid literal for int() with base 10: '10,'" when parsing the following input:
```
/dev:
total 0
drwxr-xr-x. 19 root root    system_u:object_r:device_t:s0                  3180 May 10 09:54 .
dr-xr-xr-x. 18 root root    system_u:object_r:root_t:s0                     235 May 27  2022 ..
crw-r--r--.  1 root root    system_u:object_r:autofs_device_t:s0        10, 235 May 10 09:54 autofs
drwxr-xr-x.  2 root root    system_u:object_r:device_t:s0                   140 Jun 15 16:11 block
drwxr-xr-x.  3 root root    system_u:object_r:device_t:s0                    60 May 10 09:54 bus
drwxr-xr-x.  2 root root    system_u:object_r:device_t:s0                  2680 May 10 09:55 char
crw--w----.  1 root tty     system_u:object_r:console_device_t:s0        5,   1 May 10 09:54 console
```
Because it shows major, and minor numbers when listing "/dev" instead of size.

PR #3938, opened 2023-10-16
fix: Replace "ls_lanRZ" and "ls_lanZ" with "ls_laRZ" and "ls_laZ"

huali027:
* The class "FileListing" can not parse the output of "ls" with "-nZ" now. We have to enhance the class to support it. Since the previous the "-n" argument in "ls -lanRz" and "ls -lanZ" is not meaningful, we decide to remove the "-n" in the arguments instead of make the class too complicated.

### All Pull Requests:

Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [ ] Is this PR to correct an issue?
* [x] Is this PR an enhancement?

PR #3940, opened 2023-10-17
fix: Fix bug of ls_parser when handling "major" and "minor" in e.g. "ls -lZ /dev"

huali027:
Fixes https://github.com/RedHatInsights/insights-core/issues/3939
* The base class "FileListing" of parsers for "ls -xxx" specs raises the error "ValueError: invalid literal for int() with base 10: '10,'" when parsing the output like "ls -lZ /dev"

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/core/ls_parser.py
+++ b/insights/core/ls_parser.py
@@ -114,25 +114,31 @@ def parse_rhel8_selinux(parts):
     """
 
     links, owner, group, last = parts
-
-    selinux = parts[3].split(":")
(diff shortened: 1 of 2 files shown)


huali027:
@xiangce I'll update the parser to the new parser added in https://github.com/RedHatInsights/insights-core/pull/3938
 since the old parser is deprecated.

huali027:

PR #3941, opened 2023-10-17
Add warning to yum_log exclusion

Glutexo:
### All Pull Requests:

Check all that apply:

The _yum.log_ file collected by the _yum_log_ spec can in theory leak sensitive information. A RAT scan on the real customer data show that in reality this does not happen. The warning calls for caution when adding items to the list of excluded specs.

YUM plugins can write arbitrary information to the log file, potentially creating lines containing both a filter keyword (_Erased:_, _Installed:_, _Updated:_) and an IP address, e.g. a repository URL. Also, log files are persisted in their localized form, so they can contain various strings that are separate from the code and out of developer control.


xiangce:
LGTM

PR #3945, opened 2023-10-31
feat: Add one more arg to the command "ps -eo"

huali027:
### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/ps.py
+++ b/insights/parsers/ps.py
@@ -393,26 +393,26 @@ class ContainerPsAux(ContainerParser, PsAuxww):
 @parser(Specs.ps_eo)
 class PsEo(Ps):
     """
-    Class to parse the command `ps -eo pid,ppid,comm`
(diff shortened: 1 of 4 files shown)


insights/parsers/ps.py
```diff
@@ -393,26 +393,26 @@ class ContainerPsAux(ContainerParser, PsAuxww):
...
+         3871     1 master          1
+         3886  3871 qmgr            1
+        13724  3871 pickup          1
+        15663     2 kworker/0:1     1
+        16998     2 kworker/0:3     1
+        17259     2 kworker/0:0     1
+        18294  3357 sshd            1
 
     Examples:
         >>> type(ps_eo)
```
JoySnow, line 418: Hi @huali027 , it would be better to also update the test data `PsEo_TEST_DOC` in test_ps.py , and then these following test assert lines can be updated accordingly.

xiangce (approved):

PR #3964, opened 2023-11-30
feat: update eap json report spec

chenlizhong:
### All Pull Requests:

Check all that apply:

*Add your description here*

add new spec for EAP json report


insights/core/spec_factory.py
```diff
@@ -677,6 +678,54 @@ def __call__(self, broker):
         raise ContentException("[%s] didn't match." % ', '.join(self.patterns))
 
 
+class recent_files(object):
```
xiangce, line 681: May I know if the `last_modify_hours` is a key filter for this spec? Per my understanding according to our discussion, TBH, enhancing the existing `listdir`, e.g. adding a few more arguments `sort, key, type, reverse, num`, looks like a better approach than adding this new one from my perspective.

E.g. 

The usage of the update `listdir` in your case would be:

```python
eap_json_reports = head(
    listdir("/var/tmp/insights-runtimes/uploads/", 
        sort=True, type=os.path.isfile, key=os.path.getmtime, reverse=True),
    num=30
)
```

the update to the `listdir` and `head` will also be quite simple:
```python
class listdir(object):
    def __init__(self, path, context=None, ignore=None, deps=[], sort=False, type=None, key=None, reverse=False):
...
        self.sort, self.type, self.key, self.reverse, self.num = sort, type, key, reverse


    def __call__(self, broker):
...
        try:
            result = os.listdir(p)
            if sort:
                result = sorted(filter(self.type, result), key=self.key, reverse=self.reverse)
...

class head(object):
...
    def __call__(self, lst):
        c = lst[self.dep]
        if lst:
            return c[:self.num]
```

xiangce (changes requested):
@chenlizhong, BTW, for new specs, please get it approved first in the assets repo.

chenlizhong:
> @chenlizhong, BTW, for new specs, please get it approved first in the assets repo.

@xiangce  This is one update  for existing spec. Previous, use the glob_file to collect all files under the directory. Now just collect the last 24 hours updated files.  So no need to review again as collecting less then previous only.

insights/core/spec_factory.py
chenlizhong: last_modify_hours is key as currently the requirement is to  collect files last 24 hours updated.  So if update other specs, need to add this parameter and that will make existing (head/listdir/glob_file) specs more complex. I'd like to add new one .

xiangce: As this requirement is particular, I suggest not adding such a common helper but implementing it in an individual datasource.

insights/specs/datasources/eap_reports.py
```diff
@@ -0,0 +1,46 @@
...
+
+from insights.core.context import HostContext
+from insights.core.plugins import datasource
+from insights.core.exceptions import SkipComponent
+from insights.core.spec_factory import TextFileProvider
+
+log = logging.getLogger(__name__)
+
+
+def get_recent_files(target_path, last_modify_hours):
```
xiangce, line 16: How about moving this function to a single module under the `datasources` folder, and e.g. name it as `recent_files.recent_files_of` (recent_files_of() in the recent_files module)?

insights/tests/datasources/test_eap_reports.py
```diff
@@ -0,0 +1,59 @@
...
+import tempfile
+from datetime import datetime, timedelta
+
+from insights.core import dr
+from insights.core.context import HostContext
+from insights.specs.datasources.eap_reports import eap_report_files
+
+
+@pytest.fixture
+def sample_directory(scope="module"):
```
xiangce, line 13: Instead of creating such real files for testing, and then removing them by hand, the `unittest.mock` may help for such test.

insights/specs/sos_archive.py
```diff
@@ -77,6 +77,7 @@ class SosSpecs(Specs):
     docker_list_images = simple_file("sos_commands/docker/docker_images")
     docker_storage = simple_file("/etc/sysconfig/docker-storage")
     dumpe2fs_h = glob_file("sos_commands/filesys/dumpe2fs_-h_*")
+    eap_json_reports = glob_file(r"/var/tmp/insights-runtimes/uploads/*.json")
```
xiangce, line 80: I doubt if the files collected by sos are files also modified in last `24` hours?

insights/util/autology/datasources.py
```diff
@@ -18,7 +18,7 @@
 SIMPLE_FILE_TYPE = 'simple_file'
 """ str: Literal constant for a simple_file Spec object """
 GLOB_FILE_TYPE = 'glob_file'
-""" str: Literal constant for a simple_file Spec object """
+""" str: Literal constant for a glob_file Spec object """
```
xiangce, line 21: 👍

insights/specs/datasources/eap_reports.py
```diff
@@ -0,0 +1,46 @@
...
+import os
+import time
+import logging
+
+from insights.core.context import HostContext
+from insights.core.plugins import datasource
+from insights.core.exceptions import SkipComponent
+from insights.core.spec_factory import TextFileProvider
+
+log = logging.getLogger(__name__)
```
xiangce, line 13: If `log` is not required/used by the module, better remove it.

insights/specs/insights_archive.py
```diff
@@ -65,6 +65,7 @@ class InsightsArchiveSpecs(Specs):
     docker_list_images = simple_file("insights_commands/docker_images_--all_--no-trunc_--digests")
     dotnet_version = simple_file("insights_commands/dotnet_--version")
     doveconf = simple_file("insights_commands/doveconf")
+    eap_json_reports = glob_file(r"/var/tmp/insights-runtimes/uploads/*.json")
```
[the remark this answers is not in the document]
chenlizhong: OK. will remove it.

xiangce (approved):
lgtm

PR #3965, opened 2023-12-04
fix: check status created a machine-id file

ahitacat:
### All Pull Requests:

Check all that apply:

When insights-client checked the status created a machine-id file to make some calls to the console service. That created a residual machine file if it had not access to the service (i.e. system was not registered to rhsm).

This PR refactors a bit the logic of generate_machine_id, and splits in a new function that will only get the machine id `get_machineid()`.

Checking status and `generate_machine_id` will use this new function.

Test have been changed and a new one has been created.

Resolves: CCT-161

### DIFF
--- a/insights/tests/client/connection/test_LEGACY_reg_check.py
+++ b/insights/tests/client/connection/test_LEGACY_reg_check.py
@@ -5,9 +5,10 @@ from mock.mock import MagicMock, Mock, patch
 
 
 @patch("insights.client.connection.generate_machine_id", return_value='xxxxxx')
+@patch("insights.client.connection.machine_id_exists", return_value=True)
 @patch("insights.client.connection.InsightsConnection._init_session")
 @patch("insights.client.connection.InsightsConnection.get_proxies")
-def test_registration_check_ok_reg(get_proxies, _init_session, _):
+def test_registration_check_ok_reg(get_proxies, _init_session, _machine_id_exists, _generate_machine_id):
     '''
     Request completed OK, registered
         Returns True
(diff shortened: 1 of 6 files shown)


m-horky (changes requested):
This would solve it, but I am not really a fan of some of the parts.

1. Have you considered changing the original `client.get_machine_id` to accept a `create` flag, instead of creating a new function with similar name in the utils?
https://github.com/RedHatInsights/insights-core/blob/f5479075d961d45cea25638cb001cb83aa073da3/insights/client/client.py#L307-L308
2. If that would be too complicated, I found out you are only checking if the returned value is not None. Would it make sense to rename it to something like `machine_id_exists()` returning a bool?

ahitacat:
> 1. Have you considered changing the original `client.get_machine_id` to accept a `create` flag, instead of creating a new function with similar name in the utils?
>    https://github.com/RedHatInsights/insights-core/blob/f5479075d961d45cea25638cb001cb83aa073da3/insights/client/client.py#L307-L308

I considered that, but there is also a flag to create a new file, independently if it already exists. After this fix, I was considering to create a new ticket to make some effort refactoring this utility because is a bit entangled. Ideally, discerning the functions that use this method into 2, the ones that need by any cost an id, and the ones that doesn't need an id.


> 2. If that would be too complicated, I found out you are only checking if the returned value is not None. Would it make sense to rename it to something like `machine_id_exists()` returning a bool?
Not only, if it exists it returns the value of machine-id

ahitacat:
# Changelog

- Created a method to check if the machine-id file exists, this modify back the change on generate_machine_id function.
- Function that checked registration will check first if the machine-id file exists, it will avoid to generate a machine-id file if it don't exists. Solving the issue described in CCT-161

insights/tests/client/connection/test_LEGACY_reg_check.py
```diff
@@ -42,10 +44,10 @@ def test_registration_check_ok_reg_then_unreg(get_proxies, _init_session, _):
     assert conn.api_registration_check() == '2019-04-10'
 
 
-@patch("insights.client.connection.generate_machine_id", return_value='xxxxxx')
+@patch("insights.client.connection.False", return_value=True)
```
m-horky, line 47: This is probably a bad copy-paste

m-horky (changes requested):

m-horky (approved):
LGTM

ahitacat:
test me

zhangqianqian:
Verification pass on this open PR, please refer details: https://issues.redhat.com/browse/CCT-161

ahitacat:
@xiangce This can be on the next egg release

PR #3973, opened 2023-12-11
chore: Remove the specs and parsers related to xfs_db command

qinpingli:
Running xfs_db command against the mounted device can cause segment fault, it's not recommented.

And there are no plugins need these specs and parsers too.

### All Pull Requests:

Check all that apply:

Running `xfs_db` command against the mounted device can cause segment fault. The advisor candidate using these specs and parsers is proven not a good candidate. So, removing the specs and parsers.


insights/specs/default.py
```diff
@@ -686,8 +686,6 @@ class DefaultSpecs(Specs):
     x86_pti_enabled = simple_file("sys/kernel/debug/x86/pti_enabled")
     x86_retp_enabled = simple_file("sys/kernel/debug/x86/retp_enabled")
     # Suspend the collection for xfsprogs bug INSGHTCORE-257
```
xiangce, line 688: I think this comment should also be removed

qinpingli: Good catch!

Sorry for missing it. Updated.

xiangce (approved):

PR #3976, opened 2023-12-12
fix: Remove pruning client code from rpm

ryan-blakley:
Check all that apply:

A while back datasources were added that referenced client code, this broke the rpm usage because it pruned the client code in it's build. This weirdly caused some specs to work, and some to not which caused it to go un-noticed for a bit. This change removes the prune line that removed the client code in the build process. So the rpms will be slightly larger, but will allow specs to work properly again.


xiangce (approved):
LGTM

PR #3977, opened 2023-12-14
Enhance httpd spec include

wushiqinlou:
### All Pull Requests:

Check all that apply:

This PR is used to enhance httpd_conf spec so that it can handle `Include` situation.

### DIFF
--- a/insights/specs/datasources/httpd.py
+++ b/insights/specs/datasources/httpd.py
@@ -1,7 +1,9 @@
 """
 Custom datasources related to ``httpd``
 """
+import glob
 import json
+import os
 
 from insights.combiners.ps import Ps
 from insights.core.context import HostContext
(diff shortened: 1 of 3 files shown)


insights/specs/datasources/httpd.py
```diff
@@ -55,3 +57,67 @@ def httpd_on_nfs(broker):
...
+    This datasource returns the all of httpd configuration files' path.
+
+    Returns:
+        list: the file path of httpd configuration files
+
+    Raises:
+        SkipComponent: there is no httpd configuration file
+    """
+
+    all_paths = []
```
xiangce, line 74: @wushiqinlou - your implementation definitely works  for this issue, I just implemented it in separated functions with a recursive one, since, we should also update the `httpd_conf_scl_httpd24` and `httpd_conf_scl_jbcs_httpd24`. please have a look


```python
def _get_all_include_conf(root, glob_path):
    includes = glob_path
    # In case $ServerRoot in included in the 'glob_path'
    if not glob_path.startswith(root):
        includes = os.path.join(root, glob_path)
    # In case "*.conf" is not included in the 'glob_path'
    if not includes.endswith("*.conf"):
        includes = os.path.join(includes, "*.conf")

    _paths = set()
    try:
        for conf in glob.glob(includes):
            _paths.add(conf)
            with open(conf) as cfp:
                _includes = None
                for line in cfp.readlines()[::-1]:
                    if "Include" in line or "IncludeOptional" in line:
                        _includes = line.split()[-1].strip('"\'')
                        break
                # Search "includes" after scanning the whole file
                if _includes:
                    _paths.update(_get_all_include_conf(root, _includes))
        return _paths
    except Exception:
        pass
    return _paths

def get_httpd_configuration_files(httpd_root):
    main_httpd_conf = os.path.join(httpd_root, "conf/httpd.conf")
    all_paths = set()
    try:
        with open(main_httpd_conf) as cfp:
            server_root = httpd_root
            # Add it only when it exists
            all_paths.add(main_httpd_conf)
            for line in cfp.readlines()[::-1]:
                if "ServerRoot" in line:
                    server_root = line.strip().split()[-1].strip('"\'')
                elif "Include" in line or "IncludeOptional" in line:
                    includes = line.strip().split()[-1].strip('"\'')
            # Search "includes" after scanning the whole file
            all_paths.update(_get_all_include_conf(server_root, includes))
    except Exception:
        # Skip the datasource when no such "/etc/httpd/conf/httpd.conf" file
        raise SkipComponent

    if all_paths:
        return all_paths
    raise SkipComponent


@datasource(HostContext)
def httpd_configuration_files(broker):
    """
    This datasource returns the all of httpd configuration files' path.

    Returns:
        list: the file path of httpd configuration files

    Raises:
        SkipComponent: there is no httpd configuration file
    """
    httpd_root = '/etc/httpd'
    all_paths = get_httpd_configuration_files(httpd_root)
    if all_paths:
        return all_paths
    raise SkipComponent
```

insights/specs/default.py
```diff
@@ -244,15 +244,16 @@ class DefaultSpecs(Specs):
...
-    )
+    # httpd_conf = glob_file(
+    #     [
+    #         "/etc/httpd/conf/httpd.conf",
+    #         "/etc/httpd/conf.d/*.conf",
+    #         "/etc/httpd/conf.d/*/*.conf",
+    #         "/etc/httpd/conf.modules.d/*.conf"
+    #     ]
+    # )
     httpd_conf_scl_httpd24 = glob_file(
```
xiangce, line 257: Better update the `httpd_conf_scl_httpd24` and `httpd_conf_scl_jbcs_httpd24` as well.

insights/specs/default.py
```diff
@@ -244,15 +244,16 @@ class DefaultSpecs(Specs):
     httpd24_httpd_error_log = simple_file("/opt/rh/httpd24/root/etc/httpd/logs/error_log")
     httpd_M = foreach_execute(httpd.httpd_cmds, "%s -M")
     httpd_V = foreach_execute(httpd.httpd_cmds, "%s -V")
+    httpd_conf = foreach_collect(httpd.httpd_configuration_files, "%s")
```
xiangce, line 247: it should be behind the `httpd_cert_info_in_nss`

xiangce (changes requested):

xiangce (approved):

PR #3979, opened 2023-12-19
feat: collect PCP RAW data per 'ros_collect' set in insights-client.conf

xiangce:
### All Pull Requests:

Check all that apply:

See [INSGHTCORE-276](https://issues.redhat.com/browse/INSGHTCORE-276)

### DIFF
--- a/insights/specs/datasources/pcp.py
+++ b/insights/specs/datasources/pcp.py
@@ -118,17 +118,22 @@ def pcp_raw_files(broker):
 
     pm_cand = []
     yesterday = (datetime.date.today() - datetime.timedelta(days=1)).strftime("%Y%m%d")
-    for hostname in hostnames:
+    for hostname in sorted(hostnames):
         pm_root = os.path.join("/var/log/pcp/pmlogger/", hostname)
         if os.path.exists(pm_root):
             pm_files = glob.glob(os.path.join(pm_root, "{0}.*.xz".format(yesterday)))
             pm_files.append(os.path.join(pm_root, "{0}.index".format(yesterday)))
-            pm_files = sorted(filter(lambda x: os.path.isfile(x), pm_files))
+            pm_files = sorted(set(filter(lambda x: os.path.isfile(x), pm_files)))
+            print('----', pm_files)
             if len([pf for pf in pm_files if pf.endswith(('.index', '.meta.xz', '.0.xz'))]) >= 3:
                 pm_cand.append((pm_files, os.path.getmtime(pm_files[-2])))  # timestamp of '.index'
 
     if not pm_cand:
         raise ContentException("No PCP data directory OR incomplete PCP RAW data")
 
-    # the "pm_files" with the latest yesterday that contains complete data (>=3)
-    return sorted(pm_cand, key=lambda x: x[1])[-1][0]
+    # return the "pm_files" which:
+    # 1. has latest yesterday ".index"
+    # 2. has complete data (>=3)
+    # 3. has much more data than others
+    #    - a rare case: when more pm_files that satisfy with 1 and 2 are available
+    return sorted(pm_cand, key=lambda x: (x[1], len(x[0])))[-1][0]
--- a/insights/specs/datasources/pcp.py
+++ b/insights/specs/datasources/pcp.py
@@ -124,7 +124,7 @@ def pcp_raw_files(broker):
             pm_files = glob.glob(os.path.join(pm_root, "{0}.*.xz".format(yesterday)))
             pm_files.append(os.path.join(pm_root, "{0}.index".format(yesterday)))
             pm_files = sorted(filter(lambda x: os.path.isfile(x), pm_files))
-            if len([pf.endswith(('.index', '.meta.xz', '.0.xz')) for pf in pm_files]) >= 3:
+            if len([pf for pf in pm_files if pf.endswith(('.index', '.meta.xz', '.0.xz'))]) >= 3:
                 pm_cand.append((pm_files, os.path.getmtime(pm_files[-2])))  # timestamp of '.index'
 
     if not pm_cand:
--- a/insights/specs/datasources/pcp.py
+++ b/insights/specs/datasources/pcp.py
@@ -116,21 +116,19 @@ def pcp_raw_files(broker):
     hn = broker.get(Hostname)
     hostnames.update([hn.hostname, hn.fqdn]) if hn else None
 
-    pm_root = None
-    for hostname in hostnames:
-        tmp_root = os.path.join("/var/log/pcp/pmlogger/", hostname)
-        if os.path.exists(tmp_root):
-            pm_root = tmp_root
-            break
-    if pm_root is None:
-        raise ContentException("No PCP data directory found")
-
+    pm_cand = []
     yesterday = (datetime.date.today() - datetime.timedelta(days=1)).strftime("%Y%m%d")
-    pm_files = glob.glob(os.path.join(pm_root, "{0}.*.xz".format(yesterday)))
-    pm_files.append(os.path.join(pm_root, "{0}.index".format(yesterday)))
-    pm_files = sorted(filter(lambda x: os.path.isfile(x), pm_files))
+    for hostname in hostnames:
+        pm_root = os.path.join("/var/log/pcp/pmlogger/", hostname)
+        if os.path.exists(pm_root):
+            pm_files = glob.glob(os.path.join(pm_root, "{0}.*.xz".format(yesterday)))
+            pm_files.append(os.path.join(pm_root, "{0}.index".format(yesterday)))
+            pm_files = sorted(filter(lambda x: os.path.isfile(x), pm_files))
+            if len([pf.endswith(('.index', '.meta.xz', '.0.xz')) for pf in pm_files]) >= 3:
+                pm_cand.append((pm_files, os.path.getmtime(pm_files[-2])))  # timestamp of '.index'
 
-    if len([pf.endswith(('.index', '.meta.xz', '.0.xz')) for pf in pm_files]) < 3:
-        raise ContentException("Incomplete PCP RAW data of {0}".format(yesterday))
+    if not pm_cand:
+        raise ContentException("No PCP data directory OR incomplete PCP RAW data")
 
-    return pm_files
+    # the "pm_files" with the latest yesterday that contains complete data (>=3)
+    return sorted(pm_cand, key=lambda x: x[1])[-1][0]
--- a/insights/specs/datasources/pcp.py
+++ b/insights/specs/datasources/pcp.py
@@ -6,12 +6,12 @@ import glob
 import logging
 import os
 
+from insights.combiners.hostname import Hostname
 from insights.combiners.ps import Ps
 from insights.combiners.services import Services
 from insights.core.context import HostContext
 from insights.core.exceptions import SkipComponent, ContentException
 from insights.core.plugins import datasource
-from insights.parsers.hostname import HostnameShort
 from insights.parsers.ros_config import RosConfig
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


insights/specs/default.py
```diff
@@ -473,6 +477,9 @@ class DefaultSpecs(Specs):
     password_auth = simple_file("/etc/pam.d/password-auth")
     pci_rport_target_disk_paths = simple_command("/usr/bin/find /sys/devices/ -maxdepth 10 -mindepth 9 -name stat -type f")
     pcp_metrics = simple_command("/usr/bin/curl -s http://127.0.0.1:44322/metrics --connect-timeout 5", deps=[pcp_enabled])
+    # TODO: "/var/log/pcp/pmlogger/pcpdata.tar.xz" needs to be confirmed
+    pcp_raw_data = simple_file("/var/log/pcp/pmlogger/pcpdata.tar.xz",
```
upadhyeammit, line 481: This file won't be readily available, we would need to create it out of YYYYMMDD.index, YYYYMMDD.meta.xz, YYYYMMDD.N.xz files, if this is what you are waiting on.

I am also curious if this should be simple_file or temp file, because this is not required on the system once we include it in insights-archive.

upadhyeammit:
Hello! I can see we still need to add logic to create the tarball, would that be handled in same pr or another?

insights/specs/default.py
xiangce: Hi @upadhyeammit - Thanks for reviewing and sorry for reply to you late.

`simple_file` is just a helper function for insights core collection, it's good enough to collect this file into the insights-archive. TBH, it's not so exact as in insights the `simple_file` always indicates a `text` file 😓 

> because this is not required on the system once we include it in insights-archive.
Do you mean, that you want the insights core to remove the file after the collection?

xiangce:
> Hello! I can see we still need to add logic to create the tarball, would that be handled in same pr or another?
@upadhyeammit - Do you mean, that it needs the insights-core to decompress these files into a tarball? I thought it would be done by the `sub-package` you mentioned.

upadhyeammit:
> > Hello! I can see we still need to add logic to create the tarball, would that be handled in same pr or another?
> 
> @upadhyeammit - Do you mean, that it needs the insights-core to decompress these files into a tarball? I thought it would be done by the `sub-package` you mentioned.

Opposite of what you mentioned! We need to compress the pcp generated files and create pcpdata.tar.xz out of that.

xiangce:
@upadhyeammit  - Please kindly review/test this new pcp spec.

insights/specs/datasources/pcp.py
```diff
@@ -74,3 +91,38 @@ def pmlog_summary_args(broker):
...
+        All the collected PCP RAW data files will be stored in the
+        "var/log/pcp/pmlogger" directory in the archive.
+
+    Returns:
+        list(str): list of file path
+
+    Raises:
+        ContentException: Raises when number of PCP RAW files less than 3.
+    """
+    hostname = broker[HostnameShort].hostname
```
upadhyeammit, line 113: On my system I can see directory has been created with hostname+domain and because of that we are failing to collect files,
~~~
ls /var/log/pcp/pmlogger/localhost.localdomain/
20240123.2.xz  20240130.index    20240131.index        20240201.13.27.meta

# hostname
localhost.localdomain
~~~

I can see there is another class which can be useful here,
~~~
(Pdb) broker[HostnameShort].hostname
'localhost'
(Pdb) from insights.parsers.hostname import HostnameDefault
(Pdb) broker[HostnameDefault].raw
'localhost.localdomain'
~~~

I think the best way to handle this is, we look for both full name and short name for yesterday's files. To go further here, I was checking how pcp gets the hostname and I can see its using libc gethostname https://man7.org/linux/man-pages/man2/gethostname.2.html . 
There it has mentioned that hostname can be truncated if its longer, I don't know how we can handle that case?!

insights/specs/datasources/pcp.py
```diff
@@ -74,3 +91,38 @@ def pmlog_summary_args(broker):
...
+    yesterday = (datetime.date.today() - datetime.timedelta(days=1)).strftime("%Y%m%d")
+
+    pm_root = os.path.join("/var/log/pcp/pmlogger/", hostname)
+
+    pm_files = glob.glob(os.path.join(pm_root, "{0}.*.xz".format(yesterday)))
+    pm_files.append(os.path.join(pm_root, "{0}.index".format(yesterday)))
+    pm_files = sorted(filter(lambda x: os.path.isfile(x), pm_files))
+
+    file_cnt = 0
+    for pf in pm_files:
```
upadhyeammit, line 123: Nit: Maybe before looping through files we can check the length of pm_files, and raise exception if its below 3.
```suggestion
    if len(pm_files) < 3:
      raise ContentException("Incomplete PCP RAW data of {0}".format(yesterday))
    for pf in pm_files:
      file_cnt += 1 if pf.endswith(('.index', '.meta.xz', '.0.xz')) else 0
    if file_cnt < 3:
      raise ContentException("Incomplete PCP RAW data of {0}".format(yesterday))
```

upadhyeammit (changes requested):

insights/specs/datasources/pcp.py
xiangce: Sure, I changed to use the `Hostname` parser  and checks both the short hostname and fqdn

insights/specs/datasources/pcp.py
xiangce: As there won't be too many files be found, I updated it into one line check,

insights/specs/datasources/pcp.py
```diff
@@ -74,3 +91,38 @@ def pmlog_summary_args(broker):
...
+        list(str): list of file path
+
+    Raises:
+        ContentException: Raises when no such directory or number of PCP RAW
+                          files less than 3.
+    """
+    hn = broker[Hostname]
+    pm_root = os.path.join("/var/log/pcp/pmlogger/", hn.hostname)
+    if not os.path.exists(pm_root):
+        pm_root = os.path.join("/var/log/pcp/pmlogger/", hn.fqdn)
```
upadhyeammit, line 116: I can see hn.hostname and hn.fqdn returns same output for me,
~~~
(Pdb) hn.hostname
'localhost'
(Pdb) hn.fqdn
'localhost'
~~~

However if I use hostnameDefault then I can see raw is returning the value which resembles to pcp directory naming convention,
~~~
(Pdb) hn_default.raw
'localhost.localdomain'
(Pdb) hn_default.hostname
'localhost'
~~~

Yesterday I missed to check the man page of hostname command. Below is the snippet from it, so yeah without any param hostname cmd uses the same call as pcp,
~~~
       When called without any arguments, the program displays the current names:
       hostname will print the name of the system as returned by the gethostname(2) function.
~~~

Considering this I have two thoughts,
1. We can blindly depend on hostnameDefault.raw as it resembles to what pcp is doing.
2. Maybe we should still try the both ways, I mean checking for short and long names to be safe.

I wish to go with 2nd option and use hostnameDefault, let me know your thoughts.

Thank you!

upadhyeammit (changes requested):

insights/specs/datasources/pcp.py
xiangce: Sure things

insights/specs/datasources/pcp.py
```diff
@@ -110,12 +110,20 @@ def pcp_raw_files(broker):
...
+    hostnames.add(hnd.raw) if hnd else None
+    hn = broker.get(Hostname)
+    hostnames.update([hn.hostname, hn.fqdn]) if hn else None
+
+    pm_root = None
+    for hostname in hostnames:
+        tmp_root = os.path.join("/var/log/pcp/pmlogger/", hostname)
+        if os.path.exists(tmp_root):
+            pm_root = tmp_root
+            break
```
upadhyeammit, line 124: Adding for notes:
There could be situation where the first directory we find does not have pcp archives of yesterday and hence we may never report any files from that system, because everytime the collector would run it would default to same directory.

So we need to search in every directory named by shortname, fqdn or raw to look for yesterday's files. There could be further situations,
1. Two or more direcotries having yesterday's files and both directories are having complete sets(.index, .meta, 0.xz). In this case we are going to upload the files from directory which are latest. The latest timestamp would be of .index file.
2. Two or more directories having yesterday's files however only one of that is having valid set. In that case we upload from the directory which has valid set.
3. Multiple directories having yesterday's files, and some of those are having valid set. In this case we take valid sets, and upload the latest timestamp.

upadhyeammit:
Thanks for the changes, as per the code this looks good to me. I would try this out locally few times and then approve the change.

upadhyeammit:
Hello @xiangce !

1. Tested the case where multiple directories have pcp archives, and both are having complete set. This is working correctly.
2. when two directories have yesterday's file, in this case we should collect from the directory where minimal file set is complete. This is failing, I except we collect from localhost directory in this situation,

~~~
# ll localhost.localdomain/20240206.*
-rw-r--r--. 1 root root 5027984 Feb  7 13:48 localhost.localdomain/20240206.0.xz
-rw-r--r--. 1 root root 1921276 Feb  7 13:48 localhost.localdomain/20240206.1.xz
-rw-r--r--. 1 root root   66620 Feb  7 13:48 localhost.localdomain/20240206.meta.xz

# ll localhost/*
-rw-r--r--. 1 pcp pcp 5027984 Feb  7 13:41 localhost/20240206.0.xz
-rw-r--r--. 1 pcp pcp 1921276 Feb  7 13:41 localhost/20240206.1.xz
-rw-r--r--. 1 pcp pcp   29952 Feb  7 13:41 localhost/20240206.index
-rw-r--r--. 1 pcp pcp   66620 Feb  7 13:41 localhost/20240206.meta.xz

# ll insights-localhost.localdomain-20240207134850/data/var/log/pcp/pmlogger/20240206.*
-rw-r--r--. 1 root root 5027984 Feb  7 13:48 insights-localhost.localdomain-20240207134850/data/var/log/pcp/pmlogger/20240206.0.xz
-rw-r--r--. 1 root root 1921276 Feb  7 13:48 insights-localhost.localdomain-20240207134850/data/var/log/pcp/pmlogger/20240206.1.xz
-rw-r--r--. 1 root root   66620 Feb  7 13:48 insights-localhost.localdomain-20240207134850/data/var/log/pcp/pmlogger/20240206.meta.xz
~~~

upadhyeammit:
Hello @xiangce thanks a lot for changes, I confirm the changes look correct now,
~~~
# ll insights-localhost.localdomain-20240207161825/data/var/log/pcp/pmlogger/20240206.*
-rw-r--r--. 1 root root 5027984 Feb  7 16:18 insights-localhost.localdomain-20240207161825/data/var/log/pcp/pmlogger/20240206.0.xz
-rw-r--r--. 1 root root 1921276 Feb  7 16:18 insights-localhost.localdomain-20240207161825/data/var/log/pcp/pmlogger/20240206.1.xz
-rw-r--r--. 1 root root   29952 Feb  7 16:18 insights-localhost.localdomain-20240207161825/data/var/log/pcp/pmlogger/20240206.index
-rw-r--r--. 1 root root   66620 Feb  7 16:18 insights-localhost.localdomain-20240207161825/data/var/log/pcp/pmlogger/20240206.meta.xz
~~~

I just have question on last rare use case we discussed,

1. We have multiple directories which are having valid files, I mean at least set of three
2. Out of that, 2 directories are having index files with same timesatemp
3. Out of these 2, one directory has more archive files, as example,
  - localhost directory has 20240206.0.xz, 20240206.1.xz
  - localhost.localdomain has  20240206.0.xz, 20240206.1.xz, 20240206.2.xz
4. In this case we agreed to consider localhost.localdomain as it has more samples.

You mentioned that tests for this would be challenging to write. So I leave the decision upto you,
Maybe we can cover this scenario and have tests coverage at slow pace
Or completely cover this scenario and tests later considering this would be once in decade use case

xiangce:
@upadhyeammit - the latest code of this PR already includes the rare case you mentioned, please have a test on your local, as the test is complex for unit test, I didn't make test cases for it. but I have verified it on my local that the one who has more complete data will be collected when both have the same latest yesterday.
```
❯ ll --time-style=full-iso /var/log/pcp/pmlogger/insights-release.hosts.dev.psi.pek2.redhat.com
total 0
-rw-r--r--. 1 root root 0 2024-02-07 16:38:11.616793590 +0800 20240206.0.xz
-rw-r--r--. 1 root root 0 2024-02-07 16:38:11.616793590 +0800 20240206.1.xz
-rw-r--r--. 1 root root 0 2024-02-07 19:18:31.395129315 +0800 20240206.index
-rw-r--r--. 1 root root 0 2024-02-07 16:38:11.616793590 +0800 20240206.meta.xz

❯ ll --time-style=full-iso /var/log/pcp/pmlogger/insights-release
total 0
-rw-r--r--. 1 root root 0 2024-02-07 17:59:05.602514070 +0800 20240206.0.xz
-rw-r--r--. 1 root root 0 2024-02-07 19:18:31.395129315 +0800 20240206.index
-rw-r--r--. 1 root root 0 2024-02-07 17:59:05.602514070 +0800 20240206.meta.xz

❯ ll insights-20240207191854-fb3ae0/data/var/log/pcp/pmlogger
total 0
-rw-r--r--. 1 liuxc wheel 0 Feb  7 19:18 20240206.0.xz
-rw-r--r--. 1 liuxc wheel 0 Feb  7 19:18 20240206.1.xz
-rw-r--r--. 1 liuxc wheel 0 Feb  7 19:18 20240206.index
-rw-r--r--. 1 liuxc wheel 0 Feb  7 19:18 20240206.meta.xz
```

upadhyeammit (approved):
The changes look perfect with all the use cases!

PR #3984, opened 2024-01-03
fix: Fix issue with mount parsers mount option parsing

### LINKED ISSUE #3983: Mount options can contain quoted values with embedded commas
If a mount option is quoted and contains embedded commas it will not be parsed correctly.  For example context will be split in the current [ProcMounts](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/mount.py#L280) parser:

```
tmpfs /var/lib/containers/storage/overlay-containers/ff7e79fc09c/userdata/shm tmpfs rw,nosuid,nodev,noexec,relatime,context="system_u:object_r:container_file_t:s0:c184,c371",size=64000k 0 0
```

The split options will be `[..., 'context="system_u:object_r:container_file_t:s0:c184', 'c371"',...]`.

This is actually caused by the shared function [optlist_to_dict](https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/__init__.py#L39) which doesn't handle quoted values.

bfahr:
### All Pull Requests:

Check all that apply:

* Option values in quotes with embedded commas are parsed incorrectly.
* Add a mount specific function to correctly parse mount options and use it for all mount option parsing.
* Fix #3983

### DIFF
--- a/insights/parsers/mount.py
+++ b/insights/parsers/mount.py
@@ -43,7 +43,7 @@ import os
 from insights.core import CommandParser
 from insights.core.exceptions import ParseException, SkipComponent
 from insights.core.plugins import parser
-from insights.parsers import get_active_lines, keyword_searchFix
+from insights.parsers import get_active_lines, keyword_search
 from insights.specs import Specs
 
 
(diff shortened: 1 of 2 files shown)


xiangce (approved):
@bfahr  - Thanks for catching and fixing this bug. It looks good to me.

xiangce:
@JoySnow - I'd like to have your review also as you are one of the contributors to this parser too.

JoySnow (approved):

PR #3985, opened 2024-01-04
fix: Fix problem parsing mnt opt quotes

bfahr:
### All Pull Requests:

Check all that apply:

* Fixes the problem introduced in #3984 where there was no embedded comma in the quoted value then the option would not be saved.

### DIFF
--- a/insights/parsers/mount.py
+++ b/insights/parsers/mount.py
@@ -441,7 +441,7 @@ def _parse_mount_options(mount_options):
             comb_opt = ','.join(sp_opts[start_ndx:last_ndx + 1])
             opt_name, opt_value = comb_opt.split('=', 1)
             # Remove leading and trailing quotes
-            opts[opt_name] = opt_value[1:-1]
+            opts[opt_name] = opt_value.strip('"')
             in_quote = False
 
         # Else just a normal option or option=value
--- a/insights/parsers/mount.py
+++ b/insights/parsers/mount.py
@@ -445,14 +445,11 @@ def _parse_mount_options(mount_options):
             in_quote = False
 
         # Else just a normal option or option=value
-        else:
+        elif not in_quote:
             if '=' in opt:
                 opt_name, opt_value = opt.split('=', 1)
-                # Check for quoted values and remove quotes if present
-                if not (opt_value.startswith('"') and opt_value.endswith('"')):
-                    opts[opt_name] = opt_value
-                else:
-                    opts[opt_name] = opt_value[1:-1]
+                # Remove quotes if present
+                opts[opt_name] = opt_value.strip('"')
             else:
                 opts[opt] = True
     return opts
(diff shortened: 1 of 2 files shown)


insights/parsers/mount.py
```diff
@@ -428,24 +428,31 @@ def _parse_mount_options(mount_options):
...
             opts[opt_name] = opt_value[1:-1]
+            in_quote = False
 
         # Else just a normal option or option=value
         else:
             if '=' in opt:
                 opt_name, opt_value = opt.split('=', 1)
-                opts[opt_name] = opt_value
+                # Check for quoted values and remove quotes if present
+                if not (opt_value.startswith('"') and opt_value.endswith('"')):
```
xiangce, line 452: @bfahr - I think you may have tried thinking of that, but I just want to ask: can the `opt_value.strip('"')` be used here?

insights/tests/parsers/test_mount.py
```diff
@@ -61,6 +61,7 @@
 PROC_MOUNT_QUOTES = """
 /dev/mapper/rootvg-rootlv / ext4 rw,relatime,barrier=1,data=ordered 0 0
 tmpfs /var/lib/containers/storage/overlay-containers/ff7e79fc09c/userdata/shm tmpfs rw,nosuid,nodev,noexec,relatime,context="system_u:object_r:container_file_t:s0:c184,c371",size=64000k 0 0
+tmpfs /var/lib/containers/storage/overlay-containers/aa7e79fc09c/userdata/shm tmpfs rw,nosuid,nodev,noexec,relatime,context="system_u:object_r:container_file_t:s0",size=64000k 0 0
 """
```
JoySnow, line 65: Hi @bfahr ,  
Seeing this adding  test case `...,context="system_u:object_r:container_file_t:s0",size=64000k`, I guess that maybe we can also be compatible for a mount option case like: 
`rw,nosuid,nodev,noexec,relatime,context="system_u:object_r:container_file_t:s0:c184,some_inside_quote_option,c371",size=64000k` 
What do you think ?

xiangce: I think @JoySnow might mean: if we can add a test like `rw...,context="system_u:object_r:container_file_t:s0:c184,c371,c372",size=...`

Although the `context` was parsed correctly:
```python
assert device.mount_options.context == "system_u:object_r:container_file_t:s0:c184,c371,c372"
# it's True
```
However, the second `c371` was also parsed individually in the `else` part as a Boolean option
```python
assert  hasattr(device.mount_options, "c371") is True
assert device.mount_options.c371 is True
These are also `True`, but it should be not (right?)

```

insights/parsers/mount.py
bfahr: @xiangce yes that would work, thanks for pointing that out.

insights/tests/parsers/test_mount.py
bfahr: Good catch @JoySnow!  I've added that test case and fixed the problem.

xiangce (approved):
lgtm

JoySnow (approved):

PR #3991, opened 2024-01-11
Add ls sssd dir

wushiqinlou:
### All Pull Requests:

Check all that apply:

Collect the ls information of files under `/etc/sssd` for new rules


xiangce:
@wushiqinlou - We have planned to deprecate the `Ls<SubDir>` Parsers and specs soon, and the corresponding `LS<opt>` parsers can be used as an alternative. For this requirement, you can just add a filter to the `ls_lan_filtered_dirs` spec in the rule, as below, and do not raise any spec Change to the core collection/uploader.json.
```python
add_filter(Specs.ls_lan_filtered_dirs, '/etc/sssd')


@condition(LSlanFiltered)
```

PR #4002, opened 2024-01-29
fix: resolve missing insights-client.ppid error

DuckBoss:
### All Pull Requests:

Check all that apply:

This patch attempts to resolve ABRT core dumps
caused by attempting to remove the insights-client.ppid file during program exit.

The 'write_to_disk' method is modified to
include a redundant try-except to catch
any race conditions that may cause this rare error 
and a test case is included to mock this behaviour.

Resolves: CCT-163


insights/client/utilities.py
```diff
@@ -124,7 +124,10 @@ def write_to_disk(filename, delete=False, content=get_time()):
         return
     if delete:
         if os.path.lexists(filename):
-            os.remove(filename)
+            try:
+                os.remove(filename)
+            except OSError:
+                return
```
ahitacat, line 130: As OSError could have some other causes than the file not existing, I will just check for the errno == 2.
That way it will not produce an error when it doesn't exists but it will do it if anything else happens.

ahitacat (changes requested):

insights/client/utilities.py
DuckBoss: Good point, would be safer to be specific with a `FileNotFoundError` here.

DuckBoss: Actually,  this won't work. Nevermind, `FileNotFoundError` isn't a thing in python2.6/2.7....

insights/client/utilities.py
```diff
@@ -124,7 +125,11 @@ def write_to_disk(filename, delete=False, content=get_time()):
         return
     if delete:
         if os.path.lexists(filename):
-            os.remove(filename)
+            try:
+                os.remove(filename)
+            except OSError as err:
+                if err.errno != errno.ENOENT:
```
ahitacat, line 131: I'm missing some comment here describing what is this error.

ahitacat (approved):
LGTM

xiangce:
test me

PR #4004, opened 2024-01-30
fix(compliance): RHINENG-1935 handle non-list responses from inventory

skateman:
### All Pull Requests:

Check all that apply:

The `_fetch_system_by_machine_id` might return `False` or `None` responses that weren't handled in our codebase properly. In this fix I am adding a type check on the list of returned systems to make sure that the `len()` method doesn't fail.


skateman:
@xiangce can you please take a look? Also the 3.11 failure seems irrelevant, but you might say it is 🙈

xiangce (approved):
@skateman - Thanks a lot for fixing it. Yes, the py311 failure is fixed in #4005

PR #4007, opened 2024-01-31
fix(yum_udpates): load dnf plugins

psegedy:
we need to load dnf plugins to get package updates from dnf, especially on AWS systems where amazon-id plugin substitutes substrings in repo url - RHINENG-7888

### All Pull Requests:

Check all that apply:

yum_updates collection failed on AWS systems because it tried to download repo with invalid url (rhui.REGION.aws.ce.redhat.com). `REGION` string is replaced by `amazon-id` dnf plugin, so we need to load dnf plugins. `configure_plugins` must be called after `read_all_repos` since it modifies loeaded repos.


xiangce:
py27 test is fixed in another PR #4008

PR #4013, opened 2024-02-05
fix(test): spec cleaner tests are using 'is' for comparing strings

m-horky:
This was causing the tests to fail locally.

### All Pull Requests:

Check all that apply:

On Python 2.7, these tests were failing because the strings were being compared by their identity, instead of equivalence.


xiangce:
@m-horky Thanks for fixing the issue.

PR #4022, opened 2024-02-14
fix: Make utilities.write_to_disk use current time by default

m-horky:
### All Pull Requests:

Check all that apply:

The bug in the signature makes this function write down the time the egg was loaded, not the current time when this function is actually being called.


m-horky:
@xiangce :wave: Ready for review/merge.

PR #4025, opened 2024-02-16
Deprecate pvs lvs vgs

wushiqinlou:
### All Pull Requests:

Check all that apply:

This PR is used to deprecate the parser `LvsHeadings`, `PvsHeadings`, `VgsHeadings`

### DIFF
--- a/insights/combiners/lvm.py
+++ b/insights/combiners/lvm.py
@@ -75,7 +75,7 @@ Examples:
 import copy
 from collections import namedtuple
 from insights.core.plugins import combiner
-from insights.parsers.lvm import Lvs, Pvs, Vgs
+from insights.parsers.lvm import Lvs, LvsHeadings, Pvs, PvsHeadings, Vgs, VgsHeadings
 from insights.parsers.lvm import LvsAll, PvsAll, VgsAll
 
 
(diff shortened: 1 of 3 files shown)


JoySnow:
Found that these `*vsHeadings` parsers are still used in gss-rules, and also combiner `Lvm`.

Even insights-archive no more collecting `*vs_headings`, still, for sos_archive, it has only `lvs_headings` but no `lvs_noheadings`, see this [sos_archive.py#L153](https://github.com/RedHatInsights/insights-core/blob/master/insights/specs/sos_archive.py#L153) .

Based on these facts, we should keep including these `*vsHeadings` parsers in `Lvm` combiner, at least for now.

And for deprecating the `*vsHeadings` parsers, do we intend to move these related parsers to gss-rules repo in the near future?

@wushiqinlou , @xiangce , what's your opinions?

chenlizhong:
All unused parsers should be kept.

wushiqinlou:
@JoySnow - If we do not remove unused parsers from `Lvm`, the CI of checking dependency specs(https://gitlab.cee.redhat.com/insights-rules/insights-plugins/-/blob/master/script/rules_check.py?ref_type=heads#L242) will fail when the MR using `Lvm` combiner

wushiqinlou:
@JoySnow - Per the discussion with @xiangce, we decided to keep the origin `lvm`, and add the logic to handle this case into the CI specially

JoySnow:
Thanks @wushiqinlou for the update. 
Then I guess that we can close this PR here.  

@xiangce , could you kindly help close this?  It seems I don't have related access either.

wushiqinlou:
@JoySnow @xiangce - please do not close this PR. This PR also set deprecated warning for the parsers, I will update the PR later.

wushiqinlou:
@JoySnow @xiangce - please check the update. It seems that the CI failure for python27-test is not caused by the update.

PR #4031, opened 2024-02-26
Revert "Revert "Do not create /root/.gnupg/ directory by accident" (#4021)"

xiangce:


xiangce:
Incorrect commit-id is used, close it.

PR #4047, opened 2024-03-12
refactor: generate 'canonical_facts' from Parsers instead of Specs

xiangce:
- See RHINENG-8789

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/etc_machine_id.py
+++ b/insights/parsers/etc_machine_id.py
@@ -2,6 +2,7 @@
 EtcMachineId - file ``/etc/machine-id``
 =======================================
 """
+import uuid
 
 from insights.core import Parser
 from insights.core.exceptions import SkipComponent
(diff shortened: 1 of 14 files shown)


insights/parsers/ip.py
```diff
@@ -732,3 +735,37 @@ class IpLinkInfo(IpAddr):
         TX dropped: 0
     """
     pass
+
+
+@parser(Specs.ip_addresses)
+class IPs(CommandParser):
```
JoySnow, line 741: Need to mention this new Parser in this file's docstrings part - Line#18 .

insights/parsers/etc_machine_id.py
```diff
@@ -24,10 +25,13 @@ class EtcMachineId(Parser):
...
+        '4f3fbfda-59ff-4aea-9692-55a73342d439'
     """
 
     def parse_content(self, content):
         content = get_active_lines(content)
         if not content or len(content) != 1:
             raise SkipComponent('Invalid Content')
-        self.machine_id = content[0]
+        self.id = self.machine_id = content[0]
+        self.uuid = str(uuid.UUID(self.id))
```
JoySnow, line 37: `uuid.UUID()` could throw error when failed: 
```
raise ValueError('badly formed hexadecimal UUID string')
```

Should we give it an exception handling of this `uuid.UUID()` validation?

JoySnow: Same for `insights/parsers/client_metadata.py` and `insights/parsers/subscription_manager.py` .

JoySnow (changes requested):
Hi @xiangce , the changes LGTM, except the following points:

insights/parsers/etc_machine_id.py
xiangce: The "ValueError" will also result in the parser failure, just like the "ContentException", I think it's a bit of "draw legs on a snake" if we catch the "valueError" here and re-raise it with "ContentException.

JoySnow (approved):
Thanks for the update @xiangce . The changes LGTM.

PR #4050, opened 2024-03-13
fix: ContainerMssqlApiAssessment ParseException

JoySnow:
### All Pull Requests:

Check all that apply:

-----

Get this following exception from "Insights Times and Errors Report 2024-03-05" :

| count| component| request_id| exception|
| -------- | ------- | -------- | ------- |
|271483|insights.parsers.mssql_api_assessment.ContainerMssqlApiAssessment|some_request_id|ParseException("insights.parsers.mssql_api_assessment.ContainerMssqlApiAssessment couldn't parse json.")|

An example of the file content: 
```
cat: /var/opt/mssql/log/assessments/assessment-latest: No such file or directory
```

Fix this by special handling the `CommandParser.__bad_single_lines` like output in `ContainerParser`.

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -97,21 +97,6 @@ class Parser(object):
         raise NotImplementedError(msg)
 
 
-class ContainerParser(Parser):
-    """
-    A class specifically for container parser, with the "image" name, the
-    engine provider and the container ID on the basis of ``Parser``.
-    """
-    def __init__(self, context):
-        self.image = context.image
-        """str: The image of the container."""
-        self.engine = context.engine
-        """str: The engine provider of the container."""
-        self.container_id = context.container_id
-        """str: The ID of the container."""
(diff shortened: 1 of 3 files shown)


insights/core/__init__.py
```diff
@@ -631,6 +616,21 @@ def __init__(self, context, extra_bad_lines=None):
         super(CommandParser, self).__init__(context)
 
 
+class ContainerParser(CommandParser):
```
JoySnow, line 619: From the collected content example `API_OUTPUT_BAD` (see in test file) of this Spec,  it seems that a container File collection will be collected via command `cat`.  

Based on this, both Container Command/File can be applied with `CommandParser` . 

Hi @wushiqinlou  , could you help confirm if this `CommandParser` usage here make sense in Container spec collection view ?

xiangce (approved):
good to me

wushiqinlou:
@JoySnow - Good to me.

PR #4055, opened 2024-03-18
feat: clean (obfuscate/redact) specs in memory

xiangce:
- and add more tests
- see RHINENG-8775

### All Pull Requests:

Check all that apply:

*Add your description here*


insights/core/spec_cleaner.py
```diff
@@ -336,35 +345,54 @@ def _redact_line(self, line):
...
         for line in lines:
-            # 1. Do Redaction by default, unless 'no_redact=True'
-            if line and not no_redact:
-                line = self._redact_line(line)
-            # 2. TODO: Do Filtering as per the "filters"
-            # line = self._filter_line_allow(line, filters)
-            # 3. Do Obfuscation as per the "obf_funcs"
-            line = self._obfuscate_line(line, obf_funcs or [])
+            line = _clean_line(line)
             result.append(line) if line is not None else None
```
JoySnow, line 396: This `if line is not None` checking could introduce an interesting corner case: 
   - For a collected content, after cleaner.clean_content(), it has the
`result`  `['\n', '\n']`. 

Should this be collected as the final content of the file?     
Or this file should be emptied and skipped in insights-archive?  

See the test-case example in https://github.com/RedHatInsights/insights-core/pull/4061/files .

JoySnow (changes requested):
Hi @xiangce , the changes LGTM, except the two comments below.  Please kindly let me know your thoughts.

insights/core/spec_factory.py
```diff
@@ -71,6 +70,36 @@ def stream(self):
     def _stream(self):
         raise NotImplementedError()
 
+    def _clean_content(self, content):
```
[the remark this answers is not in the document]
xiangce: Good catching. I have applied different versions of implementation before submitting the final version. One of them is to do the "cleaning" while `load()` the content, which means, in the `memory` during the collection launched by the insights-client, the `content` will be obfuscated. However, the disadvantage is, for some datasources, it's necessary to get the original IP/FQDN to collect the required data. Then I turned to do the `cleaning` when writing the content to the disk. but forget to change the `_clean_content` back to work with the `self.conent` directly. 😿

JoySnow (approved):
@xiangce  The changes LGTM now.

insights/core/spec_cleaner.py
xiangce: Per our discussion in yesterday's sync, I opened a new card to track the excluding of the empty files RHINENG-8978

insights/core/spec_factory.py
```diff
@@ -71,6 +70,30 @@ def stream(self):
     def _stream(self):
         raise NotImplementedError()
 
+    def _clean_content(self):
+        """
+        Clean (Obfuscate, Redact, and Filter) the Spec Content ONLY when doing
```
JoySnow, line 75: This `Filter` in docstrings should also be removed according to the code changes.

insights/core/spec_factory.py
```diff
@@ -268,24 +270,11 @@ class TextFileProvider(FileProvider):
...
         self.loaded = True
-        args = self.create_args()
-        if args:
-            rc, out = self.ctx.shell_out(args, keep_rc=True, env=SAFE_ENV)
-            self.rc = rc
-            return out
 
         with safe_open(self.path, "r", encoding="utf-8", errors="surrogateescape") as f:
-            return [l.rstrip("\n") for l in f]
+            return [l.rstrip("\n") for l in apply_filters(self.ds, f)]
```
JoySnow, line 277: Seeing that we are going to use `apply_filters()` to replace the `["grep", "-F", filters, self.path]` in both `TextFileProvider` and `CommandOutputProvider`. 

Just the docstrings of  func `apply_filters` mentions that: 
> Applys filters to the lines of a datasource. This function is used only in
    integration tests. Filters are applied in an equivalent but more performant
    way at run time.

So I am wondering if there could some efficiency loss.  

Any reason for this change?

JoySnow (changes requested):

insights/core/spec_factory.py
JoySnow: Discussed with @xiangce , this filter method related change was mentioned in RHINENG-7854  . Then this LGTM.

xiangce: I tested the `grep -F` will be faster than `apply_filters` line by line. so I reverted it in the latest commit. Thanks.

insights/core/spec_factory.py
xiangce: Done

JoySnow (approved):
LGTM

PR #4057, opened 2024-03-18
feat: Reuse subscription-manager identity for machine-id

m-horky:
### All Pull Requests:

Check all that apply:

* Card ID: CCT-161

insights-client sometimes incorrectly regenerates identity while it shouldn't. Because the machine-id file is created preemptively, instead of at registration time (and deleted if the registration fails or Inventory returns error status signaling the system has been deleted in there), this is the easiest way to ensure the UUID will stay semi-stable.

subscription-manager already provides a stable way to identify a system. To ensure external tooling doesn't get confused, we should try to keep them consistent.

### DIFF
--- a/insights/client/utilities.py
+++ b/insights/client/utilities.py
@@ -96,6 +96,7 @@ def write_to_disk(filename, delete=False, content=get_time()):
         return
     if delete:
         if os.path.lexists(filename):
+            logger.debug("Removing '%s'" % filename)
             try:
                 os.remove(filename)
             except OSError as err:
@@ -105,6 +106,7 @@ def write_to_disk(filename, delete=False, content=get_time()):
                 if err.errno != errno.ENOENT:
                     raise err
     else:
+        logger.debug("Writing '%s'" % filename)
         with open(filename, 'wb') as f:
             f.write(content.encode('utf-8'))
 
--- a/insights/client/utilities.py
+++ b/insights/client/utilities.py
@@ -26,6 +26,7 @@ except ImportError:
 from .. import package_info
 from .constants import InsightsConstants as constants
 from .collection_rules import InsightsUploadConf, load_yaml
+from insights.client import cert_auth
 
 from insights.core.context import Context
 from insights.parsers.os_release import OsRelease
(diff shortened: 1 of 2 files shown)


insights/client/utilities.py
```diff
@@ -105,35 +113,84 @@ def write_to_disk(filename, delete=False, content=get_time()):
...
+    if not HAS_RHSM:
+        return None
+
+    if hasattr(rhsm.config, "get_config_parser"):
+        # This function is only available since subscription-manager-1.27.1-1 (RHEL 8.3).
+        config = rhsm.config.get_config_parser()  # type: "rhsm.config.RhsmConfigParser"
+    elif hasattr(rhsm.config, "initConfig"):
+        config = rhsm.config.initConfig()  # type: "rhsm.config.RhsmConfigParser"
+    else:
+        logger.debug("")
```
ptoscano, line 136: should something be logged here?

insights/client/utilities.py
```diff
@@ -105,35 +113,84 @@ def write_to_disk(filename, delete=False, content=get_time()):
...
+        return None
+
+    rhsm_certificate_dir_path = config.get("rhsm", "consumercertdir")  # type: str
+    rhsm_certificate_path = os.path.join(rhsm_certificate_dir_path, "cert.pem")  # type: str
+
+    if not os.path.isfile(rhsm_certificate_path):
+        return None
```
ptoscano, line 143: maybe log here something too, eg to say that the system is most likely not registered?

insights/client/utilities.py
```diff
@@ -105,35 +113,84 @@ def write_to_disk(filename, delete=False, content=get_time()):
...
+    if not machine_id:
+        machine_id = _get_rhsm_identity()
+        if machine_id:
+            logger.debug("Using subscription-manager identity as machine-id: '%s'." % machine_id)
+            write_to_disk(destination_file, content=machine_id)
+
+    if not machine_id:
         machine_id = str(uuid.uuid4())
-        logger.debug("Creating %s", destination_file)
+        logger.debug("Using new machine-id: '%s'." % machine_id)
```
ptoscano, line 182: nipick: IMHO "creating" feels better than "using new"

ptoscano (commented):
generally it seems ok; maybe something for potential discussion would be whether the actual authentication method needs to be considered before trying to use the subscription-manager uuid

ptoscano:
Actually, there is already the code for doing this:
https://github.com/RedHatInsights/insights-core/blob/master/insights/client/cert_auth.py

would it be possible to use that `cert_auth` module instead? (and in case improve it as needed)

xiangce:
All the IQE test `test_authproxy_connection` failed not only for this PR, but all PRs of insights-core, let's ignore it for now.

PR #4058, opened 2024-03-19
refactor: avoid passing duplicated mainfest to insights.collect()

xiangce:
- the "client_config" (InsightsConfig) already contains the
  manifest read from the client tool, this change removes it
  from the argument list of insights.collect() and simplify
  the code

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/collect.py
+++ b/insights/collect.py
@@ -281,7 +281,11 @@ def load_manifest(data):
     """ Helper for loading a manifest yaml doc. """
     if isinstance(data, dict):
         return data
-    doc = yaml.safe_load(data)
+    if os.path.isfile(data):
+        with open(data, 'r') as f:
+            doc = yaml.safe_load(f)
+    else:
+        doc = yaml.safe_load(data)
     if not isinstance(doc, dict):
         raise Exception("Manifest didn't result in dict.")
     return doc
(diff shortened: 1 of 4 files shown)


insights/collect.py
```diff
@@ -403,35 +403,38 @@ def generate_archive_name():
...
         (str, dict): The full path to the created tar.gz or workspace.
         And a dictionary with relevant exceptions captured by the broker during
         core collection, this dictionary has the following structure:
         ``{ exception_type: [ (exception_obj, component), (exception_obj, component) ]}``.
     """
-
+    # Get the manifest from client configuration
+    manifest = default_manifest
+    if hasattr(client_config, 'manifest') and client_config.manifest:
+        manifest = client_config.manifest
```
JoySnow, line 435: @xiangce   The manifest file load part from the origin code in [insights/client/core_collector.py  Line#35](https://github.com/RedHatInsights/insights-core/pull/4058/files#diff-76352cb529a81903277207d22b270de96a23532f285893c7f37e697c6f5109afL35) is skipped here. 

```
         if self.config.app is None:
                with open(self.config.manifest, 'r') as f:
                    manifest = f.read()
```

I see the `load_manifest(manifest)` calling below doesn't do the file load either. 

Then how do we load manifest from file?  Or we don't need the file load?

xiangce: The `load_mainfest` is strong enough to handle both `dict` and `file`:
https://github.com/RedHatInsights/insights-core/blob/729a1ca28e25468f60abd707bcaccb7b75bcb300/insights/collect.py#L280-L287

xiangce: Or, we should improve it to hand a `path` to a file.

JoySnow: Yep, we still need to do the `open(mainfest, 'r')` for `load_mainfest`.

JoySnow (approved):
@xiangce  The changes LGTM.

PR #4070, opened 2024-04-03
fix: insights-client failed, when --group was used

jirihnidek:
### All Pull Requests:

Check all that apply:

* Card ID: CCT-186
* When system was registered against Satellite server, and "insights-client --register" was triggered --group="tag" CLI option, then registration failed
* Satellite also contains some bug, and it always returns status code 200 and empty list regardless existence of given group.

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -575,35 +575,83 @@ class InsightsConnection(object):
     # -LEGACY-
     def group_systems(self, group_name, systems):
         """
-        Adds an array of systems to specified group
+        Adds an array of systems to specified group. When given group_name
+        does not exist, then client tries to create a new group and adds
+        given system to the new group.
 
         Args:
             group_name: Display name of group
             systems: Array of {'machine_id': machine_id}
         """
-        api_group_id = None
+        system_ids = list(systems.values())
         headers = {'Content-Type': 'application/json'}
         group_path = self.api_url + '/v1/groups'
-        group_get_path = group_path + ('?display_name=%s' % quote(group_name))
 
+        # Description of the API is here (groups GET /groups):
+        # yeah, links with anchors do not work as expected :-/
+        # https://console.redhat.com/docs/api/inventory/v1#operations-groups-api
+        group_get_path = group_path + ('?name=%s' % quote(group_name))
         get_group = self.get(group_get_path)
+
+        api_group_id = None
         if get_group.status_code == 200:
-            api_group_id = get_group.json()['id']
+            group_info = get_group.json()
+            if isinstance(group_info, dict):
+                try:
+                    results = group_info.get("results", None)
+                    if isinstance(results, list) and len(results) >= 1:
+                        group = results[0]
+                        if isinstance(group, dict):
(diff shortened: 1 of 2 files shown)


xiangce:
FYI, please don't worry about the failure of CI/CD, it was not caused by this PR and will be fixed once the new `uploader.json` is released. Please ignore it for now.

jirihnidek:
I can see that Jenkins jobs failed, but I cannot see, why did it fail.

xiangce:
> I can see that Jenkins jobs failed, but I cannot see, why did it fail.
Please ignore it too. It's a known issue mentioned in CCT-435. The QE team will fix it soon. Please ignore the other failures as well and start the review processes.

insights/client/connection.py
```diff
@@ -573,35 +576,73 @@ def create_system(self, new_machine_id=False):
...
+            api_group_id = None
+            group_info = get_group.json()
+            if isinstance(group_info, dict):
+                try:
+                    results = group_info.get("results", None)
+                    if isinstance(results, list) and len(results) >= 1:
+                        group = results[0]
+                        if isinstance(group, dict):
+                            api_group_id = group.get("id", None)
+                    api_group_id = results
```
m-horky, line 606: You seem to be overwriting the value you have set one line above.

m-horky (changes requested):
It's a shame we had to inflate the function by this much. It doesn't seem to be breaking anything by accident, and while it all looks ugly, there may be no way to save it.

Can you please revert the formatting changes that are committed with this change on other parts of the file?

m-horky (commented):
Would it be possible to create tests for this? Mock out some example responses the server can return. To be honest, I'm kind of lost in this method now, I'd like to see the difference between hosted and Satellite.

jirihnidek:
> Would it be possible to create tests for this? Mock out some example responses the server can return. To be honest, I'm kind of lost in this method now, I'd like to see the difference between hosted and Satellite.

I will try to do that.

m-horky (changes requested):
Thank you. Would it be possible to add a test for the Satellite case as well? I'd like to have it covered as well. Just one more test that would show that response code 200 may not be enough.

jirihnidek:
> Thank you. Would it be possible to add a test for the Satellite case as well? I'd like to have it covered as well. Just one more test that would show that response code 200 may not be enough.

What exactly do you mean to add test for the Satellite case?

m-horky:
> Satellite also contains some bug, and it always returns status code 200 and empty list regardless existence of given group.

Something that would mock a response like that -- server returning 200 with empty list. The method in `connection.py` grew a lot, but the tests don't don't seem to be covering the edge case your patch was addressing.

m-horky (approved):
ACK

xiangce:
May I know if this Pr ready for merge?

m-horky:
@xiangce Hi, I believe we're waiting for QE to verify this patch. I'll ping you once it is ready.

m-horky:
@xiangce Ready to be merged :+1:

PR #4073, opened 2024-04-08
chore: print warning when 'machine_id' is skipped for redaction

xiangce:
- see RHINENG-9044

### All Pull Requests:

Check all that apply:

- A warning message should be printed to the console when the required `machine_id` is skipped for redaction by the customers.

### DIFF
--- a/insights/client/collection_rules.py
+++ b/insights/client/collection_rules.py
@@ -393,7 +393,7 @@ class InsightsUploadConf(object):
     def get_rm_conf(self):
         '''
         Try to load the the "new" version of
-        remove.conf (file-redaction.yaml and file-redaction.yaml)
+        remove.conf (file-redaction.yaml and file-content-redaction.yaml)
         '''
         rm_conf = {}
         redact_conf = self.load_redaction_file(self.redaction_file)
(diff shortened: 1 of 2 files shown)


xiangce:
Hi @Glutexo, @ptoscano, and @JoySnow , please kindly review when free.

insights/client/collection_rules.py
```diff
@@ -409,6 +409,9 @@ def get_rm_conf(self):
             # remove Nones, empty strings, and empty lists
             self.rm_conf = dict((k, v) for k, v in rm_conf.items() if v)
 
+        if self.rm_conf and ('/etc/insights-client/machine-id' in self.rm_conf.get('files', []) or
+                'insights.specs.default.DefaultSpecs.machine_id' in self.rm_conf.get('components', [])):
+            logger.warning("WARNING: Spec machine_id will be skipped for redaction; as it would cause issues, please remove it from %s.", self.redaction_file)
```
ptoscano, line 414: should we have generic lists of files and specs to look for, so in case we need to add more it is easier?

xiangce: As the candidates for the `canonical facts` are in an `optional` list, which means the canonical facts would be okay if any of them are available. For now, I have no idea if missing any of them could cause such a critical issue like RHINENG-1626. For Advisors, it would cause rules not to hit, but it's not critical enough.
https://github.com/RedHatInsights/insights-core/blob/b54b236aa75b54ccfe9684c7a8c655933b25bc2f/insights/util/canonical_facts.py#L33-L41

JoySnow (approved):
@xiangce  The changes LGTM.

insights/client/collection_rules.py
xiangce: Hi @ptoscano - Thanks for reviewing, May I know your thoughts about this? Can we just warn for this `machine_id` this time, and if we have more findings, we can improve it to a list?

ptoscano: Fine by me, although the pessimistic me thinks that that situation will happen sooner than later.

PR #4077, opened 2024-04-10
fix: obfuscate/redact the facts for legacy_upload=False

xiangce:
- RHINENG-8790

### All Pull Requests:

Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [x] Is this PR to correct an issue?
* [ ] Is this PR an enhancement?

### Complete Description of Additions/Changes:

- fix the issue of RHINENG-8772 and RHINENG-7793

PR #4084, opened 2024-04-17
Revert "fix: obfuscate/redact the facts for legacy_upload=False"

xiangce:
Reverts RedHatInsights/insights-core#4077


xiangce:
As QA verification is required for the #4077, revert it here.

PR #4089, opened 2024-04-24
feat: Add parser db2 commands

wushiqinlou:
### All Pull Requests:

Check all that apply:

Add new spec and parser for new Advisor rule


insights/parsers/db2.py
```diff
@@ -1,54 +1,110 @@
...
 """
+
 from insights.core import CommandParser
 from insights.core.exceptions import SkipComponent
 from insights.core.plugins import parser
 from insights.specs import Specs
 
 
-@parser(Specs.db2ls_a_c)
-class Db2ls(CommandParser, list):
```
xiangce, line 17: To remove existing parsers, please deprecate it first and state the version that it will be removed in the `deprecated` statement.

insights/specs/datasources/db2.py
```diff
@@ -0,0 +1,64 @@
...
+    if not ps_list:
+        raise SkipComponent("No db2 database is running")
+    for item in ps_list:
+        user_name = item.get("USER", None)
+        if user_name:
+            users_result.append(user_name)
+    if users_result:
+        users_result.sort()
+        return users_result
+    raise SkipComponent("No db2 user is available")
```
xiangce, line 36: 1. Better remove the possible duplicated users from the result, e.g. by using a `set`
2. nitpick, the following code would be a bit simpler.

```python
if ps_list:
    users_result = set(filter(None, [item.get("USER") for item in ps_list]))
    if users_result:
        return sorted(users_result)
    raise SkipComponent("No db2 user is available")
raise SkipComponent("No db2 database is running")
```

insights/specs/datasources/db2.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    result = []
+    for item in broker[LocalSpecs.db2_databases]:
+        user = item.cmd.split()[2].strip()
+        for line in item.content:
+            if "Database name" in line:
+                name = line.split("=")[-1].strip()
+                result.append((user, name))
+    if result:
+        result.sort()
```
xiangce, line 62: better remove the duplicated items, e.g. set result as a `set`

```python
result = set()
for ..
    result.add((user, name))

if result:
    return sorted(result)

**Review by xiangce — CHANGES_REQUESTED** [review]
Please deprecate the old parser instead of removing it directly, and plz check other comments as well.

File: insights/parsers/db2.py (hunk 1 above)
Comment (wushiqinlou) [reply insights/parsers/db2.py]: @xiangce - my miss, we should not remove `Db2ls`, I added it back.

File: insights/specs/datasources/db2.py (hunk 2 above)
Comment (wushiqinlou) [reply insights/specs/datasources/db2.py]: @xiangce - updated, please check

File: insights/specs/datasources/db2.py (hunk 3 above)
Comment (wushiqinlou) [reply insights/specs/datasources/db2.py]: @xiangce - updated, please check

File: insights/specs/__init__.py [hunk 4]
```diff
@@ -127,6 +127,8 @@ class Specs(SpecSet):
     date = RegistryPoint(no_obfuscate=['hostname', 'ip'])
     date_utc = RegistryPoint(no_obfuscate=['hostname', 'ip'])
     db2ls_a_c = RegistryPoint(no_obfuscate=['hostname', 'ip'])
+    db2_database_configuration = RegistryPoint(multi_output=True, no_obfuscate=['hostname', 'ip'])
+    db2_database_manager = RegistryPoint(multi_output=True, no_obfuscate=['hostname', 'ip'])
```
xiangce, line 131: The 2 new specs should be in the front of the `db2ls_a_c` per the alphabet order.

insights/parsers/db2.py
```diff
@@ -1,12 +1,18 @@
...
 
 Db2ls - command ``db2ls -a -c``
--------------------------------
+===============================
+
+Db2DatabaseConfiguration - Command ``/usr/sbin/runuser -l  <user name>  -c 'db2 get database configuration for %s'``
+====================================================================================================================
+
+Db2DatabaseManager - Command ``/usr/sbin/runuser -l  <user name>  -c 'db2 get dbm cfg'``
+========================================================================================
```
xiangce, line 13: Please use `---` line for sub titles.

wushiqinlou: @xiangce - updated, please check

insights/specs/__init__.py
wushiqinlou: @xiangce - updated, please check

insights/specs/datasources/db2.py
```diff
@@ -0,0 +1,60 @@
...
+    Raises:
+        SkipComponent: there is no db2 users running db2 databases
+    """
+    ps = broker[Ps]
+    ps_list = ps.search(COMMAND_NAME__contains="db2sysc")
+    if ps_list:
+        users_result = set(filter(None, [item.get("USER") for item in ps_list]))
+        if users_result:
+            if "root" in users_result:
+                users_result.remove("root")
```
xiangce, line 30: Just in case there is only `root` in the results, how about changing the above lines as 
```python
users_result.discard('root')
if users_result:
    return sorted(users_result)
```

wushiqinlou: @xiangce - Updated, please check

xiangce (approved):

PR #4092, opened 2024-05-03
feat: Add parser nmap

wushiqinlou:
### All Pull Requests:

Check all that apply:

Create this parser to enhance an existing Advisor rule

### DIFF
--- a/insights/parsers/nmap.py
+++ b/insights/parsers/nmap.py
@@ -65,11 +65,12 @@ class NmapSsh(CommandParser, dict):
             if not line_strip:
                 continue
 
-            if line_strip.startswith("|") and "algorithms:" in line:
-                current_key = line_strip.split(":")[0][1:].strip()
-                data[current_key] = []
-            elif line_strip.startswith("|") and line_strip.count(" ") == 7 and "algorithms:" not in line:
-                data[current_key].append(line.split()[-1])
+            if line_strip.startswith("|"):
+                if "algorithms:" in line_strip:
+                    current_key = line_strip.split(":")[0][1:].strip()
+                    data[current_key] = []
+                elif not line_strip.endswith(':'):
+                    data[current_key].append(line.split()[-1])
 
         if not data:
             raise SkipComponent("The output format is not expected")
--- a/insights/parsers/nmap.py
+++ b/insights/parsers/nmap.py
@@ -1,6 +1,6 @@
 """
 nmap - command ``nmap``
-===============================
+=======================
 
 Parser contains in this module is:
 
--- a/insights/parsers/nmap.py
+++ b/insights/parsers/nmap.py
@@ -1,0 +1,76 @@
+"""
+nmap - command ``nmap``
+===============================
+
+Parser contains in this module is:
+
+NmapSsh - command ``/usr/bin/nmap --script ssh2-enum-algos -sV -p 22 127.0.0.1``
+--------------------------------------------------------------------------------
+
+"""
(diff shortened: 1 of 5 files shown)


insights/parsers/nmap.py
```diff
@@ -0,0 +1,76 @@
+"""
+nmap - command ``nmap``
+===============================
```
JoySnow, line 3: Need to keep the same length of the title string and the `====` line.

insights/parsers/nmap.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Raises:
+        insights.core.exceptions.SkipComponent: if the output of the ``repquota -agnpuv`` command is empty or the format is not expected.
+    """
+
+    def parse_content(self, content):
+        data = {}
+        current_key = ""
+        for line in content:
+            if not line.strip():
```
JoySnow, line 64: As `line.strip()` was used more than once in this function, we can store the result of `line.strip()` into a variable and reuse it in the later code  for running time saving.

insights/parsers/nmap.py
```diff
@@ -0,0 +1,76 @@
...
+                continue
+
+            if line.strip().startswith("|") and "algorithms:" in line:
+                current_key = line.strip().split(":")[0].split("|")[-1].strip()
+                data[current_key] = []
+            if line.strip().startswith("|") and line.strip().count(" ") == 7 and "algorithms:" not in line:
```
JoySnow, line 70: As the two `if` conditions are mutually exclusive, we better use a `elif` instead of `if` for running time saving purpose.

JoySnow (changes requested):

insights/parsers/nmap.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def parse_content(self, content):
+        data = {}
+        current_key = ""
+        for line in content:
+            if not line.strip():
+                continue
+
+            if line.strip().startswith("|") and "algorithms:" in line:
+                current_key = line.strip().split(":")[0].split("|")[-1].strip()
```
JoySnow, line 68: As this `startswith("|")` was confirmed, what about taking advantage from it to replace the `.split("|")[-1]` with `[1:]`, like: 
```suggestion
                current_key = line.strip().split(":")[0][1:].strip()
```

JoySnow (approved):
@wushiqinlou  Thanks for the updates, the changes LGTM.

insights/parsers/nmap.py
```diff
@@ -0,0 +1,77 @@
...
+        current_key = ""
+        for line in content:
+            line_strip = line.strip()
+            if not line_strip:
+                continue
+
+            if line_strip.startswith("|") and "algorithms:" in line:
+                current_key = line_strip.split(":")[0][1:].strip()
+                data[current_key] = []
+            elif line_strip.startswith("|") and line_strip.count(" ") == 7 and "algorithms:" not in line:
```
xiangce, line 71: the `line_strip.count(" ") == 7` condition looks a bit strict to me, whenever there are other blank spaces appearing in the rest part of line, the parsing result would be failed.

And how about changing the logic as

```python
if line_strip.startswith("|"):
    if "algorithms:" in line_strip:
          current_key = line_strip.split(":")[0][1:].strip()
          data[current_key] = []
    elif not line_strip.endswith(':'):
          data[current_key].append(line.split()[-1])
```

xiangce (approved):

Issue #4103, opened 2024-05-14
Environment variable INSIGHTS_FILTERS_ENABLED no longer works


bfahr:
The environment variable `INSIGHTS_FILTERS_ENABLED` can be used to turn off filtering when processing archives.  Setting this env var to `False` turns off the filter checking and filtering of specs that are included in archives so that applications processing the archive using custom rules will see all of the data collected.  After [insights-core-3.3.21 release](https://github.com/RedHatInsights/insights-core/releases/tag/insights-core-3.3.21) this env var is no longer  working correctly.  If it is set to `False` then filtered specs will not be loaded and if set to `True` they will be loaded but the default filters will be applied meaning that only data such as column headings will be present in the data.  The `True` setting works as expected, but the `False` setting should enable loading all of the data from filtered specs in archives.

xiangce:
Thanks, @bfahr, for reporting this issue, the root cause is that the original validation of the `filterable` specs didn't care about this environment 
https://github.com/RedHatInsights/insights-core/blob/de92cb5b6a156c5b451d964be8fc8888cec7aad0/insights/core/spec_factory.py#L195-L198

And in #4082, the getting filters are enabled for `SerializedArchiveContext` (insights-archive from core collection).

So, in runtime against the archive collected by core collection, the process of filterable specs will be terminated due to the above validation.

I raised #4104 to fix it, please have a look as well.

Once it gets approved, I will release it this Thu.

Sorry for bring this trouble to you.

xiangce:
**Without** the #4014, the result is:
```
~/work/insights/insights-core (master) ❯ INSIGHTS_FILTERS_ENABLED=False insights-cat insights.specs.Specs.insights_client_conf /tmp/insights-20240412012625
Traceback (most recent call last):
  File "/home/liuxc/work/insights/insights-core/insights/core/plugins.py", line 101, in invoke
    return self.component(broker)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 654, in __call__
    return self.kind(ctx.locate_path(self.path), root=ctx.root,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/liuxc/work/insights/insights-core/insights/core/spec_factory.py", line 180, in __init__
    self.validate()
[...]

PR #4104, opened 2024-05-15
fix: env INSIGHTS_FILTERS_ENABLED no longer works

xiangce:
- validate the filters of 'filterable' specs only when the variable is True
  when it's False, all content will be loaded (this is not changed in this PR)
- fix: #4103

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/core/filters.py
+++ b/insights/core/filters.py
@@ -159,8 +159,13 @@ def get_filters(component):
             filters |= inner(d, filters)
         return filters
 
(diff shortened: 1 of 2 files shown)


JoySnow (approved):

bfahr (approved):
@xiangce these changes look good.  The env var INSIGHTS_FILTERS_ENABLED works as expected now.  Thanks!

PR #4122, opened 2024-06-14
feat: Enhance InstalledProductIDs to support more lines/filters

huali027:
* Add file delimiter and file path to the filters to keep the mapping between data and file

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/installed_product_ids.py
+++ b/insights/parsers/installed_product_ids.py
@@ -92,35 +92,33 @@ class InstalledProductIDs(CommandParser):
 
     Attributes:
         ids (set): set of strings of the unique IDs collected by the command
-        data(list): list of dicts of the attributes and its value
+        product_certs(list): list of dicts of the product certificates key-value pairs
 
     Examples:
         >>> type(products)
         <class 'insights.parsers.installed_product_ids.InstalledProductIDs'>
         >>> list(products.ids)
         ['69']
-        >>> products.data[0]
+        >>> products.product_certs[0]
         {'Path': '/etc/pki/product-default/69.pem', 'ID': '69'}
 
     """
     def parse_content(self, content):
         """ Parse command output """
         self.ids = set()
-        self.data = []
-        one_file_data = {}
-        valid_data = False
+        self.product_certs = []
+        one_file_data = None
         for line in content:
             # different file delimiter
             if line == 'Product Certificate':
                 if one_file_data:
-                    self.data.append(one_file_data)
+                    self.product_certs.append(one_file_data)
                 one_file_data = {}
-                valid_data = True
             else:
-                if valid_data:
+                if one_file_data is not None:
                     name, value = line.split(':', 1)
                     one_file_data[name.strip()] = value.strip()
         if one_file_data:
             # add the last file data
-            self.data.append(one_file_data)
-        self.ids = set([item['ID'] for item in self.data])
+            self.product_certs.append(one_file_data)
+        self.ids = set([item['ID'] for item in self.product_certs])
--- a/insights/parsers/installed_product_ids.py
+++ b/insights/parsers/installed_product_ids.py
@@ -123,4 +123,4 @@ class InstalledProductIDs(CommandParser):
         if one_file_data:
             # add the last file data
             self.data.append(one_file_data)
-        self.ids = {item['ID'] for item in self.data}
+        self.ids = set([item['ID'] for item in self.data])
--- a/insights/parsers/installed_product_ids.py
+++ b/insights/parsers/installed_product_ids.py
@@ -13,7 +13,7 @@ from insights.core.filters import add_filter
 from insights.specs import Specs
 from .. import parser, CommandParser
 
-add_filter(Specs.subscription_manager_installed_product_ids, 'ID:')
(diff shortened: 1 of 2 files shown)


huali027:

insights/parsers/installed_product_ids.py
```diff
@@ -88,23 +88,39 @@ class InstalledProductIDs(CommandParser):
             Brand Type:
             Brand Name:
 
-    Filters have been added to the parser so that only the ``ID`` element will
-    be collected.
+    Filters have been added to the parser so that only the filtered lines will be collected.
 
     Attributes:
         ids (set): set of strings of the unique IDs collected by the command
+        data(list): list of dicts of the attributes and its value
```
xiangce, line 95: `data` is not a good name anymore to use as attributes, as itself cannot give any hit about what kind of the `data` is. Please rename the attribute more meaningful, e.g., `product_list`, etc.

insights/parsers/installed_product_ids.py
```diff
@@ -88,23 +88,39 @@ class InstalledProductIDs(CommandParser):
...
+                one_file_data = {}
+                valid_data = True
+            else:
+                if valid_data:
+                    name, value = line.split(':', 1)
+                    one_file_data[name.strip()] = value.strip()
+        if one_file_data:
+            # add the last file data
+            self.data.append(one_file_data)
+        self.ids = set([item['ID'] for item in self.data])
```
xiangce, line 126: the code can be simplified a bit as:
```python
        one_file_data = None
        for line in content:
            line = line.strip()
            # different file delimiter
            if line == 'Product Certificate':
                one_file_data ={}
                self.data.append(one_file_data)
            elif ":" in line and one_file_data is not None:
                    name, value = line.split(':', 1)
                    one_file_data[name.strip()] = value.strip()

        self.ids = set([item['ID'] for item in self.data])


```

xiangce (changes requested):

insights/parsers/installed_product_ids.py
huali027: @xiangce sure, I'll rename it.

insights/parsers/installed_product_ids.py
huali027: @xiangce  yes, it is simple. The only concern is:
```
one_file_data ={}
self.data.append(one_file_data)
```
It adds an empty dict first. So I'm not sure if there is any case that there are no filtered data after it. But maybe we don't need to consider too much now?

insights/parsers/installed_product_ids.py
```diff
@@ -13,7 +13,7 @@
 from insights.specs import Specs
 from .. import parser, CommandParser
 
-add_filter(Specs.subscription_manager_installed_product_ids, 'ID:')
+add_filter(Specs.subscription_manager_installed_product_ids, ['ID:', 'Product Certificate', 'Path'])
```
JoySnow, line 16: IIUC, the added `'Path'` is used to filter the lines like `Path: /etc/pki/product-default/69.pem` .  If this is the case, what about adjusting the filter string to `'Path: '` to gain more accuracy?

insights/parsers/installed_product_ids.py
```diff
@@ -88,23 +88,37 @@ class InstalledProductIDs(CommandParser):
...
         for line in content:
-            if line.strip().startswith('ID:'):
-                _, id = line.strip().split(':', 1)
-                self.ids.add(id.strip())
+            # different file delimiter
+            if line == 'Product Certificate':
+                if one_file_data:
+                    self.product_certs.append(one_file_data)
+                one_file_data = {}
+            else:
```
JoySnow, line 117: When there are no ":" in `line`, this following `split()` will run into exception, eg: 
```
>>> name, value = "some line".split(':', 1)
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 2, got 1)
```

What about adding the checking of  `line` inside this `else` clause ?
```suggestion
            elif line and ":" in line:
```

insights/parsers/installed_product_ids.py
```diff
@@ -88,23 +88,37 @@ class InstalledProductIDs(CommandParser):
...
+                if one_file_data is not None:
+                    name, value = line.split(':', 1)
+                    one_file_data[name.strip()] = value.strip()
```
JoySnow, line 120: For the key variables naming, to keep a better consistency, what about converting the key to the lowercase and joined the words with low dash ? 
 
For example, put the key variables name to: 
```
    assert results.product_certs[1]['Path'] == '/etc/pki/product/479.pem'.    ----> 'path'
    assert results.product_certs[1]['ID'] == '479'                            ----> 'id'
    assert results.product_certs[1]['Brand Type'] == ''                       ----> 'brand_type'
```

JoySnow (changes requested):

insights/parsers/installed_product_ids.py
huali027: sure, I'll update it.

insights/parsers/installed_product_ids.py
huali027: yes, good to me.

xiangce (approved):

PR #4123, opened 2024-06-17
feat: include JSON format in malware detection results

michalslomczynski:
RHINENG-10594

### All Pull Requests:

Check all that apply:

In the Insights archive, malware scan results are passed as GraphQL mutation query only.
Incoming changes include additional file `malware-detection-results-raw.json`, which contains JSON representation of python dictionary that was used to compose mutation query.
New file will be consumed by the newest version of malware-detection app. Old file is still needed for backwards compatibility.

### DIFF
--- a/insights/specs/datasources/malware_detection/__init__.py
+++ b/insights/specs/datasources/malware_detection/__init__.py
@@ -231,7 +231,7 @@ class MalwareDetectionClient:
                             MALWARE_APP_URL, MALWARE_CONFIG_FILE)
 
             # This is what is uploaded to the malware backend
-            return host_scan_mutation
+            return (host_scan_mutation, self.host_scan)
         else:
             logger.error("No scans performed, no results to upload.")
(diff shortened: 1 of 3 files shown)


xiangce:
This change looks good to me. But I'd like to ask @mhuth to have a review too.

And I believe the corresponding change on the malware-detection backend should also be applied to fetch the data as per the new store way. @michalslomczynski, please let me know when the malware-detection backend is also ready for this change.

Thanks.

michalslomczynski:
Yes, we are ready. I will add changes on malware-detection-backend-v2 side after merge.

If QE are accessing `MalwareDetectionClient.run()` method directly in tests, then they will need to update it accordingly:
~~mutationQuery = mdc.run()~~
mutationQuery = mdc.run()[0]

mhuth (approved):
Simple enough change.  Looks good to me too.

PR #4125, opened 2024-06-18
feat: New spec and parser to get the expire date of ssl certificate in rsyslog

huali027:

### DIFF
--- a/insights/specs/datasources/ssl_certificate.py
+++ b/insights/specs/datasources/ssl_certificate.py
@@ -7,6 +7,7 @@ from insights.core.context import HostContext
 from insights.core.exceptions import SkipComponent
 from insights.core.plugins import datasource
 from insights.parsers.mssql_conf import MsSQLConf
+from insights.combiners.rsyslog_confs import RsyslogAllConf
 
 
 @datasource(HttpdConfTree, HostContext)
@@ -91,6 +92,7 @@ def nginx_ssl_certificate_files(broker):
 def mssql_tls_cert_file(broker):
     """
     Get the mssql tls certificate file path configured by "ssl_certificate"
+
     Arguments:
         broker: the broker object for the current session
     Returns:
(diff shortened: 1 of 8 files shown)


huali027:

insights/specs/datasources/ssl_certificate.py
```diff
@@ -102,3 +104,31 @@ def mssql_tls_cert_file(broker):
...
+    rsyslog_objs = broker[RsyslogAllConf]
+    for obj in rsyslog_objs.values():
+        for item in obj:
+            if 'DefaultNetstreamDriverCertFile' in item:
+                # basic format
+                if '$DefaultNetstreamDriverCertFile' in item:
+                    return item.split()[-1].strip()
+                else:
+                    # advanced format, it is set in global block, and the global line contains all the content in it
+                    parts = item.strip().split()
```
JoySnow, line 130: From the parser `RsyslogConf`, all lines from it had been `strip()`. Seems no need to have one more round of `strip()` here?

insights/specs/datasources/ssl_certificate.py
```diff
@@ -102,3 +104,31 @@ def mssql_tls_cert_file(broker):
...
+    """
+    rsyslog_objs = broker[RsyslogAllConf]
+    for obj in rsyslog_objs.values():
+        for item in obj:
+            if 'DefaultNetstreamDriverCertFile' in item:
+                # basic format
+                if '$DefaultNetstreamDriverCertFile' in item:
+                    return item.split()[-1].strip()
+                else:
+                    # advanced format, it is set in global block, and the global line contains all the content in it
```
JoySnow, line 129: IIUC, this `if` and `else` are describing two parallel cases: `basic format` or `advanced format` . If this is the case, it is better to have the two comment lines with the same indent.

JoySnow, line 129: BTW, just noticed that this line and also line#24 are pretty long in this file's view. Would you like help to "prettify" the lines' length in this file? Say, shorten to around length 80 ...

JoySnow (changes requested):

insights/specs/default.py
```diff
@@ -540,6 +540,7 @@ class DefaultSpecs(Specs):
     rpm_ostree_status = simple_command("/usr/bin/rpm-ostree status --json", signum=signal.SIGTERM)
     rpm_pkgs = rpm_pkgs.pkgs_with_writable_dirs
     rsyslog_conf = glob_file(["/etc/rsyslog.conf", "/etc/rsyslog.d/*.conf"])
+    rsyslog_tls_cert_enddate = command_with_args("/usr/bin/openssl x509 -in %s -enddate -noout", ssl_certificate.rsyslog_tls_cert_file)
```
JoySnow, line 543: Do we need to also add this spec to file `insights/specs/insights_archive.py` ?    @xiangce @huali027

huali027: @JoySnow I thought `command_with_args ` is just used by "core-collect", so no need to add it. @xiangce Please help do the final confirmation.

insights/specs/datasources/ssl_certificate.py
huali027: @JoySnow I adjust it to make it shorter, please help to confirm. 
We can define the standard about the max length of code and comment we must follow, then we can pay attention to it later.

huali027:
@JoySnow For other comment, I have updated the code, please help to review again, thanks.

insights/specs/default.py
xiangce: The `insights_archives.py` is a way (or workaround) for the new core collection to keep compatible with specs that is collected by the legacy collection (aka `uploader.json`)

For new specs that was not added to the `uploader.json`, like `command_with_args` kind specs, it's not required to add its `mangle_name` to the `insights_archives.py`

JoySnow: Thanks for the info and confirmation, @huali027, @xiangce .

JoySnow (approved):
Thanks for the updates.  LGTM now.

PR #4131, opened 2024-06-25
feat(pgp): Use crypto.py during Egg and Collection verification

m-horky:
### All Pull Requests:

Check all that apply:

* Card ID: CCT-131
* Card ID: CCT-405
* Card ID: RHEL-2480
* Card ID: RHEL-2482

This patch adds a self-contained and isolated GPG verification
environment. It runs GPG in an isolated environment where only selected
PGP keys are allowed to check the file signature matches its file.

GPG creates a directory `$HOME/.gnupg/` every time it performs some
operation. When run under root, but not manually (e.g. via
subscription-manager Cockpit plugin), it tries to create and write to
this directory, which pollutes user directories and/or causes SELinux
denials.

This patch utilizes the `--homedir` argument GPG supports in order to
move the GPG home directory to a temporary directory for the time of the
transaction. After the GPG action is performed, the directory is cleaned
up.

---

This PR is an improvement over previously reverted https://github.com/RedHatInsights/insights-core/pull/3930

### DIFF
--- a/insights/client/crypto.py
+++ b/insights/client/crypto.py
@@ -1,0 +1,279 @@
+import errno
+import os.path
+import logging
+import shutil
+import tempfile
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+
+class GPGCommandResult(object):
+    """Output of a GPGCommand.
+
+    :param ok: Result of an operation.
+    :type ok: bool
+    :param return_code: Return code of an operation.
+    :type return_code: int
+    :param stdout: Standard output of the command.
+    :type stdout: str
+    :param stderr: Standard error of the command.
+    :type stderr: str
+    :param _command: An optional reference to the GPGCommand object that created the result.
+    :type _command: GPGCommand | None
+    """
+    def __init__(self, ok, return_code, stdout, stderr, command):
+        self.ok = ok
+        self.return_code = return_code
+        self.stdout = stdout
+        self.stderr = stderr
+        self._command = command
+
+    def __str__(self):
+        return (
+            "<{cls} ok={ok} return_code={code} stdout={out} stderr={err}>"
+        ).format(
+            cls=self.__class__.__name__,
+            ok=self.ok, code=self.return_code,
+            out=self.stdout, err=self.stderr,
+        )
+
+
+class GPGCommand(object):
+    """GPG command run in a temporary environment.
+
+    :param command: The command to be executed.
+    :type command: list[str]
+    :param key: Path to the GPG public key to check against.
+    :type key: str
+    :param _home: Path to the temporary GPG home directory.
+    :type _home: str | None
+    :param _raw_command: The last invoked command.
+    :type _raw_command: list[str] | None
+    """
+    TEMPORARY_GPG_HOME_PARENT_DIRECTORY = "/var/lib/insights/"
+
+    def __init__(self, command, key):
+        self.command = command
+        self.key = key
+        self._home = None
+        self._raw_command = None
+
+    def __str__(self):
+        return "<{cls} _home={home} _raw_command={raw}>".format(
+            cls=self.__class__.__name__,
+            home=self._home,
+            raw=self._raw_command
+        )
+
+    def _setup(self):
+        """Prepare GPG environment.
+
+        :returns: Result of the GPG key import.
+        :rtype: GPGCommandResult
+        """
+        # The /var/lib/insights/ directory is used instead of /tmp/ because
+        # GPG needs to have RW permissions in it, and existing SELinux rules only
+        # allow access here.
+        self._home = tempfile.mkdtemp(dir=type(self).TEMPORARY_GPG_HOME_PARENT_DIRECTORY)
+
+        logger.debug("setting up gpg temporary environment in '{home}'".format(home=self._home))
+        result = self._run(["--import", self.key])  # type: GPGCommandResult
+        if not result.ok:
+            logger.warning("failed to import key '{key}': {result}".format(
+                key=self.key, result=result
+            ))
+        return result
+
+    def _supports_cleanup_socket(self):
+        """Queries for the version of GPG binary.
+
+        :returns: `True` if the gnupg is known for supporting `--kill all`.
+        :rtype: bool
+        """
+        version_process = subprocess.Popen(
+            ["/usr/bin/gpg", "--version"],
+            stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+            universal_newlines=True,
+            env={"GNUPGHOME": self._home},
+        )
+        stdout, stderr = version_process.communicate()
+        if version_process.returncode != 0:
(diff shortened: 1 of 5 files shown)


zpetrace:
From a QA perspective the pre-verification for this PR passed

insights/client/crypto.py
```diff
@@ -0,0 +1,270 @@
...
+            home=self._home,
+            raw=self._raw_command
+        )
+
+    def _setup(self):
+        """Prepare GPG environment.
+
+        :returns:
+            `True` if public GPG keys were imported into temporary environment,
+            `False` if there was an error.
```
ptoscano, line 74: the type of the return value is not correctly described here

insights/client/crypto.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+        # The /var/lib/insights/ directory is used instead of /tmp/ because
+        # GPG needs to have RW permissions in it, and existing SELinux rules only
+        # allow access here.
+        self._home = tempfile.mkdtemp(dir=type(self).TEMPORARY_GPG_HOME_PARENT_DIRECTORY)
+
+        logger.debug("setting up gpg temporary environment  in '{home}'".format(home=self._home))
+        result = self._run(["--import", self.key])  # type: GPGCommandResult
+        if not result.ok:
+            logger.debug("failed to import key '{key}': {result}".format(
```
ptoscano, line 85: perhaps I'd bump this log message to warning, as it seems a severe enough failure

insights/client/crypto.py
```diff
@@ -0,0 +1,270 @@
...
+    def evaluate(self):
+        """Run the command.
+
+        :returns: The result of the shell command.
+        :rtype: GPGCommandResult
+        """
+        try:
+            result = self._setup()
+            if not result.ok:
+                logger.debug("gpg setup failed")
```
ptoscano, line 217: perhaps I'd bump this log message to warning, as it seems a severe enough failure

insights/client/crypto.py
```diff
@@ -0,0 +1,270 @@
...
+            `True` if public GPG keys were imported into temporary environment,
+            `False` if there was an error.
+        :rtype: GPGCommandResult
+        """
+        # The /var/lib/insights/ directory is used instead of /tmp/ because
+        # GPG needs to have RW permissions in it, and existing SELinux rules only
+        # allow access here.
+        self._home = tempfile.mkdtemp(dir=type(self).TEMPORARY_GPG_HOME_PARENT_DIRECTORY)
+
+        logger.debug("setting up gpg temporary environment  in '{home}'".format(home=self._home))
```
ptoscano, line 82: nitpick: two space between "environment" and "in"

insights/client/crypto.py
```diff
@@ -0,0 +1,270 @@
...
+    :returns: Evaluated GPG command.
+    :rtype: GPGCommandResult
+    """
+    if not os.path.isfile(file):
+        logger.debug(
+            "cannot verify signature of '{file}', file does not exist".format(
+                file=file
+            )
+        )
+        raise OSError(errno.ENOENT, "File '{file}' not found".format(file=file))
```
ptoscano, line 246: I wonder whether here, rather than raising an exception, `GPGCommandResult(false)` should be returned

according to the code where now `verify_gpg_signed_file()` is used, this situation used to result in `/usr/bin/gpg` failing, and thus handled as failure for it; now this will raise an exception, and I'm not sure it will be handled well within the rest of the core code...

in case this is changed, please add a test case for this situation, together with the newly added ones

insights/client/crypto.py
```diff
@@ -0,0 +1,270 @@
...
+        logger.debug((
+            "cannot verify signature of '{file}', "
+            "signature '{signature}' does not exist"
+        ).format(file=file, signature=signature))
+        raise OSError(
+            errno.ENOENT,
+            "Signature '{sig}' of file '{file}' not found.".format(
+                sig=signature, file=file
+            )
+        )
```
ptoscano, line 258: ditto (return `GPGCommandResult(false)` rather than raising this exception)

ptoscano (changes requested):
it seems generally good; I left few notes here, with possibly changes (not complicated) to be done

don't forget to rebase on `master` when fixing/amending -- thanks!

insights/client/crypto.py
m-horky: When we bump the log message level in `_setup()`, we don't need to do it here again. We should either leave it as-is, or drop it.

ptoscano: either is fine -- my point is that kind of "important failures" ought to be shown somewhere else than the log file only

insights/client/crypto.py
```diff
@@ -0,0 +1,279 @@
...
+                version = line.split(" ")[-1]  # type: str
+                break
+        else:
+            stdout = "\n".join("stdout: {line}".format(line=line) for line in stdout.split("\n") if len(line))
+            logger.debug("could not query for gpg version: output not recognized:\n{out}".format(out=stdout))
+            return False
+
+        try:
+            version_info = tuple(int(v) for v in version.split("."))
+        except Exception as exc:
```
ptoscano, line 117: maybe one or more specific exceptions would be better than a broad `Exception` here

ptoscano (commented):
LGTM from my POV, I only left a potential note for improvements (which is fine otherwise)

@xiangce want to take a look as well?

xiangce (approved):
Per my test the new verification method looks good

```
2024-07-17 10:15:39,168    DEBUG root:219 Loaded initial egg: /var/lib/insights/last_stable.egg/insights
2024-07-17 10:15:39,168    DEBUG insights.client.crypto:271 starting gpg verification process for '/var/lib/insights/last_stable.egg'
2024-07-17 10:15:39,168    DEBUG insights.client.crypto:80 setting up gpg temporary environment in '/var/lib/insights/tmpgtyr18ke'
2024-07-17 10:15:39,177    DEBUG insights.client.crypto:200 gpg command ['--import', '/etc/insights-client/redhattools.pub.gpg']: ok
2024-07-17 10:15:39,177    DEBUG insights.client.crypto:222 running gpg in the temporary environment
2024-07-17 10:15:39,187    DEBUG insights.client.crypto:200 gpg command ['--verify', '/var/lib/insights/last_stable.egg.asc', '/var/lib/insights/last_stable.egg']: ok
2024-07-17 10:15:39,195    DEBUG insights.client.crypto:275 signature verification of '/var/lib/insights/last_stable.egg' passed
2024-07-17 10:15:39,562    DEBUG insights.client.client:145 Logging initialized
```
And when GPG check failed it can successfully roll back to load the `rpm.egg`.

@m-horky - please let me know when it's ready to merge.

insights/client/crypto.py
m-horky: There are multiple things that may go wrong there (`AttributeError`, `TypeError`, `ValueError`) depending on how malformed the line is. I think this is fine, we're logging the raw value when it fails.
I think it's fine.

m-horky:
@xiangce Ready to merge

PR #4138, opened 2024-07-03
feat: New spec "/proc/fs/cifs/DebugData" and parser

huali027:
### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/cifs.py
+++ b/insights/parsers/cifs.py
@@ -5,7 +5,7 @@ Get CIFS Info
 This module contains the following parsers:
 
 CIFSDebugData - file ``/proc/fs/cifs/DebugData``
-================================================
+------------------------------------------------
 """
 
 from insights.core import LogFileOutput
@@ -71,9 +71,9 @@ class CIFSDebugData(LogFileOutput):
         >>> from insights.core.filters import add_filter
         >>> from insights.specs import Specs
         >>> add_filter(Specs.cifs_debug_data, 'abc_def_fsf')
-        >>> type(cifg_dd_obj)
+        >>> type(cifs_dd_obj)
         <class 'insights.parsers.cifs.CIFSDebugData'>
-        >>> cifg_dd_obj.last_scan("abc_def_line", "abc_def_fsf")
+        >>> cifs_dd_obj.last_scan("abc_def_line", "abc_def_fsf")
     """
 
     time_format = None
--- a/insights/parsers/cifs.py
+++ b/insights/parsers/cifs.py
@@ -1,0 +1,79 @@
+"""
+Get CIFS Info
+=============
+
+This module contains the following parsers:
+
+CIFSDebugData - file ``/proc/fs/cifs/DebugData``
+================================================
+"""
+
+from insights.core import LogFileOutput
+from insights.specs import Specs
+from insights.core.plugins import parser
+
+
+@parser(Specs.cifs_debug_data)
+class CIFSDebugData(LogFileOutput):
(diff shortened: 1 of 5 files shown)


huali027:
@JoySnow @xiangce Please help to review, thanks.

insights/parsers/cifs.py
```diff
@@ -0,0 +1,79 @@
+"""
+Get CIFS Info
+=============
+
+This module contains the following parsers:
+
+CIFSDebugData - file ``/proc/fs/cifs/DebugData``
+================================================
```
JoySnow, line 8: Better use `---` instead of `===` for such cases, eg: https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/ip.py#L8 .

insights/tests/parsers/test_cifs.py
```diff
@@ -0,0 +1,27 @@
...
+def test_cifs_debug_data():
+    cifg_dd_obj = CIFSDebugData(context_wrap(CIFS_DEBUG_DATA_CONTENT))
+
+    assert cifg_dd_obj.test_dfs_origin_test
+    assert cifg_dd_obj.test_dfs_origin_test.get('raw_message') == "DFS origin full path: \\ad.abc.com\DFSROOT\ame1"
+
+
+def test_doc():
+    env = {
+            'cifg_dd_obj': CIFSDebugData(context_wrap(CIFS_DEBUG_DATA_CONTENT))
```
JoySnow, line 24: Seems this `cifg_dd_obj` is a typo, was mean to be `cifs_dd_obj`?

insights/parsers/cifs.py
```diff
@@ -0,0 +1,79 @@
...
+        CIFSMaxBufSize: 16384
+        Active VFS Requests: 0
+
+        Servers:
+        1) ConnectionId: 0x4 Hostname: CVXDFAVB1
+        Number of credits: 512 Dialect 0x302 signed
+        TCP status: 1 Instance: 22
+        Local Users To Server: 1 SecMode: 0x3 Req On Wire: 0
+        In Send: 0 In MaxReq Wait: 0
+        DFS origin full path: \\ad.abc.com\DFSROOT\name1
```
JoySnow, line 37: From the advisor rule's logic, we want to get the `full path` here for other usage. But if we process the hostname obfuscate for such line, the target full path will be changed, will that still usable for later logics? 

Or, if we use `no_obfuscate=['hostname', 'ip']` for this spec, will that applicable?

JoySnow (changes requested):
Hi @huali027 , Got some concerns about the hostname obfuscate thing of this spec, please check my comments inline.

insights/parsers/cifs.py
huali027: The hostname here is the remote DFS share provider's hostname. So I think obfuscation won't handle it, right?

huali027:
@JoySnow I'll update the other two comments after the obfuscation issue to see if I need to update the code.

insights/parsers/cifs.py
xiangce: Yep, currently, only the server's hostname will be obfuscated.

JoySnow (approved):

insights/parsers/cifs.py
JoySnow: Then the `full path` won't be obfuscated.  Thanks @xiangce , @huali027 .

xiangce:
Ignore the python26 test for now.

PR #4139, opened 2024-07-04
feat: New spec and parser to get rsyslog CA cert expiration date

huali027:
### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/specs/datasources/ssl_certificate.py
+++ b/insights/specs/datasources/ssl_certificate.py
@@ -167,6 +167,11 @@ def rsyslog_tls_ca_cert_file(broker):
                             else:
                                 # example: global( DefaultNetstreamDriverCAFile = "file_path" test = "abc")
                                 name_index = parts.index(part)
-                                if parts[name_index + 1].strip() == '=':
-                                    return parts[name_index + 2].strip('")')
+                                if len(parts) > name_index + 2:
+                                    if parts[name_index + 1].strip() == '=':
+                                        path = parts[name_index + 2].strip('")')
+                                        if path:
+                                            return path
+                                # no need to continue
+                                raise SkipComponent
     raise SkipComponent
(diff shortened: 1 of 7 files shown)


huali027:

insights/specs/datasources/ssl_certificate.py
```diff
@@ -134,3 +134,39 @@ def rsyslog_tls_cert_file(broker):
...
+                    for part in parts:
+                        if 'DefaultNetstreamDriverCAFile' in part:
+                            # example: global(DefaultNetstreamDriverCAFile="file_path" test="abc")
+                            if '=' in part:
+                                return part.split('=')[-1].strip('")')
+                            else:
+                                # example: global( DefaultNetstreamDriverCAFile = "file_path" test = "abc")
+                                name_index = parts.index(part)
+                                if parts[name_index + 1].strip() == '=':
+                                    return parts[name_index + 2].strip('")')
```
JoySnow, line 171: For cases like below (may not exist in real cert file, just in case ...), the `parts[name_index + 1]` access will raise IndexError. 
- `example: global( DefaultNetstreamDriverCAFile`
- `example: global( DefaultNetstreamDriverCAFile = `

Would be better if we do with a index existence check before the access.

insights/specs/datasources/ssl_certificate.py
```diff
@@ -134,3 +134,39 @@ def rsyslog_tls_cert_file(broker):
...
+            if 'DefaultNetstreamDriverCAFile' in item:
+                if '$DefaultNetstreamDriverCAFile' in item:
+                    # basic format
+                    return item.split()[-1].strip()
+                else:
+                    # advanced format
+                    # it is set in global block, and the global line contains all the content in it
+                    parts = item.split()
+                    for part in parts:
+                        if 'DefaultNetstreamDriverCAFile' in part:
```
JoySnow, line 163: And seems we can add a `break` to this `if`'s else case, to save some `parts` iteration try.

huali027: @JoySnow updated based on the two comments, please help to review again, thanks.

JoySnow (approved):
Thanks for the update, @huali027 .

PR #4146, opened 2024-07-05
fix: use ubi8 instead of centos7 in Dockerfile

chenlizhong:
### All Pull Requests:

Check all that apply:

* To fix below error when build image by Dockerfile

> Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
14: curl#6 - “Could not resolve host: [mirrorlist.centos.org](http://mirrorlist.centos.org/); Unknown error”

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 RUN microdnf update -y && microdnf -y install which procps-ng git libffi-devel file xz bzip2 tar zip python3.11 python3.11-pip && microdnf clean all
-RUN python3 -m pip install --upgrade pip
+RUN python3 -m pip install --no-cache-dir --upgrade pip
 COPY . /src
-RUN pip install /src
+RUN python3 -m pip install --no-cache-dir /src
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM centos:7
-RUN yum install -y python-devel file zip gcc libffi-devel && yum clean all
-RUN curl https://bootstrap.pypa.io/pip/2.7/get-pip.py -o get-pip.py && python get-pip.py && rm get-pip.py
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+RUN microdnf update -y && microdnf -y install which procps-ng git libffi-devel file xz bzip2 tar zip python3.11 python3.11-pip && microdnf clean all
+RUN python3 -m pip install --upgrade pip
```
xiangce, line 3: `--no-cache-dir` is recommended in Dockerfile
```
python3 -m pip install --no-cache-dir --upgrade pip 
```

Dockerfile
```diff
@@ -1,5 +1,5 @@
...
 COPY . /src
 RUN pip install /src
```
xiangce, line 5: Better update this line as the same style as the new L-3, as `python3 -m pip install --no-cache-dir /src`

Dockerfile
chenlizhong: Updated

Dockerfile
chenlizhong: Updated

PR #4151, opened 2024-07-08
chore(ci): Test on Python 3.12

m-horky:
### All Pull Requests:

Check all that apply:

This patch ensures we test on a Python present in CentOS Stream 10/ RHEL 10.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,8 +11,9 @@ jobs:
 
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false
       matrix:
-        python-versions: [3.6, 3.9, 3.11]
+        python-versions: ["3.6", "3.9", "3.11", "3.12"]
 
     steps:
     - uses: actions/checkout@v4
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ runtime = set([
     'jinja2; python_version > "2.7"',
     'pyyaml>=3.10,<=3.13; python_version < "2.7"',
     'pyyaml; python_version >= "2.7"',
+    'setuptools; python_version >= "3.12"',
 ])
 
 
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -11,8 +11,9 @@ jobs:
 
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false
       matrix:
-        python-versions: [3.6, 3.9, 3.11]
+        python-versions: ["3.6", "3.9", "3.12"]
```
xiangce, line 16: Since the python 3.11 is still available from RHEL 9.2~, I think we'd better don't remove the `3.11` from the list.
- https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/installing_and_using_dynamic_programming_languages/assembly_installing-and-using-python_installing-and-using-dynamic-programming-languages

.github/workflows/main.yml
```diff
@@ -23,6 +24,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
+        python -m pip install setuptools
```
xiangce, line 27: As the `setuptools` is a required package for insights-core to run on python 3.12, how about adding it to the `setup.py` directly, e.g. `'setuptools; python_version >= "3.12"'`?

m-horky: Good point.

setup.py
```diff
@@ -53,7 +53,8 @@ def maybe_require(pkg):
 client = set([
     'requests',
     'python-gnupg==0.4.6',
-    'oyaml'
+    'oyaml',
+    'setuptools; python_version>="3.12"',
```
xiangce, line 57: Hi @m-horky - Would you mind moving it to the `runtime` set? As the `insights-core` can also be installed via `pip` and it provides some entry-points for users to run rules or do analysis on their localhost. E.g. https://github.com/RedHatInsights/insights-core/blob/e361b98df12483270cbeb22544b328e44a822a18/setup.py#L12-L24 

And the same error will also occur in this case if py312 is used in the local env.

xiangce (approved):
Thanks @m-horky and I will add the `python 3.13 test` to the `Required` list

PR #4166, opened 2024-07-17
feat: New spec "/etc/securetty" and parser

qinpingli:
### All Pull Requests:

Check all that apply:

*Add your description here*

### DIFF
--- a/insights/parsers/securetty.py
+++ b/insights/parsers/securetty.py
@@ -3,6 +3,7 @@ Secure -  file ``/etc/securetty``
 =================================
 """
 from insights.core import Parser
+from insights.core.exceptions import SkipComponent
 from insights.core.plugins import parser
 from insights.parsers import get_active_lines
 from insights.specs import Specs
@@ -24,12 +25,18 @@ class Securetty(Parser):
         terminals(list): a list of terminal names(without leading /dev/) which are considered
                          secure for the transmission of certain authentication tokens.
 
+   Raises:
+        SkipComponent: When the content is empty.
+
     Examples::
         >>> securetty.terminals
         ['console', 'tty1', 'tty2', 'tty3']
     """
 
     def parse_content(self, content):
+        _content = get_active_lines(content)
+        if not _content:
+            raise SkipComponent("No terminals found.")
         self.terminals = []
-        for line in get_active_lines(content):
+        for line in _content:
             self.terminals.append(line)
--- a/insights/parsers/securetty.py
+++ b/insights/parsers/securetty.py
@@ -1,0 +1,36 @@
+"""
+Secure -  file ``/etc/securetty``
+=================================
+"""
+from insights.core import Parser
+from insights.core.plugins import parser
+from insights.specs import Specs
+
+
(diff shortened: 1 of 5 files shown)


qinpingli:
@xiangce @JoySnow please help review, thanks!

insights/parsers/securetty.py
```diff
@@ -0,0 +1,36 @@
...
+        >>> securetty.terminals
+        ['console', 'tty1', 'tty2', 'tty3']
+    """
+
+    def parse_content(self, content):
+        self.terminals = []
+        for line in content:
+            if not line or line.startswith('#'):
+                continue
+            self.terminals.append(line)
```
JoySnow, line 36: Just in case...  When the `self.terminals` got a empty list  `[]` at the end of this parser, Are we good to take the empty list as a result OR Do we want to raise a `SkipComponent` like  [this example](https://github.com/RedHatInsights/insights-core/blob/master/insights/combiners/cpu_vulns_all.py#L43) ?

insights/parsers/securetty.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Examples::
+        >>> securetty.terminals
+        ['console', 'tty1', 'tty2', 'tty3']
+    """
+
+    def parse_content(self, content):
+        self.terminals = []
+        for line in content:
+            if not line or line.startswith('#'):
```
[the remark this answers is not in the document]
qinpingli: Thanks! Updated.

insights/parsers/securetty.py
qinpingli: Thanks for the reminder, returning an empty list here is intentional.

xiangce: Please add corresponding comment for this intention operation, as by default we do not keep empty parser result.

xiangce: And a `warning` note in the docstring is required as well. e.g.

```
.. warnning::
     Empty parseing result will be kept for this parser, please don't take it for granted that this parser is good to trigger the condition when it's an dependency component but without empty checking. 
     
     @condition(Securetty)
     def condition_1(sec_tty):
         if sec_tty is None:   # This is required to make sure there is something being parsed
               raise SkipComponent


```

xiangce: My bad. Currently, empty files will NOT be collected at all, so it will never reach this empty logic. That is, when the file is empty, this file will not appear in the archive, and therefore, the parser will not be triggerred.

To check if a file is empty or not, the best way is to add it to one of the `Ls` Parsers.

qinpingli: @JoySnow @xiangce Thanks for the review!  Because the file will not included in the archive when it's empty. I raise exception here to make the logic is the same for the rule developers when there are no terminals found.

JoySnow: Thanks for the update @qinpingli , then the `self.terminals` would not have a empty list result here.

JoySnow (approved):

xiangce (approved):

PR #4176, opened 2024-07-29
fix: 'filterable' specs were skipped incorrectly when running plugins

xiangce:
- 'filterable' is an option for collection but not running
   It should be acceptable to run plugins that depend on "filterable"
   spec but do not "add_filter" for the spec.
   And the results of running the plugin with or without "add_filter"
   should be the same.
- This change limits the validation of filter checking to data
   collection only. That is, the filters check only works when
   collecting a 'filterable' spec but not running it.
- tests are added as well

### All Pull Requests:

Check all that apply:

- see RHINENG-11728

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -184,15 +184,18 @@ class FileProvider(ContentProvider):
         # 1. No Such File
         if not os.path.exists(self.path):
             raise ContentException("%s does not exist." % self.path)
-        # 2. No Filters for 'filterable=True' Specs
-        if (self.ds and filters.ENABLED and
-                any(s.filterable for s in dr.get_registry_points(self.ds)) and
-                not filters.get_filters(self.ds)):
-            raise NoFilterException("Skipping %s due to no filters." % dr.get_name(self.ds))
-        # 3. Customer Prohibits Collection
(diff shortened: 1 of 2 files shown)


insights/core/spec_factory.py
```diff
@@ -184,15 +184,18 @@ def validate(self):
...
-            raise NoFilterException("Skipping %s due to no filters." % dr.get_name(self.ds))
-        # 3. Customer Prohibits Collection
-        if not blacklist.allow_file("/" + self.relative_path):
-            log.warning("WARNING: Skipping file %s", "/" + self.relative_path)
-            raise BlacklistedSpec()
+        # 2. Check only when collecting
+        if isinstance(self.ctx, HostContext):
+            # 2.1 No Filters for 'filterable=True' Specs
+            if (self.ds and filters.ENABLED and
+                    isinstance(self.ctx, HostContext) and
```
JoySnow, line 191: hey @xiangce , seems we can remove this `isinstance(self.ctx, HostContext)` checking , as it is assured by the `if` in line#188.

xiangce: Done. I forgot to remove it after level it up in the L188

PR #4177, opened 2024-07-29
Update __init__.py to fix JSONParser's list parsing implementation

sferich888:
The current list implementation of the json parser doesn't seem to work with josnl data; this should fix the issue.

### All Pull Requests:

Check all that apply:

When implementing a custom Parse based on the JSONParser that parses jsonl files (or jsonl data) this parser is unable to read the data due to the following error. 

`json.decoder.JSONDecodeError: Extra data: line 2 column 1 (char ####)`

This is caused because content (is a list), but we are not properly handing the conversion of said list into a proper large json array (or object); so that it can be properly loaded.  This correct that behavior.

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -782,7 +782,7 @@ class JSONParser(Parser, LegacyItemAccess):
     def parse_content(self, content):
         try:
             if isinstance(content, list):
-                self.data = json.loads('\n'.join(content))
+                self.data = json.loads("[" + ','.join(content) +  "]")
             else:
                 self.data = json.loads(content)
         except:


bfahr:
@sferich888 you are correct that the existing JSON parser does not handle [JSONL format](https://jsonlines.org/).  It expects the content collected to be a complete, single set of JSON data.  I think it might make more sense to create a new parser that can handle the JSONL format since any rule using that format would probably expect to get a list of similar JSON inputs rather than one huge JSON input.

xiangce:
I totally agree what @bfahr mentioned that create a new parser to handle the JSONL format, as currently there are many parsers used this JSONParser as super class, and this change will break their usage. @sferich888 Would you mind adding the new base parser for JSONL by yourself? or leave it to us?

sferich888:
@bfahr / @xiangce  I have submitted a PR that adds the JSONL Parser (trying to keep is as similar to the JSONParser as possible - only altering the behavior needed to change the way the multi-lines are processed.

xiangce:
@sferich888 - Thanks a lot for making this change, there are a bit of trifles need you to have a look:

1. Please fix the DCO failure as per [here](https://github.com/RedHatInsights/insights-core/pull/4177/checks?check_run_id=28106080922)
2. Please fix the flak8 error `./insights/core/__init__.py:777:1: E302 expected 2 blank lines, found 1`
3. Would you please add a test for the new `JSONLParser` as well? See the [test for JSONParser](https://github.com/RedHatInsights/insights-core/blob/3.0/insights/tests/test_json_parser.py)

Thanks.

xiangce:
Hi @sferich888 - Sorry for bringing this trouble to you, as we just resolved an urgent bug within the repo. Please rebase the master branch of your fork of insights-core and then rebase this branch again. Please reach out to me for any problems when doing the rebase. Apologize again.

PR #4178, opened 2024-07-30
Fix and test IdM related parsers and combiners

pbrezina:
### All Pull Requests:

Check all that apply:

This pull request fix multiple issues in ipa, sssd and identity_domain. Each commit has its own description about the fix. It also adds extensive tests to verify its functionality.

### DIFF
--- a/insights/combiners/sssd_conf.py
+++ b/insights/combiners/sssd_conf.py
@@ -1,0 +1,154 @@
+"""
+SSSD Configuration
+==================
+
+Provides access to complete SSSD configuration: /etc/sssd/sssd.conf with merged
+configuration snippets from /etc/sssd/conf.d.
+"""
+
+from copy import deepcopy
+
+from insights.core.plugins import combiner
+from insights.parsers.sssd_conf import SSSDConf, SSSDConfd
+
+
+@combiner(SSSDConf, SSSDConfd)
+class SSSDConfAll(object):
+    """
+    Provides access to complete SSSD configuration: /etc/sssd/sssd.conf with
+    merged configuration snippets from /etc/sssd/conf.d.
+    """
+    def __init__(self, sssd_conf, sssd_conf_d):
+        self.config = deepcopy(sssd_conf)
+
+        for parser in sorted(sssd_conf_d, key=lambda x: x.file_name):
+            if parser.file_name.startswith("."):
+                continue
+
+            for section in parser.sections():
+                for key, value in parser.items(section).items():
+                    self.config._set(section, key, value)
+
+        self._enabled_domains = None
+
+    @property
+    def enabled_domains(self):
+        """
+        Returns the list of enabled domains.
+
+        Domains can be enabled either using the ``domains`` option in the
+        ``sssd`` section of the configuration file or using the ``enabled``
+        option in the domain configuration.
+
+        [sssd]
+        domains = a, b
+
+        [domain/a]
+        ...
+
+        [domain/b]
+        ...
+
+        [domain/c]
+        enabled = true
+
+        """
+        if self._enabled_domains is None:
+            enabled_domains = []
+
+            if self.config.has_option("sssd", "domains"):
+                domains = self.config.get("sssd", "domains")
+                enabled_domains = [domain.strip() for domain in domains.split(",")]
+
+            prefix = "domain/"
+            for section in self.config.sections():
+                # Ignore if this is not a domain configuration
+                if not section.startswith(prefix):
+                    continue
+
+                name = section[len(prefix):].strip()
+                if not name:
+                    # Invalid configuration
+                    continue
+
+                # Ignore if this is a subdomain configuration
+                # `domain/$dom/$subdom`
+                if "/" in name:
+                    continue
+
+                if self.config.has_option(section, "enabled"):
+                    enabled = self.config.getboolean(section, "enabled")
+
+                    if enabled and name not in enabled_domains:
+                        enabled_domains.append(name)
+                    elif not enabled and name in enabled_domains:
+                        enabled_domains.remove(name)
+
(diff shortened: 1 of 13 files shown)


xiangce (changes requested):
@pbrezina - Since the insights-core is a public open-source project, other than the internal projects, there might be other unknown project depends on it, we'd make sure the new changes won't break their usage.

BTW, May I know the context about why you raise this PR, e.g., a Jira card or a customer case?

pbrezina:
> @pbrezina - Since the insights-core is a public open-source project, other than the internal projects, there might be other unknown project depends on it, we'd make sure the new changes won't break their usage.

Sure, I'll look into it.

> BTW, May I know the context about why you raise this PR, e.g., a Jira card or a customer case?

I'll send the ticket to you privately, as it is not publicly accessible.

insights/specs/__init__.py
```diff
@@ -727,7 +727,8 @@ class Specs(SpecSet):
     sshd_config_d = RegistryPoint(multi_output=True, filterable=True)
     sshd_config_perms = RegistryPoint(no_obfuscate=['hostname', 'ip'])
     sshd_test_mode = RegistryPoint(filterable=True)
-    sssd_config = RegistryPoint()
+    sssd_conf = RegistryPoint()
```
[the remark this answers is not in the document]
pbrezina: I wanted to follow the actual name pattern. But I will keep `sssd_config`.

insights/specs/__init__.py
```diff
@@ -727,7 +727,8 @@ class Specs(SpecSet):
...
+    sssd_conf_d = RegistryPoint(multi_output=True)
```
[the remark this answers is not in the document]
pbrezina: https://issues.redhat.com/browse/RHINENG-11838

insights/parsers/sssd_conf.py
```diff
@@ -1,106 +1,82 @@
...
+---------------------------------------------------
 """
+
 from insights.core import IniConfigFile
 from insights.core.plugins import parser
 from insights.specs import Specs
 
 
-@parser(Specs.sssd_config)
-class SSSD_Config(IniConfigFile):
```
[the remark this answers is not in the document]
pbrezina: Done. I kept 3.6.0 as the last version, feel free to suggest different.

pbrezina:
@xiangce Please, see updated patches.

insights/parsers/sssd_conf.py
```diff
@@ -1,10 +1,86 @@
...
+            "true": True,
+            "false": False,
+        }
+
+        if val.lower() not in boolean_states:
+            raise ValueError("Not a boolean: %s" % val)
+
+        return boolean_states[val.lower()]
+
+    def set(self, section, option, value=None):
```
xiangce, line 57: I just noticed that this `set` method was added to the Parser (sorry for the late). Generally, this kind of `write-back-to-parser` operations is not recommended for insights. Since it's very common for insights to run a bunch of plugins serially in one pass of `insights-run`, if one of the plugins calls this `set` method and writes back to the Parser by mistake, it will affect the results of the following plugins that also depends on this Parser. (we have ever met such issue, and it's very difficult to debug :-().

Although, without this method, there are many ways can be used to write-back and pollution the parser indirectly,  here the `not recommended` means that we'd better do not provide such an interface publicly.

insights/combiners/sssd_conf.py
```diff
@@ -0,0 +1,142 @@
+from copy import deepcopy
```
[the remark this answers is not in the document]
pbrezina: Done.

insights/parsers/sssd_conf.py
pbrezina: I added a note and rename it to _set. I need this in the combiner to merge conf.d into the configuration. The parser is deepcopied in the container so the original parser is not changed. Let me know if you prefer different way.

xiangce (approved):

insights/parsers/sssd_conf.py
xiangce: Looks good to me. Thanks for the change!

pbrezina:
Fixed docs.

insights/parsers/sssd_conf.py
```diff
@@ -1,10 +1,88 @@
...
+
+    [sudo]
+    debug_level = 0xfff0
+
+    [domain/test]
+    id_provider = ldap
+    ldap_uri = ldap://ldap.test
+
+
+SSSDConf - files ``/etc/sssd/sssd.conf``
```
JoySnow, line 20: Seems a "typo" here, `file` would be okay.

JoySnow:
@pbrezina Just a reminder... the PR CI code-tests runs into the flake8 errors, eg. `./insights/combiners/sssd_conf.py:6:46: W605 invalid escape sequence '\*'`

JoySnow (approved):

insights/parsers/sssd_conf.py
```diff
@@ -79,6 +157,10 @@ class SSSD_Config(IniConfigFile):
         >>> 'ldap_uri' in domain
         True
     """
+    def __init__(self, context):
+        deprecated(SSSD_Config, "Please use the :class:`insights.parsers.sssd_conf.SSSDConf` instead.", "3.6.0")
```
xiangce, line 161: @pbrezina - sorry for the intermittent comment to the PR, please check if we can change suggested parser in the msg with `insights.combiners.sssd_conf.SSSDConfAll`? As you also added a new combiner for "etc/sssd/sssd.conf" and "etc/sssd/conf.d/*.conf", the combiner might be the right component for user to use.

pbrezina: Done.

PR #4182, opened 2024-08-05
fix: Run subshells with LC_ALL=C.UTF-8

m-horky:
### All Pull Requests:

Check all that apply:

* Card ID: RHINENG-11802
* Card ID: CCT-653

Ensure we get Unicode data back when we run GPG in a subshell. This will ensure we are able to parse output of the command even if the host is configured to use non-UTF-8 character set.

### DIFF
--- a/insights/client/crypto.py
+++ b/insights/client/crypto.py
@@ -95,7 +95,7 @@ class GPGCommand(object):
             ["/usr/bin/gpg", "--version"],
             stdout=subprocess.PIPE, stderr=subprocess.PIPE,
             universal_newlines=True,
-            env={"GNUPGHOME": self._home},
+            env={"GNUPGHOME": self._home, "LC_ALL": "C.UTF-8"},
         )
         stdout, stderr = version_process.communicate()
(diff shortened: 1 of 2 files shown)


ptoscano (commented):
@xiangce is `C.UTF-8` actually usable on eg RHEL 6? I'm not sure it is available by default in RHEL 6, and indeed the parsers seem to use `LC_ALL=C`; OTOH, the `subscription-manager` parsers use `LC_ALL=C.UTF-8`, and I'm not sure whether there were compliants/issues reported about that.

Otherwise this LGTM.

xiangce (commented):
Looks good to me too.

@ptoscano - After the `override_env={"LC_ALL": "C.UTF-8"}` added to the `subscription_manager_*`, seems there are no more reports of this kind of issues for these specs. And as the RHEL 6 was just fully EOL, I think we can omit the support of RHEL 6.

BTW, we planned to remove all supports of RHEL 6&py26 around this Sep. 1

xiangce (approved):

PR #4183, opened 2024-08-06
fix: Skip empty spec only when collecting

### LINKED ISSUE #4180: OSError: [Errno 90] Message too long
This issue was brought to my attention by some support engineers, it appears when there is a large amount of content exceptions logged from parsers. It eventually hits a limit, and the below exception occurs when logging exceptions to syslog with the -s parameter. I can provide more information on how to reproduce this internally.

```
$ insights run -p <rules_repo_name> path/to/must-gather/
---------
Progress:
---------
--- Logging error ---
  File "/usr/lib64/python3.11/logging/handlers.py", line 1005, in emit
    self.socket.send(msg)
OSError: [Errno 90] Message too long

During handling of the above exception, another exception occurred:

  File "/usr/lib64/python3.11/logging/handlers.py", line 1009, in emit
    self.socket.send(msg)
OSError: [Errno 90] Message too long
```

xiangce:
- When running plugins, print a debug information only for 
 empty spec, but not raise Exception
- As before, the empty spec will be collected, but the
  reason will be recorded as an ContentException in its
  meta_data JSON
- JIRA: RHINENG-11843
- And change to show the full real path instead of relative path
  in the debug message
- Fix the relevant test of compliance ds
- fix #4180

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -95,7 +95,7 @@ class ContentProvider(object):
                     obf_funcs=self.cleaner.get_obfuscate_functions(self.relative_path, no_obf),
                     no_redact=no_red)
                 if len(self._content) == 0:
-                    log.debug("Skipping %s due to empty after cleaning", "/" + self.relative_path)
+                    log.debug("Skipping %s due to empty after cleaning", self.path)
                     raise ContentException("Empty after cleaning: %s" % self.path)
             else:
                 log.debug("Skipping cleaning %s", "/" + self.relative_path)
@@ -117,7 +117,10 @@ class ContentProvider(object):
                 raise
 
         if len(self._content) == 0:
-            raise ContentException("Empty (after filtering): %s" % self.path)
+            log.debug("File is empty (after filtering): %s", self.path)
+            if isinstance(self.ctx, HostContext):
+                # Do not collect empty spec
+                raise ContentException("Empty (after filtering): %s" % self.path)
 
         return self._content
 
--- a/insights/tests/datasources/compliance/test_compliance_ds.py
+++ b/insights/tests/datasources/compliance/test_compliance_ds.py
@@ -8,7 +8,7 @@ except Exception:
 from pytest import raises
 from tempfile import NamedTemporaryFile
 
-from insights.core.exceptions import SkipComponent, ContentException
+from insights.core.exceptions import SkipComponent
 from insights.parsers.installed_rpms import InstalledRpms
 from insights.parsers.os_release import OsRelease
 from insights.parsers.redhat_release import RedhatRelease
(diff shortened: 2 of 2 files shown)


JoySnow (approved):

xiangce:
@ryan-blakley - Would you mind reviewing this PR, it would fix the issue #4180

insights/core/spec_factory.py
```diff
@@ -117,7 +117,10 @@ def content(self):
                 raise
 
         if len(self._content) == 0:
-            raise ContentException("Empty (after filtering): %s" % self.path)
+            log.debug("File is empty (after filtering): %s", self.path)
```
ryan-blakley, line 120: @xiangce It's been a while since I've worked on core stuff, but I was wondering would it be better to raise a SkipException instead here. Or would that trigger the same issue? It's already kind of hard sometimes finding out why a rule wasn't ran, now without an exception being thrown and logged. I feel like it might be even harder to determine this, but I could be thinking about it wrong and was curious about your thoughts.

xiangce: @ryan-blakley - yeah, it's a long time 😄. I think you meant `SkipComponent` (the SkipException was removed). Raising a `SkipComponent`, in most cases, it won't cause any issues. However, in our experience of rules development, it occasionally happens that, for example, it is expected that a particular keyword does not appear in a file, or that a particular directive is not explicitly configured in a "fitlterable" configuration file, but the application sets default values for those unconfigured directives, etc. In these cases, if we `Skip` these Specs in the core, the condition that depends on the Specs/Parsers will be skipped too per the dependency graph. For example:
```
@condition(Specs.A_filterable_Spec):
def no_such_line_in_file(spec):
    if "such_line" in spec.lines:
        return False
    return True

@rule(..., no_such_line_in_file, ...) 
def report(...)
```
If the above `A_filterable_Spec` is `Skipped` by the core, the condition `no_such_line_in_file` will never be triggered and hence the `report` will also be skipped.


Although we have workarounds to avoid this situation, or implement it in a `correct` way if we know the background that the spec will be Skipped when it's filtered empty, it may not mean that all rule developers are aware of this. If still implementing the `conditions` in the traditional style, we may get false positives.

ryan-blakley: Sorry, yeah I meant SkipComponent, I should have remembered that since I removed the SkipException me--. Ahhh gotcha I didn't realize the ContentException raise was recently added, thanks for the explanation, everything looks good then to me.

ryan-blakley (approved):

insights/core/spec_factory.py
xiangce: Thanks @ryan-blakley

Issue #4187, opened 2024-08-09
Enhance `simple_command` to close stdin to avoid this class of bugs in the future.


xiangce:
I'd suggest to enhance `simple_command` to close stdin to avoid this class of bugs in the future.

_Originally posted by @ekohl in https://github.com/RedHatInsights/insights-core/pull/4186#pullrequestreview-2229861161_
            
For more details see:
- https://github.com/theforeman/foreman/pull/10267
- RHINENG-11883

ShimShtein:
I think closing the stdin can also influence the caller process, which is undesirable.

Maybe the `simple_command` should use the equivalent of ruby's [popen3](https://rubyapi.org/3.3/o/open3#method-i-popen3): 
> Creates streams stdin, stdout, and stderr, which are the standard input, standard output, and standard error streams in the child process.

This way the child process would be completely isolated from the caller.

PR #4189, opened 2024-08-12
fix: pass stdin=DEVNULL to Popen to avoid eating stdin from pipes

xiangce:
- For the first cmd to run in subprocess.call, specified
  "stdin=subprocess.DEVNULL" to avoid it from eating
  the stdin behind it
- fix: #4187

### All Pull Requests:

Check all that apply:

*Add your description here*


xiangce:
Hi @bfahr, @ryan-blakley, @JoySnow, @ekohl, @ShimShtein - this is a basic change to the `util.subproc` of the core, would you please help review it whenever free, Thanks in advance.

insights/util/subproc.py
```diff
@@ -57,15 +62,19 @@ def __init__(self, *cmds, **kwargs):
     def _build_pipes(self, out_stream=PIPE):
         log.debug("Executing: %s" % str(self.cmds))
```
ekohl, line 63: Completely unrelated, but I just noticed this while reading the surrounding code. Not for this PR, but a general question.

Why isn't this using the interpolation that `logging` already has built in? That way you don't pay the `str()` and interpolation overhead if the logging level is above debug.
```suggestion
        log.debug("Executing: %s", self.cmds)
```

ekohl (approved):
The patch is harder to read because of the reformatting changes. I always like to separate those out to separate commits, making it obvious for reviewers what's going on. So the reduced patch is:
```patch
---
 insights/util/subproc.py | 9 +++++++--
 1 file changed, 7 insertions(+), 2 deletions(-)

diff --git a/insights/util/subproc.py b/insights/util/subproc.py
index 547ff0e5ca2a..d3cd46c4c25c 100644
--- a/insights/util/subproc.py
+++ b/insights/util/subproc.py
@@ -10,6 +10,11 @@ from subprocess import Popen, PIPE, STDOUT
 from insights.core.exceptions import CalledProcessError
 from insights.util import which
 
+try:
+    from subprocess import DEVNULL
+except:
+    DEVNULL = open(os.devnull, 'w')
+
 log = logging.getLogger(__name__)
 
 
@@ -57,9 +62,9 @@ class Pipeline(object):
     def _build_pipes(self, out_stream=PIPE):
         log.debug("Executing: %s" % str(self.cmds))
         if len(self.cmds) == 1:
-            return Popen(self.cmds[0], bufsize=self.bufsize, stderr=STDOUT, stdout=out_stream, env=self.env)
+            return Popen(self.cmds[0], bufsize=self.bufsize, stdin=DEVNULL, stderr=STDOUT, stdout=out_stream, env=self.env)
 
-        stdout = Popen(self.cmds[0], bufsize=self.bufsize, stderr=STDOUT, stdout=PIPE, env=self.env).stdout
+        stdout = Popen(self.cmds[0], bufsize=self.bufsize, stdin=DEVNULL, stderr=STDOUT, stdout=PIPE, env=self.env).stdout
         last = len(self.cmds) - 2
         for i, arg in enumerate(self.cmds[1:]):
             if i < last:
```

That looks ok.

However, I think the function is way too complex and I think this is the most simple version:
```python
    def _build_pipes(self, out_stream=PIPE):
        if len(self.cmds) == 0:
            return

        log.debug("Executing: %s", self.cmds)
        stdout = DEVNULL
        for arg in self.cmds[:-1]:
            stdout = Popen(arg, bufsize=self.bufsize, stdin=stdout, stderr=STDOUT, stdout=PIPE, env=self.env).stdout
        return Popen(self.cmds[-1], bufsize=self.bufsize, stdin=stdout, stderr=STDOUT, stdout=out_stream, env=self.env)
```

Or if you really want to make it obvious what the differing args are:
```python
    def _build_pipes(self, out_stream=PIPE):
        if len(self.cmds) == 0:
            return

        log.debug("Executing: %s", self.cmds)
        stdout = DEVNULL
        kwargs = {
            'bufsize': self.bufsize,
            'stderr': STDOUT,
            'env': self.env
        }
        for arg in self.cmds[:-1]:
            stdout = Popen(arg, stdin=stdout, stdout=PIPE, **kwargs).stdout
        return Popen(self.cmds[-1], stdin=stdout, stdout=out_stream, **kwargs)
```

I'm not a maintainer on the project, so take it as you prefer. Your fix is correct and resolves the original problem.

ShimShtein:
Didn't test it, but looks good for me too.

ekohl:
> ```python
>         if len(self.cmds) == 0:
> ```

And this can even be simplified further:
```python
if not self.cmds:
```

xiangce:
@ekohl and @ShimShtein - Thanks for the review. Regarding the other comments, I'd like to keep them as they were in this change. And we can update them later if necessary.

insights/util/subproc.py
xiangce: Based on your suggestion, I've un-done the previous line breaks in the new commit to keep the changes minimal for this update. Thanks again @ekohl

PR #4193, opened 2024-08-19
feat: New spec "lsattr <paths>" and its parser

huali027:
### All Pull Requests:

Check all that apply:


Lorquas:

huali027:

JoySnow (approved):
LGTM @huali027

insights/parsers/lsattr.py
```diff
@@ -0,0 +1,45 @@
...
+        'i'
+    """
+    def parse_content(self, content):
+        for line in content:
+            items = line.split(None, 1)
+            if len(items) == 2:
+                attr = items[0].strip().replace('-', '')
+                # still store the file when the attribute is empty, in case
+                # a direcotry is checked and we can not know if the file is not
+                # in the dir, or the attribute is empty if we don't store it.
```
xiangce, line 42: 1. Is a `directory` a good candidate filter for this `lsattr_files`? From the name, it seems that only `file` is acceptable. Will you want it allows `directory` to be `add_filter`ed?
2. It would be better to store the bad results with a specific `value` (e.g. "n/a"). E.g. `lsattr: No such file or directory while trying to stat a/e`. It is sometimes useful, in case the `path` doesn't exist on the host.

insights/specs/datasources/lsattr.py
```diff
@@ -0,0 +1,16 @@
...
+from insights.core.context import HostContext
+from insights.core.exceptions import SkipComponent
+from insights.core.filters import get_filters
+from insights.core.plugins import datasource
+from insights.specs import Specs
+
+
+@datasource(HostContext)
+def files_of_lsattr(broker):
+    filters = sorted(get_filters(Specs.lsattr_files))
```
xiangce, line 13: Please add a bit docstring for this method.

insights/parsers/lsattr.py
huali027: @xiangce for 1, yes, a directory is acceptable, and if a directory is used, it will list all the files in the directory.  Or I rename it to make it more readable.
For the others, I'll update it.

xiangce (approved):

PR #4195, opened 2024-08-20
remove support of RHEL 6 & python 2.6 from collection

xiangce:
- and mark the parsers for RHEL 6 only as deprecated (3.5.0)
- remove the CI/CD pipeline for python 2.6
- remove relevant unused test cases
- Jira: RHINENG-11287

### All Pull Requests:

Check all that apply:

**NOTE:** 
- This PR can only be merged and released in the week of 29, Aug


xiangce:
@ptoscano and @m-horky, please also kindly review when free, there are some changes about the test of `playbook_verifier". Thanks in advance.

JoySnow (approved):

m-horky (changes requested):
@xiangce Hi, as a representative of a team responsible for the Client I don't think this should be merged. This will cause all insights-client services on RHEL 6 to crash, which may result in many customer cases and work needed on customer side in general.

I think you should reconsider and implement the one-egg-per-rhel-major idea instead of simply dropping support. I do not like this.

PR #4203, opened 2024-08-29
Fix skipped assertions for itests

loleary:
### All Pull Requests:

Check all that apply:

Using `run_test` with `expected` parameter **DOES NOT** perform expected assertion if `expected` value evaluates to false. For example, if `expected` is an empty list or dict or str or a bool value of `False`. The result is that no assertion is performed meaning that when checking that an integration test for a `condition`, for example, returns `False` but it instead, in error, returns `True`, `run_test` does not perform the assertion giving the impression that the test passes even though it does not.

This change fixes this by using a default sentinel value for `expected` and always performing the assertion unless `expected` is the sentinel value.

This change required updating an existing integration test. Additionally, it is important to note that this change will result in existing integration tests to fail which have fallen victim to this bug as well as any calls to `run_test` that explicitly pass `expected=None` thinking that it should mean _do not perform any assertions_.


Lorquas:

xiangce:
Hi @loleary - Sorry for bringing this trouble, as we just resolved an urgent bug within the repo.   Please rebase the master branch of your fork of insights-core and then rebase this branch again. Please reach out to me for any problems when doing the rebase. Apologize again.

PR #4211, opened 2024-09-10
Catch Timeout on test connection

Glutexo:
### All Pull Requests:

Check all that apply:

The `--test-connection` command only used to catch `ConnectionError`, including `ConnectionTimeout`, but not other timeouts such as `ReadTimeoutError`. Re-used the existing `REQUEST_FAILED_EXCEPTIONS` list to process all timeouts in the same way.

There is an integration test pull request https://github.com/RedHatInsights/insights-client/pull/244 for this fix.

Card IDs:
* [CCT-506](https://issues.redhat.com/browse/CCT-506)

### DIFF
--- a/insights/tests/client/connection/test_test_connection.py
+++ b/insights/tests/client/connection/test_test_connection.py
@@ -16,18 +16,16 @@ LEGACY_URL = "https://www.example.com"
 LEGACY_URL_SUFFIXES = ["/insights/", "", "/r", "/r/insights"]
 
 EXCEPTION_MESSAGE = "request exception"
-EXPECTED_EXCEPTIONS_TEST_CONNECTION = [
+EXPECTED_EXCEPTIONS = [
     requests.exceptions.SSLError,
     requests.exceptions.ConnectTimeout,
-]
-EXPECTED_EXCEPTIONS_TEST_URLS = EXPECTED_EXCEPTIONS_TEST_CONNECTION + [
     requests.exceptions.ReadTimeout,
-]
-UNEXPECTED_EXCEPTIONS_TEST_CONNECTION = [
-    requests.exceptions.ReadTimeout,
-    UnexpectedException,
 ]
 
+parametrize_exceptions = pytest.mark.parametrize(
+    ["exception"],
+    [(exception,) for exception in EXPECTED_EXCEPTIONS],
+)
 parametrize_methods = pytest.mark.parametrize(
     ["http_method", "request_function"],
     [
@@ -35,13 +33,6 @@ parametrize_methods = pytest.mark.parametrize(
         ("POST", "insights.client.connection.InsightsConnection.post"),
     ],
 )
-
-
-def parametrize_exceptions(exceptions):
-    return pytest.mark.parametrize(
-        ["exception"],
-        [(exception,) for exception in exceptions],
-    )
 
 
 @pytest.fixture
@@ -130,7 +121,7 @@ def test_test_urls_get_success(get, temporary_file, post, insights_connection):
     post.assert_not_called()
 
 
-@parametrize_exceptions(EXPECTED_EXCEPTIONS_TEST_URLS)
+@parametrize_exceptions
 @mock.patch("insights.client.connection.InsightsConnection.post")
 @mock.patch("insights.client.connection.TemporaryFile")
 @mock.patch("insights.client.connection.InsightsConnection.get")
@@ -177,7 +168,7 @@ def test_test_urls_post_success(get, temporary_file, post, insights_connection):
     get.assert_not_called()
 
 
-@parametrize_exceptions(EXPECTED_EXCEPTIONS_TEST_URLS)
+@parametrize_exceptions
 @mock.patch("insights.client.connection.InsightsConnection.post")
 @mock.patch("insights.client.connection.TemporaryFile")
 @mock.patch("insights.client.connection.InsightsConnection.get")
@@ -213,7 +204,7 @@ def test_test_urls_no_catch(
     assert caught.value is raised
 
 
-@parametrize_exceptions(EXPECTED_EXCEPTIONS_TEST_URLS)
+@parametrize_exceptions
 @parametrize_methods
 @mock.patch("insights.client.connection.TemporaryFile")
 def test_test_urls_raise(
@@ -264,7 +255,7 @@ def test_test_urls_error_log_success(
     logger.error.assert_not_called()
 
 
-@parametrize_exceptions(EXPECTED_EXCEPTIONS_TEST_URLS)
+@parametrize_exceptions
 @parametrize_methods
 @mock.patch("insights.client.connection.logger")
 def test_test_urls_error_log_fail(
@@ -280,9 +271,9 @@ def test_test_urls_error_log_fail(
     except exception:
         pass
 
-    logger.error.assert_called_once_with(
-        "Could not successfully connect to: %s", UPLOAD_URL
-    )
+    assert logger.error.mock_calls == [
+        mock.call("Could not successfully connect to: %s", UPLOAD_URL)
+    ]
 
 
 @parametrize_methods
@@ -319,7 +310,7 @@ def test_test_urls_print_success(
     assert not err
 
 
-@parametrize_exceptions(EXPECTED_EXCEPTIONS_TEST_URLS)
+@parametrize_exceptions
 @parametrize_methods
 def test_test_urls_print_fail(
     http_method, request_function, exception, insights_connection, capsys
@@ -364,7 +355,7 @@ def test_legacy_urls_get_success(get, post, insights_connection):
     post.assert_not_called()
 
 
-@parametrize_exceptions(EXPECTED_EXCEPTIONS_TEST_URLS)
+@parametrize_exceptions
 @mock.patch("insights.client.connection.InsightsConnection.post")
 @mock.patch("insights.client.connection.InsightsConnection.get")
 def test_legacy_urls_get_one_fail(get, post, exception, insights_connection):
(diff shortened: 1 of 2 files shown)


insights/tests/client/connection/test_test_connection.py
```diff
@@ -0,0 +1,175 @@
...
+    assert not err
+
+
+@patch("insights.client.connection.InsightsConnection._test_urls")
+def test_test_connection_print_success(test_urls, insights_connection, capsys):
+    """The connection test prints a message pointing to a log file if all API calls succeed."""
+    insights_connection.test_connection()
+
+    out, err = capsys.readouterr()
+    assert out == "See %s for more details.\n" % (LOGGING_FILE,)
```
m-horky, line 129: Could you please use `.format()`-style formatting instead? This is harder to read.

Non-blocking comment.

insights/tests/client/connection/test_test_connection.py
```diff
@@ -0,0 +1,175 @@
...
+        insights_connection.test_connection()
+
+    assert caught.value is expected
+
+
+@patch("insights.client.connection.InsightsConnection._test_urls")
+def test_test_connection_return_code_success(test_urls, insights_connection):
+    """The connection test returns a zero code if all API calls succeed."""
+    result = insights_connection.test_connection()
+    assert result == 0
```
m-horky, line 165: Could this be merged with `test_test_connection_test_urls_success`?

insights/tests/client/connection/test_test_connection.py
```diff
@@ -0,0 +1,175 @@
+from logging import ERROR
```
m-horky, line 1: Instead of importing in this object, could you please use `logging.ERROR` in the tests? There are several similarly-sounding variables and it's not clear from the context that this one is a logger level (not everyone is familiar with the `caplog` API).

insights/tests/client/connection/test_test_connection.py
```diff
@@ -0,0 +1,175 @@
...
+
+@patch("insights.client.connection.InsightsConnection._test_urls")
+def test_test_connection_test_urls_no_catch(test_urls, insights_connection):
+    """The connection test is stopped in case of an unknown exception."""
+    test_urls.side_effect = Exception
+
+    try:
+        insights_connection.test_connection()
+    except Exception:
+        pass
```
m-horky, line 43: Would it make sense to define a custom `Exception` subclass only for this file? E.g. `class SideEffectException(Exception): pass`.

This way we could be sure that we're catching this mocked error and not something else from within the code that got raised by accident. This applies to all occurrences.

m-horky (changes requested):
Overall, this PR looks good, I like the tests. I left a few notes there to consider.

insights/tests/client/connection/test_test_connection.py
Glutexo: That’s very good observation! I was thinking about that and thought it would be too much of a hassle. Will do!

insights/tests/client/connection/test_test_connection.py
Glutexo: Sure thing! Probably importing just `pyrest` for `mark.parametrize` and similar would make sense too.

xiangce:
FYI, we can ignore the following failed test:
```
>               assert hasattr(DefaultSpecs, spec_name)
E               AssertionError: assert False
E                +  where False = hasattr(DefaultSpecs, 'cloud_init_cfg_run')
```
The `cloud_init_cfg_run` was just removed from core collection and as well from the `uploader.json`, however, the uploader.json on CDN will be updated after next release. Let's ignore it this time. And the `uploader.json` will be deprecated soon, as well as this test.

insights/tests/client/connection/test_test_connection.py
Glutexo: I am not sure this is a correct thing to do. The `test_test_connection_test_urls_*` tests only verify the calls to the `_test_urls` method. Hence the `test_urls` in their names. The `test_test_connection_return_code_*` ones verify the return value, which is a very different thing.

We can remove the two `return_code` tests from this pull request as they are not related to the error messages. I put it there however, to cover the case when the function doesn’t throw (or let pass uncaught) any exception.

Glutexo:
Addressed the review comments except one. See my inline comment.

insights/tests/client/connection/test_test_connection.py
Glutexo: More thoughts:

To convey the idea that the return value check is there only for the tests to have at least some meaningful `assert`, I put the `result > 0` rather than `result == 1` in the other return value test. They are there only to verify the call doesn’t fail with an error, not testing the return values precisely.

Maybe `assert True` would convey this even better. But that is implicitly tested by all other tests calling the `_test_urls` method. Seeing that I think the return value tests can be safely removed.

Glutexo:
Resolved the last review comment and rebased on the current master. Ready for a review.

Glutexo:
[pytest-catchlog](https://pypi.org/project/pytest-catchlog/) providing the _caplog_ fixture became part of Pytest since version 3.3. On previous ones it needs to be installed separately. Added the dependency to `setup.py` for Python < 2.7. Let’s see what the pipeline says.

Glutexo:
Adding the dependency to [setup.py](https://github.com/Glutexo/insights-core/blob/144dccc3aa2ab7cdd790b0e4896e69664f2e64a6/setup.py) is not enough. The dependencies are grabbed from another [repository](https://github.com/SteveHNH/jenkins-s2i-example). I created a [pull request](https://github.com/SteveHNH/jenkins-s2i-example/pull/5) there, but I am afraid it’s never going to be reviewed and merged as it’s a personal repo not owned by RedHatInsights.

I created a [card](https://issues.redhat.com/browse/CCT-768) suggesting migration of the requirements. Since [jenkins-s2i-example](https://github.com/SteveHNH/jenkins-s2i-example) is only used to provide the requirements file and its wheels, I think we can ditch the repo and move the files to this repository. We can even only adopt ci_requirements.txt and curl the wheels directly from PyPI, which is what _pip_ does for other Python versions anyways.

For now, I can either mark the tests as skipped for Python 2.6, or update them to check for function calls to the logger. The latter is suboptimal, can can do for the time being until CCT-768 is resolved. What do you think?

m-horky:
Aha, interesting. At this point I think those tests could be skipped, let's not spend more time on this than needed.

Glutexo:
The _test_connection_ method only use a single very simple error logging call. It was trivial to replace it with a logger mock. It doesn’t even use logging string interpolation, so the asserts don’t need to incorrectly test the exact arguments rather than the resulting message.

The other tests of the _\_test\_urls\__ and _\_legacy\_test\_urls_ methods use more complex logger calls, but they are no longer part of this pull request.

Glutexo:
Fixed the remaining Flake8 complaint. Now, everything should be ok. Please take a look at the alternative asserts without _caplog_.

m-horky (approved):
LGTM.

m-horky:
Hello @xiangce, this PR is ready for merge.

Glutexo:
Updated on the current _master_.

xiangce:
@m-horky, @Glutexo - So, we are okay to merge this PR, right? I received your comment tagging me from github's notfi, but didn't find it here in the PR (you deleted it?), so having this query.

m-horky:
@xiangce Hi, after I tagged you I realized it didn't go through QE preverification yet. Sorry for the noise.

Glutexo:
If this gets merged before https://github.com/RedHatInsights/insights-core/pull/4246, that pull request would bring a regression. See my [comment](https://github.com/RedHatInsights/insights-core/pull/4246#issuecomment-2410440279) there for more details.

m-horky:
QE PASSED, this patch has been verified by @zpetrace. Blocked by #4246 now.

Glutexo:
Cool! So to make things simpler, let‘s say this is blocked by https://github.com/RedHatInsights/insights-core/pull/4246 which itself is blocked by https://github.com/RedHatInsights/insights-core/pull/4248.  Merging those first will make merging easier and would significantly reduce the size of the test patch.

Glutexo:

Glutexo:
https://github.com/RedHatInsights/insights-core/pull/4248 is now merged. This is currently only blocked by https://github.com/RedHatInsights/insights-core/pull/4246.

Glutexo:
To reduce the number of conflicts, rebased on the new _master_ with https://github.com/RedHatInsights/insights-core/pull/4248 merged in. https://github.com/RedHatInsights/insights-core/pull/4246 got merged too, so going to do the second round.

insights/tests/client/connection/test_test_connection.py
```diff
@@ -457,23 +448,22 @@ def test_legacy_urls_post_all_fails(get, post, exception, insights_connection):
     get.assert_not_called()
 
 
-@parametrize_exceptions(EXPECTED_EXCEPTIONS_TEST_URLS)
```
Glutexo, line 460: This was not correct. The `no_catch` tests should have checked for unexpected exceptions.

Glutexo:
Rebased on current master with https://github.com/RedHatInsights/insights-core/pull/4246 merged in. Requesting a fresh review.

---

tl;dr

Resolving the conficts in tests, I fixed a few inconsistencies:

* A single _logger.error_ call was sometimes asserted with `assert_called_once_with(...)` and sometimes with `mock_calls == [mock.call(...)]`. Unified to the latter.
* Logger patches were sometimes listed first and sometimes last. Unified them to be first.

I made verifications of multiple exception prints in `_legacy_test_urls` more straightforward. Added a separate string list for messages instead of only having a list of exception. As a result:

* There are no more cryptic `patch.side_effect = exception_list` assignments. They do magic behind the curtain.
* Tests don’t rely on implicit conversions of exception. Plain strings are compared.

The updates to the tests all result from the fact that both `test_connection` and `_(legacy_)test_urls` now rescue from the same exception types. `ReadTimeout` is always expected.

* Parametrization for expected exceptions is always the same. 
* Parametrization for unexpected exceptions is no longer necessary. `UnexpectedException` is enough.
* The fail tests for `_(legacy_)_test_urls` don’t need to be split for a connection timeout and read timeout.

PR #4217, opened 2024-09-19
feat(client): write .last-upload.results also after non-legacy uploads

ptoscano:
### All Pull Requests:

Check all that apply:

While that file is deprecated and it is good that it is not written after non-legacy uploads, in practice it is actually used by tooling that uses insights-client.

To avoid breaking functionalities in other software when switching to non-legacy, write /etc/insights-client/.last-upload.results also after non-legacy uploads. While those softwares are fixed to not rely on that file, at least this is one small step ahead in switching the default away to non-legacy without regressions elsewhere.

Implements CCT-710

### DIFF
--- a/insights/client/client.py
+++ b/insights/client/client.py
@@ -413,6 +413,9 @@ def upload(config, pconn, tar_file, content_type, collection_duration=None):
             continue
 
         if upload.status_code in (200, 202):
+            # Write to last upload file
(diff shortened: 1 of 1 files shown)


m-horky (approved):
LGTM

m-horky:
@xiangce Hello, this PR is ready for merge.

PR #4230, opened 2024-09-28
fix: Update value format of CupsBrowsedConf

wushiqinlou:
### All Pull Requests:

Check all that apply:

Update the value format of this parser to adjust the Advisor rule.
- Per the manual and tests of cups_browsed.conf, 
  different directives will be handled differently. E.g, 
  for `BrowseAllow`, all its configured values will be kept, 
  but for `BrowseRemoteProtocols`, only its last one will work.

### DIFF
--- a/insights/tests/parsers/test_cups_confs.py
+++ b/insights/tests/parsers/test_cups_confs.py
@@ -132,6 +132,7 @@ CUPS_BROWSED_CONF = """
 # Can use DNSSD and/or CUPS and/or LDAP, or 'none' for neither.
 
 BrowseRemoteProtocols dnssd cups
+BrowseRemoteProtocols none
 BrowseAllow 192.168.0.1
 BrowseAllow 192.168.0.255
 BrowseAllow cups.example.com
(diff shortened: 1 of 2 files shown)


insights/tests/parsers/test_cups_confs.py
```diff
@@ -176,8 +177,8 @@ def test_cups_browsed_files_conf():
     result = CupsBrowsedConf(context_wrap(CUPS_BROWSED_CONF))
     assert len(result) == 2
     assert 'BrowseRemoteProtocols' in result
-    assert result['BrowseRemoteProtocols'] == ['dnssd', 'cups']
-    assert sorted(result['BrowseAllow']) == sorted(['192.168.0.1', '192.168.0.255', 'cups.example.com'])
+    assert result['BrowseRemoteProtocols'] == ['dnssd cups', 'none']
+    assert result['BrowseAllow'] == ['192.168.0.1', '192.168.0.255', 'cups.example.com', '192.168.0.255']
```
xiangce, line 181: Better add a special description to the parser in the docstring, like:
```
.. note::
  Values of each directives will be store to a list with the original order, and duplicated values will be kept without de-duplication.

```

xiangce (approved):

PR #4236, opened 2024-10-07
feat: New datasource "ld_library_path_of_global" and its parser

qinpingli:
Signed-off-by: Ping Qin <piqin@redhat.com>

rh-pre-commit.version: 2.3.1

### All Pull Requests:

Check all that apply:

* [X] No Sensitive Data in this change?

Add this datasource for new rule.


Lorquas:

insights/specs/datasources/env.py
```diff
@@ -0,0 +1,95 @@
...
+        "/root/.tcshrc"
+    ]
+    data = []
+
+    for item in config_files:
+        if (os.path.exists(item) and
+                os.path.isfile(item) and
+                _is_env_exported(item, env_name)):
+            data.append(item)
+        elif os.path.exists(item) and os.path.isdir(item):
```
xiangce, line 65: Better use `os.walk` to recursively read the files here in case the dir is structured in multiple levels

insights/parsers/ld_library_path.py
```diff
@@ -64,3 +64,31 @@ def parse_content(self, content):
             raise SkipComponent("LD_LIBRARY_PATH not set.")
 
         self.extend(llds)
+
+
+@parser(Specs.ld_library_path_of_global)
+class GlobalLdLibraryPath(Parser):
```
xiangce, line 70: The class name `GlobalLdLibraryPath` looks more like that it will provide the value of the `LD_LIBRARY_PATH`, however, it just returns the files that `exported` this variable, but not its actual value.

Would you mind renaming it more relevant?

insights/specs/datasources/env.py
```diff
@@ -0,0 +1,95 @@
...
+import os
+
+from insights.core.context import HostContext
+from insights.core.dr import SkipComponent
+from insights.core.plugins import datasource
+from insights.core.spec_factory import DatasourceProvider
+
+
+@datasource(HostContext)
+def ld_library_path_of_global(broker):
```
xiangce, line 14: The same comment as above, would you rename this method more relevant?

insights/specs/datasources/env.py
qinpingli: @xiangce Thanks for the review! I intent to use `os.listdir` here. Per the following test:

```
[root@preserve-storage-server1 profile.d]# cat test_1.sh 
export TEST_FILE="/testfile"
[root@preserve-storage-server1 profile.d]# cat test/test.sh 
export TEST_DIR="test_dir"
[root@preserve-storage-server1 profile.d]# env|grep TEST
TEST_FILE=/testfile
```
Looks like only the files under `/etc/profile.d` or `/etc/env.d` directly will be executed.

qinpingli: I added a comment here to explain why `listdir` is used here.

insights/parsers/ld_library_path.py
qinpingli: Thanks! Updated.

insights/specs/datasources/env.py
qinpingli: Thanks! Updated.

JoySnow:
@qinpingli  Out of curious, from the related advisor rule's logic, it also mentioned for `2, there is no "unset LD_LIBRARY_PATH" from the above files`. Do we planned to cover that in a separate PR?

qinpingli:
> @qinpingli Out of curious, from the related advisor rule's logic, it also mentioned for `2, there is no "unset LD_LIBRARY_PATH" from the above files`. Do we planned to cover that in a separate PR?

Good question, I'll collect this information in this PR too.

insights/specs/datasources/env.py
```diff
@@ -0,0 +1,129 @@
...
+    # 1. export LD_LIBRARY_PATH=/test/path
+    # 2. export     LD_LIBRARY_PATH
+    # 3. export LD_LIBRARY_PATH=/test/path TEST=/path/to/test
+    # There might be other syntaxs, they will not be collected.
+    pattern = r"\b{0}\b".format(env_name)
+
+    with open(file_path, 'r') as file:
+        for line in file.readlines():
+            line = line.strip()
+            if not line or line.startswith("#") or line.split()[0] != "export":
```
JoySnow, line 110: To handle the comment lines, `line.startswith("#")` here is to skip the whole commented lines. The inline comment is another case that better be covered here, eg.   
```
export  TEST=/path/to/test   #  LD_LIBRARY_PATH=/test/path 
```
See the `Inline comment # comment`  handling in `get_active_lines()` .

insights/specs/datasources/env.py
```diff
@@ -0,0 +1,129 @@
...
+
+def _is_env_unset(file_path, env_name):
+    # if one line with "unset " + LD_LIBRARY_PATH
+    # returns True
+    pattern = r"\b{0}\b".format(env_name)
+
+    with open(file_path, 'r') as file:
+        for line in file.readlines():
+            line = line.strip()
+            if not line or line.startswith("#"):
```
JoySnow, line 125: Same inline comment case here .

qinpingli: Thanks, updated.

insights/specs/datasources/env.py
qinpingli: Thanks, updated.

JoySnow:
```
    with open(file_path, 'r') as file:
        active_lines = list(filter(None, (line.split("#", 1)[0].strip() for line in file.readlines())))
        for line in active_lines:
            if "export " in line and re.search(pattern, line):
                return True
```
@qinpingli Based on your newest implementation, what about calling this `get_active_lines()` func directly instead of rewrite the code?

qinpingli: Hmm, I intend not calling `get_active_lines` to avoid potential circle import here, because looks like the `parsers` is a higher level package of `specs`,  it will import from `specs` package, so here I avoid to import `parsers` from `specs` package.

insights/specs/datasources/env.py
xiangce: It's because of the following bash scripts written in the /etc/profile:
For general Linux distribution, it looks like 
```bash
for script in /etc/profile.d/*.sh ; do
        if [ -r $script ] ; then
                . $script
        fi
done
```
For RHEL, it is:
```bash
for i in /etc/profile.d/*.sh /etc/profile.d/sh.local ; do
    if [ -r "$i" ]; then
        if [ "${-#*i}" != "$-" ]; then
            . "$i"
        else
            . "$i" >/dev/null
        fi
    fi
done
```
I'd to say, currently the `listdir` works. Whenever there is some change RHEL officially or customized by customer, things will be changed.

But it doesn't seem that it is worth monitoring such changes for us.

insights/parsers/ld_library_path.py
```diff
@@ -64,3 +64,46 @@ def parse_content(self, content):
...
+        >>> type(global_ld_library_path_conf)
+        <class 'insights.parsers.ld_library_path.GlobalLdLibraryPathConf'>
+        >>> global_ld_library_path_conf.export_files
+        '/etc/environment,/etc/env.d/test.conf,/root/.bash_profile'
+        >>> global_ld_library_path_conf.unset_files
+        '/etc/profile'
+
+    Attributes:
+        export_files(str): a list of files that define the global LD_LIBRARY_PATH environment.
+        unset_files(str): a list of files that unset the global LD_LIBRARY_PATH environment.
```
xiangce, line 91: With introducing this `unset_files`, I have such questions, e.g. in some files (or even the same file), there are `unset` and `export` lines for the `LD_LIBRARY_PATH` variable. How does the plugins to identify which one works finally? If it's unset at last, the plugin will hit false positive.

xiangce (commented):
As mentioned before, setting/loading ENV variables is open to users in Linux, and is very flexible. This will lead the Advisor plugin easily resulting in false positive. How about evaluating the candidate again?

insights/parsers/ld_library_path.py
xiangce: Talked with @wushiqinlou - we have workaround (ignore host that have ever `unset` this ENV Var in any listed files). It's good to me.

xiangce (approved):

xiangce (approved):

PR #4191, opened 2024-08-14
Catch Timeout on test connection

Glutexo:
### All Pull Requests:

Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [x] Is this PR to correct an issue?
* [ ] Is this PR an enhancement?

### Complete Description of Additions/Changes:

The `--test-connection` command only used to catch `ConnectionError`, including `ConnectionTimeout`, but not other timeouts such as `ReadTimeoutError`. Re-used the existing `REQUEST_FAILED_EXCEPTIONS` list to process all timeouts in the same way.

There is an integration test pull request https://github.com/RedHatInsights/insights-client/pull/244 for this fix.

Card IDs:
* [CCT-506](https://issues.redhat.com/browse/CCT-506)
* [CCT-724](https://issues.redhat.com/browse/CCT-724)

PR #4237, opened 2024-10-07
Remove double exception print

Glutexo:
### All Pull Requests:

Check all that apply:

In case of a connection error, an exception description was printed on the standard output twice: once by the outer test_connection method and once by the inner _\_(legacy\_)test\_urls_ method. The legacy version tries multiple URLs and prints the exception description on each failure. Thus, removed the outer print that only duplicates the last caught exception.

Card IDs:

* [CCT-724](https://issues.redhat.com/browse/CCT-724)

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -349,7 +349,7 @@ class InsightsConnection(object):
                 else:
                     logger.info("Connection failed")
                     return False
-            except requests.ConnectionError as exc:
+            except REQUEST_FAILED_EXCEPTIONS as exc:
                 last_ex = exc
                 logger.error(
                     "Could not successfully connect to: %s", test_url + ext)
@@ -381,13 +381,11 @@ class InsightsConnection(object):
             else:
                 logger.info("Connection failed")
                 return False
-        except requests.ConnectionError as exc:
-            last_ex = exc
+        except REQUEST_FAILED_EXCEPTIONS as exc:
             logger.error(
                 "Could not successfully connect to: %s", url)
             print(exc)
-        if last_ex:
-            raise last_ex
+            raise exc
 
     def test_connection(self, rc=0):
         """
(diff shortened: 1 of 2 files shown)


Glutexo:
This pull request contains changes [requested](https://github.com/RedHatInsights/insights-core/pull/4191#pullrequestreview-2260061356) by Matyáš. I extracted them from the original PR https://github.com/RedHatInsights/insights-core/pull/4191.

Both pull requests introduce changes to the now untested test connection mechanic and both deal with exception handling. Because of that, general unit tests for sub-method calls are included in both of them. After one PR will have been merged I will resolve the conflict in the other one.

As for the _caplog_ fixture, that is not supported on Python ≤2.7, I’m aware that the suggestion was to mark the tests as skipped. The change to logger asserts was however trivial. Moreover, I already used those in the sister pull request https://github.com/RedHatInsights/insights-core/pull/4211. Such asserts are less semantic, but working. They need to be replaced after fixing the pipeline, but so do the skip marks.

---

Crafting the pull request, I identified multiple fixes it introduces. I am thus going to split it to even smaller chunks, so they can be merged and reviewed separately. Even though the changes to the runtime code are tiny, each comes with quite an amount of tests. Hence the will to split.

- Removing the additional print from _test_connection_, duplicating the ones from _\_(legacy\_)test\_urls_. The inner ones are the ones to keep, because in the legacy version, multiple requests can be issued. **This has been the original intention of this pull request.**
- Catching read timeouts in addition to connection timeouts in _\_(legacy\_)test\_urls_. https://github.com/RedHatInsights/insights-core/pull/4211 does this only for the test_connection method, preventing the traceback to be printed.
- Removing a redundant raise statement, not changing behavior in any way.

I am opening the pull request for review though, so I can apply any comments to the split ones. @m-horky

Glutexo:
Extracted the originally requested fix to a new pull request https://github.com/RedHatInsights/insights-core/pull/4246. Closing this one and will put the other tiny fixes in their own PRs.

PR #4239, opened 2024-10-08
New spec "/etc/rhc/config.toml" and parser

huali027:
### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/parsers/rhc_conf.py
+++ b/insights/parsers/rhc_conf.py
@@ -1,0 +1,49 @@
+"""
+RHCConf - file ``/etc/rhc/config.toml``
+=======================================
+"""
+
+from insights import Parser, parser
+from insights.parsers import get_active_lines
+from insights.specs import Specs
+from insights.core.dr import SkipComponent
+
+
+@parser(Specs.rhc_conf)
+class RHCConf(Parser, dict):
+    """
+    Class to parse the ``/etc/rhc/config.toml`` configuration file.
+
+    It's better to use the built-in "tomllib" lib to parse it after abandoning the support
+    of versions before python3.11.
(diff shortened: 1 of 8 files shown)


Lorquas:

huali027:
@JoySnow @xiangce Please help to review, thanks.

insights/parsers/rhc_conf.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+
+from insights import Parser, parser
+from insights.parsers import get_active_lines
+from insights.specs import Specs
+from insights.core.dr import SkipComponent
+
+
+@parser(Specs.rhc_conf)
+class RHCConf(Parser, dict):
```
xiangce, line 13: A nitpick, but would you think about renaming it to `RhcConf` a CamelCase naming style?

xiangce (commented):
And how about renaming the python module as just `rhc` instead of `rhc_conf`, in cases we may add more RHC relevant parsers without adding new modules?

insights/parsers/rhc_conf.py
huali027: @xiangce RHC is an abbreviation of "Remote Host Connection", so I used uppercase for all of them. But it's good to me to update it to "RhcConf" to avoid two uppercase "C" in the middle, I have updated it. And it seems there is only one configuration file for rhc client, so I just use "RhcConf" without "Toml". Please help to review again, thanks.

huali027:
> And how about renaming the python module as just `rhc` instead of `rhc_conf`, in cases we may add more RHC relevant parsers without adding new modules?

Updated, please help to review again.

JoySnow (approved):

xiangce (approved):

PR #4242, opened 2024-10-09
fix: encoding issue in python2 env

xiangce:
- Originally, simple file content was encoded when loading in py2 env
  but was not encoded when writing to archive file.
  When there are unicode characters in simple file, it causes 
  UnicodeEncodeError and results in file cannot be collected.
- fix RHINENG-13636

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -44,7 +44,7 @@ A minimal set of environment variables for use in subprocess calls
 if "LANG" in os.environ:
     SAFE_ENV["LANG"] = os.environ["LANG"]
 
-safe_open = open if six.PY3 else codecs.open
+safe_open, encoding = (open, "utf-8") if six.PY3 else (codecs.open, None)
 
 
 class ContentProvider(object):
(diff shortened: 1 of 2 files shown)


xiangce:
Regarding the Missing line in Codecov report, As 
1. in this PR, no new lines are introduced
2. we have another particular task RHINENG-12098 to enhance its coverage. 

Let's ignore this Missing for now.

chenlizhong (approved):
Looks good to me

xiangce:
@chenlizhong - as talked today, I added an unittest for this change. Without this change the test will fail as below.

```
self = <closed file '/tmp/insights_unittest_unicode_text', mode 'w+b' at 0x491cdb0>, object = '\xe2\x80\x9c\xef\xbc\x81\xe2\x80\xa6\xe2\x80\xa6\xe2\x80\x9d\nliuxc-rhel6-test'

    def write(self, object):

        """ Writes the object's contents encoded to self.stream.
            """
>       data, consumed = self.encode(object, self.errors)
E       UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 0: ordinal not in range(128)
```

chenlizhong (approved):

PR #4246, opened 2024-10-11
Remove double exception print

Glutexo:
### All Pull Requests:

Check all that apply:

In case of a connection error, an exception description was printed on the standard output twice: once by the outer test_connection method and once by the inner _\_(legacy\_)test\_urls_ method. The legacy version tries multiple URLs and prints the exception description on each failure. Thus, removed the outer print that only duplicates the last caught exception.

Card IDs:

* [CCT-724](https://issues.redhat.com/browse/CCT-724)

### DIFF
--- a/insights/tests/client/connection/test_test_connection.py
+++ b/insights/tests/client/connection/test_test_connection.py
@@ -15,11 +15,19 @@ UPLOAD_URL = "https://www.example.com/insights"
 LEGACY_URL = "https://www.example.com"
 LEGACY_URL_SUFFIXES = ["/insights/", "", "/r", "/r/insights"]
 
-EXPECTED_EXCEPTIONS = [
+EXCEPTION_MESSAGE = "request exception"
+EXPECTED_EXCEPTIONS_TEST_CONNECTION = [
     requests.exceptions.SSLError,
     requests.exceptions.ConnectTimeout,
+]
+EXPECTED_EXCEPTIONS_TEST_URLS = EXPECTED_EXCEPTIONS_TEST_CONNECTION + [
     requests.exceptions.ReadTimeout,
 ]
+UNEXPECTED_EXCEPTIONS_TEST_CONNECTION = [
+    requests.exceptions.ReadTimeout,
+    UnexpectedException,
+]
+
 parametrize_methods = pytest.mark.parametrize(
     ["http_method", "request_function"],
     [
@@ -122,7 +130,7 @@ def test_test_urls_get_success(get, temporary_file, post, insights_connection):
     post.assert_not_called()
 
 
-@parametrize_exceptions(EXPECTED_EXCEPTIONS)
+@parametrize_exceptions(EXPECTED_EXCEPTIONS_TEST_URLS)
 @mock.patch("insights.client.connection.InsightsConnection.post")
 @mock.patch("insights.client.connection.TemporaryFile")
 @mock.patch("insights.client.connection.InsightsConnection.get")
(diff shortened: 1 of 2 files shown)


Glutexo:
This pull request contains changes [requested](https://github.com/RedHatInsights/insights-core/pull/4191#pullrequestreview-2260061356) by Matyáš. First, I [extracted](https://github.com/RedHatInsights/insights-core/pull/4237) the changes from https://github.com/RedHatInsights/insights-core/pull/4191 that didn’t belong to that PR. Then, I only kept the requested changes, replacing an intermediary PR https://github.com/RedHatInsights/insights-core/pull/4237.

Both pull requests introduce changes to the now untested test connection mechanic and both deal with exception handling. Because of that, general unit tests for sub-method calls are included in both of them. After one PR will have been merged I will resolve the conflict in the other one.

As for the _caplog_ fixture, that is not supported on Python ≤2.7, I’m aware that the suggestion was to mark the tests as skipped. The change to logger asserts was however trivial. Moreover, I already used those in the sister pull request https://github.com/RedHatInsights/insights-core/pull/4211. Such asserts are less semantic, but working. They need to be replaced after fixing the pipeline, but so do the skip marks.

Opening the pull request for review. @m-horky

Glutexo:
If https://github.com/RedHatInsights/insights-core/pull/4211 gets merged first, this pull request would bring a regression. The other PR extends the set of exceptions caught by the outer _test\_connection_ method, but this one removes the redundant exception print from the same method.

That means that after merging https://github.com/RedHatInsights/insights-core/pull/4211, exceptions will be properly printed for read timeouts too (currently they are not), but then after merging this PR, that print will get removed. As a result, the read timeouts will be caught and would not dump a traceback, but wouldn’t have a nice print.

To fix that, exception handling in the inner _\_(legacy\_)test\_urls_ methods needs to be extended too. Doing so, read timeouts would reach the proper inner print and the outer one won’t be necessary – which was the original intention of this PR. I will create a new pull request to unblock this one.

insights/tests/client/connection/test_test_connection.py
m-horky, line 1: I believe these tests shouldn't be part of the PR?

m-horky (changes requested):
ACK. One note.

insights/tests/client/connection/test_test_connection.py
Glutexo: That would be true if this was based on another PR from its family. I based it on _master_, so it can be merged even alone and still contain all the relevant tests.

In this pull request there are both tests for _(\_legacy)\_test\_urls_ and _test\_connection_, because both layers contain exception handling, error logging and printing on stdout. They verify what’s not output by the removed print and also what is printed/logged by the one that was not removed.

If any of the sister PRs gets merged, the tests patch will be significantly reduced.

Glutexo:
To make things simpler, let‘s say this is blocked by https://github.com/RedHatInsights/insights-core/pull/4246 and is a blocker of https://github.com/RedHatInsights/insights-core/pull/4211. Merging https://github.com/RedHatInsights/insights-core/pull/4246 first will make further reviewing and merging easier and would significantly reduce the size of the test patch.

Glutexo:

Glutexo:
https://github.com/RedHatInsights/insights-core/pull/4248 got merge. There are now conflict. That’s because both PRs contained same tests. Will resolve and re-request a review.

Glutexo:
Rebased on the current _master_ with https://github.com/RedHatInsights/insights-core/pull/4248 merged. Resolved conflicts in the tests:

- Different exceptions are expected in _test\_connection_ and in _\_(legacy\_)_test_urls_.
- Covered cases when Read Timeout caught by _\_(legacy\_)_test_urls_, but not _test\_connection_.

This pull request only changes what is print on standard output. It is a part of what’s communicated to the user though. Because of that there are error log tests too. Prints and logging are always happen at the same time. They are on adjacent lines.

insights/tests/client/connection/test_test_connection.py
Glutexo: https://github.com/RedHatInsights/insights-core/pull/4248 got merged. The amount of tests dropped to about a half.

m-horky (approved):
ACK. Hi @xiangce, good to merge!

PR #4248, opened 2024-10-14
Catch Timeout on test urls

Glutexo:
### All Pull Requests:

Check all that apply:

The internal methods used by the `--test-connection` command only used to catch `ConnectionError`, including `ConnectionTimeout`, but not other timeouts such as `ReadTimeoutError`. As a result, in case of a read timeout,  the failure is not fully handled. That means:

* an error message containing the failed URL is not logged;
* and in case of legacy upload, the URL fallback mechanism is interrupted.

Re-used the existing `REQUEST_FAILED_EXCEPTIONS` list to process all timeouts in the same way.

Card IDs:

* [CCT-859](https://issues.redhat.com/browse/CCT-859)

### DIFF
--- a/insights/tests/client/connection/test_test_connection.py
+++ b/insights/tests/client/connection/test_test_connection.py
@@ -1,0 +1,616 @@
+import pytest
+import requests
+from mock import mock
+
+from insights.client.config import InsightsConfig
+from insights.client.connection import InsightsConnection
+
+
+class UnexpectedException(Exception):
+    pass
+
+
+LOGGING_FILE = "logging-file.log"
+UPLOAD_URL = "https://www.example.com/insights"
+LEGACY_URL = "https://www.example.com"
+LEGACY_URL_SUFFIXES = ["/insights/", "", "/r", "/r/insights"]
+
+EXPECTED_EXCEPTIONS = [
+    requests.exceptions.SSLError,
+    requests.exceptions.ConnectTimeout,
+    requests.exceptions.ReadTimeout,
+]
+parametrize_methods = pytest.mark.parametrize(
+    ["http_method", "request_function"],
+    [
+        ("GET", "insights.client.connection.InsightsConnection.get"),
+        ("POST", "insights.client.connection.InsightsConnection.post"),
+    ],
+)
+
+
+def parametrize_exceptions(exceptions):
+    return pytest.mark.parametrize(
+        ["exception"],
+        [(exception,) for exception in exceptions],
+    )
+
+
+@pytest.fixture
+@mock.patch("insights.client.connection.InsightsConnection._init_session")
+@mock.patch("insights.client.connection.InsightsConnection.get_proxies")
+def insights_connection(get_proxies, init_session):
+    config = InsightsConfig(
+        base_url="www.example.com",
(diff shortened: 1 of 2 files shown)


Glutexo:
I extracted this fix from https://github.com/RedHatInsights/insights-core/pull/4237 and originally from https://github.com/RedHatInsights/insights-core/pull/4211. https://github.com/RedHatInsights/insights-core/pull/4211 only changes exception handling in the outer `test_connection` method, because that is the minimal patch to fix the reported issue of a stack trace being dumped on the standard output. This new PR, on the other hand, fixes error logging and fallback mechanism in case of a read timeout.

Moreover, this would unblock https://github.com/RedHatInsights/insights-core/pull/4246 in case of https://github.com/RedHatInsights/insights-core/pull/4211 being merged first of the two. See my [comment](https://github.com/RedHatInsights/insights-core/pull/4246#issuecomment-2410440279) in https://github.com/RedHatInsights/insights-core/pull/4246 for more details.

Glutexo:
This patch only updates the inner _\_test\_urls_ and _\_legacy\_test\_urls_ methods, not the outer _test\_connection_ method. It doesn’t even change any printing or logging logic. Therefore, unlike__ https://github.com/RedHatInsights/insights-core/pull/4211 and https://github.com/RedHatInsights/insights-core/pull/4246, there are no tests of _test\_connection_ or integration tests.

The error log and standard out tests are the only way how to find out whether the _except_ block fired as intended. That’s why there are log/stdout tests in place. They, however, don’t test the contents of the messages as that’s not the subject of this pull request. That will change after merging with the tests from other related PRs.a

Glutexo:
Open for review. @m-horky

insights/tests/client/connection/test_test_connection.py
m-horky, line 1: Are all these tests for the patch of `connection.py` in this PR?

m-horky (commented):
I don't have objections, I'll just need to see in context of the other PRs.

insights/tests/client/connection/test_test_connection.py
Glutexo: Not entirely. As I mentioned in my [comment](https://github.com/RedHatInsights/insights-core/pull/4248#issuecomment-2413773334), there are only tests for the _\_test\_urls_ and _\_legacy\_test\_urls_. There are no tests for _test\_connection_.

Some notes for tests that may seem over the scope:

* The log and print tests are there to check that the _except_ block executed. They only assert the calls, not the contents.
* The _get_/_post_ methods are the ones that raise the caught exceptions. Hence the call asserts for those.

The other pull requests contain more tests or make some of the existing ones more precise. The latter applies for the log/stdout messages. I will resolve the merge conflict, by which the new patch in the following PR will be nice and clear. The test names are the same, so this will be easy.

Glutexo:
To make things simpler, let‘s say this is a blocker of https://github.com/RedHatInsights/insights-core/pull/4246, which is a blocker of https://github.com/RedHatInsights/insights-core/pull/4211. Merging this first will make things easier.

Glutexo:

m-horky:
Hi, @xiangce, good to merge!

PR #4249, opened 2024-10-15
fix: do not obfuscate in runtime but in archive only

xiangce:
- For 'persist' specs, they might be used during collection,
  to avoid failure collection of dependencies, they shouldn't
  be obfuscated in runtime.  But they should be obfuscated when
  writing to archive. See the tests
- Jira: RHINENG-13591

### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -128,9 +128,10 @@ class ContentProvider(object):
     def write(self, dst):
         fs.ensure_path(os.path.dirname(dst))
         # Clean Spec Content when writing it down to disk before uploading
+        content = "\n".join(self._clean_content())
+        content = content.encode("utf-8") if six.PY3 else content
         with open(dst, "wb") as f:
-            content = "\n".join(self._clean_content())
-            f.write(content.encode('utf-8') if six.PY3 else content)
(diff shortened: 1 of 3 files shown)


chenlizhong (approved):
Looks good to me

insights/core/spec_factory.py
```diff
@@ -126,10 +127,9 @@ def content(self):
 
     def write(self, dst):
         fs.ensure_path(os.path.dirname(dst))
-        # Clean Spec Content when writing it down to disk and before uploading
-        self._clean_content()
+        # Clean Spec Content when writing it down to disk before uploading
         with open(dst, "wb") as f:
-            content = "\n".join(self.content)
+            content = "\n".join(self._clean_content())
```
JoySnow, line 132: @xiangce What about keeping this `self._clean_content()` calling before the file open phrase? When `ContentException` be raised during `self._clean_content()`, it can save a file open action somehow.

xiangce: Nice catching @JoySnow. It would also prevent creating an empty file. updated in the new commit.

JoySnow (approved):

PR #4254, opened 2024-10-17
Simplify re-raise for non-legacy

Glutexo:
### All Pull Requests:

Check all that apply:

To avoid confusion for developers, removed the unnecessary conditional for exception re-raise.

* It is unnecessary to set last_ex on non-legacy, because there is no fallback to other URLs. Only one URL is tried and if it fails, there is only a single exception.
* The `if last_ex` condition can never be false, because if an exception is not raised, the function hits a return clause earlier. It only gets to the conditional if an exception has been caught.

A simple `raise` is enough. Without an argument, the last caught exception is re-raised by default. It’s thus equivalent to `raise exc`.

Card IDs:

* [CCT-867](https://issues.redhat.com/browse/CCT-867)

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -382,12 +382,10 @@ class InsightsConnection(object):
                 logger.info("Connection failed")
                 return False
         except REQUEST_FAILED_EXCEPTIONS as exc:
-            last_ex = exc
             logger.error(
                 "Could not successfully connect to: %s", url)
             print(exc)
-        if last_ex:
-            raise last_ex
+            raise
 
     def test_connection(self, rc=0):
         """


insights/tests/client/connection/test_test_connection.py
m-horky, line 1: Aren't these tests part of some other PR as well?

m-horky (changes requested):
Makes sense. One note.

insights/tests/client/connection/test_test_connection.py
Glutexo: They are. I made the PRs standalone. Each one is based on master and only contains relevant tests. Many of those tests are the same for all the PRs in the family.

I will resolve conflicts after merging. The test names are the same, so it will be very easy.

insights/tests/client/connection/test_test_connection.py
```diff
@@ -449,9 +449,7 @@ def test_legacy_urls_post_all_fails(get, post, exception, insights_connection):
 
 
 @parametrize_methods
-def test_legacy_urls_no_catch(
-    http_method, request_function, insights_connection
-):
+def test_legacy_urls_no_catch(http_method, request_function, insights_connection):
```
Glutexo, line 452: This is only a result of reformatting by _black_.

insights/tests/client/connection/test_test_connection.py
Glutexo: All other pull requests in the family have been merged. That means all tests are already in place. This pull request does not change behavior. Thus, the tests remain unchanged.

Glutexo:
Rebased on the current master. Ready for review.

insights/tests/client/connection/test_test_connection.py
m-horky: Please revert this part of the file. Let's not introduce additional changes that aren't related to the main problem. This should be done in a separate PR, if required.

Glutexo: Sure thing!

Glutexo: In https://github.com/RedHatInsights/insights-core/pull/4211, however, unrelated [pieces](https://github.com/RedHatInsights/insights-core/pull/4211) got merged actually:

* Two kinds of inconsistencies in existing tests.
* A fix of an existing incorrect tests.

I agree though that those should have been and be separate.

Glutexo:
Removed the unrelated reformatting. Ready for re-review/merge.

What about reformatting the codebase with Black? Then, we could add it to the CI and reject non-conforming code. JIRA card?

m-horky (approved):
LGTM! Hello @xiangce, it's ready :)

> What about reformatting the codebase with Black?

I'm generally in favor, but 1) we don't own the repository, 2) I'm afraid Black/ruff may introduce formatting that's not compatible with Python 2.6/2.7. I think it should be on a long-term roadmap, but there are more difficult problems to be solved first.

PR #4255, opened 2024-10-17
FEAT: new custom filter 'fstab_mounted.dirs' to ls_lan spec

TZ3070:
### All Pull Requests:

Check all that apply:

- add a special filter keyword `fstab_mounted.dirs` to the `ls_lan` spec
  it's not a real directory existing in any system, but a `switch` like keyword,
  to enable the `ls_lan_dirs` to involve the mounted points in the default `/etc/fstab`
  configuration file as the `ls -lan` candidates.


xiangce (changes requested):
Please check my comments to the relevant Jira card.

insights/specs/datasources/ls.py
```diff
@@ -52,7 +54,13 @@ def list_with_la_filtered(broker):
...
 @datasource(HostContext)
 def list_with_lan(broker):
-    return _list_items(Specs.ls_lan_dirs)
+    filters = _list_items(Specs.ls_lan_dirs)
+    if 'specs.fstab_mounted' in filters and FSTab in broker:
+        filter_list = filters.split()
+        for mntp in broker[FSTab].mounted_on.keys():
+            mnt_point = os.path.dirname(mntp)
+            filter_list.extend([mnt_point] if mnt_point and mnt_point not in filter_list else [])
+    return ' '.join(filter_list)
```
JoySnow, line 63: There could be a case that there are more than 1k mount entries from mount command. But not known how's the mount entries count in /etc/fstab. If the mount_point got a quite large count than avg, say, 100, will this `ls -lan` command work?

insights/tests/datasources/test_ls.py
```diff
@@ -59,8 +71,12 @@ def test_la_filterd():
 
 
 def test_lan():
-    ret = list_with_lan({})
-    assert ret == '/ /boot'
+    fstab = FSTab(context_wrap(FSTAB_CONTEXT))
+    broker = {
+        FSTab: fstab
+    }
```
JoySnow, line 77: Better keep the original `test_lan` case with empty broker. 
```
def test_lan():
    ret = list_with_lan({})
    assert ret == '/ /boot'
```
And add this with `FSTab` one as a new test case.

insights/collect.py
```diff
@@ -241,6 +241,10 @@
       enabled: true
     - name: insights.components.selinux.SELinuxEnabled
       enabled: true
+
+    # needed for the 'ls lan' spec
```
xiangce, line 245: Better explain it a bit clearer, e.g. `needed for the 'spec.fstab_mounted' to the 'ls_lan' spec`

insights/specs/datasources/ls.py
```diff
@@ -52,7 +54,13 @@ def list_with_la_filtered(broker):
 
 @datasource(HostContext)
 def list_with_lan(broker):
-    return _list_items(Specs.ls_lan_dirs)
+    filters = _list_items(Specs.ls_lan_dirs)
+    if 'specs.fstab_mounted' in filters and FSTab in broker:
+        filter_list = filters.split()
```
xiangce, line 59: This seems a redundant operation, better update the `_list_items` and let it return a `sorted` list instead. (you may want to update all other `datasources` to `join` the new value returned by `_list_items`

insights/specs/datasources/ls.py
```diff
@@ -52,7 +54,13 @@ def list_with_la_filtered(broker):
...
+        for mntp in broker[FSTab].mounted_on.keys():
+            mnt_point = os.path.dirname(mntp)
+            filter_list.extend([mnt_point] if mnt_point and mnt_point not in filter_list else [])
```
xiangce, line 62: declear the `filter_list` as a set first to avoid duplicate element by nature. 
```python
filter_list = set(_list_items(Specs.ls_lan_dirs)) # assume the _list_items returns `list` here already
for mntp in broker[FSTab].mounted_on.keys():
    filter_list.add(os.path.dirname(mntp))


```

insights/specs/datasources/ls.py
```diff
@@ -52,7 +54,13 @@ def list_with_la_filtered(broker):
 
 @datasource(HostContext)
 def list_with_lan(broker):
-    return _list_items(Specs.ls_lan_dirs)
+    filters = _list_items(Specs.ls_lan_dirs)
+    if 'specs.fstab_mounted' in filters and FSTab in broker:
```
xiangce, line 58: As per the Jira, for now, this filter is only for the SAP rule, I think we could add the `sap_hana_instance` as one dependency here.
```python
from insights.specs.datasources.sap import LocalSpecs as SAPSpecs

if 'specs.fstab_mounted' in filters and SAPSpecs.sap_hana_instance in broker and FSTab in broker:
```

insights/specs/datasources/ls.py
TZ3070: Testing result with 5003 mount in /etc/fstab:
```
{"name": "insights.specs.Specs.ls_lan", "exec_time": 0.0006165504455566406, "errors": [], "results": {"type": "insights.core.spec_factory.CommandOutputProvider", "object": {"rc": null, "cmd": "...
```
the ls command work well

insights/specs/datasources/ls.py
```diff
@@ -36,50 +38,55 @@ def _list_items(spec):
...
 
 @datasource(HostContext)
 def list_with_lan(broker):
-    return _list_items(Specs.ls_lan_dirs)
+    filters = set(_list_items(Specs.ls_lan_dirs))
+    if 'specs.fstab_mounted' in filters and FSTab in broker:
+        for mntp in broker[FSTab].mounted_on.keys():
+            mnt_point = os.path.dirname(mntp)
+            filters.add(mnt_point) if mnt_point else None
+    return ' '.join(sorted(list(filters)))
```
JoySnow, line 62: @TZ3070  We can directly sort on a `set` type, this `list(filters)` converting can be skipped here.

JoySnow (approved):
The code LGTM !

xiangce (approved):

PR #4274, opened 2024-11-20
fix(client): checkin for unregistered hosts should fail

pkoprda:
### All Pull Requests:

Check all that apply:

* Card ID: CCT-891
* Card ID: RHINENG-13984

The `insights-client --checkin` command currently succeeds on hosts that are not registered with `insights-client`, but are registered with `subscription-manager`. This fix addresses the issue to ensure the `--checkin` command properly fails as expected when the client is not registered.

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -1127,6 +1127,10 @@ class InsightsConnection(object):
         '''
         logger.info("Checking in...")
 
+        if not self._fetch_system_by_machine_id():
+            logger.error("This host is not registered. To register, run 'insights-client --register'.")
+            return False
+
         try:
             canonical_facts = get_canonical_facts()
             canonical_facts = self._clean_facts(canonical_facts)
--- a/insights/tests/client/connection/test_checkin.py
+++ b/insights/tests/client/connection/test_checkin.py
@@ -24,6 +24,7 @@ def _get_canonical_facts_response(canonical_facts):
 @patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
 @patch(
     "insights.client.connection.InsightsConnection._init_session", Mock())
+@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=True)
 @patch(
     "insights.client.connection.get_canonical_facts",
     return_value=_get_canonical_facts_response({"subscription_manager_id": str(uuid4())})
@@ -33,7 +34,7 @@ def _get_canonical_facts_response(canonical_facts):
     **{"return_value.status_code": codes.CREATED}
 )
 @patch("insights.client.connection.InsightsConnection.get_proxies")
-def test_canonical_facts_request(get_proxies, post, get_canonical_facts, rm_conf):
+def test_canonical_facts_request(get_proxies, post, get_canonical_facts, fetch_system_by_machine_id, rm_conf):
     """
     A POST requests to the check-in endpoint is issued with correct headers and
     body containing unobfuscated Canonical Facts.
@@ -54,6 +55,7 @@ def test_canonical_facts_request(get_proxies, post, get_canonical_facts, rm_conf
 @patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
 @patch(
     "insights.client.connection.InsightsConnection._init_session", Mock())
+@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=True)
 @patch(
     "insights.client.connection.get_canonical_facts",
     return_value=_get_canonical_facts_response({"subscription_manager_id": str(uuid4())})
@@ -63,7 +65,7 @@ def test_canonical_facts_request(get_proxies, post, get_canonical_facts, rm_conf
     **{"return_value.status_code": codes.CREATED}
 )
 @patch("insights.client.connection.InsightsConnection.get_proxies")
-def test_canonical_facts_request_cleaned(get_proxies, post, get_canonical_facts, rm_conf):
+def test_canonical_facts_request_cleaned(get_proxies, post, get_canonical_facts, fetch_system_by_machine_id, rm_conf):
     """
     A POST requests to the check-in endpoint is issued with correct headers and
     body containing obfuscateed Canonical Facts.
@@ -85,6 +87,7 @@ def test_canonical_facts_request_cleaned(get_proxies, post, get_canonical_facts,
 @patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
 @patch(
     "insights.client.connection.InsightsConnection._init_session", Mock())
+@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=True)
 @patch("insights.client.connection.generate_machine_id", return_value=str(uuid4()))
 @patch("insights.client.connection.get_canonical_facts", side_effect=RuntimeError())
 @patch(
@@ -92,7 +95,7 @@ def test_canonical_facts_request_cleaned(get_proxies, post, get_canonical_facts,
     **{"return_value.status_code": codes.CREATED}
 )
 @patch("insights.client.connection.InsightsConnection.get_proxies")
-def test_insights_id_request(get_proxies, post, get_canonical_facts, generate_machine_id, rm_conf):
+def test_insights_id_request(get_proxies, post, get_canonical_facts, generate_machine_id, fetch_system_by_machine_id, rm_conf):
     """
     A POST requests to the check-in endpoint is issued with correct headers and
     body containing only an Insights ID if Canonical Facts collection fails.
@@ -113,13 +116,14 @@ def test_insights_id_request(get_proxies, post, get_canonical_facts, generate_ma
 @mark.parametrize(("exception",), ((ConnectionError,), (Timeout,)))
 @patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
 @patch("insights.client.connection.InsightsConnection._init_session", Mock())
+@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=True)
 @patch(
     "insights.client.connection.get_canonical_facts",
     return_value=_get_canonical_facts_response({"subscription_manager_id": "notauuid"})
 )
 @patch("insights.client.connection.InsightsConnection.post")
 @patch("insights.client.connection.InsightsConnection.get_proxies")
-def test_request_http_failure(get_proxies, post, get_canonical_facts, rm_conf, exception):
+def test_request_http_failure(get_proxies, post, get_canonical_facts, fetch_system_by_machine_id, rm_conf, exception):
     """
     If the checkin-request fails, None is returned.
     """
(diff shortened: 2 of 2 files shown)


insights/tests/client/connection/test_checkin.py
```diff
@@ -191,3 +191,14 @@ def test_response_failure(get_proxies, post, get_canonical_facts, rm_conf, statu
...
     with raises(Exception):
         connection.checkin()
+
+
+def test_unregistered_host():
+    """
+    If the host is not registered, the checkin fails.
+    """
+    config = InsightsConfig(legacy_upload=False, base_url="www.example.com")
+    connection = InsightsConnection(config)
```
ptoscano, line 201: Hmm this test is not mocking things like all the other tests in this file; this probably means that it will actually try to get the facts, read the config, and get the machine-id of the system, thus potentially failing the test if run on a system registered with insights-client.

insights/client/connection.py
```diff
@@ -1127,6 +1127,10 @@ def checkin(self):
         '''
         logger.info("Checking in...")
 
+        if not self.config.legacy_upload and not self.api_registration_check():
```
ptoscano, line 1130: I see that using `--checking` switches automatically to non-legacy upload; regardless, I'd remove the explicit check here to reduce the checks for legacy upload.
Rather than using `api_registration_check()` directly, use `insights.client.support.registration_check()`, as it should take care of whatever upload mode (legacy vs non-legacy).

insights/tests/client/connection/test_checkin.py
```diff
@@ -191,3 +191,14 @@ def test_response_failure(get_proxies, post, get_canonical_facts, rm_conf, statu
 
     with raises(Exception):
         connection.checkin()
+
+
+def test_unregistered_host():
```
Glutexo, line 196: Let’s patch the function that checks whether the host is registered.

```suggestion
@patch("insights.client.connection.InsightsConnection.api_registration_check", return_value=false)
def test_unregistered_host(api_registration_check):
```

or 

```suggestion
@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=false)
def test_unregistered_host(fetch_system_by_machine_id):
```

if you accepted my suggestion above.

All tests in the file patch `_init_session`, because it verifies RHSM certificates on the host machine. That can cause false errors.

```suggestion
@patch("insights.client.connection.InsightsConnection._init_session", Mock())
def test_unregistered_host():
```

They patch `InsightsUploadConf.get_rm_conf` too. It seems not required: it may be a mistake or a relic. Let’s go the safe way and keep the patches consistent. They can be removed in bulk separately.

```suggestion
@patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
def test_unregistered_host(rm_conf):
```

Combined:

```suggestion
@patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
@patch("insights.client.connection.InsightsConnection._init_session", Mock())
@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=false)
test_unregistered_host(fetch_system_by_machine_id, rm_conf):
```

---

The other tests need to be updated too. A patch to api_registration_check` / `_fetch_system_by_machine_id` with a truth return value needs to be added for several reasons:

* The non-patched original method would issue real requests and tamper with the machine ID on the host.
* The tests need to ensure that the check passes, so they can check what happens after the conditional.
* If there was a test with a patched `get` method, assertions for its calls would break.

```python
...
@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=True)
...
def test_canonical_facts_request(get_proxies, post, fetch_system_by_machine_id, get_canonical_facts, rm_conf):
```

insights/tests/client/connection/test_checkin.py
```diff
@@ -191,3 +191,14 @@ def test_response_failure(get_proxies, post, get_canonical_facts, rm_conf, statu
...
+    """
+    If the host is not registered, the checkin fails.
+    """
+    config = InsightsConfig(legacy_upload=False, base_url="www.example.com")
+    connection = InsightsConnection(config)
+
+    result = connection.checkin()
+    assert result is False
```
Glutexo, line 204: I think there should be one more test verifying that the unregistered state of the host actually causes the checkin function to immediately return. This can be achieved by checking that `get_canonical_facts` and `post` aren’t called. Maybe split into two test cases. 🤷🏻‍♂️

```suggestion
    assert result is False


@patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
@patch("insights.client.connection.InsightsConnection._init_session", Mock())
@patch("insights.client.connection.post")
@patch("insights.client.connection.get_canonical_facts")
def test_unregistered_host_no_checkin(get_canonical_facts, post, rm_conf)
    config = InsightsConfig(legacy_upload=False, base_url="www.example.com")
    connection = InsightsConnection(config)

    connection.checkin()
    get_canonical_facts.assert_not_called()
    post.assert_not_called()
```

Glutexo (changes requested):

insights/tests/client/connection/test_checkin.py
pkoprda: I think one test for this is enough. However, I added there two asserts to ensure `get_canonical_facts` and `post` functions are not called :+1:

pkoprda:
Thank you @ptoscano and @Glutexo for review! I hope I fulfilled all your requests. Now when the host is not registered to insights-client, the `--checkin` command should fail for legacy and non-legacy mode :+1:

insights/tests/client/connection/test_checkin.py
```diff
@@ -24,6 +24,7 @@ def _get_canonical_facts_response(canonical_facts):
 @patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
 @patch(
     "insights.client.connection.InsightsConnection._init_session", Mock())
+@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=True)
```
Glutexo, line 27: I always tried to keep the order of patches matching the order of first call in the production code. In this case, `_fetch_system_by_machine_id` is the first thing that is called in the tested `checkin` method.

As I see, however, the order of the existing patches doesn’t follow this rule: `post` is called after `get_canonical_facts`. 🤷🏻‍♂️ So no need to worry, just noting.

```python
@patch(
    "insights.client.connection.InsightsConnection.post",
    **{"return_value.status_code": codes.CREATED}
)
@patch(
    "insights.client.connection.get_canonical_facts",
    return_value=_get_canonical_facts_response({"subscription_manager_id": str(uuid4())})
)
@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=True)
```

insights/tests/client/connection/test_checkin.py
```diff
@@ -24,6 +24,7 @@ def _get_canonical_facts_response(canonical_facts):
 @patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
```
Glutexo, line 24: I rêmembered! `get_rm_conf` is called from the depths of `get_canonical_facts`. Getting the canonical facts is not as trivial as it might sound like and it initializes a Kraken-sized component.

insights/tests/client/connection/test_checkin.py
```diff
@@ -33,7 +34,7 @@ def _get_canonical_facts_response(canonical_facts):
     **{"return_value.status_code": codes.CREATED}
 )
 @patch("insights.client.connection.InsightsConnection.get_proxies")
-def test_canonical_facts_request(get_proxies, post, get_canonical_facts, rm_conf):
+def test_canonical_facts_request(get_proxies, post, get_canonical_facts, fetch_system_by_machine_id, rm_conf):
```
Glutexo, line 37: In an ideal world… 🤷🏻‍♂️ But no problem, let’s keep it.

```suggestion
def test_canonical_facts_request(get_proxies, fetch_system_by_machine_id, get_canonical_facts, post, rm_conf):
```

insights/tests/client/connection/test_checkin.py
```diff
@@ -54,6 +55,7 @@ def test_canonical_facts_request(get_proxies, post, get_canonical_facts, rm_conf
 @patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
 @patch(
     "insights.client.connection.InsightsConnection._init_session", Mock())
```
Glutexo, line 57: This is another thing I did wrong when writing the original tests. 😢 

```suggestion
    "insights.client.connection.InsightsConnection._init_session")
```

would be the correct notation. It’s not related to your pull request. I just noticed it and it made me sad.

insights/tests/client/connection/test_checkin.py
```diff
@@ -63,7 +65,7 @@ def test_canonical_facts_request(get_proxies, post, get_canonical_facts, rm_conf
     **{"return_value.status_code": codes.CREATED}
 )
 @patch("insights.client.connection.InsightsConnection.get_proxies")
-def test_canonical_facts_request_cleaned(get_proxies, post, get_canonical_facts, rm_conf):
+def test_canonical_facts_request_cleaned(get_proxies, post, get_canonical_facts, fetch_system_by_machine_id, rm_conf):
```
Glutexo, line 68: If the above was true, the function signature would be different.

```suggestion
def test_canonical_facts_request_cleaned(get_proxies, post, get_canonical_facts, fetch_system_by_machine_id, init_session, rm_conf):
```

Just saying to finish the story, nothing related to your code again.

insights/tests/client/connection/test_checkin.py
Glutexo: Whatever. 🤷🏻‍♂️

insights/tests/client/connection/test_checkin.py
```diff
@@ -191,3 +198,22 @@ def test_response_failure(get_proxies, post, get_canonical_facts, rm_conf, statu
 
     with raises(Exception):
         connection.checkin()
+
+
+@patch('insights.client.connection.InsightsUploadConf.get_rm_conf', return_value={})
+@patch("insights.client.connection.InsightsConnection._init_session", Mock())
```
Glutexo, line 204: I like the consistency!

insights/tests/client/connection/test_checkin.py
```diff
@@ -191,3 +198,22 @@ def test_response_failure(get_proxies, post, get_canonical_facts, rm_conf, statu
...
+@patch("insights.client.connection.InsightsConnection._fetch_system_by_machine_id", return_value=False)
```
Glutexo, line 205: Here too. 👍🏻

insights/tests/client/connection/test_checkin.py
```diff
@@ -191,3 +198,22 @@ def test_response_failure(get_proxies, post, get_canonical_facts, rm_conf, statu
...
+@patch("insights.client.connection.get_canonical_facts")
+@patch("insights.client.connection.InsightsConnection.post")
+@patch("insights.client.connection.InsightsConnection.get_proxies")
+def test_unregistered_host(get_proxies, post, get_canonical_facts, fetch_system_by_machine_id, rm_conf):
+    """
+    If the host is not registered, the checkin fails.
```
Glutexo, line 211: I can imagine a bit more descriptive description like 

```suggestion
    If the host is not found in the inventory, the checkin doesn't happen.
```

This is a test, not a message for the user, so we can be more technically precise about what’s going on.

No need to block the PR because of this, however.

insights/tests/client/connection/test_checkin.py
```diff
@@ -191,3 +198,22 @@ def test_response_failure(get_proxies, post, get_canonical_facts, rm_conf, statu
...
+    """
+    config = InsightsConfig(base_url="www.example.com")
+    connection = InsightsConnection(config)
+
+    result = connection.checkin()
+    assert result is False
```
Glutexo, line 217:
```suggestion
    assert result is False
    
```

Glutexo (approved):
:shipit:

LGTM

zpetrace (approved):
Looks good from the QE POV, verification comment is in the related card - nice work Peter!

PR #4301, opened 2024-12-07
feat: Print config on connection test

Glutexo:
### All Pull Requests:

Check all that apply:

For non-legacy, `--test-connection` dumps a user-friendly connection configuration.

First, the authentication information is printed starting with type. For `BASIC`, username is printed. For `CERT`, certificate and key paths are verified and printed. In case of missing files or credentials, the connection test fails immediately.

Second, tested URLs (base, Ingress, Inventory, API cast) are listed and server type (production, staging, Satellite) is determined. HTTPS proxy information is included.

Card IDs:
* [CCT-963](https://issues.redhat.com/browse/CCT-963)

### DIFF
--- a/insights/client/client.py
+++ b/insights/client/client.py
@@ -326,7 +326,7 @@ def collect(config):
     pc = InsightsUploadConf(config)
     dc = CoreCollector(config)
 
-    logger.info('Starting to collect Insights data for %s' % determine_hostname(config.display_name))
+    logger.info('* Starting to collect Insights data for %s' % detsermine_hostname(config.display_name))
 
     dc.run_collection(pc.get_rm_conf(),
                       get_branch_info(config),
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -388,6 +388,76 @@ class InsightsConnection(object):
             raise
 
     def test_connection(self, rc=0):
+        logger.info("Running Connection Tests...")
+        if self.config.legacy_upload:
+            return self._legacy_test_connection(rc)
+
+        authentication_failures = []
+        if self.authmethod == "BASIC":
+            logger.info("\nAuthentication: login credentials ({})".format(self.authmethod))
+
+            for credential_description, var, placeholder in [
+                ("Username", self.username, None),
+                ("Password", self.password, "********"),
+            ]:
+                if not var:
+                    authentication_failure = "{} not set.".format(credential_description)
+                    authentication_failures.append(authentication_failure)
+
+                val = placeholder or var if var else "not set"
+                logger.info("  %s: %s", credential_description, val)
+        elif self.authmethod == "CERT":
+            logger.info("\nAuthentication: identity certificate ({})".format(self.authmethod))
+
+            for path_description, path_func in [
+                ("Certificate", rhsmCertificate.certpath),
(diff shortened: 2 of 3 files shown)


m-horky (changes requested):
Reviewing without running the code first. The methods seem reasonably split, that's a good indicator of it being maintainable, good job on that.

My biggest problem with this approach is that you are using logs as public API. Your output, to be pretty-printed, is prefixed with a bunch of space characters, and the fact that the stdout is your API makes your tests very verbose and difficult to handle. There is no separation between the business logic and the presentation.

Instead, I'd much rather see the tests return some kinds of objects (class instances, named tuples, exceptions, whatever) that would contain information about what has gone wrong.

For example, in `_test_url_config()`, we could do something like
```py
try:
    parsed = urllib3.util.url.parse_url(url)
except urllib3.exceptions.LocationParseError:
    return ConfigurationIssue(
        cause="Malformed URL",
        url=url,
        description="API URL"
    )
```
what would get pretty-printed somewhere else. This would make your tests much simpler and smaller.

I know the Core codebase does it, but please don't misuse logging for standard output. For this output-heavy command, let's use `print`s instead. I'd like to see logs as well, for both ok and failed findings, to allow support and engineers debug network issues, though

```
14:37:01.001 ... test_connection OK: test=proxy-url, result=https://proxy.internal
14:37:01.003 ... test_connection OK: test=proxy-connection, result=can establish connection
```

that could be presented as text
```
Running tests...
- Proxy configuration is valid: OK
- Proxy is reachable: OK
  ...
```

Glutexo:
Copying my JIRA [comment](https://issues.redhat.com/browse/CCT-963?focusedId=26503443&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26503443):

To focus on another task, I am leaving this pull request to Petter. I pushed everything I have done so far.

Matyáš’s review comments need to be addressed. Most importantly his request to split the presentation (error messages) from the logic (the actual connection tests). The test URLs method should return an object, which is then separately translated to the output. Doing so, the function can be mocked and tests are less brittle. (I was thinking about putting this in a follow-up card, keeping the current Insights Client style for now. Matyáš doesn’t want to merge without the split though.)

Other things that need to be done are:
- The messages suggesting to check Red Hat Status page should only be printed if the Client connects to Red Hat servers, not Satellite.
- Existing tests (currently marked as skipped) need to be either fixed or removed.
- The pull request description and the commit message should describe its content. Currently, it’s an unedited squash of multiple outdated commit messages.

I thought it would be appropriate to change all the non-debug test connection log messages to the `INFO` level: even a failure is not an error, just an expected output. Doing so would remove log level mixing and simplify the tests a lot: a simple function expanding a list of messages into caplog record tuples would save a lot of repetition.

Personally, I would think parametrize the `insights_connection` fixture configured by the `insights_config` marker. That would allow using the fixture in the few tests that currently create the object manually. See [Parametrizing fixtures](https://docs.pytest.org/en/latest/how-to/fixtures.html#parametrizing-fixtures) in Pytest documentation.

pkoprda:
Since for now we don't want to introduce any further regression into the client functionality, this PR can be closed.
@xiangce can you close this PR?

PR #4309, opened 2024-12-17
feat: add a base parser LazyLogFileOutput

xiangce:
- Doesn't like the LogFileOutput, this LazyLogFileOutput doesn't load the
 content during initialization.  Its content will be loaded later whenever
 the parser instance being used.
- It's useful for the cases where need to load thousands of files that
  belong to one single Spec in one pass of running.
- For implementation's convenience, converted the `"scanners"` from `list`
  `dict` in which the `result_key` is key and `scanner` is value.  And
  removed the `scanner_keys`.
- Updated the LsPci accordingly.
- see RHINENG-13810

### All Pull Requests:

Check all that apply:

- As some passive updates done by `black` pre-commit
  were also committed in this PR, please majorly check this
  [code slice](https://github.com/RedHatInsights/insights-core/pull/4309/files#diff-3f51854c15b54335dc91cd8acf82eff62d9aad50b6d3d413317f82f7b00479d7R1379-R1410) which is for this change.

### DIFF
--- a/insights/core/__init__.py
+++ b/insights/core/__init__.py
@@ -77,7 +77,9 @@ class Parser(object):
     """
 
     def __init__(self, context):
-        self.file_path = os.path.join("/", context.relative_path) if context.relative_path is not None else None
+        self.file_path = (
+            os.path.join("/", context.relative_path) if context.relative_path is not None else None
+        )
         """str: Full context path of the input file."""
         self.file_name = os.path.basename(context.path) if context.path is not None else None
         """str: Filename portion of the input file."""
@@ -150,6 +152,7 @@ def flatten(docs, pred):
                 if c.children:
                     c.children = inner(c.children, stack)
         return results
+
     return inner(docs, [])
 
 
@@ -340,6 +343,7 @@ class ConfigParser(Parser, ConfigComponent):
     Raises:
         SkipComponent: When input content is empty.
     """
+
     def parse_content(self, content):
         if not content:
             raise SkipComponent('Empty content.')
@@ -359,6 +363,7 @@ class ConfigCombiner(ConfigComponent):
     include directives for supplementary configuration files. httpd and nginx
     are examples.
     """
+
     def __init__(self, confs, main_file, include_finder):
         self.confs = confs
         self.main = self.find_main(main_file)
@@ -396,6 +401,7 @@ class ContainerConfigCombiner(ConfigCombiner):
     files with include directives for supplementary configuration files.
     httpd and nginx are examples.
     """
+
     def __init__(self, confs, main_file, include_finder, engine, image, container_id):
         self.image = image
         """str: The image of the container."""
@@ -516,8 +522,7 @@ class SysconfigOptions(Parser, LegacyItemAccess):
 
             # Either only one thing or line or rest starts with comment
             # but either way we need to have an equals in the first word.
-            if (len(words) == 1 or (len(words) > 1 and words[1][0] == '#')) \
-                    and '=' in words[0]:
+            if (len(words) == 1 or (len(words) > 1 and words[1][0] == '#')) and '=' in words[0]:
                 key, value = words[0].split('=', 1)
                 result[key] = value
             # Only store lines if they aren't comments or blank
@@ -527,7 +532,7 @@ class SysconfigOptions(Parser, LegacyItemAccess):
         self.unparsed_lines = unparsed_lines
 
     def keys(self):
-        """ Return the list of keys (in no order) in the underlying dictionary."""
+        """Return the list of keys (in no order) in the underlying dictionary."""
         return self.data.keys()
 
 
@@ -544,11 +549,11 @@ class CommandParser(Parser):
     """
 
     __bad_single_lines = [
-            "no such file or directory",
-            "not a directory",
-            "command not found",
-            "no module named",
-            "no files found for",
+        "no such file or directory",
+        "not a directory",
+        "command not found",
+        "no module named",
+        "no files found for",
     ]
     """
     This variable contains filters for bad responses of the single line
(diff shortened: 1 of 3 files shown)


insights/core/__init__.py
```diff
@@ -1161,6 +1181,29 @@ class parser :py:class:`TextFileOutput`.
...
+        """
+        for scanner in self.scanners:
+            scanner(self)
+
+    @property
+    def lines(self):
+        if self._lines is None:
+            # one-shot load all lines here
+            self._lines = self._context.content
+        return self._lines
```
qinpingli, line 1205: @xiangce Seems the following code is enough:

```
@property
    def lines(self):
         return self._context.content
```
It can ensure the data is loaded once and all lines passed the filters.

insights/core/__init__.py
```diff
@@ -1161,6 +1181,29 @@ class parser :py:class:`TextFileOutput`.
...
+        self.parse_content(None)
+
+    def parse_content(self, content=None):
+        """
+        Use all the defined scanners to search the log file, setting the
+        properties defined in the scanner.  However, do not pre-load the lines.
+
+        Keep None as a fake keyword argument for compatibility.
+        """
+        for scanner in self.scanners:
```
qinpingli, line 1197: @xiangce Because `_handle_content` calls `parse_content`, so it means `parse_content` is called when parsers are initialized.

So, it seams that files are loaded in the parser's initialization?

insights/core/__init__.py
```diff
@@ -1161,6 +1181,29 @@ class parser :py:class:`TextFileOutput`.
     * A None value when there is no timestamp info in the log file
     """
 
+    def _handle_content(self, context):
```
qinpingli, line 1184: Question: From the implementation, it looks like that `LogFileOutput` only supports lazy content loading, is this correct?

insights/core/__init__.py
xiangce: Agree, but put such an explicit operation here would be safe and no hurt.

insights/core/__init__.py
xiangce: Nope, it depends.
If no re-implementation of the `parse_content` in the child Parsers, this one will be called without any content being loaded.
Unless the "self.lines` is used (or called) in the re-implemented `parse_content`, the content will never be loaded in initialization.

insights/core/__init__.py
xiangce: Yup, I don't think it's required to add another one base class for this feature, as it doesn't affect the rules' execution on insights-archvies.

insights/core/__init__.py
qinpingli: In the scanners, the `self.lines` are called. Such as:
```
@classmethod
    def token_scan(cls, result_key, token, check=all):
        """ 
        Define a property that is set to true if the given token is found in
        the log file.  Uses the __contains__ method of the log file.
        
        Parameters:
            result_key(str): the scanner key to register
            token(str or list): one or more strings to search for
            check(func): built-in function ``all`` or ``any`` applied to each line
        
        Returns:
            (bool): the property will contain True if a line contained (any
            or all) of the tokens given.
        """
        def _scan(self):
            search_by_expression = self._valid_search(token, check)
            return any(search_by_expression(l) for l in self.lines)
        
        cls.scan(result_key, _scan)

Comment (xiangce) [reply insights/core/__init__.py:1197]: Nope, the `*scan` functions just register the `result_key` and its scan function `_scan` to the `self.scanners`, but never calls it. The `_scan` will be only actually called when the registered `result_key` being called.
E.g.

```python
>>> MyTexter.token_scan('find_more', 'more')
>>> my_texter.find_more  # <--<< Here the `self.lines` will be finally called.
```

xiangce: Hm.., I might be wrong. Let me have a double check.

xiangce: @qinpingli - The former implementation is incorrect, I add a new `LazyLogFileOutput` which needs to load the content by hand explicitly. Let's have a talk about this one.

insights/tests/test_lazylogfileoutput.py
```diff
@@ -0,0 +1,362 @@
...
+    log.do_scan('puppet_master_logs')
+    assert hasattr(log, 'puppet_master_logs')
+    log.do_scan('kernel_logs')
+    assert hasattr(log, 'kernel_logs')
+    log.do_scan('middleware_exception_present')
+    assert hasattr(log, 'middleware_exception_present')
+    log.do_scan('cron_present')
+    assert hasattr(log, 'cron_present')
+    log.do_scan('lost_messages')
+    assert hasattr(log, 'lost_messages')
```
JoySnow, line 119: [nitpick] 
The codecov reports for [partial coverage lines](https://app.codecov.io/gh/RedHatInsights/insights-core/pull/4309?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=RedHatInsights#e3ed9166e35444a62975de12bf0d55b7-R1429) on:
```
+     def do_scan(self, result_key=None):
+         if result_key:
+             if result_key not in self._scanned and result_key in self.scanners:     <-- this line
+                 self.scanners[result_key](self)
+                 self._scanned.add(result_key)
+         else:
+             for key, scanner in self.scanners.items():
+                 if key not in self._scanned:                                        <-- and this line
```

@xiangce  Would it be possible to add a few more test lines to trigger this `key in self._scanned`?  And `result_key not in self.scanners` if needed?

xiangce: Sure things, will add them

xiangce: Done, but the Codecov cannot sync the result back to the PR so quickly. It's covered, see https://app.codecov.io/gh/RedHatInsights/insights-core/pull/4309?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=RedHatInsights#e3ed9166e35444a62975de12bf0d55b7-R1429

JoySnow (approved):

PR #4341, opened 2025-01-20
fix: 'tail' filterable text files before filtering with 'grep'

chenlizhong:
### All Pull Requests:

Check all that apply:

As some logs files are not with log rotate setting, these files contain old logging and more than 100 million lines. Use the 'tail' command to get the last 1 million only.

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -268,7 +268,10 @@ class TextFileProvider(FileProvider):
         if self._filters:
             log.debug("Pre-filtering %s", self.relative_path)
             args.append(
-                ["grep", "-F", "\n".join(sorted(self._filters.keys(), reverse=True)), self.path]
+                ["tail", "-n" ,"1000000", self.path]
+            )
+            args.append(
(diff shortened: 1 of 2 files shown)


insights/core/spec_factory.py
```diff
@@ -268,7 +268,10 @@ def create_args(self):
         if self._filters:
             log.debug("Pre-filtering %s", self.relative_path)
             args.append(
-                ["grep", "-F", "\n".join(sorted(self._filters.keys(), reverse=True)), self.path]
+                ["tail", "-n", "1000000", self.path]
```
xiangce, line 271: It would be good to declare a GLOBAL constant for the line value, it's useful for documentation and future update.

xiangce (commented):
The update looks good. Please check if it's possible to add a unit-test, like this one https://github.com/RedHatInsights/insights-core/blob/master/insights/tests/specs/test_specs_filters.py#L178-L180

xiangce (approved):

PR #4345, opened 2025-01-29
Support non-RHEL systems for Core updates

m-horky:
### All Pull Requests:

Check all that apply:

* Card ID: CCT-1148

This PR adds support for Fedora for egg update process.

First, tests were added for `utilities.os_release_info()`, since there were none.

Second, method `utilities.get_rhel_version()` was created to map RHEL, CentOS and Fedora versions to a single integer representing the RHEL major version the Insights Core targets.

Third, this code was used in `client.__init__` to wrap original `os_release_info()` call in distribution-aware wrapper.

### DIFF
--- a/insights/client/__init__.py
+++ b/insights/client/__init__.py
@@ -19,7 +19,7 @@ from .utilities import (write_data_to_file,
                         get_tags,
                         write_tags,
                         migrate_tags,
-                        os_release_info,
+                        get_rhel_version,
                         get_parent_process)
 
 NETWORK = constants.custom_network_log_level
@@ -131,8 +131,7 @@ class InsightsClient(object):
         self.tmpdir = tempfile.mkdtemp()
         atexit.register(self.delete_tmpdir)
         try:
-            _, os_release = os_release_info()
-            rhel_major = os_release.split('.')[0]
+            rhel_major = get_rhel_version()
             # set egg name as 'insights-core.el#.egg' per RHEL #
             egg_name = 'insights-core.el{0}.egg'.format(rhel_major)
         except Exception:


insights/client/utilities.py
```diff
@@ -471,7 +476,45 @@ def os_release_info():
...
+        raise ValueError("Could not determine version of {}.".format(distribution))
+
+    _distribution = distribution.lower()  # type: str
+    version = 0  # type: int
+    for family_name in ("red hat enterprise linux", "centos"):
+        if family_name in _distribution:
+            version = int(release.split(".")[0])
+    for family_name in ("fedora", ):
+        if family_name in _distribution:
+            version = 10
```
pkoprda, line 510: What about using `.startswith()` instead of for loops? Something like:
```py
if _distribution.startswith("red hat enterprise linux") or _distribution.startswith("centos"):
    ...
elif _distribution.startswith("fedora"):
    ....
else:
    # raise a ValueError
```

insights/client/utilities.py
```diff
@@ -471,7 +476,45 @@ def os_release_info():
...
+    this method returns `10` for all Fedora versions.
+
+    :returns: Major RHEL version (or an equivalent for non-RHEL systems)
+    :raises ValueError: Version could not be determined.
+    """
+    distribution, release = os_release_info()
+    if distribution == "Unknown":
+        raise ValueError("Could not determine distribution family.")
+    if release == "":
+        raise ValueError("Could not determine version of {}.".format(distribution))
```
pkoprda, line 501: Could you write some tests cases for these scenarios?

pkoprda (changes requested):
Thank you @m-horky! Generally looks good, I just left some comments with possible improvements.

Also please rebase.

insights/client/utilities.py
m-horky: I wanted to keep it scalable, but perhaps it's just overengineering. Will update.

pkoprda (approved):
LGTM :+1: let's wait for QE

zhangqianqian:
test me

zpetrace (approved):
LGTM, verification passed as well as the tests, I will leave the comment in the CCT card and this can be merged!:)

m-horky:
I've just rebased on top of master. @xiangce, when the CI is done, we can merge this :)

xiangce:
@m-horky - Thanks for making this change. I believe this PR should also go to the Egg without "el#" postfixes, to ensure the old `insights-core.egg` could guide to the right next update.

PR #4346, opened 2025-02-03
Several speed improvements for parsing ls directory output

PaulWay:

- Avoid re-splitting the line if it's already been split.
- Handle a few failure modes in parsing links and size on '?' and other bad input.
- Each directory is all one 'mode' - normal, or SELinux on RHEL 6/7 or 8+. Instead of trying to detect the mode of each line, detect the mode once and then use that for all lines in this directory.  Difficult to extrapolate that to all listings in a parse - let's see how this goes for now.
- Simplify how we pick up directory names and store previously processed directories.
- Add a few comments to make it easier to see what's being parsed where.

### All Pull Requests:

Check all that apply:

* [Y] No Sensitive Data in this change?
* [Y] Is this PR an enhancement?


xiangce (commented):
@PaulWay - Left a tiny suggestion, please take a look. The Change looks good to me, and per my tests, it improves the parsing significantly when the ls result is large. 👍

xiangce (approved):
It looks good to me

PR #4350, opened 2025-02-11
fix(client): Change registration endpoint to /host_exists

pkoprda:
### All Pull Requests:

Check all that apply:

* Card ID: CCT-721
* Card ID: RHINENG-11711

Currently, the client is using the `/hosts` endpoint to check if the host is registered. This change updates the client to use the `/host_exists` endpoint instead.

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -471,6 +471,8 @@ class InsightsConnection(object):
                    rhsm_hostname != 'subscription.rhsm.redhat.com'):
                     logger.error('Please enable Insights on Satellite server '
                                  '%s to continue.', rhsm_hostname)
+            if req.status_code == 404 or req.status_code == 409:
+                return False
             if req.status_code == 412:
                 try:
                     unreg_date = req.json()["unregistered_at"]
@@ -741,9 +743,9 @@ class InsightsConnection(object):
         try:
             # [circus music]
             if self.config.legacy_upload:
-                url = self.base_url + '/platform/inventory/v1/hosts?insights_id=' + machine_id
+                url = self.base_url + '/platform/inventory/v1/host_exists?insights_id=' + machine_id
             else:
-                url = self.inventory_url + '/hosts?insights_id=' + machine_id
+                url = self.inventory_url + '/host_exists?insights_id=' + machine_id
             res = self.get(url)
         except REQUEST_FAILED_EXCEPTIONS as e:
             _api_request_failed(e)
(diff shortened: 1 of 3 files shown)


m-horky (approved):
LGTM. Let's wait for QE.

pkoprda:
Updated the code. The change of `_fetch_system_by_id()` would break functionality of the options `--display-name`,  `--ansible-host` and the method [show_inventory_deep_link()](https://github.com/RedHatInsights/insights-core/blob/d48f19d9c4e3d76edc7d9992185080d2a4ad0c97/insights/client/__init__.py#L564). 

@m-horky can you do re-review please?

insights/client/connection.py
```diff
@@ -752,14 +754,16 @@ def _fetch_system_by_machine_id(self):
...
             res_json = json.loads(res.content)
+            res_json['insights_id'] = machine_id
         except ValueError as e:
             logger.error(e)
             logger.error('Could not parse response body.')
             return None
-        if res_json['total'] == 0:
-            logger.debug('No hosts found with machine ID: %s', machine_id)
+        if res.status_code == 404:
+            logger.debug(res.json()['detail'])
```
m-horky, line 763: Are we sure the response will always have `'detail'` key? The [official documentation](https://console.redhat.com/docs/api/inventory/v1#operations-hosts-api\.host\.get_host_exists) doesn't contain any schema for 404 response, I'd like to wrap this in `try/except` or otherwise ensure that we don't error out if the response is empty or doesn't contain that key.

insights/tests/client/connection/test_reg_check.py
m-horky, line 1: Would you mind creating a test for responses 404 and 409?

m-horky (changes requested):
I found more stuff :)

insights/client/connection.py
pkoprda: I would rather have there original debug message saying `No hosts found with machine ID: ...` without using `try/expect` block

m-horky: It is valuable to have the response in the log file. We can try to log the content and then always say we didn't find the host there, no matter whether we `except`ed or not.

m-horky (approved):
LGTM. QE time.

zpetrace:
Seems fine from QE side, the verification passed and I have put the comment to the CCT card, this can be merged:)

pkoprda:
@xiangce we can merge this :rocket:

PR #4365, opened 2025-02-26
fix: limit loading content when processing data

xiangce:
- instead of "tail .| grep" commands using python read and apply filters dynamically according to the add_filter
- add a staticmethod to cleaner.filters to do the filtering
- update corresponding test
- see RHINENG-15986

### All Pull Requests:

Check all that apply:

*Add your description here*


insights/core/spec_factory.py
```diff
@@ -291,7 +290,11 @@ def load(self):
             return out
 
         with safe_open(self.path, "r", encoding=encoding, errors="surrogateescape") as f:
-            return [l.rstrip("\n") for l in f]
+            valid_content = [l.rstrip("\n") for l in f][-MAX_CONTENT_LINES:]
```
chenlizhong, line 293: According to this line, all lines of one file will be loaded into memory. For big file (more than 1GB), it will lead to out of memory.   For the command tail,  It does this without reading the entire file first, so it doesn't matter how large the file is.

xiangce: re-do it by loading the last 200MB part of files larger then 200MB.

insights/core/spec_factory.py
```diff
@@ -290,8 +289,19 @@ def load(self):
...
+            if fsize > MAX_CONTENT_SIZE:
+                # read the last ``MAX_CONTENT_SIZE`` MB only
+                f.seek(fsize - MAX_CONTENT_SIZE)
+                log.debug("Extra-huge file is truncated %s", self.relative_path)
+                content = [l.rstrip("\n") for l in f][1:]  # discard the first line which is broken
+            else:
+                content = [l.rstrip("\n") for l in f]
+            if not isinstance(self.ctx, HostContext) and self._filters:
+                # Post-filtering ONLY when processing data
+                return AllowFilter.filter_content(content[::-1], self._filters)[::-1]
```
JoySnow, line 303: For large `content`, the `[::-1]` list reverse action would construct a new list object and cost quite some memory. Is it possible to pass the content as itself without reverse to `AllowFilter.filter_content`, and do a bottom-up filter check in it.

insights/cleaner/filters.py
```diff
@@ -32,3 +32,26 @@ def parse_line(self, line, **kwargs):
...
+        :return: list of lines
+        """
+        result = []
+        for line in lines:
+            for a_key in list(allowlist.keys()):  # copy keys to avoid RuntimeError
+                if a_key in line:
+                    allowlist[a_key] -= 1
+                    # stop checking it when enough lines contain the key were found
+                    allowlist.pop(a_key) if allowlist[a_key] == 0 else None
+                    result.append(line)
```
JoySnow, line 56: Seems we would need just one time of this `result.append(line)` be called for each content line, right?

JoySnow (approved):

xiangce:
test me

PR #4369, opened 2025-02-28
In `from_dict` set fields from ls parser dict correctly

PaulWay:
### All Pull Requests:

Check all that apply:

* [Y] No Sensitive Data in this change?
* [N] Is this PR an enhancement?

The FilePermissions class has a `from_dict` method.  Unfortunately, this does
not return an instance of the class.  Class methods get the class itself as
their first argument, whereas the `from_dict` method called that `self`, so
the `from_dict` method actually returned the class, annotated with properties.

This change correctly invokes the `__new__` class method to generate an
unpopulated class instance, and then populates it with properties from the
given dict.  It also makes sure that the `path` value set by the line regular
expression parsing is copied from the `name` value given in the dict.

This change also provides the `type` property as the first character of the
permissions block, which was previously discarded and had to be reconstructed
from the line.  This is `d`, `b`, `c`, `l` or `-`.


insights/util/file_permissions.py
```diff
@@ -79,27 +79,31 @@ def __init__(self, line):
         self.line = line
         r = self._PERMISSIONS_PATTERN.search(self.line)
         if r:
-            (self.perms_owner, self.perms_group, self.perms_other,
+            (self.type, self.perms_owner, self.perms_group, self.perms_other,
```
xiangce, line 82: @PaulWay would you mind adding the new `self.type` to the above `Attributes` list in docstring?

PaulWay: Good point there @xiangce - will do.

PR #4370, opened 2025-02-28
feat: Add datasource and spec for dirs/files number

wushiqinlou:
rh-pre-commit.version: 2.3.1

### All Pull Requests:

Check all that apply:

Add this datasource to enhance an Advisor rule


wushiqinlou:
@xiangce - please review

xiangce (commented):
How about enhancing the parser to support calculating the count of both `dirs` and `files`? in case their are request to count the dirs?

xiangce (commented):
Looks great to me, but please discard the unrelevant update to malware_detecation before merging.

insights/specs/datasources/files_dirs_number_of.py
```diff
@@ -0,0 +1,35 @@
...
+                item = item + "/"
+            if os.path.exists(item):
+                files_number = 0
+                dirs_number = 0
+                result[item] = {}
+                for name in os.listdir(item):
+                    if os.path.isfile(item + name):
+                        files_number = files_number + 1
+                    else:
+                        dirs_number = dirs_number + 1
```
JoySnow, line 31: To check for the dir type, what about making it more restrict by using `os.path.isdir()`? 

Like, for a file `crw-rw-rw-  1 root   wheel      0x400007c Feb 18 18:30 ttywc` , we would not want to include that as a dir:
```
>>> os.path.isfile("/dev/ttywc")
False
>>> os.path.isdir("/dev/ttywc")
False
```

insights/specs/datasources/files_dirs_number_of.py
```diff
@@ -0,0 +1,35 @@
...
+                result[item]["files_number"] = files_number
+                result[item]["dirs_number"] = dirs_number
+        return DatasourceProvider(content=json.dumps(result, sort_keys=True), relative_path='files_dirs_number')
```
JoySnow, line 34: For the `filters` has value but the final `result` is empty case, the code will still return a `DatasourceProvider` with `{}`. 
According to our usage, I agree that the current handling is more appropriate that raising a `SkipComponent`. 

Just to make it more easy to maintain, what about adding a comment message to note about this case?

insights/parsers/files_dirs_number_of_dirs.py
```diff
@@ -0,0 +1,41 @@
...
+        >>> filesnumberofdir.data["/var/spool/postfix/maildrop/"]["files_number"]
+        5
+        >>> filesnumberofdir.dirs_number_of("/var/spool/clientmqueue/")
+        1
+    """
+
+    def dirs_number_of(self, dir):
+        """Return the number of dirs under specified `dir`"""
+        if dir in self.data:
+            return self.data[dir]["dirs_number"]
```
xiangce, line 36: How about raising an `KeyError` Exception when no such `dir` in `data`?
```
        if dir in self.data:
            return self.data[dir]["dirs_number"]
        raise KeyError("No such %s dir collected", dir)
```
How do you think?

BTW, please add a test for "no such dir".

insights/parsers/files_dirs_number_of_dirs.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def files_number_of(self, dir):
+        """Return the number of files under specified `dir`"""
+        if dir in self.data:
+            return self.data[dir]["files_number"]
```
xiangce, line 41: Same comment as above

xiangce (approved):

JoySnow (approved):

PR #4377, opened 2025-03-06
fix(client): Always set rhsmCertificate.PATH attribute

m-horky:
### All Pull Requests:

Check all that apply:

Some tests may fail when the attribute is missing. We should always be able to test whether an identity certificate exists or not.

### DIFF
--- a/insights/client/cert_auth.py
+++ b/insights/client/cert_auth.py
@@ -23,7 +23,7 @@ class rhsmCertificate:
     try:
         PATH = RHSM_CONFIG.get('rhsm', 'consumerCertDir')
     except:
-        pass
+        PATH = "/etc/pki/consumer/"
 
     KEY = 'key.pem'
     CERT = 'cert.pem'


insights/client/cert_auth.py
```diff
@@ -23,7 +23,7 @@ class rhsmCertificate:
     try:
         PATH = RHSM_CONFIG.get('rhsm', 'consumerCertDir')
     except:
-        pass
+        PATH = "/etc/pki/consumer/"
```
pkoprda, line 26: Maybe I would add here some `logger.warning()` that it failed to get `consumerCertDir` from `RHSM_CONFIG`

Non-blocking

m-horky: That would only happen if you didn't have subscription-manager installed, which isn't a supported configuration. Other parts of the code already raise errors about it:

https://github.com/RedHatInsights/insights-core/blob/53b5ff55ce3ef6b47b444089f47c7357694a566a/insights/client/connection.py#L151-L157

pkoprda: Okay, then just please rebase

PR #4388, opened 2025-03-12
feat(client): Disable API registration check

pkoprda:
### All Pull Requests:

Check all that apply:

* Card ID: CCT-869

This update checks registration status using the `.registered` file instead of querying Inventory, which improves performance by reducing API calls. If the file exists, an upload proceeds.

### DIFF
--- a/insights/client/phase/v1.py
+++ b/insights/client/phase/v1.py
@@ -230,12 +230,7 @@ def post_update(client, config):
         return
 
     # check registration status before anything else
-    if isfile(constants.machine_id_file):
-        reg_check = client.get_registration_status()
-        if reg_check is None:
-            sys.exit(constants.sig_kill_bad)
-    else:
-        reg_check = False
+    reg_check = isfile(constants.registered_files[0])
 
     # --status
     if config.status:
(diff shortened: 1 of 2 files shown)


m-horky:
I'm surprised all unit tests passed, they are failing locally for me. Integration tests are very unhappy about it as well. We'll need to sort them out before we release this.

insights/client/phase/v1.py
```diff
@@ -230,12 +230,7 @@ def post_update(client, config):
         return
 
     # check registration status before anything else
-    if isfile(constants.machine_id_file):
-        reg_check = client.get_registration_status()
-        if reg_check is None:
-            sys.exit(constants.sig_kill_bad)
-    else:
-        reg_check = False
+    reg_check = isfile(constants.machine_id_file) and isfile(constants.registered_files[0])
```
m-horky, line 233: The `machine-id` file is created proactively, we'll need to only check for the `.registered`.

m-horky (approved):
Other than the comment, it's looking fine. I've played with it and I didn't discover any weird behavior that would be caused by this change.

zpetrace:
The verification PASSED, more info is in the corresponding CCT card:) I think we can merge @xiangce

PR #4386, opened 2025-03-12
fix: handle unexpected lines in parser LvmFullReport

JoySnow:
### All Pull Requests:

Check all that apply:

* [x] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [x] No Sensitive Data in this change?
* [x] Is this PR to correct an issue?
* [ ] Is this PR an enhancement?

### Complete Description of Additions/Changes:

 - the input of parser LvmFullReport may contains unexpected WARNING message lines in the middle of json format content. Enhance the parser to pre-handling those warning lines
 - return the index instead of actual line in the shared function find_warnings() to speed up the later content filtering

serve RHINENG-16590

PR #4391, opened 2025-03-18
fix: warnings keywords over-matching in LvmFullReport

JoySnow:
### All Pull Requests:

Check all that apply:

  The former added warning keywords filtering breaks one of the testcase
  in extraction-rules repo, eg. "duplicate" is a keyword from the lvm
  report json, and should not be filtered out as warning messages.
  
  ~~Drop the whole find_warnings() keywords in LvmFullReport, to avoid
  any hidden over-matching, instead, filter only on the known "WARNING:"
  lines in the middle of the json. And copyied that testcase data here.~~

  Consider all the keywords in find_warnings() are not just single word,
  except this "duplicate".  Fix by changing the keyword to " duplicate ".
  This also brings the consistency on warning messages handling
  among the lvm commands.

The former LvmFullReport change: https://github.com/RedHatInsights/insights-core/pull/4386

### DIFF
--- a/insights/parsers/lvm.py
+++ b/insights/parsers/lvm.py
@@ -86,7 +86,7 @@ def find_warnings(content):
             "failed.",
             "Invalid metadata",
             "response failed",
-            "duplicate",
+            " duplicate ",
             "not found",
             "Missing device",
             "Internal error",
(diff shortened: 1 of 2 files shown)


JoySnow:
@chenlizhong Updated based on the test data that you found in extraction-rules repo. Please help double check this.

xiangce (commented):
From my perspective, the issue can simply be fixed by replace the key work `"duplicate"` in the list with `" duplicate "`.

JoySnow:
@bfahr Yep for this case.  Glad that @chenlizhong found the test data from yours in another repo.  
This `pv_duplicate` is really a keyword of lvm, https://github.com/lvmteam/lvm2/blob/master/lib/report/values.h#L83

JoySnow:
@chenlizhong @xiangce The fix approach was updated accordingly. Would you help review again?

PR #4402, opened 2025-03-21
feat: add new parser openshift-kube logs

mohitkumarrh:
### All Pull Requests:

Check all that apply:

*A parser is needed to close the Jira ticket [INSIGHTOCP-2134](https://issues.redhat.com/browse/INSIGHTOCP-2134). This Jira is a sub-task for an internal CCX rule, which requires these two parsers during the rule development.*


mohitkumarrh:
@xiangce - can you please review this PR, Thanks !

insights/parsers/openshift_kube_log.py
```diff
@@ -0,0 +1,57 @@
...
+
+ApiServerLog - file ``/var/log/pods/openshift-kube-apiserver*/*/*.log``
+-----------------------------------------------------------------------
+
+.. note::
+    Please refer to the super-class :class:`insights.core.LogFileOutput`
+"""
+
+from insights.specs import Specs
+from .. import LogFileOutput, parser
```
xiangce, line 19: Please use the uniform style for these importings.
```
from insights.core import LogFileOutput, parser
```
Thanks.

xiangce (changes requested):
@mohitkumarrh please double check the typical content of the two log files. From your PR, I can see they are different from the content you used in test cases. If the test case is the right one, please set the https://github.com/RedHatInsights/insights-core/blob/master/insights/core/__init__.py#L1183 correspondingly.

insights/parsers/openshift_kube_log.py
mohitkumarrh: updated as `from insights import LogFileOutput, parser` instead of `from insights.core import LogFileOutput, parser
`

xiangce:
> @xiangce - Thanks for the review, the data used in the test file is correct, as I have verified it from a real sosreport. I have also updated the time_format accordingly. Please review and let me know if any further changes are needed.

@mohitkumarrh - Thanks for the confirmation and update the code. Please update the typical content used in the docstring as well.

mohitkumarrh:
@xiangce - I have updated the content used in the docstring, please have a look and review, Thanks !

xiangce (approved):

PR #4409, opened 2025-03-31
feat: introduce a unique obfuscation configure entry: obfuscation_list

xiangce:
- Deprecate the old options: obfuscate and obfuscation
- Data collection will only work with the new obfuscation_list Old options will be convert to new option correspondingly The old options should work as they were for backward compatibility
- Update relevant tests
- Jira: RHINENG-17075

### All Pull Requests:

Check all that apply:

*Add your description here*


xiangce:
The `smoke` tests will be failed because that some existing tests which is newly added for IPv6/MAC should be updated according to the new option.

xiangce:
@JoySnow, please kindly review.

@m-horky, Since the obfuscation part of the `client._validate_options` is updated in this PR, please take a look too.

xiangce:
Just added another condition for printing the deprecation warning: it will only be printed when Egg version is newer than 3.6.0 (T.B.D) on RHEL 9+

JoySnow (approved):
@xiangce The code LGTM.  And,  the deprecation warning is target for Egg version >= 3.6.0 on RHEL 8+ in code, are we going to change it to RHEL 9+ ?

xiangce:
> @xiangce The code LGTM. And, the deprecation warning is target for Egg version >= 3.6.0 on RHEL 8+ in code, are we going to change it to RHEL 9+ ?

My bad, it's a language expression issue, I take it for granted the `RHEL 8+` includes all RHEL 8 and other higer RHEL major versions e.g. RHEL 9/10. 

I'll update the relevant comment in the code soon.

insights/client/config.py
```diff
@@ -681,6 +682,52 @@ def _validate_options(self):
...
         Make sure there are no conflicting or invalid options
         '''
+
+        def _validate_obfuscation_options():
+            # When old switches are set explicitly, even set as False
+            if self.obfuscate is not None or self.obfuscate_hostname is not None:
+                # Warning deprecation only on RHEL 8+ and from egg v 3.6.0 (planned)
+                if self._print_errors and get_rhel_version() > 7:
+                    egg_ver = get_version_info()['core_version'].split('-')[0]
+                    if [int(i) for i in egg_ver.split('.')] >= [3, 6, 0]:
```
pkoprda, line 692: Can you maybe export this code into some function (e.g. create a function in `utilities.py`) which would return a tuple? For example something like this:
```python
def get_egg_version():
    core_version = get_version_info()['core_version'].split('-')[0]
    return tuple(int(i) for i in min_version.split('.'))
````

File: insights/client/config.py [hunk 2]
```diff
@@ -681,6 +682,52 @@ def _validate_options(self):
...
+                        sys.stdout.write(
+                            'WARNING: `obfuscate` and `obfuscate_hostname` are deprecated, please use `obfuscation_list` instead.\n'
+                        )
```
pkoprda, line 695: I would rather to use here `logging.warning()` instead of `sys.stdout.write("WARNING: ...")`. Then we can see the logs in structured way in a log file.

insights/client/config.py
```diff
@@ -681,6 +682,52 @@ def _validate_options(self):
...
+                if self.obfuscation_list is not None:
+                    # When BOTH new and old switches are set explicitly
+                    if self._print_errors:
+                        sys.stdout.write(
+                            'WARNING: Conflicting options: `obfuscation_list` and `obfuscate`, using: "obfuscation_list={0}"\n'.format(
+                                self.obfuscation_list
+                            )
```
pkoprda, line 702: ditto here

insights/client/config.py
```diff
@@ -681,6 +682,52 @@ def _validate_options(self):
...
+                obf_opt = self.obfuscation_list or list()
+                if isinstance(obf_opt, six.string_types):
+                    obf_opt = set(ow.strip() for ow in obf_opt.strip('\'"').split(','))
+                invalid_opt = set(ow for ow in obf_opt if ow not in DEFAULT_OBFUSCATIONS)
+                self.obfuscation_list = sorted(set(obf_opt) - invalid_opt)
+                if invalid_opt and self._print_errors:
+                    sys.stdout.write(
+                        'WARNING: ignoring invalid obfuscate options: `{0}`, using: "obfuscation_list={1}"\n'.format(
+                            '`, `'.join(invalid_opt), ','.join(self.obfuscation_list)
+                        )
```
pkoprda, line 725: ditto here

pkoprda (changes requested):

insights/client/config.py
xiangce: Yup, using `sys.stdout.write` was just to keep the uniform style in this `insights.client.config` python module, as it was used in other places as well. If you are good with that, I'm happy to change all the new addeds and the existing others to `logger.warning`.

insights/client/config.py
xiangce: It's totally okay to me to wrap it with a function. As thinking here were the only usage point, didn't do that. Will do it. Thanks.

insights/client/config.py
pkoprda: I understand that it may not be consistent with the style in this file, but IMO we are trying to have `logging` module instead of just printing it to console by using `sys.stdout`.

insights/client/config.py
```diff
@@ -681,6 +681,51 @@ def _validate_options(self):
...
+                    self.obfuscation_list.append('hostname') if self.obfuscate_hostname else None
+                # discard old options
+                self.obfuscate = self.obfuscate_hostname = None
+            # Re-format the new obfuscation_list
+            if self.obfuscation_list is not None and type(self.obfuscation_list) is not list:
+                obf_opt = self.obfuscation_list or list()
+                if isinstance(obf_opt, six.string_types):
+                    obf_opt = set(ow.strip() for ow in obf_opt.strip('\'"').split(','))
+                invalid_opt = set(ow for ow in obf_opt if ow not in DEFAULT_OBFUSCATIONS)
+                self.obfuscation_list = sorted(set(obf_opt) - invalid_opt)
```
pkoprda, line 718: You are creating a set for variable, which might be already a set. Also `ow` is a confusing variable. What about something like this:
```python
if isinstance(obf_raw, six.string_types):
    obf_items = set(item.strip() for item in obf_opt.strip('\'"').split(','))
else:
    obf_items = set(obf_opt)

# Separate valid and invalid entries
self.obfucation_list = sorted(obf_items & DEFAULT_OBFUSCATIONS)   # Set intersection
invalid_items = obf_items - DEFAULT_OBFUSCATIONS
```

pkoprda (changes requested):

insights/client/config.py
```diff
@@ -1,11 +1,15 @@
...
-import os
-import logging
+
 import argparse
 import copy
+import logging
+import os
 import six
-import sys
+
```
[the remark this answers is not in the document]
xiangce: I removed it and updated all the relevant tests.

insights/client/archive.py
```diff
@@ -139,12 +143,7 @@ def copy_dir(self, path):
         return path
 
     def get_compression_flag(self, compressor):
-        return {
-            "gz": "z",
-            "xz": "J",
-            "bz2": "j",
-            "none": ""
-        }.get(compressor, "z")
+        return {"gz": "z", "xz": "J", "bz2": "j", "none": ""}.get(compressor, "z")
```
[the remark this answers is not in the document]
xiangce: Yes, these are done by the pre-commit-hook (black) in my local, however as it's not easy to split them out from this PR as they are mixed the PR together.

As there are just uniform format updates which won't impact the running, please ignore them this time.

And we'll soon add the `pre-commit` (includes `black`) soon to the PR's CI/CD.

insights/client/config.py
xiangce: Yup, it's true, however, I don't think set(set()) would hurt the code, I'd keep it as it was.

pkoprda (approved):
LGTM, thank you, I think we can merge this @xiangce :rocket:

xiangce:
> LGTM, thank you, I think we can merge this @xiangce 🚀

@pkoprda Thanks a lot for your review. It helps a lot! Once it passes our QE's verification, I'll merge it.

PR #4414, opened 2025-04-03
fix: improve wording for unassignable Policy

romanblanco:
https://issues.redhat.com/browse/RHINENG-10847

When Compliance tries to assign Host to a Policy, it returns either 202 in case of success or 404 in any other case.

When one tries to assign a Host that already has a Policy assigned, the Policy is not anymore in the list of available options, hence the process can not succeed with creating the link and returns 404.

That in the logic of insights-client seems as failure, and it returns non-zero error status with the confusing message `"Policy ... does not exist"`.

This commit is to improve the message sent in such case to make it less misleading.

### DIFF
--- a/insights/specs/datasources/compliance/__init__.py
+++ b/insights/specs/datasources/compliance/__init__.py
@@ -235,10 +235,14 @@ class ComplianceClient:
         logger.debug("Content of the response {0} - {1}".format(response, response.content))
 
         if response.status_code == 202:
-            logger.info("Operation completed successfully.\n")
+            logger.info("System successfully assigned to policy.\n")
             return 0
         else:
-            logger.error("Policy ID {0} does not exist.".format(policy_id))
+            logger.error(
+                "Policy ID {0} can not be assigned. "
+                "Refer to the /var/log/insights-client/insights-client.log for more details."
+                .format(policy_id)
+            )
             return constants.sig_kill_bad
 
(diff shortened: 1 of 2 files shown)


insights/specs/datasources/compliance/__init__.py
```diff
@@ -227,7 +227,7 @@ def policy_link(self, policy_id, opt):
             logger.info("Operation completed successfully.\n")
```
Mishrasubha, line 227: can we also change `Operation completed successfully` to something more specific like `Policy successfully assigned`?

xiangce:
@romanblanco, Please also update the expected assert string in tests. Thanks.

romanblanco:
> as there are might be 2 cases:
> 
>  1. Policy doesn't exist
>  2. Policy exists but host can't be assigned due to already existing association
> 
> For the 2nd case we might want to add additional request to else block: systems/{ID}/policies?limit=100&ids_only=true and check if policy id is already there - then we return message "Policy already associated to host"
> And if not - "Policy doesn't exist"
> 
> It's not cool to have another request for that but it will be beneficial for all of us + this request is easy one for backend.

Agreed with @LightOfHeaven1994 to add this change in a separate PR.

LightOfHeaven1994:
@xiangce can you help us get it merged, please?

PR #4415, opened 2025-04-06
feat: new datasources: host_key_files

shlao:
### All Pull Requests:

Check all that apply:

Add datasource to check why sshd couldn’t load host key.

### DIFF
--- a/insights/specs/datasources/host_key_files.py
+++ b/insights/specs/datasources/host_key_files.py
@@ -1,0 +1,50 @@
+"""
+Custom datasources relate to `/etc/*dsa*_key`
+"""
+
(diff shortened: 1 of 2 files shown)


xiangce (commented):
It's not a good idea to read the `/var/log/messages` directly in datasource as it would be huge.
If you have to do that, please use the existing spec, as it's filtered.

Finally, if the file on customer side is huge even after filtering, this datasource will be time/memory consuming.

xiangce:
@shlao as mentioned, we need to reconsider if it's worth adding this spec to core collection, do you have a Jira card to get this spec being approved first?

PR #4426, opened 2025-04-16
feat: Enhance parser IlabConfigShow

wushiqinlou:
### All Pull Requests:

Check all that apply:

* [ ] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [ ] No Sensitive Data in this change?
* [ ] Is this PR to correct an issue?
* [ ] Is this PR an enhancement?

### Complete Description of Additions/Changes:

Enhance this parser as `YAMLParser` function can not handle the command output directly

PR #4429, opened 2025-04-16
enhance: YAMLParser can skip specified lines

xiangce:
- by adding the first words of the lines, the YAMLPasers can skip them

### All Pull Requests:

Check all that apply:

*Add your description here*


xiangce:
#4426 depends on this PR

insights/core/__init__.py
```diff
@@ -765,9 +765,29 @@ class YAMLParser(Parser, LegacyItemAccess):
...
+    * All keywords added in this list must be in lowercase.
+    """
+
     def parse_content(self, content):
+        def _ignore_lines(content, ignore_lines):
+            idxs = []
+            if content and ignore_lines:
+                for idx, line in enumerate(content):
+                    # Skipping lines start with any keyword of ignore_lines
+                    if line.strip().lower().startswith(tuple(ignore_lines)):
```
JoySnow, line 780: What about making this `tuple(ignore_lines)` conversion ahead of this for loop? so that it can be reused inside the loop.

insights/core/__init__.py
```diff
@@ -765,9 +765,29 @@ class YAMLParser(Parser, LegacyItemAccess):
...
 
+    ignore_lines = []
+    """
+    The first keywords of the lines that need to ignored in the parsing process.
+    * All keywords added in this list must be in lowercase.
+    """
+
     def parse_content(self, content):
+        def _ignore_lines(content, ignore_lines):
+            idxs = []
```
JoySnow, line 776: From the usage of the `idxs`, what about using it as `idxs = set()`, to help speed up a little bit in the item adding and `in` checking below?

insights/core/__init__.py
xiangce: Nice catching.

insights/core/__init__.py
JoySnow: For both "item adding" and "x in s" operations, it takes O(1) time complexity (on average case) for `set()` type , and takes O(n) for `list()` type.  

As we do "item adding" inside  `_ignore_lines()`, and do the `in` checking at `for i, l in enumerate(content) if i not in idxs]`, both can benefit from the `set()` type than `list()`, just, benefit a little bit as we don't likely have a long `idxs` typically.

xiangce: Hm... I cannot agree with you about this. As `set` stores items with a hash algorithm, no matter long or short, it cost more than a `list` when storing one item. The only good point I can think for a `set` is when doing a `search` which is not used in this case.

xiangce: No worries, I removed the pre-searching of the `indexes`, change to skip the lines directly which will save one pass of `content` iteration.

JoySnow (approved):

PR #4436, opened 2025-04-22
fix: support filters that start with minus sign

jholecek-rh:
The issue was the root cause of RHINENG-15287.

Commit d1065f5ec25ece259acd38c9cef78ac0721da863 resolved the issue using reversed sorting, making a silent assumption that each component would have at least one filter message that starts with a character with lower precedence than `-`.

This commit uses the `grep` command line option `--` which causes all following arguments to be interpreted as operands (positional arguments) even if they start with `-`. See `info grep`. This works even if a component has a single filter message and the filter message starts with `-`.

### All Pull Requests:

Check all that apply:

Call `grep` with `--` command line options delimiter.

## Summary by Sourcery

Fix grep filtering for filter messages that start with a minus sign by using the `--` delimiter

Bug Fixes:
- Resolve an issue with grep filtering when filter messages start with a minus sign, which could previously cause incorrect filtering or empty content

Tests:
- Add a test case to verify filtering works correctly for filter messages starting with a minus sign

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -275,7 +275,7 @@ class TextFileProvider(FileProvider):
             # Pre-filtering ONLY when collecting data
             log.debug("Pre-filtering %s", self.relative_path)
             args.append(
-                ["grep", "-F", "\n".join(sorted(self._filters.keys(), reverse=True)), self.path]
+                ["grep", "-F", "--", "\n".join(self._filters.keys()), self.path]
             )
 
         return args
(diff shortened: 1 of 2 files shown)


xiangce (approved):

xiangce:
@jholecek-rh This `--` option to the grep command is quite new to me. I'm wondering why such an important option is no illustrated in `man grep`, but we have to dig deep into the `info grep` which is not so handy 😅. Thanks a lot.

jholecek-rh:
The `--` option is a standard [GNU `getopt` feature ](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html#Command_002dLine-Interfaces). :shrug: I know for a fact that the default `grep` on Mac (BSD grep, afaik) supports it as well.

PR #4444, opened 2025-04-28
feat: new LSlHFiles parser for ls -lH <files> spec

xiangce:
- move the FilePermissions from insights.util to the insights.ls_parser
- RHINENG-16907

### All Pull Requests:

Check all that apply:

*Add your description here*


insights/parsers/ls.py
```diff
@@ -399,3 +403,37 @@ class LSlaZ(FileListing):
...
+        True
+        >>> ls_files["/etc/redhat-release"].all_zero()
+        False
+    """
+
+    def parse_content(self, content):
+        for line in content:
+            try:
+                line = line.strip()
+                self[line.rsplit(None, 1)[-1]] = FilePermissions(line)
```
JoySnow, line 435: Looks like we are mean to parse for `FilePermissions` info of the given files here, and unlike the other ls parsers to inherit from `FileListing` directly. Will it make more sense to name this parser with `permission` related wording, say `LSlnHFilesPermissions` ?

xiangce: Good question. However, I don't think so, the reason didn't inherit it from the `FileListing` is that, 1. `FileListing` only works for listing files under specified directories; 2. In most cases, listing a specific file, it's not the only purpose to check its permissions, 'size', 'owner`, `date` would be also candidate checkpoints. I prefer to leave it as `LSlnHFiles`.

JoySnow: I see, so this parser name `LSlHFiles` ends with `Files` which is diff from other dir listing ones. Then I am good with it.

JoySnow (approved):

PR #4474, opened 2025-06-05
fix: FileProvider.validate() leaked files outside root

jholecek-rh:
### All Pull Requests:

Check all that apply:

Fixes two issues and adds two improvements:

* FileProvider would read files outside the root if the root were a prefix of the file path string-wise. Consider a root path `/tmp/sample/root` and a relative path `../root.txt`, resulting in reading a file `/tmp/sample/root.txt`.

* The "inside root" check was performed after checking if file existed. This would expose information about files outside the root through the exception. While an exploit would be difficult (if possible at all), it is better to fix this weakness anyway.

* The exception type is changed from the most generic `Exception` to `ValueError` to make it easier to distinguish it from `ContentException` that is raised in other cases.

* The `"/"` literal in path operations is changed to `os.sep` to improve portability.

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -215,7 +215,22 @@ class FileProvider(ContentProvider):
 
         self.validate()
 
+    def _is_inside_root(self):
+        """Checks that `self.relative_path` does not point outside `self.root`."""
+        resolved = os.path.realpath(self.path)
+
+        # pathlib.Path.is_relative_to() has been added only in Python 3.9
+        resolved_root = os.path.realpath(self.root)
+        if not resolved_root.endswith(os.sep):
+            resolved_root += os.sep
+
+        return resolved.startswith(resolved_root)
+
     def validate(self):
+        if not self._is_inside_root():
+            msg = "Relative path points outside the root: %s"
+            raise ValueError(msg % (self.path))
+
         # 1. No Such File
         if not os.path.exists(self.path):
             raise ContentException("%s does not exist." % self.path)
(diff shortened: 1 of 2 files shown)


insights/core/spec_factory.py
```diff
@@ -214,7 +214,22 @@ def __init__(self, relative_path, root="/", save_as=None, ds=None, ctx=None, cle
 
         self.validate()
 
+    def _is_inside_root(self):
```
xiangce, line 217: As this method is not used anywhere out the validate, better make it an `inner funciton` of the `validate`, Otherwise, we may need to add particular tests for this method (although it's a private one)

jholecek-rh: P.S. The `FileProvider.validate()` method is not used only in `FileProvider.__init__`. Should it be defined as a nested function in the `__init__` method?

xiangce: Sorry, I didn't get your point

1. Although the `validate()` is called only in the `__ini__`, but for different child classes, it has different implementation. It should not be designed as inner function.
2. `inner function`, only works for function which is very simple (just like this one) and sometimes a bit common (can be called times) for its parent function.


If its function is common enough, and is useful not only for its parent caller. We can upgrade it as you did or might need to consider moving it to the `insights.util` as a utility function.

Because I didn't get from the code and feel hard to imagine the possibility that it will be used by other cases in this class, I asked you move it to a `inner function` or just expand it in the caller.

xiangce (approved):
Per our F2F discussion, this PR is ready to merge. @jholecek-rh - please rebase it firstly and open another PR to the `3.0_egg` branch for the same change. Thanks.

PR #4482, opened 2025-06-19
feat: add Disable_Remote_Resource_Access variable for RemoteResource …

chenlizhong:
…class

### All Pull Requests:

Check all that apply:

* [ ] No Sensitive Data in this change?

 Add Disable_Remote_Resource_Access variable  for clase RemoteResource
Disable_Remote_Resource_Access is False, it can call the remote resource.
Disable_Remote_Resource_Access is true, it will raise exception and does not access remote resource.

*Add your description here*

### DIFF
--- a/insights/tests/test_remote_resource.py
+++ b/insights/tests/test_remote_resource.py
@@ -1,6 +1,7 @@
 from cachecontrol.cache import DictCache
 from insights.core.remote_resource import RemoteResource, CachedRemoteResource
 from insights.tests.mock_web_server import TestMockServer
+from insights import apply_configs
 
 GOOD_PAYLOAD = b'Successful return from Mock Service'
 NOT_FOUND = b'{"error":{"code": "404", "message": "Not Found"}}'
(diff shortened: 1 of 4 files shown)


xiangce (changes requested):
As mentioned, we should make this setting configurable on the insights-engine side. IMHO, there are multiple ways for use to achieve this. However, adding one new option into the `plugins/service` section in the "manifest" would be a much easier one. e.g 
- https://github.com/RedHatInsights/insights-core/blob/4ca6a02894cb6c2d794c47f44eda572da4c06097/insights/specs/manifests.py#L56

On engine side it's [here](https://gitlab.cee.redhat.com/insights-platform/insights-engine/-/blame/master/deploy/clowdapp.yaml)


Besides that, more corresponding updates are required to let the RemoteResource to read this option.

insights/tests/test_remote_resource.py
```diff
@@ -24,10 +24,20 @@ def test_get_remote_resource_not_found(self):
         rtn = rr.get(url)
         assert rtn.content == NOT_FOUND
 
+    # Test disable RemoteResource
+    def test_get_disable_remote_resource(self):
+        RemoteResource.Disable_Remote_Resource_Access = True
```
xiangce, line 29: After implemented the configurable option, the test of this flag can be implemented via reading a testing `manifest`

xiangce:
> adding one new option into the plugins/service section in the "manifest" would be a much easier one.

Just thought of using the `".insights.yaml"` (kind of [yamls](https://github.com/RedHatInsights/insights-core/blob/4ca6a02894cb6c2d794c47f44eda572da4c06097/insights/settings.py#L19-L21)) supported by the `insights.settings` would be easier than the `plugins/service section in the "manifest"`, it requires nothing update in the online insights-engine side, but just adding one `.insights.yaml` to the IoP engine image

```
# /etc/insights.yaml
# ~/.localinsights.yaml
# .insights.yaml
defaults:
  remote_resource_access: false
```

PR #4483, opened 2025-06-20
feat: Collect "corosync-cmapctl" on RHEL10

huali027:
### All Pull Requests:

Check all that apply:

### DIFF
--- a/insights/specs/datasources/corosync.py
+++ b/insights/specs/datasources/corosync.py
@@ -3,13 +3,14 @@ Custom datasources related to ``corosync``
 """
 import os
 
-from insights.components.rhel_version import IsRhel7, IsRhel8, IsRhel9
+from insights.components.rhel_version import IsRhel7, IsRhel6
+from insights.combiners.redhat_release import RedHatRelease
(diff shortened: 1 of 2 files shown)


insights/specs/datasources/corosync.py
```diff
@@ -3,13 +3,14 @@
...
-from insights.components.rhel_version import IsRhel7, IsRhel8, IsRhel9
+from insights.components.rhel_version import IsRhel7, IsRhel6
+from insights.combiners.redhat_release import RedHatRelease
 from insights.core.context import HostContext
 from insights.core.exceptions import SkipComponent
 from insights.core.plugins import datasource
 
 
-@datasource(HostContext, [IsRhel7, IsRhel8, IsRhel9])
+@datasource(HostContext, [IsRhel6, IsRhel7, RedHatRelease])
```
xiangce, line 13: As both `IsRhel*` and `RedHatRelease` are used here, and the `IsRhel*` is also depending on `RedHatRelease`, how about just use `RedHatRelease` only?

PR #4499, opened 2025-07-17
draft: : Add 3rd party yara rules integration

mtclinton:
### All Pull Requests:

Check all that apply:

* [ ] No Sensitive Data in this change?

*Add your description here*


insights/specs/datasources/malware_detection/config.py
```diff
@@ -0,0 +1,246 @@
...
+        except RuntimeError as e:
+            logger.error(str(e))
+            exit(constants.sig_kill_bad)
+        except Exception as e:
+            logger.error(
+                "Error getting the version of the specified yara binary %s: %s",
+                yara,
+                str(e),
+            )
+            exit(constants.sig_kill_bad)
```
abuckta, line 231: Is there any chance we can make the caller handle the failure here? 

I.e.
  ```
def find_yara():
      ... some code some code... 

      logger.error("No compatible YARA installation found")
      return None, None


  yara_binary, yara_version = find_yara()
  if not yara_binary:
      logger.error("Malware detection requires YARA >= %s", MIN_YARA_VERSION)
      return {"error": "yara_not_found", "required_version": MIN_YARA_VERSION}

```

abuckta (commented):
Looks pretty good Max! I added a few comments. I also just want to start a discussion on the following: 
1. should we add metrics for rule download success and failure? 
3. should we validate the Yara rule syntax before using them?

insights/specs/datasources/malware_detection/config.py
mtclinton: yes, it could easily be altered to that. Does it help though. Purpose is to test if yara exists on server and if not exit immediatley

abuckta: Yes, I think it is better. In general, having exits in the caller allows for easier debugging and testing. It also makes the code more readable.

PR #4500, opened 2025-07-21
Replace pkg_resources with importlib.metadata

hroncok:
Also, don't put the name of the project to core_version variable first, it's rather confusing.

Fixes https://issues.redhat.com/browse/CCT-1528

This requires Python 3.8+.
If support of 2.7 and similarly old Python 3 versions is desired, fallback imports of importlib_metadata and a conditional dependency on https://pypi.org/project/importlib-metadata/ is needed. (I can do that, just let me know.)

### All Pull Requests:

Check all that apply:

This is untested and provided as proof of concept.

### DIFF
--- a/insights/client/connection.py
+++ b/insights/client/connection.py
@@ -233,13 +233,14 @@ class InsightsConnection(object):
         """
         Generates and returns a string suitable for use as a request user-agent
         """
-        import pkg_resources
-        core_version = "insights-core"
-        pkg = pkg_resources.working_set.find(pkg_resources.Requirement.parse(core_version))
-        if pkg is not None:
-            core_version = "%s %s" % (pkg.project_name, pkg.version)
-        else:
+        import importlib.metadata
+
+        try:
+            pkg = importlib.metadata.distribution("insights-core")
+        except ModuleNotFoundError:
             core_version = "Core %s" % package_info["VERSION"]
+        else:
+            core_version = "%s %s" % (pkg.name, pkg.version)
 
         try:
             from insights_client import constants as insights_client_constants
@@ -253,10 +254,12 @@ class InsightsConnection(object):
         else:
             parent_process = "unknown"
 
-        requests_version = None
-        pkg = pkg_resources.working_set.find(pkg_resources.Requirement.parse("requests"))
-        if pkg is not None:
-            requests_version = "%s %s" % (pkg.project_name, pkg.version)
+        try:
+            pkg = importlib.metadata.distribution("requests")
+        except ModuleNotFoundError:
+            requests_version = None
+        else:
+            requests_version = "%s %s" % (pkg.name, pkg.version)
 
         python_version = "%s %s" % (platform.python_implementation(), platform.python_version())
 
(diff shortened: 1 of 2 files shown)


insights/client/connection.py
```diff
@@ -233,13 +233,14 @@ def user_agent(self):
         """
         Generates and returns a string suitable for use as a request user-agent
         """
-        import pkg_resources
-        core_version = "insights-core"
-        pkg = pkg_resources.working_set.find(pkg_resources.Requirement.parse(core_version))
-        if pkg is not None:
-            core_version = "%s %s" % (pkg.project_name, pkg.version)
-        else:
+        import importlib.metadata
```
m-horky, line 236: Hi. This branch still has to support Python 2.7, `importlib.metadata` have been [introduced in 3.8](https://docs.python.org/3/library/importlib.metadata.html#:~:text=Added%20in%20version), this won't work :/

pkoprda: Hi, insights-core has only one branch which is targeting RHEL7/8/9/10, so your code needs to be in Python2.7

hroncok: Understood. Amedned to fallback to pkg_resources, as importlib_metadta might not be available in RHEL 7.

Happy to change the try-import to if sys.version_info check.

insights/client/connection.py
```diff
@@ -77,6 +77,25 @@ def _api_request_failed(exception, message='The Insights API could not be reache
...
+    else:
+        try:
+            pkg = importlib.metadata.distribution(name)
+            return pkg.name, pkg.version
+        except ModuleNotFoundError:
+            pass
+    return None, None
+
+
+
```
pkoprda, line 98: Probably too many empty lines, this would not be accepted by `flake8`/`black` pipeline. 2 empty lines is enough

pkoprda (changes requested):
Looks good, left one comment.

Also please squash the commits into one and update the PR description.

hroncok:
Squashed, reworded the commit message to reflect the current state, and removed the superfluous empty line.

hroncok:
In [python27-test](https://github.com/RedHatInsights/insights-core/actions/runs/16550825609/job/46839373286?pr=4500#logs) I see:

```
./insights/client/connection.py:93:16: F821 undefined name 'ModuleNotFoundError'
```

This is on Python 2, but that code will never be reached on Python 2. Shall I replace that with ImportError instead? That would arguably make the code less readable for the sake of pleasing a linter. Or alternatively, add some noqa comment?

---

In [code-test (3.9)](https://github.com/RedHatInsights/insights-core/actions/runs/16550825609/job/46839373266?pr=4500#logs) I see:

```
ERROR:insights.specs.datasources.compliance.compliance_ds:Unexpected exception in compliance: 'PathDistribution' object has no attribute 'name'
```

It seems that importlib.metadata on 3.9 is not good enough for this. Shall I replace the try-import with `if sys.version_info >= (3, 10)`?

pkoprda (approved):
Looks good to me :+1:

Please, rebase and squash the commits. Then @zpetrace will do QE verification

hroncok:
Squashed without changes.

pkoprda:
@xiangce can you trigger the CI jobs? Then you can merge this if the tests will not fail

PR #4763, opened 2026-03-30
[CCXDEV-15098] Fix memory leak - cleanup broker exception traceback

matysek:
### All Pull Requests:

Check all that apply:

This PR fixes a memory leak in `insights/core/dr.py` that affects long-running services using insights-core. When components raise exceptions during `run_components()`, Python attaches a live traceback object to the exception. This traceback holds a reference back to the stack frame's local variables — including the `broker` — creating a circular reference chain that prevents garbage collection via reference counting. Over time, this causes steady memory growth (observed in production).

The fix clears `exception.__traceback__` immediately after the formatted traceback string has been captured via `traceback.format_exc()`. This breaks the circular reference without losing any debugging information, since all existing code in the codebase accesses tracebacks through `broker.tracebacks` (the formatted string), not `exception.__traceback__`.

In `run_components()`, caught exceptions are stored in `broker.exceptions` along with their formatted traceback string via `broker.add_exception(component, ex, tb)`. However, the exception's `__traceback__` attribute still holds a reference to the live stack frame, which references all local variables — including `broker`. This creates a circular reference chain:

```
broker.exceptions -> exception -> __traceback__ -> frame -> local vars (broker) -> broker
```

CPython's reference-counting collector cannot break cycles. The cyclic GC may eventually collect them, but in long-running services this is unreliable and the memory grows steadily before collection occurs.

## Summary by Sourcery

Prevent memory leaks in long-running services by clearing stored exception tracebacks after component execution and documenting the change.

Bug Fixes:
- Break circular references by clearing exception __traceback__ after capturing formatted tracebacks in run_components, covering generic, skip, and blacklisted spec exceptions.

Documentation:
- Note the traceback cleanup behavior and memory leak fix in the changelog.

Tests:
- Add regression tests ensuring exception __traceback__ is cleared, tracebacks remain available as strings, and brokers are garbage-collected without cyclic GC.

### DIFF
--- a/insights/core/dr.py
+++ b/insights/core/dr.py
@@ -1093,6 +1093,9 @@ def run_components(ordered_components, components, broker):
             for x in get_registry_points(component):
                 BLACKLISTED_SPECS.append(str(x).split('.')[-1])
             broker.add_exception(component, bs, traceback.format_exc())
+            # CCXDEV-15098: Clear traceback to break circular reference.
+            # See the comment in the ``except Exception`` block below.
+            bs.__traceback__ = None
         except MissingRequirements as mr:
             if log.isEnabledFor(logging.DEBUG):
                 name = get_name(component)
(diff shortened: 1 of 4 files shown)


joselsegura (approved):

insights/core/dr.py
```diff
@@ -1103,14 +1106,38 @@ def run_components(ordered_components, components, broker):
...
+                # CCXDEV-15098: Clear traceback to break circular reference.
+                # See the comment in the ``except Exception`` block below.
+                sc.__traceback__ = None
         except Exception as ex:
             log.debug(ex)
             tb = traceback.format_exc()
             broker.add_exception(component, ex, tb)
             for reg_spec in get_registry_points(component):
                 broker.add_exception(reg_spec, ex, tb)
+            # CCXDEV-15098: Clear the traceback reference to break a circular
```
jholecek-rh, line 1118: Nit: With the exception of to-do items, I do not think that internal Red Hat trackers belong to the code of an open-source project. All information that is relevant for long-term maintenance should be available in the code directy. Additional information (including any external trackers) can be included in the commit message. It is already the case, so the tracker ID in the code is useless.

insights/core/dr.py
```diff
@@ -1103,14 +1106,38 @@ def run_components(ordered_components, components, broker):
...
+            # reference chain that prevents the broker from being GC'd.
+            #
+            # When Python catches an exception, it attaches the live
+            # traceback object to ``ex.__traceback__``.  That traceback
+            # holds a reference to the stack frame, which in turn
+            # references all local variables — including ``broker``.
```
jholecek-rh, line 1124: How does the proposed change affect integration with error tracking systems like GlitchTip? IIUC, GlitchTip works best if it has access to the live traceback object. I suspect that denying GlitchTip the option to create an event from the live traceback will reduce the actionability of GlitchTip events. As someone who is responsible for monitoring rule performance and quality, I do not like this prospect.

Separately, I do not like that this cleanup is done independently in every `except` clause of of the `try`-`except`-`finally` statement. This is error-prone - developers have to remember to add/keep it whenever the except clauses change. A better way would be to do it in the `add_exception` method, but this solution would not address the first issue.

Yet a better solution would be to do the cleanup for all exceptions raised by a component in the `finally` clause, after running all the observers. This would do the cleanup at a single place in the code while allowing applications to access the live traceback through observers. (It would require changes to the applications. I would like to investigate how difficult this would be in our use case before we commit to this option.)

PR #4765, opened 2026-04-01
feat: new datasource spec azure_instance_compute_metadata

JoySnow:
### All Pull Requests:

Check all that apply:

Jira: RHINENG-25366

## Summary by Sourcery

Add a new Azure instance compute metadata datasource and parser, wiring it into the specs system with Azure-specific filtering and coverage tests.

New Features:
- Introduce azure_instance_compute_metadata datasource that retrieves and filters Azure Instance Metadata Service compute fields.
- Add AzureInstanceComputeMetadata parser based on JSONParser to expose filtered compute metadata to clients.

Enhancements:
- Register the new azure_instance_compute_metadata spec in the default specs and manifests, enabling it only when running on Azure.
- Extend Specs registry to support a filterable, non-obfuscated azure_instance_compute_metadata entry with relevant network identifiers.

Tests:
- Add parser tests for AzureInstanceComputeMetadata, including doc examples integration.
- Add datasource tests for azure_instance_compute_metadata covering normal, no-filter, and abnormal output scenarios.

### DIFF
--- a/insights/parsers/azure_instance.py
+++ b/insights/parsers/azure_instance.py
@@ -13,8 +13,12 @@ AzureInstanceType - 'vmSize' of Azure Instance
 AzureInstancePlan - 'plan' of Azure Instance
 --------------------------------------------
 
+AzureInstanceComputeMetadata - 'compute' metadata of Azure Instance
+--------------------------------------------------------------------
+
 AzurePublicIpv4Addresses - list of public IPv4 addresses
 --------------------------------------------------------
+
 """
 
 import json
@@ -22,9 +26,18 @@ import json
 from uuid import UUID
 
 from insights.core import CommandParser
+from insights.core import JSONParser
 from insights.core.exceptions import ParseException, SkipComponent
+from insights.core.filters import add_filter
 from insights.core.plugins import parser
 from insights.specs import Specs
+from insights.util import deprecated
+
+# TODO: Need to migrate the `add_filters` to filter-requester.
+# Note. Please pay attention to the sensitive data when adding new fields to the filter.
+add_filter(
+    Specs.azure_instance_compute_metadata, ['licenseType', 'vmSize', 'vmId', 'plan', 'offer']
+)
 
 
 def validate_content(content):
@@ -54,6 +67,14 @@ class AzureInstanceID(CommandParser):
         >>> azure_id.id
         'f904ece8-c6c1-4b5c-881f-309b50f25e50'
     """
+
+    def __init__(self, *args, **kwargs):
+        deprecated(
+            AzureInstanceID,
+            "Please use the :py:class:`AzureInstanceComputeMetadata` instead.",
+            "3.9.0",
+        )
+        super(AzureInstanceID, self).__init__(*args, **kwargs)
 
     def parse_content(self, content):
         validate_content(content)
@@ -107,6 +128,14 @@ class AzureInstanceType(CommandParser):
         >>> azure_type.raw
         'Standard_L64s_v2'
     """
+
+    def __init__(self, *args, **kwargs):
+        deprecated(
+            AzureInstanceType,
+            "Please use the :py:class:`AzureInstanceComputeMetadata` instead.",
+            "3.9.0",
+        )
+        super(AzureInstanceType, self).__init__(*args, **kwargs)
 
     def parse_content(self, content):
         validate_content(content)
@@ -164,6 +193,14 @@ class AzureInstancePlan(CommandParser):
         True
     """
 
+    def __init__(self, *args, **kwargs):
+        deprecated(
+            AzureInstancePlan,
+            "Please use the :py:class:`AzureInstanceComputeMetadata` instead.",
+            "3.9.0",
+        )
+        super(AzureInstancePlan, self).__init__(*args, **kwargs)
+
     def parse_content(self, content):
         validate_content(content)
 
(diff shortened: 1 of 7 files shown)


insights/parsers/azure_instance.py
```diff
@@ -13,18 +13,30 @@
...
 
 from insights.core import CommandParser
+from insights.core import JSONParser
 from insights.core.exceptions import ParseException, SkipComponent
+from insights.core.filters import add_filter
 from insights.core.plugins import parser
 from insights.specs import Specs
+from insights.util import deprecated
+
+# Note. Please pay attention to the sensitive data when adding new fields to the filter.
```
xiangce, line 37: As we have agreed to move these `adding filters` to the actual filter requesters in the near future, would you mind adding a `TODO` reminder here, e.g.
```
TODO: Need to migrate the `add_filters` to filter-requester.
```

insights/parsers/azure_instance.py
```diff
@@ -183,6 +219,57 @@ def __repr__(self):
...
+
+    The datasource collects filtered fields from the Azure Instance Metadata Service endpoint:
+    ``http://169.254.169.254/metadata/instance/compute?api-version=2021-12-13&format=json``
+
+    This parser extends JSONParser and provides dictionary-style access to the filtered metadata fields:
+    - ``licenseType`` - Azure Hybrid Benefit license type (e.g., "RHEL_BYOS", "Windows_Server", or empty)
+    - ``offer`` - Marketplace offer identifier (e.g., "RHEL", "UbuntuServer")
+    - ``vmSize`` - VM size (e.g., "Standard_B1s", "Standard_L64s_v2")
+    - ``vmId`` - VM instance UUID
+    - ``plan`` - Publisher plan information (dict with name, product, publisher)
```
xiangce: Since the `filters` is dynamic, it's not a good idea to hardcode these keys here. Better make the description a bit open, like `more fields would be included according to the custom filters`

insights/parsers/azure_instance.py
```diff
@@ -183,6 +219,57 @@ def __repr__(self):
...
+
+    Examples:
+        >>> azure_instance_compute_metadata['licenseType']
+        ''
+        >>> azure_instance_compute_metadata['offer']
+        'RHEL'
+        >>> azure_instance_compute_metadata['vmSize']
+        'Standard_B1s'
+        >>> azure_instance_compute_metadata['vmId']
+        '3c29e210-0669-496f-812a-279123beb99f'
```
xiangce: I'm sure this Id is from your test env, but it would be better to obfuscate some parts of it explicitly.

insights/specs/datasources/azure.py
```diff
@@ -0,0 +1,72 @@
...
+                        ctx=broker.get(HostContext),
+                        cleaner=broker.get("cleaner"),
+                    )
+                raise SkipComponent("No matching filters found in metadata")
+            raise SkipComponent("Invalid JSON format")
+        raise SkipComponent("No filters defined")
+    except SkipComponent:
+        raise
+    except Exception as e:
+        raise SkipComponent("Failed to parse metadata: {0}".format(e))
```
xiangce, line 59: Here the process of `Skip` is a bit complex. How about refactoring as 
```python
filters = get_filters(Specs.azure_instance_compute_metadata)
if not filters:
    raise SkipComponent("No filters defined")

raw_content = broker[LocalSpecs.azure_instance_compute_metadata_raw].content
if not raw_content or 'curl: ' in raw_content[0]:
    raise SkipComponent("Empty content or curl error")

try:
    content = json.loads(raw_content[0])
    result = {key: content[key] for key in filters if key in content}
except Exception as ex:
    raise SkipComponent(ex)

if result:
    return DatasourceProvider(
        content=json.dumps(result, sort_keys=True),
        relative_path='insights_datasources/azure_instance_compute_metadata',
        ds=Specs.azure_instance_compute_metadata,
        ctx=broker.get(HostContext),
        cleaner=broker.get("cleaner"),
    )  
raise SkipComponent("No matching filters found in metadata")
```

insights/specs/manifests.py
```diff
@@ -123,6 +123,10 @@
     - name: insights.components.ceph.IsCephMonitor
       enabled: true
 
+    # needed for azure specs
+    - name: insights.components.cloud_provider.IsAzure
+      enabled: true
+
```
xiangce: This sub-item `IsAzure` is already covered by the above `insights.components.cloud_provider`.

insights/parsers/azure_instance.py
```diff
@@ -183,6 +219,57 @@ def __repr__(self):
...
+            "plan": {"name": "", "product": "", "publisher": ""},
+            "vmId": "3c29e210-0669-496f-812a-279123beb99f",
+            "vmSize": "Standard_B1s"
+        }
+
+    Raises:
+        SkipComponent: When content is empty, curl error occurs, or no filters defined.
+
+    Attributes:
+        data (dict): The parsed JSON data containing the filtered metadata fields
```
xiangce: Just a suggestion, since `data` is not a recommended attribute name to parsers, and in the examples, it's not used either, can we do not write down this section here?

insights/parsers/azure_instance.py
JoySnow: updated

insights/specs/datasources/azure.py
JoySnow: updated

insights/parsers/azure_instance.py
```diff
@@ -183,6 +220,48 @@ def __repr__(self):
...
+        {
+            "fieldKey1": "Field value 1",
+            "fieldKey2": "Field value 2"
+        }
+
+    Raises:
+        SkipComponent: When content is empty, curl error occurs, or no filters defined.
+
+    Attributes:
+        ... : The value of the ... field
```
xiangce: Hm… I don't think so. For the JSONParser, other than the `data` which is the one we don't recommend to use, no other attributes are exposed. It depends on whether it's a `list [...]` or `dict {...}` JSON. In this case, it's a dict JSON, where also no keys are set as attributes.

insights/specs/datasources/azure.py
```diff
@@ -0,0 +1,75 @@
...
+        return DatasourceProvider(
+            content=json.dumps(result, sort_keys=True),
+            relative_path='insights_datasources/azure_instance_compute_metadata',
+            ds=Specs.azure_instance_compute_metadata,
+            ctx=broker.get(HostContext),
+            cleaner=broker.get("cleaner"),
+        )
+
+    except SkipComponent:
+        raise
```
xiangce: I'm wondering why set this `try...catch` scope so wide? Other than the calling of `json.loads()`, all other operations are taken care enough by the safety if checks.

If you insist on extending the `try...catch` to the whole body, how about just keep the `except Exception as e` only? As for both existing `except`, they both raise SkipComponent.

insights/parsers/azure_instance.py
JoySnow: yep, i shall remove this `Attributes` here, and had mentioned about `dictionary-style access` in the above docstrings

insights/specs/datasources/azure.py
JoySnow: see your point here, updated to wrap for `json.loads()` only

xiangce (approved):

xiangce:
test dp

koalakangaroo:
test egg

qinpingli:
Create a new PR per this: https://github.com/RedHatInsights/insights-core/pull/4772.

So, I'll close this one.

PR #4771, opened 2026-05-25
feat(Compliance): fail scans when SSG version mismatches

RoamingNoMaD:
### All Pull Requests:

Check all that apply:

**Fail compliance scan on SSG version mismatch**

Before running an OpenSCAP scan, we now fetch the tailoring metadata for each assigned policy from `/compliance/v2/policies/{id}/tailorings/{os_minor}` and read its `security_guide_version` field. That version is compared against the locally installed `scap-security-guide` RPM. If they differ, the scan is aborted with a clear error message telling the user which version to install.

The metadata fetch retries up to 3 times (with 10s delays) on transient errors — consistent with how tailoring file downloads already behave. A non-200 response after all retries, or a response missing the version field, also aborts the scan.

## Summary by Sourcery

Enforce compatibility between Compliance backend tailoring metadata and the locally installed scap-security-guide before running scans.

New Features:
- Add retrieval of tailoring metadata security_guide_version from the Compliance service for each policy.
- Abort compliance scans when the backend SSG version differs from the installed scap-security-guide RPM version.

Enhancements:
- Implement retry logic and strict validation for fetching tailoring metadata, treating missing or invalid SSG versions as fatal errors.

Tests:
- Add unit tests covering tailoring SSG version retrieval, retry/error handling, and scan behavior when SSG versions match or mismatch, including obfuscation scenarios.


mtclinton:
the souceryai comment on exception handling is true, but not a big deal and has been in code for a while.
looks good, thanks for creating pr :)

RoamingNoMaD:
We need to solve this in a more customer facing way, I'll leave this in draft for the time being

romanblanco (changes requested):
> We need to solve this in a more customer facing way, I'll leave this in draft for the time being

Confirming what @RoamingNoMaD wrote.

Since we have an official support matrix for SSG/RHEL versions, I don't think it's a good idea to force customer to upgrade his packages in order to be able to run OSCAP scan.

We should _somehow_ resolve the SSG version inconsistency on our side.

RoamingNoMaD:
This is not the way.

PR #4778, opened 2026-06-18
feat: add spec and parser for mdadm_detail_platform (RHINENG-27037)

sabhasin:
### All Pull Requests:

Check all that apply:

* [x ] Have you followed the guidelines in our Contributing document, including the instructions about commit messages?
* [x ] No Sensitive Data in this change?
* [x ] Is this PR an enhancement?

This PR adds support for collecting and parsing the output of:

mdadm --detail-platform

The command is required for RHINENG-27037 to identify systems that use Intel(R) Rapid Storage Technology.

Changes included:

    - Added a new spec mdadm_detail_platform
    - Added command collection for mdadm --detail-platform
    - Added MDAdmDetailPlatform parser
    - Added tests for the new parser

The parser stores the command output as key/value pairs for use by customers.
If the command returns no output, the parser skips processing by raising SkipComponent.

## Summary by Sourcery

Add collection and parsing support for the `mdadm --detail-platform` command output.

New Features:
- Introduce the `mdadm_detail_platform` spec and command collection for `mdadm --detail-platform`.
- Add an `MDAdmDetailPlatform` parser that exposes the detail-platform output as key/value data.

Tests:
- Add unit tests covering normal parsing and empty-output handling for the new `MDAdmDetailPlatform` parser.

### DIFF
--- a/insights/parsers/mdadm.py
+++ b/insights/parsers/mdadm.py
@@ -64,6 +64,7 @@ class MDAdmMetadata(CommandParser, dict):
         >>> mdadm["Events"]
         60
     """
+
     def parse_content(self, content):
         mdadm_dev = "/mdadm_-E_.dev."
         if mdadm_dev in self.file_path:
@@ -88,6 +89,7 @@ class MDAdmDetailDevice(dict):
     The md device's properties in <property name> : <property value> format
     will be stored seprately, and are accessable via <property name>.
     """
+
     def parse_device(self, content, device_start_index, table_start_index, index):
 
         self['device_name'] = content[device_start_index].split(':')[0]
@@ -235,3 +237,29 @@ class MDAdmDetail(CommandParser, list):
         # Empty prased data
         if len(self) < 1:
             raise SkipComponent('Empty parsed device')
+
+
+@parser(Specs.mdadm_detail_platform)
+class MDAdmDetailPlatform(CommandParser, dict):
+    """
+    Parser for output of ``mdadm --detail-platform``.
+
+    Sample output::
+
+        Platform : Intel(R) Rapid Storage Technology
+        Version : 14.8.0.2377
+        RAID Levels : raid0 raid1 raid10 raid5
+        Chunk Sizes : 4k 8k 16k 32k 64k 128k
+        2TB volumes : supported
+        2TB disks : supported
+        Max Disks : 7
+        Max Volumes : 2 per array, 4 per controller
+        I/O Controller : /sys/devices/pci0000:00/0000:00:1f.2 (SATA)
+        Port4 : /dev/sdd (ZDEEX7A8)
+    """
+
+    def parse_content(self, content):
+        if len(content) == 0:
+            raise SkipComponent("Empty content of command output")
+        for key, val in split_kv_pairs(content, split_on=':').items():
+            self[key] = val
--- a/insights/parsers/mdadm.py
+++ b/insights/parsers/mdadm.py
@@ -256,10 +256,26 @@ class MDAdmDetailPlatform(CommandParser, dict):
         Max Volumes : 2 per array, 4 per controller
         I/O Controller : /sys/devices/pci0000:00/0000:00:1f.2 (SATA)
         Port4 : /dev/sdd (ZDEEX7A8)
+
+    Examples:
+        >>> mdadm_plat["Platform"]
+        'Intel(R) Rapid Storage Technology'
+        >>> mdadm_plat["Version"]
+        '14.8.0.2377'
+        >>> mdadm_plat["Max Disks"]
+        '7'
+
+    Raises:
+        SkipComponent: When content is empty or no platform data is found.
     """
 
     def parse_content(self, content):
         if len(content) == 0:
             raise SkipComponent("Empty content of command output")
+
         for key, val in split_kv_pairs(content, split_on=':').items():
             self[key] = val
+
+        # If no data was parsed, skip
+        if len(self) == 0:
+            raise SkipComponent("No platform data found")
--- a/insights/specs/__init__.py
+++ b/insights/specs/__init__.py
@@ -477,7 +477,7 @@ class Specs(SpecSet):
     md5chk_files = RegistryPoint(multi_output=True)
     mdadm_D = RegistryPoint()
     mdadm_E = RegistryPoint(multi_output=True)
-    mdadm_detail_platform = RegistryPoint()
+    mdadm_detail_platform = RegistryPoint(no_obfuscate=['hostname', 'ipv4', 'ipv6', 'mac'])
     mdatp_managed = RegistryPoint()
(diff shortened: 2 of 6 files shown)


insights/specs/__init__.py
```diff
@@ -477,6 +477,7 @@ class Specs(SpecSet):
     md5chk_files = RegistryPoint(multi_output=True)
     mdadm_D = RegistryPoint()
     mdadm_E = RegistryPoint(multi_output=True)
+    mdadm_detail_platform = RegistryPoint()
```
qinpingli: IIUC, when obfuscation is enabled, the obfuscation will be done for all specs. Since you there is no need obfuscation for:

- IPv4
- IPv6
- Hostname
- Mac

So, I suggest to update this registry point to: mdadm_detail_platform = RegistryPoint(no_obfuscate=['hostname', 'ipv4', 'ipv6', 'mac'])

insights/specs/sos_archive.py
```diff
@@ -168,6 +168,7 @@ class SosSpecs(Specs):
     manila_conf = first_file(["/var/lib/config-data/puppet-generated/manila/etc/manila/manila.conf", "/etc/manila/manila.conf"])
     mdadm_D = simple_file("sos_commands/md/mdadm_-D_.dev.md")
     mdadm_E = glob_file("sos_commands/md/mdadm_-E_*")
+    mdadm_detail_platform = simple_file("sos_commands/md/mdadm_--detail-platform")
```
qinpingli: @sabhasin could you share with me how sosreport collects this data?

insights/parsers/mdadm.py
```diff
@@ -235,3 +237,29 @@ def _handle_device(device_start_index, table_start_index, index):
...
+        Platform : Intel(R) Rapid Storage Technology
+        Version : 14.8.0.2377
+        RAID Levels : raid0 raid1 raid10 raid5
+        Chunk Sizes : 4k 8k 16k 32k 64k 128k
+        2TB volumes : supported
+        2TB disks : supported
+        Max Disks : 7
+        Max Volumes : 2 per array, 4 per controller
+        I/O Controller : /sys/devices/pci0000:00/0000:00:1f.2 (SATA)
+        Port4 : /dev/sdd (ZDEEX7A8)
```
qinpingli, line 258: Could  you add `Example` about how to use this parser?

insights/parsers/mdadm.py
```diff
@@ -235,3 +237,29 @@ def _handle_device(device_start_index, table_start_index, index):
...
+    """
+
+    def parse_content(self, content):
+        if len(content) == 0:
+            raise SkipComponent("Empty content of command output")
```
qinpingli, line 274: Hmm, since `keep_rc=True` is defined in the spec, should we consider the scenario when `/usr/sbin/mdadm --detail-platform` running failed. And there is some information is stderr?

insights/tests/parsers/test_mdadm.py
```diff
@@ -382,6 +382,31 @@ def test_mdadm_d_exceptions():
...
+    assert mdadm["Chunk Sizes"] == "4k 8k 16k 32k 64k 128k"
+    assert mdadm["Max Disks"] == "7"
+
+
+def test_mdadm_detail_platform_empty():
+    with pytest.raises(SkipComponent) as exc:
+        MDAdmDetailPlatform(context_wrap(""))
+    assert "Empty content of command output" in str(exc)
+
+
```
qinpingli, line 409: Could you add document test for the new parser?

insights/specs/sos_archive.py
sabhasin: It appears that sosreport has an md plugin that collects mdadm-related information.

insights/parsers/mdadm.py
sabhasin: Thanks for pointing it out. I have added the example.

insights/parsers/mdadm.py
sabhasin: Yes, I have now included the scenario where the command failed and returned error messages, as well as the scenario where no data was parsed.

insights/tests/parsers/test_mdadm.py
sabhasin: I have added the document test

insights/specs/__init__.py
sabhasin: Sure, I have added this and also noted it for the future.

insights/specs/sos_archive.py
qinpingli: Hmm, I didn't see the `mdadm --detail-platform` is collected in https://github.com/sosreport/sos/blob/main/sos/report/plugins/md.py. Could you point me the current link?

insights/parsers/mdadm.py
qinpingli: Thanks for adding the exception process. IMO, some exception process is introduced by `keep_rc=True` in the following definition:
```
mdadm_detail_platform = simple_command("/usr/sbin/mdadm --detail-platform", keep_rc=True)
```

Seems `keep_rc=True` is no helpful for the data collection, but added the complication of data processing. So how about updating the spec definition to:
```
mdadm_detail_platform = simple_command("/usr/sbin/mdadm --detail-platform")

File: insights/specs/sos_archive.py (hunk 2 above)
Comment (sabhasin) [reply insights/specs/sos_archive.py]: It seems that sosreport is not collecting this data, hence i have removed it from sos_archives.

File: insights/parsers/mdadm.py (hunk 4 above)
Comment (sabhasin) [reply insights/parsers/mdadm.py:274]: Thank you for the review. As suggested, we have removed the `keep_rc` for `mdadm_detail_platform`. The parser will handle command failures appropriately by raising `SkipComponent` exceptions.

File: insights/parsers/mdadm.py [hunk 6]
```diff
@@ -235,3 +237,51 @@ def _handle_device(device_start_index, table_start_index, index):
...
+
+    MDADM_ERROR_MSG_PREFIX = "mdadm: "
+
+    def parse_content(self, content):
+        if len(content) == 0:
+            raise SkipComponent("Empty content of command output")
+
+        # Check if the command failed and returned error messages
+        if content[0].strip().startswith(self.MDADM_ERROR_MSG_PREFIX):
+            raise SkipComponent("Command Failed: {0}".format(content[0].strip()))
```
qinpingli: Since we do not set `keep_rc=True`, IIUC, the error handling part is no needed?

sabhasin: Yes, the spec doesn't set keep_rc=True. When the command fails (returns a non-zero exit code), the insights framework will not even call the parser; the command output will be skipped entirely at the collection level. Hence, i have removed the logic from code as well as its test case.

qinpingli:
I'll close and reopen this PR to test if ``code/snyk` ci job can be triggered.

qinpingli (approved):

PR #4782, opened 2026-07-08
feat(container): Replace podman exec with host-side RPM queries for container RPM collection (RHINENG-27038)

sabhasin:
Check all that apply:

  * [x] Have you followed the guidelines in our Contributing document, including the
  instructions about commit messages?
  
  **RHINENG-27038: Replace podman exec with host-side RPM queries for container RPM collection.**
  
  #### Problem Statement

 The current `container_installed_rpms` spec uses `podman exec` to run `rpm -qa`  inside each container. This approach consumes container resources (~15MB memory spike per exec), which is problematic for containers with tight memory limits. For example, a container with a 100MB limit already using 50MB sees usage increase to  ~65MB when exec runs.

  #### Solution Overview

This PR replaces the `podman exec` approach with host-side RPM queries using `podman inspect` to extract the container's overlay filesystem `MergedDir` path, then running `rpm -qa --root <MergedDir>` on the host. This eliminates container resource usage while maintaining identical output format.

  #### Implementation Details

  **1. New Datasource: `container_merged_dirs`**
  - File: `insights/specs/datasources/container/merged_dirs.py`
  - Extracts `GraphDriver.Data.MergedDir` from container inspect output.
  - Returns tuples: `(image, engine, container_id, merged_dir)`
  - Validates paths exist and are readable.
  - Gracefully skips containers without overlay storage.
  - Includes comprehensive error handling and logging.

  **2. New Spec Factory: `container_foreach_execute`**
  - File: `insights/core/spec_factory.py` (new class).
  - Extends `foreach_execute` to preserve container metadata.
  - Substitutes `merged_dir` into command template.
  - Wraps output in `HostBasedContainerCommandProvider`
  - Sets `context.image`, `context.engine`, `context.container_id` for parsers.
  - Handles edge cases (missing rpm command, invalid paths).

  **3. Updated Spec: `container_installed_rpms`**
  - File: `insights/specs/default.py`
  - Changed from `container_execute` to `container_foreach_execute`.
  - Uses `container_merged_dirs` datasource as provider.
  - Command: `/usr/bin/rpm -qa --root %s` (runs on host, not in container).

  **4. Comprehensive Testing**
  - File: `insights/tests/datasources/container/test_merged_dirs.py`
  - 13 unit tests covering:
    - Valid overlay with MergedDir extraction.
    - Missing GraphDriver, Data, or MergedDir fields.
    - Malformed JSON and invalid data.
    - Path validation (non-existent, unreadable).
    - Multiple containers with mixed validity.
    - Tuple reordering for command formatting.

  #### Backward Compatibility

   **Parser Interface:** No changes to `ContainerInstalledRpms` parser required:
  - Metadata (`.image`, `.engine`, `.container_id`) preserved via `HostBasedContainerCommandProvider`
  - RPM output format identical (same `_rpm_format`, same `rpm -qa` command)

   **Dependent Rules:** All 23 rules continue working unchanged:
  - Container rules: `container_product_eol_check`.
  - MSSQL rules: 17 rules using `ContainerMssqlVersion` combiner.
  - Nginx rules: 5 rules using `ContainerInstalledNginx` combiner.

  #### Benefits

  1. **Eliminates Container Resource Consumption**
     - Before: ~15MB memory spike per container during collection.
     - After: 0MB (runs on host, not in container).

  2. **Scalability**
     - For 10 containers: 150MB total spike eliminated.
     - Especially beneficial for containers with tight memory limits.

  3. **Identical Output**
     - Same RPM list format.
     - Same parser compatibility.
     - No changes needed to downstream consumers.

  4. **Robust Edge Case Handling**
     - Gracefully skips non-overlay storage drivers
     - Validates paths before use
     - Logs issues for debugging
     - Raises SkipComponent when no valid containers

 #### Testing Performed

 - All new unit tests pass (13/13).
 - Parser integration tests pass.
 - Mock data processing tested.
 - Edge cases covered.

## Summary by Sourcery

Replace in-container RPM execution with host-side queries against container overlay filesystems to reduce resource usage while preserving existing RPM collection behavior.

New Features:
- Collect container RPM inventories by querying container overlay filesystems from the host instead of executing RPM inside each container.
- Preserve container metadata and archive output for host-based container commands.

Enhancements:
- Add resilient extraction of overlay MergedDir paths from container inspection data, skipping unsupported or incomplete containers.
- Retain existing RPM output and downstream parser compatibility while avoiding container resource consumption.

Documentation:
- Update the custom datasource documentation index.

Tests:
- Add coverage for host-based container command providers, command expansion, metadata preservation, and container inspection edge cases.

### DIFF
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -535,6 +535,61 @@ class ContainerCommandProvider(ContainerProvider):
         return 'ContainerCommandProvider("%r")' % self.cmd
 
 
+class HostBasedContainerCommandProvider(CommandOutputProvider):
+    """
+    Provider for commands executed on the host that produce container-related data.
+
+    Unlike ContainerCommandProvider which parses 'podman exec' commands, this class runs commands
+    directly on the host using container metadata (e.g., overlay paths). The image, engine, and container_id
+    are set explicitly rather than parsed from the command string.
+
+    Inherits from CommandOutputProvider directly to keep the hierarchy shallow and separate from in-container providers.
+    Uses root="insights_containers" to honor the existing archive convention for container-related data.
+    """
+
+    def __init__(
+        self,
+        cmd_path,
+        ctx,
+        image=None,
+        engine=None,
+        container_id=None,
+        args=None,
+        split=True,
+        keep_rc=False,
+        ds=None,
+        timeout=None,
+        inherit_env=None,
+        override_env=None,
+        signum=None,
+        cleaner=None,
+    ):
+        super(HostBasedContainerCommandProvider, self).__init__(
+            cmd_path,
+            ctx,
+            root="insights_containers",
+            save_as=None,
+            args=args,
+            split=split,
+            keep_rc=keep_rc,
+            ds=ds,
+            timeout=timeout,
+            inherit_env=inherit_env,
+            override_env=override_env,
+            signum=signum,
+            cleaner=cleaner,
+        )
+        self.image = image
+        self.engine = engine
+        self.container_id = container_id
+        self.relative_path = os.path.join(
+            container_id, "insights_commands", mangle_command(self.cmd)
+        )
+
+    def __repr__(self):
+        return 'HostBasedContainerCommandProvider("%r")' % self.cmd
+
+
 class RegistryPoint(object):
     # Marker class for declaring that an element of a `SpecSet` subclass
     # is a registry point against which further subclasses can register
@@ -1409,6 +1464,120 @@ class container_execute(foreach_execute):
         raise ContentException("No results found for [%s]" % self.cmd)
 
 
+class container_foreach_execute(foreach_execute):
+    """
+    Execute a command on the host for each container using data from a provider datasource.
+
+    This is similar to container_execute, but instead of running commands inside containers
+    with 'podman exec', it runs commands on the host using container-related data (e.g, overlay
+    filesystem paths) from the provider datasource.
+
+    The provider datasource should return a list of tuples where:
+    - First 3 elements are (image, engine, container_id)
+    - Fourth element is: the value to substitute into command (e.g, merged_dir path)
+    - Remaining elements (if any) are: additional arguments for command substitution
+
+    Args:
+        provider (list): A datasource that returns a list of tuples with format (image, engine, container_id,
+                        substitution_value, <optional_args>)
+        cmd (str): A command template with %s placeholder(s). The first %s will be replaced with the substitution_value
+                (4th tuple element), and any additional %s will be replaced with optional_args.
+        context (ExecutionContext): The context under which the datasource should run.
+        split (bool): Whether the output should be split into a list of lines.
+        keep_rc (bool): Whether to return the error code. If False, non-zero return codes raise CalledProcessError. If
+                        True, return code and output are returned.
+        timeout (int): Seconds to wait for command completion. None for infinite.
+        inherit_env (list): Environment variables to override from the calling process.
+        override_env (dict): Environment variables to override in the calling process.
+
+    Returns:
+        function: A datasource that returns a list of HostBasedContainerCommandProvider outputs, each preserving
+        container metadata (image, engine, container_id).
+
+    Example:
+        If provider returns: ("rhel8", "podman", "/var/lib/.../merged")
+        AND cmd is: "/usr/bin/rpm -qa --root %s"
+        Then the_cmd becomes: "/usr/bin/rpm -qa --root /var/lib/.../merged"
+    """
+
+    def __call__(self, broker):
+        result = []
+        source = broker[self.provider]
+        cleaner = broker.get('cleaner')
+        ctx = broker[self.context]
+
+        if isinstance(source, ContentProvider):
+            source = source.content
+        if not isinstance(source, (list, set)):
+            source = [source]
+
+        for e in source:
+            try:
+                # e = (image, engine, container_id, substitution_value, <optional_args>)
+                image, engine, cid = e[0], e[1], e[2]
+
+                # Fourth element is the value to substitute into command
+                # (e.g, merged_dir for rpm --root command)
+                substitution_value = e[3]
+
+                # Any remaining elements are additional args for the command
+                additional_args = e[4:] if len(e) > 4 else ()
+
+                # Build args tuple: (substitution_value, <additional_args>)
+                all_agrs = (substitution_value,) + additional_args if additional_args else (substitution_value,)
+
+                # Substitute all args at once
+                # Note: we must handle rpm format strings that contain %{...} carefully
+                # The self.cmd should only have one %s (for merged_dir), but may have %{Name} etf for rpm
+                if "%s" in self.cmd:
+                    try:
+                        the_cmd = self.cmd % all_agrs
+                    except (TypeError, ValueError):
+                        # If % substitution fails (e.g, due to %{...} in rpm format),
+                        # fall back to simple string replacement for the first %s only
+                        the_cmd = self.cmd.replace("%s", substitution_value, 1)
+                        # Then handle additional args if any
+                        for arg in additional_args:
+                            the_cmd = the_cmd.replace("%s", str(arg), 1)
+                else:
+                    the_cmd = self.cmd
+
+                cmd_exec = self.cmd.split(None, 1)[0]
+                wrapped_cmd = 'sh -c "command -v {0} > /dev/null && {1}"'.format(cmd_exec, the_cmd)
+
+                # Create HostBasedContainerCommandProvider to preserve metadata
+                # This provider doesn't parse the command string like ContainerCommandProvider.
+                # so it works with host-side commands that don't follow 'podman exec' format
+
+                ccp = HostBasedContainerCommandProvider(
+                    wrapped_cmd,
+                    ctx,
+                    image=image,
+                    engine=engine,
+                    container_id=cid,
+                    args=e,
+                    split=self.split,
+                    keep_rc=self.keep_rc,
+                    ds=self,
+                    timeout=self.timeout,
+                    inherit_env=self.inherit_env,
+                    override_env=self.override_env,
+                    signum=self.signum,
+                    cleaner=cleaner,
+                )
+
+                result.append(ccp)
+
+            except NoFilterException as nfe:
+                raise nfe
+            except Exception:
+                log.debug(traceback.format_exc())
+
+        if result:
+            return result
+        raise ContentException("No results found for [%s]" % self.cmd)
+
+
 class container_collect(foreach_execute):
     """
     Collects the files at the resulting path in running containers.
@@ -1742,6 +1911,40 @@ def serialize_container_command(obj, root):
 
 @deserializer(ContainerCommandProvider)
 def deserialize_container_command(_type, data, root, ctx, ds):
+    rel = data["relative_path"]
+    res = SerializedOutputProvider(rel, root=root, ctx=ctx, ds=ds)
+    res.rc = data["rc"]
+    res.cmd = data["cmd"]
+    res.args = data["args"]
+    res.image = data["image"]
+    res.engine = data["engine"]
+    res.container_id = data["container_id"]
+    return res
+
+
+@serializer(HostBasedContainerCommandProvider)
+def serialize_oob_container_command(obj, root):
+    rel = os.path.join("insights_containers", obj.relative_path)
+    if obj.save_as:
+        rel = os.path.join("insights_containers", obj.save_as)
+        if obj.save_as.endswith("/"):
+            rel = os.path.join(rel, os.path.basename(obj.relative_path))
+    dst = os.path.join(root, rel)
+    rc = obj.write(dst)
+    return {
+        "rc": rc,
+        "cmd": None,
+        "args": obj.args,
+        "save_as": bool(obj.save_as),
+        "relative_path": rel,
+        "image": obj.image,
+        "engine": obj.engine,
+        "container_id": obj.container_id,
+    }
+
+
+@deserializer(HostBasedContainerCommandProvider)
+def desdeserialize_oob_container_command(_type, data, root, ctx, ds):
     rel = data["relative_path"]
     res = SerializedOutputProvider(rel, root=root, ctx=ctx, ds=ds)
     res.rc = data["rc"]
--- a/insights/core/spec_factory.py
+++ b/insights/core/spec_factory.py
@@ -535,6 +535,61 @@ class ContainerCommandProvider(ContainerProvider):
         return 'ContainerCommandProvider("%r")' % self.cmd
 
 
+class HostBasedContainerCommandProvider(CommandOutputProvider):
+    """
+    Provider for commands executed on the host that produce container-related data.
+
+    Unlike ContainerCommandProvider which parses 'podman exec' commands, this class runs commands
+    directly on the host using container metadata (e.g., overlay paths). The image, engine, and container_id
+    are set explicitly rather than parsed from the command string.
+
+    Inherits from CommandOutputProvider directly to keep the hierarchy shallow and separate from in-container providers.
+    Uses root="insights_containers" to honor the existing archive convention for container-related data.
+    """
+
+    def __init__(
+        self,
+        cmd_path,
+        ctx,
+        image=None,
+        engine=None,
+        container_id=None,
+        args=None,
+        split=True,
+        keep_rc=False,
+        ds=None,
+        timeout=None,
+        inherit_env=None,
+        override_env=None,
+        signum=None,
+        cleaner=None,
+    ):
+        super(HostBasedContainerCommandProvider, self).__init__(
+            cmd_path,
+            ctx,
+            root="insights_containers",
+            save_as=None,
+            args=args,
+            split=split,
+            keep_rc=keep_rc,
+            ds=ds,
+            timeout=timeout,
+            inherit_env=inherit_env,
+            override_env=override_env,
+            signum=signum,
+            cleaner=cleaner,
+        )
+        self.image = image
+        self.engine = engine
+        self.container_id = container_id
+        self.relative_path = os.path.join(
+            container_id, "insights_commands", mangle_command(self.cmd)
+        )
+
+    def __repr__(self):
+        return f"HostBasedContainerCommandProvider('{self.cmd}')"
+
+
 class RegistryPoint(object):
     # Marker class for declaring that an element of a `SpecSet` subclass
     # is a registry point against which further subclasses can register
@@ -1409,6 +1464,102 @@ class container_execute(foreach_execute):
         raise ContentException("No results found for [%s]" % self.cmd)
 
 
+class container_foreach_execute(foreach_execute):
+    """
+    Execute a command on the host for each container using data from a provider component.
+
+    This is similar to container_execute, but instead of running commands inside containers
+    with 'podman exec', it runs commands on the host using container-related data (e.g, overlay
+    filesystem paths) from the provider component.
+
+    The provider component should return a list of tuples where:
+    - First 3 elements are (image, engine, container_id)
+    - Remaining elements are: command arguments for %s substitution in cmd
+
+    Args:
+        provider: A component that returns a list of tuples with format (image, engine, container_id,
+                  command_args...). Each tuple provides metadata and arguments for one container command.
+        cmd (str): A command template with %s placeholder(s) that will be replaced with the command arguments.
+                   **Note:** Commands containing format strings (e.g., RPM ``%{NAME}``) must escape the
+                   percent sign as ``%%{NAME}`` to avoid conflicts with Python's % operator.
+        context (ExecutionContext): The context under which the component should run.
+        split (bool): Whether the output should be split into a list of lines.
+        keep_rc (bool): Whether to return the error code. If False, non-zero return codes raise CalledProcessError. If
+                        True, return code and output are returned.
+        timeout (int): Seconds to wait for command completion. None for infinite.
+        inherit_env (list): Environment variables to override from the calling process.
+        override_env (dict): Environment variables to override in the calling process.
+
+    Returns:
+        function: A component that returns a list of HostBasedContainerCommandProvider outputs, each preserving
+        container metadata (image, engine, container_id).
+
+    Example:
+        If provider returns: ("rhel8", "podman", "abc123", "/var/lib/.../merged")
+        AND cmd is: "/usr/bin/rpm -qa --root %s --qf '%%{NAME}-%%{VERSION}'"
+        Then the_cmd becomes: "/usr/bin/rpm -qa --root /var/lib/.../merged --qf '%{NAME}-%{VERSION}'"
+    """
+
+    def __call__(self, broker):
+        result = []
+        source = broker[self.provider]
+        cleaner = broker.get('cleaner')
+        ctx = broker[self.context]
+
+        if isinstance(source, ContentProvider):
+            source = source.content
+        if not isinstance(source, (list, set)):
+            source = [source]
+
+        for e in source:
+            try:
+                # e = (image, engine, container_id, command_args...)
+                image, engine, cid = e[0], e[1], e[2]
+                cmd_args = e[3:]  # Extract only the command arguments
+
+                if "%s" in self.cmd:
+                    try:
+                        the_cmd = self.cmd % tuple(cmd_args)
+                    except (TypeError, ValueError):
+                        # TypeError: not enough args for placeholders
+                        # ValueError: incomplete format or other % operator issues
+                        raise ValueError(
+                            "Not enough command arguments for placeholders in: %s" % self.cmd
+                        )
+                else:
+                    the_cmd = self.cmd
+
+                ccp = HostBasedContainerCommandProvider(
+                    the_cmd,
+                    ctx,
+                    image=image,
+                    engine=engine,
+                    container_id=cid,
+                    args=e,
+                    split=self.split,
+                    keep_rc=self.keep_rc,
+                    ds=self,
+                    timeout=self.timeout,
+                    inherit_env=self.inherit_env,
+                    override_env=self.override_env,
+                    signum=self.signum,
+                    cleaner=cleaner,
+                )
+
+                result.append(ccp)
+
+            except NoFilterException as nfe:
+                raise nfe
+            except ContentException as ce:
+                log.debug(ce)
+            except Exception:
+                log.debug(traceback.format_exc())
+
+        if result:
+            return result
+        raise ContentException("No results found for [%s]" % self.cmd)
+
+
 class container_collect(foreach_execute):
     """
     Collects the files at the resulting path in running containers.
@@ -1742,6 +1893,40 @@ def serialize_container_command(obj, root):
 
 @deserializer(ContainerCommandProvider)
 def deserialize_container_command(_type, data, root, ctx, ds):
+    rel = data["relative_path"]
+    res = SerializedOutputProvider(rel, root=root, ctx=ctx, ds=ds)
+    res.rc = data["rc"]
+    res.cmd = data["cmd"]
+    res.args = data["args"]
+    res.image = data["image"]
+    res.engine = data["engine"]
+    res.container_id = data["container_id"]
+    return res
+
+
+@serializer(HostBasedContainerCommandProvider)
+def serialize_hb_container_command(obj, root):
+    rel = os.path.join("insights_containers", obj.relative_path)
+    if obj.save_as:
+        rel = os.path.join("insights_containers", obj.save_as)
+        if obj.save_as.endswith("/"):
+            rel = os.path.join(rel, os.path.basename(obj.relative_path))
+    dst = os.path.join(root, rel)
+    rc = obj.write(dst)
+    return {
+        "rc": rc,
+        "cmd": None,
+        "args": obj.args,
+        "save_as": bool(obj.save_as),
+        "relative_path": rel,
+        "image": obj.image,
+        "engine": obj.engine,
+        "container_id": obj.container_id,
+    }
+
+
+@deserializer(HostBasedContainerCommandProvider)
+def deserialize_hb_container_command(_type, data, root, ctx, ds):
     rel = data["relative_path"]
     res = SerializedOutputProvider(rel, root=root, ctx=ctx, ds=ds)
     res.rc = data["rc"]
--- a/insights/specs/datasources/container/containers_inspect.py
+++ b/insights/specs/datasources/container/containers_inspect.py
@@ -3,6 +3,7 @@ Custom datasources for containers inspect information
 """
 
 import json
+import logging
 import os
 
 from insights.core.context import HostContext
@@ -12,6 +13,8 @@ from insights.core.plugins import datasource
 from insights.core.spec_factory import DatasourceProvider, foreach_execute
 from insights.specs import Specs
 from insights.specs.datasources.container import running_rhel_containers
+
+log = logging.getLogger(__name__)
 
 
 @datasource(running_rhel_containers, HostContext)
@@ -101,4 +104,59 @@ def containers_inspect_data_datasource(broker):
                 )
     except Exception as e:
         raise ContentException("Unexpected content exception:{e}".format(e=str(e)))
+    raise SkipComponent
+
+
+@datasource(LocalSpecs.containers_inspect_data_raw, running_rhel_containers, HostContext)
+def container_merged_dirs(broker):
+    """
+    This datasource extracts the MergedDir path from container inspect output for containers using overlay
+    storage driver.
+
+    The MergedDir is the unified view of all container layers and is used to run host-side RPM queries without
+    entering the container, eliminating container resource consumption.
+
+    Uses ``LocalSpecs.containers_inspect_data_raw`` so that ``/usr/bin/docker|podman inspect <container ID>``
+    runs only once per container, shared with ``containers_inspect_data_datasource``.
+
+    Returns:
+        list: List of tuples ``(image, engine, container_id, merged_dir)``
+              where merged_dir is the path to the container's overlay merged directory.
+
+    Raises:
+        SkipComponent: When no containers have valid overlay MergedDir or any exception occurs.
+    """
+    image_map = {}
+    for container in broker[running_rhel_containers]:
+        image, engine, container_id = container
+        image_map[container_id] = image
+
+    results = []
+
+    for item in broker[LocalSpecs.containers_inspect_data_raw]:
+        try:
+            inspect_data = json.loads(''.join(item.content))[0]
+
+            if not hasattr(item, 'args') or not item.args:
+                continue
+
+            engine, container_id = item.args
+            image = image_map.get(container_id, '')
+
+            merged_dir = inspect_data.get("GraphDriver", {}).get("Data", {}).get("MergedDir")
+            if not merged_dir:
+                continue
+
+            if not os.path.exists(merged_dir):
+                log.debug("MergedDir path does not exist: %s for container %s", merged_dir, container_id)
+                continue
+
+            results.append((image, engine, container_id, merged_dir))
+
+        except (json.JSONDecodeError, KeyError, IndexError, ValueError, AttributeError):
+            continue
+
+    if results:
+        return results
+
     raise SkipComponent
--- a/insights/specs/datasources/container/containers_inspect.py
+++ b/insights/specs/datasources/container/containers_inspect.py
@@ -3,6 +3,7 @@ Custom datasources for containers inspect information
 """
 
 import json
+import logging
 import os
 
 from insights.core.context import HostContext
@@ -12,6 +13,8 @@ from insights.core.plugins import datasource
 from insights.core.spec_factory import DatasourceProvider, foreach_execute
 from insights.specs import Specs
 from insights.specs.datasources.container import running_rhel_containers
+
+log = logging.getLogger(__name__)
 
 
 @datasource(running_rhel_containers, HostContext)
@@ -101,4 +104,50 @@ def containers_inspect_data_datasource(broker):
                 )
     except Exception as e:
         raise ContentException("Unexpected content exception:{e}".format(e=str(e)))
+    raise SkipComponent
+
+
+@datasource(LocalSpecs.containers_inspect_data_raw, HostContext)
+def container_merged_dirs(broker):
+    """
+    This datasource extracts the MergedDir path from container inspect output for containers using overlay
+    storage driver.
+
+    The MergedDir provides a unified view of all container layers. It enables host-side queries without
+    entering the container, eliminating container resource consumption.
+
+    Uses ``LocalSpecs.containers_inspect_data_raw`` so that ``/usr/bin/docker|podman inspect <container ID>``
+    runs only once per container, shared with ``containers_inspect_data_datasource``.
+
+    Returns:
+        list: List of tuples ``(image, engine, container_id, merged_dir)``
+              where merged_dir is the path to the container's overlay merged directory.
+
+    Raises:
+        SkipComponent: When no containers have valid overlay MergedDir or any exception occurs.
+    """
+    results = []
+
+    for item in broker[LocalSpecs.containers_inspect_data_raw]:
+        try:
+            inspect_data = json.loads(''.join(item.content))[0]
+
+            if not hasattr(item, 'args') or not item.args:
+                continue
+
+            engine, container_id = item.args
+            image = inspect_data.get("ImageName", "")
+
+            merged_dir = inspect_data.get("GraphDriver", {}).get("Data", {}).get("MergedDir")
+            if not merged_dir:
+                continue
+
+            results.append((image, engine, container_id, merged_dir))
+
+        except (json.JSONDecodeError, KeyError, IndexError, ValueError, AttributeError):
+            continue
+
+    if results:
+        return results
+
     raise SkipComponent
--- a/insights/tests/core/spec_factory/test_container_foreach_execute.py
+++ b/insights/tests/core/spec_factory/test_container_foreach_execute.py
@@ -1,0 +1,326 @@
+from unittest.mock import patch
+
+from insights import dr
+from insights.core.context import HostContext
+from insights.core.exceptions import NoFilterException
+from insights.core.plugins import datasource
+from insights.core.spec_factory import (
+    DatasourceProvider,
+    HostBasedContainerCommandProvider,
+    container_foreach_execute,
+    foreach_execute,
+)
+
+
+MOCK_WHICH = "insights.core.spec_factory.which"
+MOCK_BLACKLIST = "insights.core.spec_factory.blacklist.allow_command"
+
+SAMPLE_IMAGE = "registry.access.redhat.com/ubi8/ubi:latest"
+SAMPLE_ENGINE = "podman"
+SAMPLE_CID = "abc123def456"
+SAMPLE_MERGED_DIR = "/var/lib/containers/storage/overlay/abcdef123456/merged"
+
+MULTIPLE_CONTAINERS_DATA = [
+    ("image1", "podman", "cid111111111", "/merged/1"),
+    ("image2", "podman", "cid222222222", "/merged/2"),
+    ("image3", "docker", "cid333333333", "/merged/3"),
+]
+
+
+def test_container_foreach_execute_inherits_from_foreach_execute():
+    """Verify container_foreach_execute inherits from foreach_execute."""
+    assert issubclass(container_foreach_execute, foreach_execute), \
+        "container_foreach_execute must inherit from foreach_execute"
+
+
+# Test datasources - defined at module level like in production
+@datasource(HostContext)
+def mock_single_container(broker):
+    return [(SAMPLE_IMAGE, SAMPLE_ENGINE, SAMPLE_CID, SAMPLE_MERGED_DIR)]
+
+
+@datasource(HostContext)
+def mock_multiple_containers(broker):
+    return MULTIPLE_CONTAINERS_DATA
+
+
+@datasource(HostContext)
+def mock_empty_containers(broker):
+    return []
+
+
+# Define specs at module level - mimics production usage
+container_rpms_single = container_foreach_execute(
+    mock_single_container,
+    "/usr/bin/rpm -qa --root %s",
+    context=HostContext
+)
+
+container_rpms_multiple = container_foreach_execute(
+    mock_multiple_containers,
+    "/usr/bin/rpm -qa --root %s",
+    context=HostContext
+)
+
+container_rpms_with_format = container_foreach_execute(
+    mock_single_container,
+    "/usr/bin/rpm -qa --root %s --qf '%%{NAME}-%%{VERSION}'",
+    context=HostContext
+)
+
+container_rpms_empty = container_foreach_execute(
+    mock_empty_containers,
+    "/usr/bin/rpm -qa --root %s",
+    context=HostContext
+)
+
+container_rpms_two_placeholders = container_foreach_execute(
+    mock_single_container,
+    "/usr/bin/rpm --root %s --dbpath %s",
+    context=HostContext
+)
+
+
+def _run(spec):
+    """Run spec through the framework like in production."""
+    broker = dr.Broker()
+    broker[HostContext] = HostContext()
+    return dr.run(dr.get_dependency_graph(spec), broker)
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_cfe_simple_substitution(mock_which, mock_bl):
+    """Test basic %s substitution with merged directory path."""
+    broker = _run(container_rpms_single)
+    result = broker[container_rpms_single]
+
+    assert len(result) == 1
+    provider = result[0]
+    assert isinstance(provider, HostBasedContainerCommandProvider)
+    assert SAMPLE_MERGED_DIR in provider.cmd
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_cfe_preserves_container_metadata(mock_which, mock_bl):
+    """Test that container metadata (image, engine, container_id) is preserved."""
+    broker = _run(container_rpms_single)
+    result = broker[container_rpms_single]
+
+    provider = result[0]
+    assert provider.image == SAMPLE_IMAGE
+    assert provider.engine == SAMPLE_ENGINE
+    assert provider.container_id == SAMPLE_CID
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_cfe_passes_command_directly(mock_which, mock_bl):
+    """Test that the command is substituted correctly."""
+    broker = _run(container_rpms_single)
+    result = broker[container_rpms_single]
+
+    provider = result[0]
+    assert provider.cmd == "/usr/bin/rpm -qa --root %s" % SAMPLE_MERGED_DIR
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_cfe_rpm_format_strings(mock_which, mock_bl):
+    """Test that RPM format strings like %{NAME} are preserved."""
+    broker = _run(container_rpms_with_format)
+    result = broker[container_rpms_with_format]
+
+    assert len(result) == 1
+    assert SAMPLE_MERGED_DIR in result[0].cmd
+    assert "%{NAME}-%{VERSION}" in result[0].cmd
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_cfe_multiple_containers(mock_which, mock_bl):
+    """Test processing multiple containers."""
+    broker = _run(container_rpms_multiple)
+    result = broker[container_rpms_multiple]
+
+    assert len(result) == 3
+    for i, provider in enumerate(result):
+        assert provider.image == MULTIPLE_CONTAINERS_DATA[i][0]
+        assert provider.engine == MULTIPLE_CONTAINERS_DATA[i][1]
+        assert provider.container_id == MULTIPLE_CONTAINERS_DATA[i][2]
+        assert MULTIPLE_CONTAINERS_DATA[i][3] in provider.cmd
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_cfe_stores_original_tuple_as_args(mock_which, mock_bl):
+    """Test that the original tuple is stored in args."""
+    broker = _run(container_rpms_single)
+    result = broker[container_rpms_single]
+
+    entry = (SAMPLE_IMAGE, SAMPLE_ENGINE, SAMPLE_CID, SAMPLE_MERGED_DIR)
+    assert result[0].args == entry
+
+
+def test_cfe_raises_content_exception_on_empty_source():
+    """Test that ContentException is raised when no containers are provided."""
+    broker = _run(container_rpms_empty)
+
+    # When empty source, ContentException is raised and caught by the framework,
+    # so the spec doesn't get added to broker
+    assert container_rpms_empty not in broker
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_cfe_no_shell_wrapping(mock_which, mock_bl):
+    """Verify the command is passed directly without sh -c wrapping."""
+    broker = _run(container_rpms_with_format)
+    result = broker[container_rpms_with_format]
+
+    assert len(result) == 1
+    assert "sh -c" not in result[0].cmd
+    assert SAMPLE_MERGED_DIR in result[0].cmd
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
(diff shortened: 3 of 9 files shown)


insights/specs/datasources/container/merged_dirs.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Returns:
+        list: List of (engine, container_id, image) tuples
+    """
+    containers_info = []
+    for container in broker[running_rhel_containers]:
+        image, engine, container_id = container
+        # Reorder to (engine, container_id, image) so command formatting uses first 2 elements.
+        containers_info.append((engine, container_id, image))
+    return containers_info
```
qinpingli: Seems `running_rhel_containers` returns ` a list of tuple of (image, <podman|docker>, container_id) of the running rhel containers.`. It can be used in `container_merged_dirs` directly?

insights/specs/datasources/container/merged_dirs.py
```diff
@@ -0,0 +1,132 @@
...
+
+            merged_dir = driver_data.get('MergedDir')
+            if not merged_dir:
+                # Skip containers without MergedDir (Non-overlay storage or missing field)
+                continue
+
+            # Edgecase: Validate that MergedDir path exists on the filesystem
+            if not os.path.exists(merged_dir):
+                log.debug("MergedDir path does not exist: %s for container %s", merged_dir, container_id)
+                continue
```
qinpingli: How about updating the `merged_dir` check to:
```
merged_dir = inspect_data.get("GraphDriver", {}).get("Data", {}).get("MergedDir")
if not merged_dir:
    continue

# Edgecase: Validate that MergedDir path exists on the filesystem
if not os.path.exists(merged_dir):
    log.debug("MergedDir path does not exist: %s for container %s", merged_dir, container_id)
    continue

File: insights/specs/datasources/container/merged_dirs.py [hunk 3]
```diff
@@ -0,0 +1,132 @@
...
+
+            # Edgecase: Validate that MergedDir path is accessible (readable)
+            if not os.access(merged_dir, os.R_OK):
+                log.debug("MergedDir path is not accessible: %s for container %s", merged_dir, container_id)
+                continue
```
qinpingli: Hmm, for this part, is it necessary based on `insights-client` is run with root user?

insights/core/spec_factory.py
```diff
@@ -1409,6 +1470,124 @@ def __call__(self, broker):
...
+                # Use self.cmd (original) to extract cmd_exec to avoid issues with % in substituted values
+
+                cmd_exec = self.cmd.split(None, 1)[0]
+                wrapped_cmd = 'sh -c "command -v %s > /dev/null && %s"' % (cmd_exec, the_cmd)
+
+                # Create HostBasedContainerCommandProvider to preserve metadata
+                # This provider doesn't parse the command string like ContainerCommandProvider.
+                # so it works with host-side commands that don't follow 'podman exec' format
+
+                ccp = HostBasedContainerCommandProvider(
```
qinpingli, line 1532: Hmm, Just a thought, does the following code work?
```
for e in source:
    try:
        image, engine, cid = e[:3]
        if len(e) > 3:
            the_cmd = self.cmd % e[3:]
        else:
            the_cmd = self.cmd
         cop = CommandOutputProvider(
                    the_cmd,
                    ctx,
                    args=e,
                    split=self.split,
                    keep_rc=self.keep_rc,
                    ds=self,
                    timeout=self.timeout,
                    inherit_env=self.inherit_env,
                    override_env=self.override_env,
                    signum=self.signum,
                    cleaner=cleaner,
                )
....

File: insights/specs/datasources/container/merged_dirs.py (hunk 1 above)
Comment (sabhasin) [reply insights/specs/datasources/container/merged_dirs.py]: Thank you for reviewing it, `running_rhel_containers` returns `(image, engine, container_id)`, but `foreach_execute` uses positional `%s` substitution for the command ``/usr/bin/%s inspect %s`, which needs `engine` and `container_id` as the first two elements. That's why we reorder to `(engine, container_id, image)` so the command formats correctly, and `image` is still preserved in `item.args` for later use in `container_merged_dirs`.

File: insights/specs/datasources/container/merged_dirs.py (hunk 2 above)
Comment (sabhasin) [reply insights/specs/datasources/container/merged_dirs.py]: Thank you for the suggestion. I have applied the suggestion to the code.

File: insights/specs/datasources/container/merged_dirs.py (hunk 3 above)
Comment (sabhasin) [reply insights/specs/datasources/container/merged_dirs.py]: I think we can remove this edgecase, as os.access check will never fail, i will remove this from the code.

File: insights/core/spec_factory.py (hunk 4 above)
Comment (sabhasin) [reply insights/core/spec_factory.py:1532]: Using `CommandOutputProvider` here would unfortunately lose the container metadata `(.image, .engine, and .container_id)` that downstream parsers rely on.

Specifically, the `ContainerInstalledRpms` parser expects these attributes to be present on the provider, whereas CommandOutputProvider does not populate them.

The existing `ContainerCommandProvider` is also not a suitable alternative in this case because its `_misc_settings()` implementation extracts the metadata by parsing commands in the `podman exec <container_id> ... ` format. Our host-side command `(rpm -qa --root <merged_dir>) `does not follow that pattern, so the parsing would fail.

This is the reason `HostBasedContainerCommandProvider` exists. It inherits from `ContainerProvider`, ensuring that parsers still have access to `.image, .engine, and .container_id`, while explicitly setting these attributes instead of attempting to derive them from the command string.

File: insights/specs/datasources/container/merged_dirs.py (hunk 1 above)
Comment (qinpingli) [reply insights/specs/datasources/container/merged_dirs.py]: Thanks for the explanation! 

How about moving `container_merged_dirs` datasource to `specs.datasources.container.containers_inspect`. The reason is:
1. container_merged_dirs ds also gets from container inspect information, we can see `specs.datasources.container.containers_inspect`  module contains the datasources related to container inspect information
2. `specs.datasources.container.containers_inspect。LocalSpecs.containers_inspect_data_raw` component can be used for `container_merged_dirs` ds, then `/usr/bin/podman inspect <container_id>` command will be run once.

File: insights/core/spec_factory.py (hunk 4 above)
Comment (qinpingli) [reply insights/core/spec_factory.py:1532]: Thanks for the explanation!

`HostBasedContainerCommandProvider` provider's implementation and name  are not easy to understand, so I wanted to remove this provider. (Naming is difficult for me. :( )

Hmm, let's back to add a new provider.

File: insights/core/spec_factory.py [hunk 5]
```diff
@@ -535,6 +535,67 @@ def __repr__(self):
         return 'ContainerCommandProvider("%r")' % self.cmd
 
 
+class HostBasedContainerCommandProvider(ContainerProvider):
```
qinpingli: [Question] In the future, is it possible use one provider to finished the data extraction from the outside of container? The detail is: when we want to get a file in container?

If no, how about updating class name to `OutOfBandContainerCommandProvider`? Is it better?

And how about inheriting from CommandOutputProvider?

insights/specs/datasources/container/merged_dirs.py
sabhasin: Thanks for the suggestion! That makes complete sense. Having `container_merged_dirs` in `containers_inspect` prevents `podman inspect` from running twice for the same containers.

Here's the approach:

- Move `container_merged_dirs` into `specs.datasources.container.containers_inspect`.

- Make it depend on `LocalSpecs.containers_inspect_data_raw` (reusing the existing inspect output) along with `running_rhel_containers` (to get the `image` field, since `containers_inspect_data_raw` only carries `(engine, container_id)` in `item.args`).

- Remove `merged_dirs.py` entirely; no more separate `running_rhel_containers_with_image` datasource or duplicate `LocalSpecs`.

This ensures `/usr/bin/podman inspect <container_id>` runs only once per container.

@qinpingli  Do let me know if this looks good to you; I will start working on this approach.

insights/core/spec_factory.py
sabhasin: Completely agree, the naming is always the hard part. 

Here are a few alternatives we could consider:

1. ContainerHostCommandProvider — "container command, run on host"
2. ExternalContainerCommandProvider — "external to the container"
3. ContainerRootfsCommandProvider — describes the actual use case (querying the container's rootfs from the host). 

Happy to go with whichever reads best to you, or if you have a different suggestion.

insights/core/spec_factory.py
sabhasin: Yes, the `MergedDir` approach could potentially be extended to file collection in the future, instead of `podman cp`, we could read files directly from `<MergedDir>/path/to/file`. That would be a separate provider, though (file-based, not command-based), so something we could explore in a follow-up.

Regarding naming - `OutOfBandContainerCommandProvider` works. I will update the class name with this once the above-mentioned approach looks good to you. 

On inheritance - we need to inherit from `ContainerProvider` rather than `CommandOutputProvider` because `ContainerProvider` defines the `.image, .engine, .container_id `interface that downstream parsers like `ContainerInstalledRpms` depend on. Inheriting from `CommandOutputProvider` directly would lose that interface.

insights/specs/datasources/container/merged_dirs.py
qinpingli: Thanks! It looks good to me.

insights/core/spec_factory.py
qinpingli: > ContainerProvider defines the .image, .engine, .container_id interface

Yes, this is what I thought, ContainerProvider defines all required interfaces related to container datasources. We should inherit from it. But seems only image there (https://github.com/RedHatInsights/insights-core/blob/master/insights/core/spec_factory.py#L490)?  It more like a provider for image.

If we want to follow this design - ContainerProvider defines all required interfaces related to container datasources. It looks like a good idea. Since it's related to basic providers change, I'd like to invite another maintainer to have more discussion.

Hi @jholecek-rh  ,

We need your opinion here.

apimplap:
test dp

apimplap:
test dp

apimplap:
test dp

apimplap:
test dp

apimplap:
test dp

apimplap:
test dp

apimplap:
test dp

apimplap:
test dp

insights/core/spec_factory.py
jholecek-rh: As far as I can tell, provider class hierarchy has little (if anything) to do with interfaces. I think that the hierarchy is an ad-hoc product of the desire to minimize code duplication at every step (change) leading to the current state. The `ContainerProvider` class and the `_misc_settings` method are good examples of that.

Provider classes serve as values of "spec" implementations. The most important interface that they need to provide is the one consumed by higher-level components (parsers, typically) - this is the abstraction that specs are meant to provide. Apart from the access to the line-oriented content, the container-related provided classes are expected to provide `image`, `engine` and `container_id` attributes. The implementations can be widely different. For example, if the sosreport started gathering installed RPM from running containers, `SosSpecs.container_installed_rpms` would have to build a value from files in the sosreport that would provide the same effective interface as `DefaultSpecs.container_installed_rpms` does. There would be hardly any reason for the provider class used by `SosSpecs.container_installed_rpms` to inherit from `CommandOutputProvider`.

Provider classes that are used by specs collected by the insights-client have another important interface to consider: the insights-client archive structure. The hardcoded string `"insights_containers"` in `ContainerProvider` is a directory name in the archive to store container-related data. IIUC, there are no hard requirements for the new provider class to use the same directory. It could  use any as long as it avoids collisions with other datasources. However, it would be probably reasonable for the new provider class to honor the existing convention of putting container-related data to the `insights_containers` directory.

IIUC, the purpose of the `ContainerProvider` class is to:
1. Avoid the need for `ContainerFileProvider` and `ContainerCommandProvider` to avoid the need to repeat the lengthy `__init__` signature and passing all the arguments to the parent `__init__` method, utilizing the ill-designed `_misc_settings` method.
2. Have both `ContainerFileProvider`  and `ContainerCommandProvider` store data in the same place in the insights-client archive (the `insights_containers` directory).
3. Store the `image` attribute. I personally think this does more harm than good because the actual interfaces become less obvious. 

The new provider class will have to define its own `__init__` method anyway so the first reason does not apply. The third reason is a non-reason (for me). 

So the base class decision boils down to:

* `CommandOutputProvider`
  * Shallower class hierarchy is easier to troubleshoot.
  * Container-related attributes defined in a single space.
  * The hierarchy separates the new provider from ones that do in-container stuff.
* `ContainerProvider`
  * The insights-client directory for container-related data defined in one place (for now! future changes may prove this hierarchy insustainable).
  * All container-related providers have a shared parent (helps categorization).

Having said that, it does not matter much. I am slightly in favor of inheriting directly from `CommandOutputProvider`.

insights/core/spec_factory.py
```diff
@@ -535,6 +535,67 @@ def __repr__(self):
...
+        timeout=None,
+        inherit_env=None,
+        override_env=None,
+        signum=None,
+        cleaner=None,
+    ):
+        # Store engine and container_id before calling parent __init__
+        # because _misc_setting() needs them
+        self._engine = engine
+        self._container_id = container_id
```
jholecek-rh: This is not a way to alter the behavior of a parent method. Let's call the parent `__init__` method and then update the attributes as needed. The `_misc_settings` method is not generic enough to be useful in this case. We don't have to (should not) use it, especially if using it makes the code more convoluted.

insights/core/spec_factory.py
jholecek-rh: `HostBasedContainerCommandProvider` is fine, IMHO. The main challenge with the naming in this case is that we need a name that would contrast with the ambiguous `ContainerCommandProvider`. No name we choose for the new provider class will make `ContainerCommandProvider` more specific.

As for getting replacing the new provider class with something completely different, that is a change beyond the scope of this PR. As I hinted at in my other comment, there's a lot to clean up regarding interfaces (not only) in this part of the code base.

jholecek-rh:
@sabhasin Please resolve all threads from Sourcery. It is up to you if you review it for any meaningful hints or if you discard it outright. If you happen to review its suggestions, I would be interested in your feedback. My experience with Sourcery has been pretty bad so far, and skimming the comments in this PR suggests that this is a new low.

jholecek-rh:
@sabhasin If you have not done an end-to-end test, please do one:

1. Provision a RHEL system with running containers
2. Have an insights-client to create an archive using the new code
3. Locate the data collected by the spec in the archive
4. Have a rule using the `ContainerInstalledRpms` spec hit when executed on the insights-client archive using `insights-run`
5. Share details of the test environment, a generated archive with the collected data and the sample rule with us

It might be difficult to trigger an existing rule in a way that would demonstrably use the spec. I would recommend writing a sample rule depending just on the `ContainerInstalledRpms` parser.

Feel free to ask if you are not sure how to run a test like that. I do not know exactly either, but we can figure it out together. :slightly_smiling_face: 

Bottom line: I am 99% confident that the change is incomplete. I think that the archive will not contain the data. If I am right, a test and a sample archive will make it easier for me to explain. If I am wrong, I will have an example to correct my understanding. Either way, conducting a manual end-to-end tests using a development version of the code will be a very useful experience for us. We've lost a lot of working knowledge in the team recently.

apimplap:
test egg

apimplap:
test egg

apimplap:
test egg

insights/core/spec_factory.py
sabhasin: Agreed, using `_misc_settings()` with pre-stored temp variables was an unnecessary indirection. Updated to simply call the parent `__init__` and then set `image`, `engine`, `container_id` and `relative_path` directly afterward. The `_misc_settings()` override is removed entirely.

insights/core/spec_factory.py
sabhasin: Thank you for the detailed analysis of the provider hierarchy. 
ContainerProvider's main value (avoiding `__init__`duplication) doesn't apply here since we define our own `__init__` anyway. And keeping the hierarchy shallow makes the code easier to follow and troubleshoot.  As suggested, I have updated the `OutOfBandContainerCommandProvider` to inherit directly from `CommandOutputProvider`:

- `image`, `engine`, and `container_id` are set as direct attributes - all container-related state is defined in one place.
- root="insights_containers" is passed explicitly to `CommandOutputProvider`. `__init__` to honor the existing archive directory convention.
- MRO is now `OutOfBandContainerCommandProvider` → `CommandOutputProvider` → `ContentProvider`, cleanly separated from the in-container provider branch.

The effective interface for parsers (`.image`, `.engine`, `.container_id`, line-oriented `.content`) remains the same.

insights/core/spec_factory.py
sabhasin: `HostBasedContainerCommandProvider` is fine. The naming ambiguity stems from `ContainerCommandProvider` being too generic, and no rename on our side would fix that. Keeping the original name. The broader interface cleanup is beyond the scope of this PR.

insights/specs/datasources/container/merged_dirs.py
sabhasin: Performed the suggested changes.

insights/core/spec_factory.py
```diff
@@ -535,6 +535,67 @@ def __repr__(self):
...
+class HostBasedContainerCommandProvider(ContainerProvider):
+    """
+    A CommandOutputProvider subclass for commands executed on the host (not inside containers) but
+    that produce container-related data.
+
+    Unlike ContainerCommandProvider which prases 'podman exec' commands, this class is used when the
+    command runs directly on the host using container metadata (e.g, overlay paths).
+
+    The engine and container_id must be set explicitly after initialization, as they cannot be parsed
+    from the command string.
```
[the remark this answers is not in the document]
sabhasin: 1. `/usr/bin%s` missing slash — This is no longer an issue. We removed `merged_dirs.py` and its separate `LocalSpecs.container_inspect_raw`. The `container_merged_dirs` datasource now reuses `LocalSpecs.containers_inspect_data_raw` from containers_inspect.py, which already has the correct /usr/bin/%s inspect %s template.

2. Broad except Exception in container_foreach_execute - Acknowledged. This follows the same pattern
used by container_execute and foreach_execute elsewhere in the codebase, but we can look at narrowing it or raising the log level in a follow-up.

3. Unit tests for `HostBasedContainerCommandProvider` - Good idea. Note that the suggested implementation doesn't apply to the current code - we've since changed the inheritance to `CommandOutputProvider` (not `ContainerProvider`), removed `_misc_settings()` entirely, and set attributes directly after `super().__init__().` We'll add tests that match the current implementation.

4. Unit tests for `container_foreach_execute` — Agree this would be valuable. We'll add tests covering command formatting (including the `%` fallback path for rpm format strings), host-side wrapping, metadata preservation, and the `ContentException` path.

jholecek-rh:
@sabhasin I'm struggling to understand the PR history. When were the `@serializer` and `@deserializer` definitions added to the PR? Commit 47532e6 didn't have them. I believe this was the commit that I reviewed, and the missing definitions were the reason why I [said](https://github.com/RedHatInsights/insights-core/pull/4782#issuecomment-5035307132) it wouldn't work.

I do not see when the branch was updated from that commit. The next push message says that the branch was already at commit f2a541a and this commit has the definitions. Is it the commit that added the definitions?

I was really confused when you shared working test results. I thought that I missed something and it worked all along. Could you clarify, please? I would like to know for sure whether my analysis was correct or flawed.

sabhasin:
Hello @jholecek-rh , 

Your analysis was correct - commit 47532e6 did not include the `@serializer` and `@deserializer`  definitions. Without them, the archive would not contain the data. That was a real gap.

Here's what happened (from the git reflog):

47532e61  RHINENG-27038@{4}: commit (amend) — this is what you reviewed (no serializer/deserializer, used merged_dirs.py)
      ↓  (amend)
f2a541a9  RHINENG-27038@{2}: commit (amend) — added serializer/deserializer, moved to  containers_inspect.py, added tests
      ↓  (amend)
dbe68de4  RHINENG-27038@{1}: commit (amend)
      ↓  (rebase onto updated master)
60dc5e5a  RHINENG-27038@{0}: current HEAD — same content as f2a541a, rebased

The serializer/deserializer were added in the amend that produced f2a541a. Since all these were amends to a single commit (not separate commits), there's no visible "when" in the PR push history - the branch jumped from 47532e6 to f2a541a in a force-push. That's why the push message said the branch was "already at" f2a541a — it had been force-pushed with the additions.

So to answer directly: your analysis was correct. The commit you reviewed (47532e6) was missing the serializer/deserializer, and the archive would not have contained the data. I added them afterward and force-pushed (amended), which is why the timeline was confusing. I should have made that explicit rather than sharing test results that made it look like it worked all along. I apologize for the confusion.

---
Changes performed today:

- Moved container_merged_dirs into specs/datasources/container/containers_inspect.py. It depends on  LocalSpecs.containers_inspect_data_raw (reusing the existing inspect output) along with running_rhel_containers (to get the image field, since containers_inspect_data_raw only carries (engine, container_id) in item.args). Removed merged_dirs.py entirely - no more separate running_rhel_containers_with_image datasource or duplicate LocalSpecs. This ensures /usr/bin/podman inspect <container_id> runs only once per container.

 - Updated HostBasedContainerCommandProvider (previously OutOfBandContainerCommandProvider) to inherit directly from CommandOutputProvider as suggested. ContainerProvider's main value (avoiding __init__  duplication) doesn't apply here since we define our own __init__ anyway, and keeping the hierarchy shallow makes the code easier to follow and troubleshoot:
    - image, engine, and container_id are set as direct attributes — all container-related state is
  defined in one place
    - root="insights_containers" is passed explicitly to CommandOutputProvider.__init__ to honor the
  existing archive directory convention
    - MRO is now HostBasedContainerCommandProvider → CommandOutputProvider → ContentProvider, cleanly
  separated from the in-container provider branch
    - The effective interface for parsers (.image, .engine, .container_id, line-oriented .content)
  remains the same
  - Added @serializer(HostBasedContainerCommandProvider) and
  @deserializer(HostBasedContainerCommandProvider) - these were missing in the commit you reviewed
  (47532e6) and are required for the archive to actually persist the collected container data
  ---

dipayandutta:
test rpm

dipayandutta:
test rpm

apimplap:
test rpm

dipayandutta:
test rpm

dipayandutta:
test rpm

dipayandutta:
test rpm

dipayandutta:
test rpm

apimplap:
test egg

dipayandutta:
test rpm

dipayandutta:
test rpm

apimplap:
test egg

apimplap:
test egg

apimplap:
test egg

dipayandutta:
test rpm

apimplap:
test egg

apimplap:
test egg

dipayandutta:
test rpm

dipayandutta:
test rpm

apimplap:
test egg

dipayandutta:
test rpm

apimplap:
test egg

dipayandutta:
test rpm

koalakangaroo:
test egg

insights/core/spec_factory.py
```diff
@@ -1409,6 +1464,120 @@ def __call__(self, broker):
...
+
+                # Fourth element is the value to substitute into command
+                # (e.g, merged_dir for rpm --root command)
+                substitution_value = e[3]
+
+                # Any remaining elements are additional args for the command
+                additional_args = e[4:] if len(e) > 4 else ()
+
+                # Build args tuple: (substitution_value, <additional_args>)
+                all_agrs = (substitution_value,) + additional_args if additional_args else (substitution_value,)
```
jholecek-rh: The implementation dedicates a lot of code and documentation text to making a distinction between the first command argument and additional arguments. What is the reason for making the distinction when we merge them into `all_args` eventually anyway? The only reason I can come up with is that at least one command argument has to be provided. I am not sure if `container_foreach_execute` needs to enforce that.  Even if it does, there is an easier and more explicit way to do that, imho.

```suggestion
                # e = (image, engine, container_id, command_args...)
                image, engine, cid, *all_args = e

                if not all_args:
                    raise ValueError("No command arguments")   # stub message, needs more love
```

insights/core/spec_factory.py
```diff
@@ -1409,6 +1464,120 @@ def __call__(self, broker):
...
+                        the_cmd = self.cmd % all_agrs
+                    except (TypeError, ValueError):
+                        # If % substitution fails (e.g, due to %{...} in rpm format),
+                        # fall back to simple string replacement for the first %s only
+                        the_cmd = self.cmd.replace("%s", substitution_value, 1)
+                        # Then handle additional args if any
+                        for arg in additional_args:
+                            the_cmd = the_cmd.replace("%s", str(arg), 1)
+                else:
+                    the_cmd = self.cmd
```
jholecek-rh, line 1530: Why is a command with no placeholder allowed when at least one argument has to be provided (the current implementation would raise `IndexError` if it were not)?

We really need to make our mind whether `container_foreach_execute` should allow commands with no argument placeholders and implement the chosen behavior consistently.

insights/core/spec_factory.py
```diff
@@ -1409,6 +1464,120 @@ def __call__(self, broker):
...
+                if "%s" in self.cmd:
+                    try:
+                        the_cmd = self.cmd % all_agrs
+                    except (TypeError, ValueError):
+                        # If % substitution fails (e.g, due to %{...} in rpm format),
+                        # fall back to simple string replacement for the first %s only
+                        the_cmd = self.cmd.replace("%s", substitution_value, 1)
+                        # Then handle additional args if any
+                        for arg in additional_args:
+                            the_cmd = the_cmd.replace("%s", str(arg), 1)
```
jholecek-rh: Somewhat related to the topic of mandatory arguments/placeholders that I mentioned in another comment, we need to decided the desired behavior if the number of placeholders is different from the number of provided arguments.

The current implementation:

* Allows fewer placeholders than arguments (ignoring extraneous arguments)
* Allows more placeholders than arguments (calling commands with literal `%s` arguments instead of real ones)

The latter is definitely undesirable. It is a data model bug; the datasource created by the `container_foreach_execute` factory should fail with an exception if that happens to increase chances of early discovery and to simplify troubleshooting.

The former is less clear. One could argue that allowing users to use just the first few values from a tuple returned by a more generic provider datasource fosters reusability (of provider datasources). I would argue that if it is a desired feature (it is), users should be able select arbitrary values to use from all the values returned by a generic provider datasource. Since `foreach_execute` works the same 😞, I will not insist on changing this behavior. However, implementing this change would allow us to stop distinguishing between the first three values (image, engine, container ID) and command arguments. All would become container metadata to be used in commands freely.

insights/core/spec_factory.py
```diff
@@ -1409,6 +1464,120 @@ def __call__(self, broker):
...
+                else:
+                    the_cmd = self.cmd
+
+                cmd_exec = self.cmd.split(None, 1)[0]
+                wrapped_cmd = 'sh -c "command -v {0} > /dev/null && {1}"'.format(cmd_exec, the_cmd)
```
jholecek-rh: Could you help me understand this construction?

insights/specs/datasources/container/containers_inspect.py
```diff
@@ -102,3 +105,58 @@ def containers_inspect_data_datasource(broker):
...
+            engine, container_id = item.args
+            image = image_map.get(container_id, '')
+
+            merged_dir = inspect_data.get("GraphDriver", {}).get("Data", {}).get("MergedDir")
+            if not merged_dir:
+                continue
+
+            if not os.path.exists(merged_dir):
+                log.debug("MergedDir path does not exist: %s for container %s", merged_dir, container_id)
+                continue
```
jholecek-rh: IIUC, this check is pointless. The container and its merge path can disappear between this check and a `HostBasedContainerCommandProvider` instance accessing the merge path. The provider class (and the datasource) has to be ready to handle a case when the merged path does not exist. 

I would like to ask you to investigate what happens in such a case and describe the behavior in a response to this comment.

insights/core/spec_factory.py
sabhasin: You're right, the distinction was unnecessary. Simplified the unpacking from:

```
image, engine, cid = e[0], e[1], e[2]
substitution_value = e[3]
additional_args = e[4:] if len(e) > 4 else ()
all_agrs = (substitution_value,) + additional_args if additional_args else (substitution_value,)
to:
image, engine, cid, *all_args = e
```

Updated the docstring accordingly to remove the artificial distinction between `substitution_value` and `optional_args`.


**Your question** ""I am not sure if container_foreach_execute needs to enforce that [at least one argument]. Even if it does, there is an easier and more explicit way to do that.""

Agreed. Removed the explicit `if not all_args: raise ValueError(...)` guard. If the command has `%s` placeholders and no args are provided, Python's `%` operator will raise a `TypeError` naturally. No need to enforce it ourselves.

insights/core/spec_factory.py
sabhasin: Good point on the inconsistency. Resolved by removing the explicit argument count enforcement entirely. The behavior is now consistent `container_foreach_execute` doesn't enforce argument/placeholder counts itself. If `cmd` has `%s` and no args, Python's % raises naturally. If `cmd` has no `%s`, extra args are harmlessly unused (the `image/engine/cid` metadata still flows through). This is consistent with how the parent `foreach_execute` works.

insights/core/spec_factory.py
sabhasin: Agreed, running a command with literal `%s` in it is a silent bug. Added a post-substitution check:
```
if "%s" in the_cmd:
     raise ValueError(
            "Not enough command arguments for placeholders in: %s" % self.cmd
            )
```
This runs after both substitution paths (the `%` operator and the `.replace()` fallback), so unfilled placeholders are always caught. Fewer placeholders than arguments is allowed (extra args are ignored), consistent with `foreach_execute`.

insights/core/spec_factory.py
sabhasin: Good catch. The `sh -c "command -v ... && ..."` wrapping was copied from `container_execute`, where it makes sense because the command runs inside a container via `podman exec`. This needs to verify the binary exists within the container's filesystem before executing it.

But `container_foreach_execute` runs commands on the host, not inside a container. The wrapping was unnecessary here, the parent class `foreach_execute` also passes commands directly without any shell wrapping. So I have removed it and verified all tests pass.

insights/specs/datasources/container/containers_inspect.py
sabhasin: Thank you for reviewing it. The `os.path.exists` check is a TOCTOU race and doesn't provide any real safety. Removed it.

I investigated what happens when the merged path disappears between datasource resolution and command execution. Here's the full chain:

1. `container_merged_dirs` returns tuples like `(image, engine, container_id, merged_dir)`, now without any path existence check.

2. `container_foreach_execute.__call__` iterates these tuples and creates `HostBasedContainerCommandProvider` instances. No command is executed at this point, execution is lazy, deferred until `.content` is accessed.

3. The `parser.invoke()` method `(plugins.py:160)` receives the list of providers and iterates over it, instantiating a parser for each provider.

4. Parser instantiation accesses `.content (spec_factory.py:120)`, which triggers `CommandOutputProvider.load()` -> `ctx.shell_out()` -> `subproc.call()`

5. The command (e.g., `rpm -qa --root /nonexistent/merged`) fails with a non-zero exit code because the rpmdb cannot be found at the given path.

6. Since `keep_rc=False` (the default), `subproc.call()` raises `CalledProcessError`.

7. Back in `parser.invoke()` (`plugins.py:197`), `CalledProcessError` is caught:
```
except CalledProcessError as cpe:
        log.debug(cpe)
        broker.add_exception(self.component, cpe, traceback.format_exc())
        if not self.continue_on_error:
             exception = True
             break
```

8. Since `continue_on_error=True` (the default, `plugins.py:157`), the failure is logged, and the loop continues to the next provider. Only the successfully parsed containers end up in the results list.


So the framework already handles per-provider failures gracefully, a disappearing merged path causes a `CalledProcessError` for that one container, which is caught and logged, while all other containers proceed normally. The `os.path.exists` pre-check added no real protection over what the framework already provides.

insights/core/spec_factory.py
```diff
@@ -1409,6 +1464,101 @@ def __call__(self, broker):
...
+    def __call__(self, broker):
+        result = []
+        source = broker[self.provider]
+        cleaner = broker.get('cleaner')
+        ctx = broker[self.context]
+
+        if isinstance(source, ContentProvider):
+            source = source.content
+        if not isinstance(source, (list, set)):
+            source = [source]
```
qinpingli, line 1512: Per following docstring:
```
provider (list): A datasource that returns a list of tuples with format (image, engine, container_id,
                        command_args...)
```

Why line 1507~1508 is needed?

And docstring for provder, is datasource needed? Based it's a public interface, we may should not introduce datasource here?

insights/core/spec_factory.py
qinpingli: Hmm, the new code doesn't look not robust enough for me.

The mainly reason is: Is the placeholder `%s` reliable？ Whether there are some commands that contain `xx%sxx` string?

insights/core/spec_factory.py
```diff
@@ -1409,6 +1464,101 @@ def __call__(self, broker):
...
+                    inherit_env=self.inherit_env,
+                    override_env=self.override_env,
+                    signum=self.signum,
+                    cleaner=cleaner,
+                )
+
+                result.append(ccp)
+
+            except NoFilterException as nfe:
+                raise nfe
```
qinpingli, line 1552: IIUC, here when `NoFilterException` is raised, we will stop the check for all containers. Why there is different process for other exceptions raised in `HostBasedContainerCommandProvider`, for example:
- ContentException
- BlacklistedSpec

insights/core/spec_factory.py
```diff
@@ -1751,3 +1901,37 @@ def deserialize_container_command(_type, data, root, ctx, ds):
     res.engine = data["engine"]
     res.container_id = data["container_id"]
     return res
+
+
+@serializer(HostBasedContainerCommandProvider)
+def serialize_oob_container_command(obj, root):
```
qinpingli: Update the method name from `serialize_oob_container_command` to `serialize_hb_container_command`?

insights/core/spec_factory.py
```diff
@@ -1751,3 +1901,37 @@ def deserialize_container_command(_type, data, root, ctx, ds):
...
+        "save_as": bool(obj.save_as),
+        "relative_path": rel,
+        "image": obj.image,
+        "engine": obj.engine,
+        "container_id": obj.container_id,
+    }
+
+
+@deserializer(HostBasedContainerCommandProvider)
+def desdeserialize_oob_container_command(_type, data, root, ctx, ds):
```
qinpingli: Update the method name from `desdeserialize_oob_container_command` to `desdeserialize_hb_container_command`?

insights/specs/datasources/container/containers_inspect.py
```diff
@@ -102,3 +105,54 @@ def containers_inspect_data_datasource(broker):
...
+
+
+@datasource(LocalSpecs.containers_inspect_data_raw, running_rhel_containers, HostContext)
+def container_merged_dirs(broker):
+    """
+    This datasource extracts the MergedDir path from container inspect output for containers using overlay
+    storage driver.
+
+    The MergedDir is the unified view of all container layers and is used to run host-side RPM queries without
+    entering the container, eliminating container resource consumption.
```
qinpingli, line 117: Nit: IIUC, this datasource is not only used for run RPM queries, right?

How about updating it to:
```
The MergedDir provides a unified view of all container layers. It enables host-side queries without entering the container, eliminating container resource consumption.
```

insights/specs/datasources/container/containers_inspect.py
```diff
@@ -102,3 +105,54 @@ def containers_inspect_data_datasource(broker):
...
+              where merged_dir is the path to the container's overlay merged directory.
+
+    Raises:
+        SkipComponent: When no containers have valid overlay MergedDir or any exception occurs.
+    """
+    image_map = {}
+    for container in broker[running_rhel_containers]:
+        image, engine, container_id = container
+        image_map[container_id] = image
+
```
qinpingli: Hmm, why `running_rhel_containers` is needed here? IIUC, we can get image information from `podman/docker inspect <container_id>`.

insights/specs/datasources/container/containers_inspect.py
```diff
@@ -102,3 +105,54 @@ def containers_inspect_data_datasource(broker):
     except Exception as e:
         raise ContentException("Unexpected content exception:{e}".format(e=str(e)))
     raise SkipComponent
+
+
+@datasource(LocalSpecs.containers_inspect_data_raw, running_rhel_containers, HostContext)
+def container_merged_dirs(broker):
```
qinpingli, line 111: Could add the new datasource into documentation file: docs/custom_datasources_index.rst?

insights/tests/core/spec_factory/test_host_based_container.py
```diff
@@ -0,0 +1,244 @@
+import os
```
qinpingli: How about updating the test file name to `test_host_based_container_command_provider.py`?

insights/tests/core/spec_factory/test_host_based_container.py
```diff
@@ -0,0 +1,244 @@
...
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_host_provider_repr(mock_which, mock_bl):
+    provider = HostBasedContainerCommandProvider(
+        SAMPLE_CMD, _make_ctx(),
+        image=SAMPLE_IMAGE, engine=SAMPLE_ENGINE, container_id=SAMPLE_CID,
+    )
+    r = repr(provider)
+    assert r.startswith('HostBasedContainerCommandProvider(')
+    assert SAMPLE_CMD in r
```
qinpingli: How about updating
```
    r = repr(provider)
    assert r.startswith('HostBasedContainerCommandProvider(')
    assert SAMPLE_CMD in r
```
to
```
assert repr(provider) == f"HostBasedContainerCommandProvider('{SAMPLE_CMD}')"

File: insights/tests/core/spec_factory/test_host_based_container.py (hunk 20 above)
Comment (qinpingli) [reply insights/tests/core/spec_factory/test_host_based_container.py]: Hmm, seems this file also contains the test code for container_foreach_execute. Is it possible to track the test for container_foreach_execute in another file?

File: insights/tests/core/spec_factory/test_container_foreach_execute.py [hunk 22]
```diff
@@ -0,0 +1,244 @@
...
+        cfe.provider: source,
+        cfe.context: ctx,
+    }[key]
+    broker.get = lambda key, default=None: None
+    return broker
+
+
+@patch(MOCK_BLACKLIST, return_value=True)
+@patch(MOCK_WHICH, return_value="/usr/bin/rpm")
+def test_cfe_simple_substitution(mock_which, mock_bl):
```
qinpingli, line 93: Hmm, seems the following test cases test `_make_cfe` not `container_foreach_execture`?

insights/core/spec_factory.py
sabhasin: The initial `.replace()` approach had issues:

```
  # PROBLEMATIC APPROACH:
  try:
      the_cmd = self.cmd % tuple(all_args)
  except (TypeError, ValueError):
      # Fallback to replace - NOT ROBUST
      the_cmd = self.cmd
      for arg in all_args:
          the_cmd = the_cmd.replace("%s", str(arg), 1)
```

 Problems with `.replace()` fallback:
 
  - Not escape-aware: would incorrectly match `%s` inside `%%s` or other escaped contexts
  - Could corrupt command strings that legitimately contain `%s` as text
  - The root cause isn't addressed: why did the `%` operator fail?



Robust Solution Implemented:

The real issue is that Python's `%` operator treats RPM format strings like `%{NAME}` as malformed placeholders. The robust fix is to temporarily escape those patterns before substitution:

 ```
 if "%s" in self.cmd:
      # Temporarily escape %{...} patterns (e.g., RPM format strings like %{NAME})
      # so Python's % operator treats them as literals instead of format placeholders
      cmd_escaped = re.sub(r'%\{', r'%%{', self.cmd)
      try:
          the_cmd = cmd_escaped % tuple(all_args)
      except (TypeError, ValueError) as e:
          # TypeError: not enough args for placeholders
          # ValueError: incomplete format or other % operator issues
          raise ValueError(
              "Not enough command arguments for placeholders in: %s" % self.cmd
          )
  else:
      the_cmd = self.cmd
```


Why this is robust:

  1. Addresses the root cause: Escapes `%{NAME}` -> `%%{NAME}` so Python's `%` operator doesn't choke on RPM format strings
  2. Preserves the format: After substitution, `%%{NAME}` -> `%{NAME}` in the final command
  3. No fallback needed: The `%` operator now works reliably for all cases
  4. Clear error handling: If substitution fails (wrong number of args), we raise a clear `ValueError` instead of silently continuing
  5. No accidental matches: Uses Python's built-in `%` operator which properly handles escaping

File: insights/core/spec_factory.py (hunk 13 above)
Comment (sabhasin) [reply insights/core/spec_factory.py:1512]: Changes Made:

  1. Removed misleading type hint `(list)` from `provider` parameter
     - Before: `provider (list): A datasource that returns a list...`
     - After: `provider`: `A component that returns a list...`
     - Why: `provider` is a component class, not a list. The defensive code at lines 1507-1510 is needed because broker[self.provider] can return various
       types at runtime.
  2. Replaced framework jargon with user-friendly terminology
     - Before: "datasource" (internal framework term)
     - After: "component" (general, user-friendly term)
     - Why: Public APIs should avoid exposing internal implementation terminology
  3. Updated all references consistently
     - Class docstring: "provider datasource" -> "provider component"
     - Args description: "A datasource that returns" -> "A component that returns"
     - Returns description: "A datasource that returns" -> "A component that returns"


The Defensive Code Remains:

  Lines 1507-1510 are necessary and unchanged because at runtime `broker[self.provider]` can be:
  - A `ContentProvider` object -> needs `.content` extracted
  - A single tuple -> needs wrapping in `[tuple]`
  - A list or set -> ready to use
  
The docstring now accurately reflects the actual behavior.
  
The `(list)` type hint was misleading and removed, and the term "datasource" was replaced with "component" throughout to avoid exposing internal framework terminology in the public API.

File: insights/core/spec_factory.py (hunk 14 above)
Comment (sabhasin) [reply insights/core/spec_factory.py:1552]: You're right to question why `NoFilterException` gets special treatment.

Updated the code to explicitly handle `ContentException` like the parent class does:
```
  except NoFilterException as nfe:
      raise nfe
  except ContentException as ce:
      log.debug(ce)
  except Exception:
      log.debug(traceback.format_exc())
```

  - `NoFilterException`: System-wide configuration issue → fail fast for all containers
      - If a filterable spec has no filters defined, it's a global problem that should stop the entire operation rather than silently skipping all containers
  - `ContentException`, `BlacklistedSpec`: Per-container issues -> log and continue to next container
      - `ContentException`: Command doesn't exist in that container's environment
      - `BlacklistedSpec`: Command is blacklisted for that specific container
      - One container failing shouldn't prevent others from being processed

The explicit ContentException handler makes this intent clearer than relying on the generic `Exception` catch-all. This now matches the parent `foreach_execute` pattern for consistency.

File: insights/core/spec_factory.py (hunk 15 above)
Comment (sabhasin) [reply insights/core/spec_factory.py]: Updated both the serializer

File: insights/core/spec_factory.py (hunk 16 above)
Comment (sabhasin) [reply insights/core/spec_factory.py]: Updated both the deserializer

File: insights/specs/datasources/container/containers_inspect.py (hunk 17 above)
Comment (sabhasin) [reply insights/specs/datasources/container/containers_inspect.py:117]: Updated the docstring as suggested.

File: insights/specs/datasources/container/containers_inspect.py (hunk 18 above)
Comment (sabhasin) [reply insights/specs/datasources/container/containers_inspect.py]: Removed the unnecessary dependency:
  - `running_rhel_containers` dependency removed from `container_merged_dirs`
  - `mage_map` building loop removed
  - Image now extracted directly: `image = inspect_data.get("ImageName", "")`
  
This is simpler and more efficient. The `LocalSpecs.containers_inspect_data_raw` already depends on `running_rhel_containers_id` (which depends on `running_rhel_containers`), so the containers are already filtered to RHEL containers through that chain. 

Updated test data to include the `ImageName` field to match real podman/docker inspect output. All tests pass.

File: insights/specs/datasources/container/containers_inspect.py (hunk 19 above)
Comment (sabhasin) [reply insights/specs/datasources/container/containers_inspect.py:111]: Added `container_merged_dirs` to the documentation.

File: insights/tests/core/spec_factory/test_host_based_container.py (hunk 20 above)
Comment (sabhasin) [reply insights/tests/core/spec_factory/test_host_based_container.py]: Split the tests into two separate files for better organization:

1. `test_host_based_container_command_provider.py` - 7 tests for `HostBasedContainerCommandProvider` class
2. `test_container_foreach_execute.py` - 8 tests for `container_foreach_execute` class

All spec_factory tests pass.

File: insights/tests/core/spec_factory/test_host_based_container.py (hunk 21 above)
Comment (sabhasin) [reply insights/tests/core/spec_factory/test_host_based_container.py]: Updated both the repr implementation and the test for better clarity:

**repr implementation:** Changed from `'HostBasedContainerCommandProvider("%r")' % self.cmd` to `f"HostBasedContainerCommandProvider('{self.cmd}')"`

**test assertion:** Changed to exact match: `assert repr(provider) == f"HostBasedContainerCommandProvider('{SAMPLE_CMD}')"`

All tests pass.

File: insights/tests/core/spec_factory/test_container_foreach_execute.py (hunk 22 above)
Comment (sabhasin) [reply insights/tests/core/spec_factory/test_container_foreach_execute.py:93]: You're absolutely right to question whether `_make_cfe()` tests the actual class or just a mock. The helper uses `object.__new__()` which bypasses `__init__`, so these tests focus on the `__call__` method in isolation rather than the full class lifecycle.

However, this is intentional for good reasons:

1. Avoids global registry pollution
     - The real constructor calls `__init__` (inherited from `foreach_execute`), which registers each instance in the global `dr.DELEGATES` registry via the `@datasource` decorator
     - Using the real constructor would pollute the registry: 0→1→2→3... with each test
     - This breaks test isolation and can cause state to leak between tests

2. These are unit tests, not integration tests
     - They specifically test the `__call__` method logic (tuple unpacking, command substitution, provider creation, exception handling)
     - `__init__` is inherited from the parent `foreach_execute` class and already tested elsewhere
     - The `@datasource` decorator registration is framework behavior, not what we're testing here
 
3. The alternative would require complex cleanup
```
  # Would need cleanup after every test:
  def test_something():
      cfe = container_foreach_execute(provider, cmd, context=HostContext)
      # ... test ...
      # Cleanup: remove from dr.DELEGATES, dr.COMPONENTS, etc.

```
What is being tested:
  - Command substitution with `%s` placeholders
  - RPM format string handling (`%{NAME}`)
  - Container metadata preservation (image, engine, container_id)
  - Multiple container handling
  - Exception handling (ContentException, empty source)
  - Provider creation (HostBasedContainerCommandProvider)
  
What is not tested:
  - `__init__` behavior (inherited, tested in parent class tests)
  - `@datasource` registration (framework behavior)
  - Dependency resolution (framework behavior)
  
Conclusion: The `object.__new__()` approach is the correct pattern for these unit tests. It provides clean test isolation while thoroughly testing the actual business logic in `__call__`.

File: insights/core/spec_factory.py (hunk 13 above)
Comment (qinpingli) [reply insights/core/spec_factory.py:1512]: Thanks for the update!

File: insights/specs/datasources/container/containers_inspect.py (hunk 19 above)
Comment (qinpingli) [reply insights/specs/datasources/container/containers_inspect.py:111]: Thanks for the update!

File: insights/core/spec_factory.py (hunk 14 above)
Comment (qinpingli) [reply insights/core/spec_factory.py:1552]: > ContentException: Command doesn't exist in that container's environment

Hmm, IIUC, for `container_forech_execute` all the commands are the host command, for example for `container_installed_rpms` spec, command doesn't exist means on the host `/usr/bin/rpm` command doesn't exist.

File: insights/specs/datasources/container/containers_inspect.py (hunk 18 above)
Comment (qinpingli) [reply insights/specs/datasources/container/containers_inspect.py]: Thanks for the update!

File: insights/tests/core/spec_factory/test_host_based_container.py (hunk 20 above)
Comment (qinpingli) [reply insights/tests/core/spec_factory/test_host_based_container.py]: Thanks for the update!

File: insights/tests/core/spec_factory/test_host_based_container.py (hunk 21 above)
Comment (qinpingli) [reply insights/tests/core/spec_factory/test_host_based_container.py]: Thanks for the update!

File: insights/tests/core/spec_factory/test_container_foreach_execute.py (hunk 22 above)
Comment (qinpingli) [reply insights/tests/core/spec_factory/test_container_foreach_execute.py:93]: Thanks for the explain! `container_foreach_execute` as a new class, only test `__call__` looks not enough, for example: if the code is changed, it's parent class is not `foreach_execute` anymore, the current test can not find this change.

File: insights/core/spec_factory.py (hunk 14 above)
Comment (sabhasin) [reply insights/core/spec_factory.py:1552]: For `container_foreach_execute`, all commands execute on the HOST (not inside containers), so the exception handling rationale is different:

- `NoFilterException`: System-wide configuration issue → fail fast for all containers
  - If a filterable spec has no filters defined, it's a global problem that should stop the entire operation
- `ContentException`: Typically means the command doesn't exist on the host system → log and continue
  - Example: `/usr/bin/rpm` is not installed on the host
  - Since the command runs on the host (not in the container), if it's missing, it will fail for ALL containers
  - However, we still log and continue because:
    i. The provider might have already created some successful results before hitting this error.
    ii. Consistent with the parent `foreach_execute` pattern.
    iii. The final check `if result: return result` ensures we raise `ContentException` if NO containers were processed successfully.
- `BlacklistedSpec` and other exceptions: Per-processing-attempt issues → log and continue
  - Individual processing failures shouldn't stop iteration through remaining containers.
  
Unlike `container_execute` (which runs commands INSIDE containers), `container_foreach_execute` runs host-based commands using container data (like merged directories). The "per-container" aspect refers to iterating through container data, not container-specific command environments.

The explicit `ContentException` handler matches the parent class pattern for consistency across the codebase.

File: insights/tests/core/spec_factory/test_container_foreach_execute.py (hunk 22 above)
Comment (sabhasin) [reply insights/tests/core/spec_factory/test_container_foreach_execute.py:93]: Thanks for pointing it out.  Added an inheritance test to verify the class hierarchy. 

Now we have both:
1. Inheritance test: Catches if the parent class changes.
2. `__call__` unit tests (with `object.__new__()`): Test method logic without registry pollution.

Test results: All 39 spec_factory tests pass (including the new inheritance test).

File: insights/core/spec_factory.py (hunk 14 above)
Comment (qinpingli) [reply insights/core/spec_factory.py:1552]: Okay, do you mean in the data collection process, the runtime might be changed? And you want to collect the data as much as possible?

It's okay for me.

Comment (sabhasin) [reply insights/core/spec_factory.py:1552]: Yes, exactly. We want to collect data from as many containers as possible. If processing one container fails (e.g., `ContentException` on container #2), we continue to containers #3, #4, etc., to maximize data collection.

Thank you for reviewing and approving this approach!

File: insights/tests/datasources/container/test_containers_inspect.py [hunk 23]
```diff
@@ -1060,3 +1061,258 @@ def test_containers_inspect_datasource_NG_output_2():
...
+def test_container_merged_dirs_no_mergeddir():
+    """Test that containers without MergedDir field are skipped"""
+    inspect_data = Mock()
+    inspect_data.content = INSPECT_MERGED_NO_MERGEDDIR.splitlines()
+    inspect_data.cmd = "/usr/bin/podman inspect 38fb57be8bb2"
+    inspect_data.args = ("podman", "38fb57be8bb2")
+
+    broker = {
+        LocalSpecs.containers_inspect_data_raw: [inspect_data],
+        running_rhel_containers: CONTAINERS_FOR_MERGED,
```
qinpingli: Seems `running_rhel_containers` is not removed in tests.

insights/tests/core/spec_factory/test_container_foreach_execute.py
```diff
@@ -0,0 +1,197 @@
...
+def mock_single_container(broker):
+    return [(SAMPLE_IMAGE, SAMPLE_ENGINE, SAMPLE_CID, SAMPLE_MERGED_DIR)]
+
+
+@datasource(HostContext)
+def mock_multiple_containers(broker):
+    return [
+        ("image1", "podman", "cid111111111", "/merged/1"),
+        ("image2", "podman", "cid222222222", "/merged/2"),
+        ("image3", "docker", "cid333333333", "/merged/3"),
```
qinpingli: How about create a constant for multiple containers data, so we don't need to define them again in `test_cfe_multiple_containers`.

insights/tests/core/spec_factory/test_container_foreach_execute.py
```diff
@@ -0,0 +1,197 @@
...
+    entry = (SAMPLE_IMAGE, SAMPLE_ENGINE, SAMPLE_CID, SAMPLE_MERGED_DIR)
+    assert result[0].args == entry
+
+
+def test_cfe_raises_content_exception_on_empty_source():
+    """Test that ContentException is raised when no containers are provided."""
+    broker = _run(container_rpms_empty)
+
+    # When empty source, the spec doesn't get added to broker
+    # (ContentException is logged and spec is skipped)
```
qinpingli: Nit: for the comment, IIUC, the exception is not log by `log.debug(ce)`?

insights/tests/core/spec_factory/test_container_foreach_execute.py
```diff
@@ -0,0 +1,197 @@
+from unittest.mock import patch
```
qinpingli, line 1: Lines 1508, 1510, 1530, 1552, 1554 are not tested, could you try to test them?

insights/core/spec_factory.py
qinpingli: Hmm, just a thought:
how about having a specific placeholder for this spec factory, for example `%container_paths`, `%container_dirs`?

With this, seems we can make `container_foreach_execute` clearer. For example, it can show this spec factory is running command against container_paths.

[Questions]:
1. For the current container related specs, I didn't find multiple args spec (only container path), is there multiple args spec requirement in the future?
2. With this, maybe we can introduce a specific data structure for container_paths, is it safer?

@jholecek-rh @sabhasin WDYT?

sabhasin: Current state:
- Only one usage in production: `container_installed_rpms` with a single `merged_dir` path
- No current plans for multiple args (the tuple has `image, engine, container_id, merged_dir` but we only substitute the path).

Tradeoff:
- Generic `%s`: More flexible (if we need multiple paths later), consistent with `foreach_execute` parent class, but less clear.
- Specific placeholder: Self-documenting, safer, but requires custom implementation and differs from parent pattern


Since we have only one use case and no known future requirement for multiple args, I'm open to using a specific placeholder like `%{merged_dir}` for clarity. However, this would be a breaking API change.

Given that `container_foreach_execute` is newly introduced (this PR), now is the right time to make this decision.

@jholecek-rh - Do you have input on whether we should use specific placeholders vs generic %s? 

@qinpingli @koalakangaroo Any known future requirements I'm missing?

insights/tests/datasources/container/test_containers_inspect.py
sabhasin: Thank you for highlighting. I have removed this.

insights/tests/core/spec_factory/test_container_foreach_execute.py
sabhasin: Thanks for highlighting, as it will eliminate code duplication. I have made the changes.

insights/tests/core/spec_factory/test_container_foreach_execute.py
sabhasin: Earlier, I misunderstood the flow. Looking at the code:
- For an empty source, `ContentException` is raised at line 1560 (outside the try/except block)
- It's caught by the framework (`dr.run()`), not logged within the function
- The `log.debug(ce)` at line 1554 only logs ContentExceptions that occur inside the loop (for individual containers with problems)

Fixed the comment:
- When empty source, a ContentException is raised and caught by the framework, so the spec doesn't get added to the broker.

insights/tests/core/spec_factory/test_container_foreach_execute.py
sabhasin: I've added comprehensive tests for all those lines.

New tests added:

1. `test_cfe_contentprovider_source` - Tests line 1508
   - When datasource returns a `ContentProvider`, it extracts `.content`
2. `test_cfe_single_tuple_source` - Tests line 1510
   - When source is a single tuple (not list/set), it wraps it in a list
3. `test_cfe_command_without_placeholder` - Tests line 1530
   - When command has no `%s` placeholder, uses command directly (no substitution)
4. `test_cfe_contentexception_inside_loop` - Tests line 1554
   - When `ContentException` is raised inside the loop, it's logged via `log.debug(ce)`
   - Uses mixed valid/invalid data to trigger exception for one container
5. `test_cfe_nofilterexception_propagates` - Tests line 1552
   - When `NoFilterException` is raised, it's re-raised (not caught)
   - Verifies spec is not in broker when exception propagates

apimplap:
test egg

apimplap:
test egg

insights/core/spec_factory.py
qinpingli: If we still want to use `%s`, IMO, here using the same way `self.cmd %e` as foreach_execute might be the best choice for me.

sabhasin: Removed framework escaping logic:

 Simplified __call__ method:
`image, engine, cid = e[0], e[1], e[2]`
`cmd_args = e[3:]`
`the_cmd = self.cmd % tuple(cmd_args)`

Escaping now handled by spec:
- Updated rpm.py: `%{NAME}` --> `%%{NAME}`
- Updated docstring to document escaping requirement
- Updated tests with pre-escaped format strings

apimplap:
test dp

PR #4805, opened 2026-08-18
feat(conftest): Add ALLOW_TYPE_CLASS doctest flag for Py2/Py3 type repr

qinpingli:
Normalize <type 'name'> to <class 'name'> when comparing doctest output so type() examples work when doctests are cherry-picked between Python 2 and 3.

Co-authored-by: Cursor <cursoragent@cursor.com>

rh-pre-commit.version: 2.3.2

### All Pull Requests:

Check all that apply:

* [X] No Sensitive Data in this change?

*Add your description here*

## Summary by Sourcery

Enable doctests to compare Python 2 and Python 3 type representations consistently.

New Features:
- Add cross-version doctest support for treating Python 2 and Python 3 type representations as equivalent.

Enhancements:
- Configure project doctests to use the enhanced checker and runner with the new comparison option enabled by default.

Tests:
- Add coverage for compatible and incompatible type representations and doctest.testmod integration.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,8 +1,13 @@
+import doctest
 import logging
 import pytest
 import warnings
 from insights.tests import run_test
+from insights.tests.doctest_support import INSIGHTS_DOCTEST_OPTIONFLAGS, InsightsDocTestRunner
 warnings.simplefilter('always', DeprecationWarning)
+
+# testmod() constructs DocTestRunner internally; replace the class for all doctest.testmod() calls.
+doctest.DocTestRunner = InsightsDocTestRunner
 
 
 def pytest_addoption(parser):
@@ -14,6 +19,8 @@ def pytest_addoption(parser):
 def pytest_configure(config):
     level = logging.DEBUG if config.getoption("--appdebug") else logging.ERROR
     logging.basicConfig(level=level)
+    optionflags = getattr(config.option, 'doctest_optionflags', 0)
+    config.option.doctest_optionflags = optionflags | INSIGHTS_DOCTEST_OPTIONFLAGS
 
 
 @pytest.fixture
--- a/insights/tests/doctest_support.py
+++ b/insights/tests/doctest_support.py
@@ -1,0 +1,45 @@
+"""
+Py2/Py3 compatible doctest checker and runner for parser documentation tests.
+"""
+
+import doctest
+import re
+
+DOCTEST_ALLOW_TYPE_CLASS = doctest.register_optionflag('ALLOW_TYPE_CLASS')
+INSIGHTS_DOCTEST_OPTIONFLAGS = DOCTEST_ALLOW_TYPE_CLASS
+
+_TYPE_CLASS_RE = re.compile(r"<type '([^']+)'>")
+
+
+def _normalize_type_class_repr(text):
+    return _TYPE_CLASS_RE.sub(r"<class '\1'>", text)
+
+
+_StandardOutputChecker = doctest.OutputChecker
+
+
+class InsightsDoctestChecker(_StandardOutputChecker):
+    def check_output(self, want, got, optionflags):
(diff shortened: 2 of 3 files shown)


qinpingli:
@jholecek-rh could you help review? Thanks!

jholecek-rh:
@qinpingli 

**Proposed Solution**

As discussed in person, I am not thrilled about this solution.

Pros:

- Minimal code changes.

Cons:

- Global patch of  `doctest.DocTestRunner` could cause difficult-to-troubleshoot issues if we ever start using `doctest` in other areas of the codebase.
- Hidden implementation. If a developer tries to run the test manually, they will not be able to explain why the test passes. The actual return value will be different from the value in the docstring. Neither the parser code nor the test code will indicate that there is some hidden logic that transforms the values. The code that does the transformations gets loaded only when the tests are executed using `pytest`.

Overall, I think that this solution increases the already massive technical debt of the insights-core codebase. :( 

**How doctest Test Work Now**

All parsers that want to use a doctest test (some parsers don't, it is probably a gap) need to define a test manually. All these tests use the same template that looks like this:

```python
def test_doc_examples():
    # define the environment; this is the key part
    env = {
        'audit_rules': AuditRules(context_wrap(AUDIT_RULES_OUTPUT2)),
        'auds': AuditStatus(context_wrap(NORMAL_AUDS_RHEL7))
    }
    # boilerplate that is the same in doctest tests for (almost) all parsers
    failed, total = doctest.testmod(auditctl, globs=env)
    assert failed == 0
```

**Alternative 1**

We could introduce our own module, e.g. `insights.test.doctest` with a `testmod()` function. For Python 3, it would just explain why we need it and import `doctest.testmod`. For Python 2.7, it would provide a custom `testmod()` that would do the required transformations.

Pros:

- Tests stay the same, only the import statement changes.
- Test code points to the custom `insights.tests.doctest` module that documents/explains what is going on.
- No global modifications of third-party objects.

Cons:

- About 400 import statements need to be updated.

**Alternative 2**

We could eliminate the boilerplate and centralize what the doctest test does with a little bit of decorator magic:

```python
from insigths.tests.doctest import parser_doctest
@parser_doctest
def test_doc_examples():
    return {
        'audit_rules': AuditRules(context_wrap(AUDIT_RULES_OUTPUT2)),
        'auds': AuditStatus(context_wrap(NORMAL_AUDS_RHEL7))
    }
```

Pros:

- Test code points to the definition of the actual test which documents/explains the automatic transformations.
- No global modifications of third-party objects.
- Eliminates boilerplate code duplication. The test logic is managed centrally.
- Provides a structured way to enforce that all parsers need to have a doctest test.

Cons:

- Slightly bigger changes to about 400 existing doctest tests (change import statement, add decorator, remove boilerplate code).

**My Preferences**

I obviously like Alternative 2 the most, but I understand that it could be seen as scope creep. I think that Alternative 1 would still be a better solution than the current proposal with no real downsides (the import statements can be changed by one string substitution command).

jholecek-rh:
@qinpingli Are you aware of the code in `insights/tests/parser/__init__.py`? 

* The `Py23DocChecker` class was probably used to achieve something similar in the past. It does not seem to be used in `insights-core` and it seems unlikely that any downstream application would still be using it. Perhaps we can deprecate and eventually remove the class?
* The `ic_testmod` is used in one test, `insights/tests/parsers/test_checkin_conf.py`. I did not try to reverse-engineer the requirements that led to its creation. At the first glance, it seems as a use case for Alternative 2 from my previous comment where the lack of a central test definition and unwillingness to update existing tests led to this custom function and code fragmentation.
  *  Note that the function uses either `DebugRunner` or `DocTestRunner` to run the test depending on external configuration. I suspect that your solution would not work when `DebugRunner` were used.

Could you clarify how this code relates to the proposed change, please?

PR #4807, opened 2026-08-20
fix(parsers): support platform_firmware_information in dmidecode

amepatil-git:
### All Pull Requests:

Check all that apply:

dmidecode 3.6+ renamed "BIOS Information" to "Platform Firmware Information", causing KeyError and missing properties on newer systems.

~~~
- Support both 'platform_firmware_information' and 'bios_information' in properties.
- Add forward/backward compatibility of both keys access in parse_content.
- Add platform_firmware, platform_firmware_vendor, and platform_firmware_date properties.
- Update unit tests with dmidecode 3.6+ test case.

rh-pre-commit.version: 2.4.0
~~~

*Add your description here*

Due to this the downstream `extraction_rules` had failure when it attempted to access the keys inside property `bios` due to which it resulted in exception.
This change will help access `bios` or `platform_firmware` keys interchangeably and will not cause exception for accessing any of  its properties. If the section for `bios` or `platform_firmware` and key in them does not exist, it will return `None`

## Summary by Sourcery

Support renamed platform firmware sections in dmidecode while preserving compatible BIOS data access.

New Features:
- Add platform firmware aliases for BIOS information, vendor, and release date properties.

Bug Fixes:
- Maintain BIOS data access across dmidecode versions that use either BIOS Information or Platform Firmware Information, preventing missing-key failures.

Enhancements:
- Return None safely when firmware sections or requested properties are unavailable.

Documentation:
- Update dmidecode parser documentation and examples for platform firmware naming.

Tests:
- Add coverage for dmidecode 3.6+ output and backward- and forward-compatible BIOS/platform firmware access.

### DIFF
--- a/insights/parsers/dmidecode.py
+++ b/insights/parsers/dmidecode.py
@@ -23,9 +23,13 @@ The common information retrieved from dmidecode is available in several
 convenience properties:
 
 * **system_info** - Information about the machine itself
-* **bios** - the BIOS information
-* **bios_vendor** - the BIOS's 'Vendor' attribute
-* **bios_date** - the BIOS's 'Release Date' attribute
+* **bios** - the BIOS / Platform Firmware information
+* **bios_vendor** - the BIOS's / Platform Firmware's 'Vendor' attribute
+* **bios_date** - the BIOS's / Platform Firmware's 'Release Date' attribute
+* **platform_firmware** - alias of ``bios``, for dmidecode 3.6+ where the
+  'BIOS Information' section was renamed to 'Platform Firmware Information'
+* **platform_firmware_vendor** - alias of ``bios_vendor``
+* **platform_firmware_date** - alias of ``bios_date``
 * **processor_manufacturer** - the processor's 'Manufacturer' attribute
 * **is_present** - this indicates whether dmidecode information was found.
 
@@ -84,7 +88,11 @@ Examples:
     1
     >>> dmi['bios_information'][0]['vendor']
     'HP'
+    >>> dmi['platform_firmware_information'][0]['vendor']
+    'HP'
     >>> dmi.bios_vendor
+    'HP'
+    >>> dmi.platform_firmware_vendor
     'HP'
     >>> dmi.bios_date
     datetime.date(2008, 9, 27)
@@ -113,6 +121,17 @@ class DMIDecode(CommandParser, LegacyItemAccess):
     def parse_content(self, content):
         self.data = parse_dmidecode(content, pythonic_keys=True)
 
+        # If bios_info exists one of the keys was present in parsed data,
+        # assign its value to the missing key for access with both keys.
+        bios_info = self._get_bios_info()
+        if bios_info:
+            self.data.setdefault("platform_firmware_information", bios_info)
+            self.data.setdefault("bios_information", bios_info)
+
+    def _get_bios_info(self):
+        """(list): Helper to fetch the BIOS or Platform Firmware Information block list."""
+        return self.data.get("platform_firmware_information") or self.data.get("bios_information")
+
     @property
     def system_info(self):
         """(str): Convenience method to get system information"""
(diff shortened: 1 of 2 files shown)


amepatil-git:
Added Sign-off since the greeting test had failed.

insights/parsers/dmidecode.py
```diff
@@ -112,6 +120,17 @@ class DMIDecode(CommandParser, LegacyItemAccess):
...
+        if (
+            "platform_firmware_information" in self.data
+            and "bios_information" not in self.data
+        ):
+            self.data["bios_information"] = self.data["platform_firmware_information"]
+        elif (
+            "bios_information" in self.data
+            and "platform_firmware_information" not in self.data
+        ):
+            self.data["platform_firmware_information"] = self.data["bios_information"]
```
qinpingli: If `platform_firmware_information` and `bios_information` both exist, what will happen?

insights/parsers/dmidecode.py
```diff
@@ -127,21 +146,45 @@ def system_uuid(self):
 
     @property
     def bios(self):
-        """(str): Convenience method to get BIOS information"""
-        return self["bios_information"][0] if "bios_information" in self else None
+        """(dict): Convenience method to get BIOS / Platform Firmware information"""
+        for key in ("platform_firmware_information", "bios_information"):
```
qinpingli: Since we already added the logic in self.data, how about using self.data here?

insights/parsers/dmidecode.py
amepatil-git: When we have both `platform_firmware_information` and `bios_information`  as keys in  `self.data`,  it should retain their original values and not assign each other their values.

However ideally I would expect only one of them is present. Since the `platform_firmware_information` was only added in rpm versions dmidecode-3.6-2.el9.x86_64 on RHEL 9.8 onward, dmidecode-3.6-5.el10.x86_64 on RHEL 10.2 onward and that field replaced the `bios_information` so they going forward the `bios_information` field will be absent, but since we are using properties  (and downstream rules) which use the `bios_information` I have added this check here where if any one of the key is present it also assigns value to other key so that both old and new properties can access this information.

Please let me know what you think and if any change is needed.

insights/parsers/dmidecode.py
amepatil-git: Yes I had initially used for to check for both the key one by one for `platform_firmware_information` and `bios_information` in the self, since this was the same loop I added in extraction_rules.  So I copied the same idea here. However since here are already using assigning the values of key platform_firmware_information` and `bios_information` to each other in `parse_content()` I think we can skip this for loop. The sourcery-ai  bot also suggested me this same and I am thinking that this check is redundant and we could remove this and I could just do something like.

 ```python
return self["bios_information"][0] if "bios_information" in self and self["bios_information"] else None
```

Here instead of `self.data["bios_information"]` , I am using `self["bios_information"]` since previous code was also using  `self["bios_information"][0]` in this property so I was reusing that.

Do let me know what you think and if you think we should remove this for loop and just use this `return` statement or we could keep both as fallback just to be safe ? Let me know what you think.

File: insights/parsers/dmidecode.py (hunk 1 above)
Comment (qinpingli) [reply insights/parsers/dmidecode.py]: But from the docstr, seems `platform_firmware_information` always is same with `bios_information`.  And it's also related to the next comment. If as you said `When we have both platform_firmware_information and bios_information as keys in self.data, it should retain their original values and not assign each other their values.`, in `self.bios`, when both data exists, the data in `platform_firmware_information` part will be returned.

Comment (amepatil-git) [reply insights/parsers/dmidecode.py]: As per my understanding, only one of the field will be present, if old version of dmidecode then `bios_information` field will be present, and if new version of dmidecode it would be `platform_firmware_information`, I think ideally its should not happen that both fields are present at once, 
It can only be possible if dmidecode data output is malformed or if dmidecode authors create some new field in future and again name it `bios_information` for some other data. Due to which I was checking if one is present and other is absent assign the data to the absent key.

Both the fields being present would be a mistake given on the current dmidecode command output. I had earlier not considered assigning the fields value to anyone if both of them are present but I see that it does affect the properties which consume this data downstream. Maybe I can add a `SkipComponent` when both of them are present and also add unit test for this condition ?

In the other code part you highlighted, in `self.bios` I was trying to do computation in all the `bios*`property and return the same values in `platform_firmware*` properties. Do you think I should do independent computation separately in in those `platform_firmware*` fields ?

Comment (amepatil-git) [reply insights/parsers/dmidecode.py]: I have now updated code to use a `_get_bios_info()` where it decide the logic for getting the key.
And `self.bios` as well as `parse_content()` consume this. Please let me know what you think ?

File: insights/parsers/dmidecode.py [hunk 3]
```diff
@@ -113,6 +121,22 @@ class DMIDecode(CommandParser, LegacyItemAccess):
...
+            if "bios_information" not in self.data:
+                self.data["bios_information"] = bios_info
+
+    def _get_bios_info(self):
+        """(list): Helper to fetch the BIOS or Platform Firmware Information block list."""
+        for key in ("platform_firmware_information", "bios_information"):
+            if key in self.data and self.data[key]:
+                return self.data[key]
+        return None
+
```
qinpingli, line 134: How about updating it to something like:
```
self._bios_info = (
    self.data.get("platform_firmware_information")
    or self.data.get("bios_information")
)

# Sync self.data to ensure the information can be used by dmi[key]
if self._bios_info:
    self.data.setdefault("platform_firmware_information", self._bios_info)
    self.data.setdefault("bios_information", self._bios_info)
```

The reason is even one of key doesn't exist in self.data, we still need to sync the data and ensure all keys are the same.

amepatil-git: Thanks for the suggestion, yes this makes it more easy to understand then the my earlier changes with double `if` statement. I have made these change and pushed again.

qinpingli (approved):

PR #4811, opened 2026-08-27
feat(RHINENG-30469, RHINENG-30402): Collect rootless containers

fstavela:
### Jira

[RHINENG-29756](https://redhat.atlassian.net/browse/RHINENG-29756)
[RHINENG-30469](https://redhat.atlassian.net/browse/RHINENG-30469)
[RHINENG-30402](https://redhat.atlassian.net/browse/RHINENG-30402)

### All Pull Requests:

Check all that apply:

#### Summary

insights-client runs as root, so the existing podman ps collection only sees root-owned containers and misses the now-common rootless deployments. This adds rootless collection and a merged view, and shares the (potentially expensive) user lookup with the existing rpm datasource.

Container detection for AAP and Satellite is done directly in puptoo (https://github.com/RedHatInsights/insights-puptoo/pull/924), so this PR only provides the collection, parser, and merged combiner it consumes.

- Rootless collection — a new datasource discovers local users with a rootless podman storage directory (`~/.local/share/containers/storage`), runs `podman ps --all --no-trunc --format=json` as each user via runuser, and aggregates the output into one flat list (spec podman_ps_all_json_rootless, parser PodmanPsAllJsonRootless). The Labels key is stripped from every container (it may hold sensitive data), and the owning username is deliberately not persisted (it may be PII).
- PodmanContainers combiner — merges rootful (PodmanPsAllJson) + rootless (PodmanPsAllJsonRootless) into one searchable view via the shared PodmanContainerSearch mixin.
- Shared pwd.getpwall() lookup — pwd.getpwall() can be expensive on hosts with many users (e.g. LDAP-backed systems). The rpm datasource has called it since 2022; rather than have rootless collection call it a second time, a new internal all_users datasource performs the lookup once (broker-cached), and both pkgs_with_writable_dirs and the rootless podman datasource depend on it. all_users is not a registered spec and returns a plain list, so it is never written to the archive.

#### Tests

- Datasource: user discovery/filtering, per-user container aggregation, Labels stripping, unparseable-output, and empty SkipComponent paths
- all_users: single-lookup behavior, empty SkipComponent, and a guard asserting it is never persisted to the archive
- rpm: shell-filtered user selection now driven by the shared lookup (behavior-preserving)
- Parsers: PodmanPsAllJsonRootless (flat data, inherited search, doctests) + existing PodmanPsAllJson search helpers
- PodmanContainers: rootful-only, rootless-only, merged, and search across both

[RHINENG-29756]: https://redhat.atlassian.net/browse/RHINENG-29756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Collect and unify rootful and rootless Podman containers while safely sharing user discovery across datasources.

New Features:
- Collect rootless Podman containers from eligible users and expose them through a dedicated parser and unified searchable view with rootful containers.

Bug Fixes:
- Prevent unsafe usernames from being interpolated into commands executed as root.

Enhancements:
- Centralize and cache passwd database lookups for datasources that need user information while keeping the internal result out of collected archives.
- Remove potentially sensitive container labels and user ownership information from rootless container data.

Documentation:
- Document the new rootless Podman datasource and shared Podman containers combiner.

Tests:
- Add coverage for rootless user discovery, collection and sanitization, shared user lookup, parsing, combining, searching, and unsafe username filtering.

[RHINENG-30469]: https://redhat.atlassian.net/browse/RHINENG-30469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[RHINENG-30402]: https://redhat.atlassian.net/browse/RHINENG-30402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fstavela:
@sourcery-ai review

fstavela:
@sourcery-ai review

insights/specs/datasources/podman.py
```diff
@@ -0,0 +1,111 @@
...
+        raise SkipComponent("No users with rootless podman storage found")
+    return users
+
+
+class LocalSpecs(Specs):
+    """Local specs used only by the rootless podman datasource."""
+
+    podman_ps_rootless_raw = foreach_execute(
+        podman_rootless_users,
+        "/usr/sbin/runuser -s /bin/bash %s -c 'cd && /usr/bin/podman ps --all --no-trunc --format=json'",
```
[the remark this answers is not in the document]
fstavela: From `runuser` manual:

> For backward compatibility, runuser defaults to setting only the environment variables HOME and SHELL (plus USER and LOGNAME if the target user is not root).

So the `runuser` does set HOME even without `-l`, so the command `cd && ...` will work. If we used `-l`, PAM login-session restrictions could reject service accounts without shells or profile scripts.

insights/specs/datasources/podman.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    result = []  # type: List[dict]
+    for item in broker[LocalSpecs.podman_ps_rootless_raw]:
+        try:
+            containers = json.loads("\n".join(item.content))
+        except ValueError:
+            continue
+        for container in containers:
+            # Labels may contain sensitive information; never persist them.
+            container.pop("Labels", None)
```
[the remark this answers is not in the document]
fstavela: Fixed in the latest commit

insights/specs/datasources/podman.py
```diff
@@ -0,0 +1,117 @@
...
+        raise SkipComponent("No users with rootless podman storage found")
+    return users
+
+
+class LocalSpecs(Specs):
+    """Local specs used only by the rootless podman datasource."""
+
+    podman_ps_rootless_raw = foreach_execute(
+        podman_rootless_users,
+        "/usr/sbin/runuser -s /bin/bash %s -c 'cd && /usr/bin/podman ps --all --no-trunc --format=json'",
```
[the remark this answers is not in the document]
fstavela: This is not an issue, see https://github.com/RedHatInsights/insights-core/pull/4811#discussion_r3978511274

insights/specs/datasources/podman.py
```diff
@@ -0,0 +1,107 @@
...
+    users = _get_rootless_podman_users(broker[all_users])
+    if not users:
+        raise SkipComponent("No users with rootless podman storage found")
+    return users
+
+
+class LocalSpecs(Specs):
+    """Local specs used only by the rootless podman datasource."""
+
+    podman_ps_rootless_raw = foreach_execute(
```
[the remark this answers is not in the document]
fstavela: Fixed in the latest commit

insights/specs/__init__.py
```diff
@@ -641,6 +641,7 @@ class Specs(SpecSet):
     podman_list_containers = RegistryPoint()
     podman_list_images = RegistryPoint()
     podman_ps_all_json = RegistryPoint()
+    podman_ps_all_json_rootless = RegistryPoint()
```
qinpingli, line 644: IIUC, the content of `podman_ps_all_json_rootless` does not contain IP/MAC/hostname data, so how about updating it to:
```
podman_ps_all_json_rootless = RegistryPoint(no_obfuscate=['hostname', 'ipv4', 'ipv6', 'mac'])
```

qinpingli: Sorry, maybe command feild contain this kind of data. Let's ignore this.

insights/specs/datasources/user_group.py
```diff
@@ -5,13 +5,42 @@
...
+    :class:`insights.core.spec_factory.DatasourceProvider`, so its output is
+    never written to the archive.
+
+    Raises:
+        SkipComponent: When no users are found.
+    """
+    users = pwd.getpwall()
+    if not users:
+        raise SkipComponent("No users found")
+    return users
```
qinpingli, line 41: How about adding some logic (pwd.getpwnam) to get all valid users?

qinpingli: Sorry, seems you already added a `is_safe_username` method, how about using this method here to do the user check, and return safe users?

insights/specs/datasources/user_group.py
```diff
@@ -5,13 +5,42 @@
...
+
 from insights.core.context import HostContext
 from insights.core.exceptions import SkipComponent
 from insights.core.filters import get_filters
 from insights.core.plugins import datasource
 from insights.specs import Specs
 
 
+@datasource(HostContext)
+def all_users(broker):
```
qinpingli, line 19: How about adding this to docs/custom_datasources_index.rst?

insights/specs/datasources/user_group.py
fstavela: If you want, I can use it here, but I wouldn't. I think that the `all_users` datasource should return "all users", and the filtering should happen later.
`pkgs_with_writable_dirs` gets the data from `all_users`, but it uses it only to compare the username strings in the RPM ownership, it never interpolates it into a command, so there is no risk there. `db2.py` which runs `runuser` doesn't use `all_users` at all, it gets the users via `ps` command, so adding the check to `all_users` wouldn't help here either. That's why I had to add the `is_safe_username` usage directly into `db2_users`.

insights/specs/datasources/user_group.py
fstavela: Nice catch, done

qinpingli: Thanks!

insights/specs/datasources/user_group.py
qinpingli: >  I think that the all_users datasource should return "all users", and the filtering should happen later.

Yes, I thought this too. But another thought is: what does `all users` mean for insights-core?

Should we have a unique definition for `all users` in insights-core? If not: using insights-core to get different data for all users may have different results?

fstavela: I've moved the `is_safe_username` check directly to `all_users` datasource in https://github.com/RedHatInsights/insights-core/pull/4811/changes/076b6d3e54a82404c227cf8832d2e3c663098f3b

qinpingli: Thanks for the update!

qinpingli (approved):

PR #4818, opened 2026-09-10
Fix missing Host header in proxy CONNECT requests

### LINKED ISSUE #4817: Session doesn't explicitly set Host header for proxy connection
Some web proxies expect a `Host` header to be present in the `CONNECT` method. 
When the [Session is initialized](Proxy is being set in insight-core there: https://github.com/RedHatInsights/insights-core/blob/7b397128fabfaf200e05d25ebd5a6257687720ae/insights/client/connection.py#L171) and the proxy configuration set, it relies on the default Python requests module logic of not setting the `Host` header.

This will lead  insight clients behind some web proxy to fail as the proxy will not forward the requests.

Anakintano:
Hi,

Fixes #4817.

Adds the destination Host header to HTTPS proxy CONNECT requests while preserving existing proxy headers.

Tests: 6 focused tests passed.

## Summary by Sourcery

Ensure HTTPS proxy CONNECT requests identify their destination with a valid Host header.

Bug Fixes:
- Include the destination Host header in HTTPS proxy CONNECT requests while preserving existing proxy headers.

Enhancements:
- Apply the CONNECT header behavior only to HTTPS traffic using HTTP(S) proxies, with IPv6 authorities formatted correctly.

Tests:
- Add focused coverage for destination Host headers, proxy-header preservation, IPv6 formatting, and proxy adapter selection.


insights/client/connection.py
```diff
@@ -47,6 +50,37 @@
...
+class _ProxyHostHTTPSConnection(HTTPSConnection):
+    """Add the destination authority to HTTPS proxy CONNECT requests."""
+
+    def set_tunnel(self, host, port=None, headers=None, *args, **kwargs):
+        tunnel_headers = dict(headers or {})
+        if not any(key.lower() == 'host' for key in tunnel_headers):
+            authority_host = host
+            if ':' in authority_host and not authority_host.startswith('['):
+                authority_host = '[%s]' % authority_host
+            authority_port = self.default_port if port is None else port
```
zpetrace, line 64: Using `self.default_port` when `port is None` matches HTTPS CONNECT (443). That’s good.

If you'd add another test case, `port=None` → `Host: example.com:443` it would document this behavior explicitly. Probably just a nit though.

insights/tests/client/connection/test_init_session.py
```diff
@@ -138,3 +140,18 @@ def test_get_rhsm_and_env(logger):
     with patch.dict(os_environ, {"NO_PROXY": "redhat.com"}, clear=True):
         connection.get_proxies()
     assert connection.proxies == {"https": config.proxy}
+
+
+@patch("insights.client.connection.HTTPSConnection.set_tunnel")
+def test_https_proxy_tunnel_includes_host_header(set_tunnel):
```
zpetrace, line 146: Optional coverage: call `set_tunnel("dest.example.com", 443, {"Host": "custom"})` and assert we **do not** override an existing `Host` (matches the guard on line 59 in `connection.py`).

insights/tests/client/connection/test_init_session.py
```diff
@@ -138,3 +140,18 @@ def test_get_rhsm_and_env(logger):
...
+
+    connection.set_tunnel(
+        "2001:db8::1", 8443, {"Proxy-Authorization": "Basic token"})
+
+    set_tunnel.assert_called_once_with(
+        "2001:db8::1", 8443,
+        {"Host": "[2001:db8::1]:8443", "Proxy-Authorization": "Basic token"})
+    adapter = _ProxyHostHTTPAdapter()
+    manager = adapter.proxy_manager_for("http://proxy.example.com")
+    assert manager.pool_classes_by_scheme["https"] is _ProxyHostHTTPSConnectionPool
```
zpetrace, line 157: Nit: the adapter/pool assertion is a separate concern from tunnel headers. Splitting into `test_https_proxy_tunnel_includes_host_header` and `test_https_proxy_manager_uses_custom_pool` would make failures easier to read — totally fine to leave as-is if you prefer one file-local test.

zpetrace (changes requested):
Thanks for fixing that! Tests well cover the IPv6 + auth header path and adapter wiring. Just a couple of small test/doc nits below; nothing blocking from my side once rebased on `master` though.

insights/client/connection.py
```diff
@@ -47,6 +50,37 @@
...
+
+class _ProxyHostHTTPSConnection(HTTPSConnection):
+    """Add the destination authority to HTTPS proxy CONNECT requests."""
+
+    def set_tunnel(self, host, port=None, headers=None, *args, **kwargs):
+        tunnel_headers = dict(headers or {})
+        if not any(key.lower() == 'host' for key in tunnel_headers):
+            authority_host = host
+            if ':' in authority_host and not authority_host.startswith('['):
+                authority_host = '[%s]' % authority_host
```
qinpingli, line 63: Nit: IIUC, line 61~62 is used to check if the `authority_host` is an ipv6 address. How about adding a comment to help understand?

Anakintano: Hi .  Added a short comment explaining the IPv6 bracket formatting. 👍

qinpingli: Thanks for the update!

qinpingli (approved):

qinpingli:
test dp

